PR #520, opened 2014-04-27
Improved the snippet generation by auto-importing needed classes

### LINKED ISSUE #517: Generated snippets should allow addign needed use statements
Currently, the snippet generated with a TableNode or a PyStringNode relies on the fact that the corresponding use statement already exists in the context file.
This is true when generating the context class thanks to the template, but it may not be true anymore later if the user cleans the unused use statements.

On the other hand, the use statement for the PendingException is added by the snippet appender to ensure it is always there (without knowing whether the generated snippet actually needs it).

I think the Snippet should contain the information about which use statements it needs, and then the SnippetAppender would ensure they are present (with the logic existing for the PendingException).

stof:
Fixes #517

### DIFF
--- a/features/append_snippets.feature
+++ b/features/append_snippets.feature
@@ -402,6 +402,178 @@ Feature: Append snippets option
       }
       """
 
+  Scenario: Append snippets to main context with auto use PendingException
+    Given a file named "features/bootstrap/FeatureContext.php" with:
+      """
+      <?php
+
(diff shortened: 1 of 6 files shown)


ciaranmcnulty:
I would still like this feature please (nearly 1 year on :-P )

ciaranmcnulty:
@stof Are you going to address @everzet's suggestion? If not I think it's still usable as it is and should be merged

PR #568, opened 2014-06-22
Translate keywords in definition printers

### LINKED ISSUE #544: -dl doesn't use translation
When executing ``behat -dl --lang=nl`` I would expect to see the step types to be translated (as is also shown when doing ``behat --story-syntax --lang=nl``), but instead only the english version is shown. It would be great if they could be translated too.

Tested with Behat 3

WouterJ:
Fixes #544

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -35,6 +35,7 @@
 3.0.9 / 2014-06-20
 ==================
 
+  * [BC break] Added `KeywordsInterface` as second parameter to `ConsoleDefinitionPrinter`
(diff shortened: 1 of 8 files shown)


WouterJ:
Sorry @everzet, seems like I forgot this after my holiday. I've rebased now

stof:
This is broken

stof:
Both Travis and Scrutinizer are showing it (even though we don't have failure conditions configured on Scrutinizer so it sets the status as green)

WouterJ:
fixed (if I'm correct)

PR #617, opened 2014-08-24
Allow users to specify execution order of features

ciaranmcnulty:
This feature adds the --order flag to the CLI to allow the user to select an algorithm that determines what order features are run in (currently only feature-level sorting is supported).

Included are two default prioritisers, 'reverse' and 'random'. Both are useful for detecting dependencies between features.

bin/behat --order=random

This feature also allows future extension by registering new orderers via tagged service definitions. This would enable more sophisticated prioritisation algorithms e.g. history-based, diff-based etc.

(based on #611 with comments addressed + moved to separate Extension, dependencies moved into Testwork and away from Behat\Behat)

### DIFF
--- a/src/Behat/Testwork/Ordering/ServiceContainer/OrderingExtension.php
+++ b/src/Behat/Testwork/Ordering/ServiceContainer/OrderingExtension.php
@@ -1,0 +1,156 @@
+<?php
+
+/*
+ * This file is part of the Behat.
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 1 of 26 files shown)


src/Behat/Testwork/Ordering/ServiceContainer/OrderingExtension.php
```diff
@@ -0,0 +1,156 @@
...
+
+    /**
+     * Defines default orderers
+     *
+     * @param ContainerBuilder $container
+     */
+    private function loadDefaultOrderers(ContainerBuilder $container)
+    {
+        $definition = new Definition('Behat\Testwork\Ordering\Orderer\ReverseOrderer');
+        $definition->addTag(self::ORDERER_TAG, array('priority' => -9999));
```
stof: the priority is useless, as orderers are not sorted (they are accessed by name)

PR #618, opened 2014-08-26
Formatter JUnit

james75:
The work on this Junit formatter at https://github.com/Behat/Behat/pull/527 seems to have stalled. I have take what was done by WouterJ done a refactor on the xml creation and completed the work. I have also brought the base inline with master so it can be merged.

### DIFF
--- a/features/junit_format.feature
+++ b/features/junit_format.feature
@@ -120,6 +120,82 @@ Feature: JUnit Formatter
       """
     And the file "junit/default.xml" should be a valid document according to "junit.xsd"
 
+  Scenario: Multiple Features
+    Given a file named "features/bootstrap/FeatureContext.php" with:
+    """
+      <?php
+
+      use Behat\Behat\Context\CustomSnippetAcceptingContext,
+          Behat\Behat\Tester\Exception\PendingException;
+
+      class FeatureContext implements CustomSnippetAcceptingContext
+      {
+          private $value;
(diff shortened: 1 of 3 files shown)


src/Behat/Behat/Output/Statistics/Statistics.php
```diff
@@ -65,6 +65,14 @@
      */
     public function __construct()
     {
+        $this->resetAllCounters();
+
+        $this->timer = new Timer();
+        $this->memory = new Memory();
+    }
+
+    public function resetAllCounters()
```
everzet: If base `Statistics` doesn't work for JUnit formatter I'd rather JUnit had it's own version of `Statistics` than everyone else to depend on interface they don't fully use.

src/Behat/Testwork/Output/Printer/ConsoleOutputPrinter.php
```diff
@@ -215,7 +215,7 @@ protected function createOutputStream()
      *
      * @uses createOutputStream()
      */
-    final protected function createOutputConsole($stream = null)
+    protected function createOutput($stream = null)
```
everzet: A better way to be able to extend this point.

I'd move the whole `createOutput()` or `createOutputConsole()` behaviour to the external Factory instead of relying on override of protected methods :)

Issue #337, opened 2013-10-09
Get a list of scenarios


patxi1980:
I will find very usefull to have an option for formatting behat output only to list the scenarios that i have in the application.

The progress option is fine, but if i want only the list of scenarios i have it is useless.

The pretty option has the scenarios, but also have the steps, so it have too much information.

I understand that i have to have control of which scenarios i have and which not, but nowadays i have more than 200 scenarios (and growing), and sometimes it is difficult to know which ones are missing, and which ones are obsolete.

netraagal:
Hi,
Is this solution developped on the actual version ?

PR #635, opened 2014-09-14
[POC] Draft of a digest formatter

gquemener:
After reading https://github.com/Behat/Behat/issues/337,
I came up with the idea of implementing that formatter to get a list of scenarios.

My goal is to make it available through `behat --digest` (like `behat --config-reference`) and to make it usefull to quickly find a scenario file/line to execute this particular one.

For the moment, one must use `behat --dry-run -f digest --no-snippets`
Here's the result:
screenshot from 2014-09-15 12 17 23

In my opinion, if you think it's better to keep it as a formatter (than a --digest option), then test result should be displayed next to the scenario title with a tick, an "X", or anything that show the actual scenario execution result (one might still run with --dry-run to quicky view the list).

This is open to discussion, I used "digest" instead of "terse" cause it simply makes more sense to me.

### DIFF
--- a/features/digest_format.feature
+++ b/features/digest_format.feature
@@ -1,0 +1,54 @@
+Feature: Digest Formatter
+  In order to have an overview of the project features
+  As a feature writer
+  I need to have digest formatter
+
+  Scenario: Complex
+    Given a file named "features/World.feature" with:
+      """
+      Feature: World consistency
+        In order to maintain stable behaviors
+        As a features developer
+        I want, that "World" flushes between scenarios
+
+        Background:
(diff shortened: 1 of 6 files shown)


ciaranmcnulty:
I like this formatter! It's debatable whether it should be in core, I think that's a decision for @everzet how many formatters he wants to ship with.

everzet:

stof:
@gquemener making it a formatter means it would look very similar to a TAP output (it would just be a matter of putting the execution result on the left). So I think a better feature would be to implement a TAP formatter.

PR #647, opened 2014-10-19
allow suite settings with null values to exist

docteurklein:


ciaranmcnulty:
The CI issues here seem to just need a re-run.

The change seems sensible where values can accept a `null` :+1:

everzet:

docteurklein:
@everzet glad it landed in latest RC :)
great work ! \o/

PR #654, opened 2014-11-14
Added support for row table transformations

PurpleBooth:
When testing single things with large numbers of values it's often handy to lay them out as a row hash table, rather than the usual column hash table. 

This patch adds support for allowing transformations on those tables, allowing us to make these steps more re-usable like we already can for column based tables

Here's a little example:

The PHP for the context
```php
/** @Transform rowtable:username,age */
public function createUserFromRowTable(TableNode $table) {
    $hash     = $table->getRowsHash();
    $username = $hash['username'];
    $age      = $hash['age'];

    return new User($username, $age);
}
```

Example input Gherkin
```gherkin
Scenario:
  Given I am user:
    | username | ever.zet |
    | age      | 22       |
  Then Username must be "ever.zet"
  And Age must be 22
```

Depends on: Behat/Gherkin#89

[Review on Reviewable](https://reviewable.io/reviews/behat/behat/654)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
     "require": {
         "php":                           ">=5.3.3",
         "ext-mbstring":                  "*",
-        "behat/gherkin":                 "~4.3",
+        "behat/gherkin":                 "~4.4",
         "behat/transliterator":          "~1.0",
         "symfony/console":               "~2.1",
         "symfony/config":                "~2.3",
--- a/src/Behat/Behat/Transformation/Transformer/RepositoryArgumentTransformer.php
+++ b/src/Behat/Behat/Transformation/Transformer/RepositoryArgumentTransformer.php
@@ -15,6 +15,7 @@ use Behat\Behat\Definition\Pattern\PatternTransformer;
 use Behat\Behat\Transformation\Call\TransformationCall;
 use Behat\Behat\Transformation\Transformation;
 use Behat\Behat\Transformation\TransformationRepository;
+use Behat\Gherkin\Exception\NodeException;
 use Behat\Gherkin\Node\ArgumentInterface;
 use Behat\Gherkin\Node\TableNode;
 use Behat\Testwork\Call\CallCenter;
@@ -165,6 +166,8 @@ final class RepositoryArgumentTransformer implements ArgumentTransformer
     /**
      * Checks if provided transformation is applicable table transformation.
      *
+     * Supports both tables with columns as the headings and rows as the headings.
+     *
      * @param Transformation $transformation
      * @param mixed          $value
      *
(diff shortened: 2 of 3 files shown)


ciaranmcnulty:
I like this feature but it needs rebasing or merging.

What happens when a table has >1 row (+ header) and this transform is used?

PurpleBooth:
I'll write some tests for that scenario tomorrow and rebase this branch, I hadn't thought of that.

src/Behat/Behat/Transformation/Transformer/RepositoryArgumentTransformer.php
```diff
@@ -180,6 +200,35 @@ private function isApplicableTableTransformation(Transformation $transformation,
...
+     *
+     * @return Boolean
+     */
+    private function isApplicableRowTableTransformation(Transformation $transformation, $value)
+    {
+        if (!$value instanceof TableNode) {
+            return false;
+        };
+
+        try {
```
PurpleBooth: This logic is a bit weird. Should I move this back to the NodeTable in Gherkin, and add a "Column count" method or is it okay as it is?

PurpleBooth:
@ciaranmcnulty I wrote some tests for the scenario where the columns are empty, and discovered weird things were happening, so I've changed it to ignore tables that don't have 1 key column, 1 values column.

I haven't committed that test though because it doesn't match the style of any other tests that were in there. I can add it if need be.

matteomoretti:
+1

pfrenssen:
:+1:

composer.json
```diff
@@ -16,7 +16,7 @@
     "require": {
         "php":                           ">=5.3.3",
         "ext-mbstring":                  "*",
-        "behat/gherkin":                 "~4.3",
+        "behat/gherkin":                 "~4.4@dev",
```
pfrenssen: Gherkin 4.4 has been released, so the `@dev` can be removed.

src/Behat/Behat/Transformation/Transformer/RepositoryArgumentTransformer.php
pfrenssen: It reads a bit strange but the logic is fine. Maybe just add a line of documentation to the second try/catch to explain it?

PurpleBooth:
Fixed those suggested changes.

I am sort of thinking about closing this PR at the moment. It's nearly a year old and it still hasn't been merged in, I don't really want to keep maintaining the branch forever if it's not going to be merged.

pfrenssen:
@PurpleBooth thanks for your work, I understand it can be demotivating if it takes a long time for it to be accepted. If you want I can take over maintenance of this branch if you can add me as a collaborator. We are depending on this for our test suites at the European Commission.

@stof, do you think this is something that could be worthwhile to add to Behat? Is there anything you think needs to be done?

PR #681, opened 2015-01-05
Fixed deprecation on setFallbackLocale from symfony/translation

chillu:
For the 2.5 branch. Similar error and fix to #680.

Original error: Deprecated: The setFallbackLocale() method is deprecated since version 2.3 and will be removed in 3.0. Use setFallbackLocales() instead. in [...]/vendor/symfony/translation/Symfony/Component/Translation/Translator.php on line 163

Same as #680, we could also solve this by a more specific, lower `symfony/translation` requirement. Also, there'll likely be 5 unrelated test failures here on the Travis CI run.

### DIFF
--- a/src/Behat/Behat/DependencyInjection/config/behat.xml
+++ b/src/Behat/Behat/DependencyInjection/config/behat.xml
@@ -305,8 +305,10 @@
                 <service class="%behat.translator.message_selector.class%" />
             </argument>
 
-            <call method="setFallbackLocale">
-                <argument>en</argument>
+            <call method="setFallbackLocales">
+                <argument type="collection">
+                    <argument>en</argument>
+                </argument>
             </call>
 
             <!-- Translation loaders -->
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,12 +13,6 @@ matrix:
     - php: hhvm
   include:
     - php: 5.5
-      env: SYMFONY_VERSION='2.0.*'
-    - php: 5.5
-      env: SYMFONY_VERSION='2.1.*'
-    - php: 5.5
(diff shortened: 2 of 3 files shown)


src/Behat/Behat/DependencyInjection/config/behat.xml
```diff
@@ -305,8 +305,10 @@
                 <service class="%behat.translator.message_selector.class%" />
             </argument>
 
-            <call method="setFallbackLocale">
-                <argument>en</argument>
+            <call method="setFallbackLocales">
```
stof: This requires bumping the ``symfony/translation`` requirement to 2.3+

chillu: OK, bumped to `~2.3`.  Do you think it makes sense to bump the other symfony requirements explicitly as well? `symfony/config` and `symfony/yaml` are implicitly bumped due to https://github.com/symfony/Translation/blob/2.3/composer.json

stof: no they are not. the Translation component defines dev requirements on them only, not requirements

chillu: Ah heh, didn't read the JSON properly - all good then.

stof:
Can you rebase your branch as I fixed the tests of the 2.5 branch

stof:
@chillu ping

chillu:
@stof The branch didn't need rebasing. I've re-run the tests locally and they all pass. I've removed the 2.0, 2.1 and 2.2 Symfony build targets from `travis.yml`, since that conflicts with the new composer settings.

chillu:
Tests are passing now as well.

PR #683, opened 2015-01-10
Improved exceptions

dantleech:
- Show registered formatters in "formatter not found" exception
- All exceptions use "`" to delimit literals

### DIFF
--- a/src/Behat/Behat/Gherkin/ServiceContainer/GherkinExtension.php
+++ b/src/Behat/Behat/Gherkin/ServiceContainer/GherkinExtension.php
@@ -355,9 +355,9 @@ final class GherkinExtension implements Extension
         }
 
         throw new ExtensionException(sprintf(
-            '`%s` filter is not supported by the `filters` option of gherkin extension. Supported types are %s.',
+            '`%s` filter is not supported by the `filters` option of gherkin extension. Supported types are `%s`.',
             $type,
-            implode(', ', array('`role`', '`name`', '`tags`'))
+            implode('`, `', array('role', 'name', 'tags'))
         ), 'gherkin');
     }
 }
--- a/src/Behat/Behat/Gherkin/Specification/LazyFeatureIterator.php
+++ b/src/Behat/Behat/Gherkin/Specification/LazyFeatureIterator.php
@@ -168,10 +168,10 @@ final class LazyFeatureIterator implements SpecificationIterator
         }
 
         throw new SuiteConfigurationException(sprintf(
-            '`%s` filter is not supported by the `%s` suite. Supported types are %s.',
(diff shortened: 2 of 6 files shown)


WouterJ:
:+1:

src/Behat/Behat/Gherkin/ServiceContainer/GherkinExtension.php
```diff
@@ -355,9 +355,9 @@ private function createFilterDefinition($type, $filterString)
         }
 
         throw new ExtensionException(sprintf(
-            '`%s` filter is not supported by the `filters` option of gherkin extension. Supported types are %s.',
+            '`%s` filter is not supported by the `filters` option of gherkin extension. Supported types are `%s`.',
             $type,
-            implode(', ', array('`role`', '`name`', '`tags`'))
+            implode('`, `', array('`role`', '`name`', '`tags`'))
```
everzet: I'm not sure this one will result in what you expect it would :)

src/Behat/Behat/Gherkin/Specification/LazyFeatureIterator.php
```diff
@@ -168,10 +168,10 @@ private function createFilter($type, $filterString, Suite $suite)
         }
 
         throw new SuiteConfigurationException(sprintf(
-            '`%s` filter is not supported by the `%s` suite. Supported types are %s.',
+            '`%s` filter is not supported by the `%s` suite. Supported types are `%s`.',
             $type,
             $suite->getName(),
-            implode(', ', array('`role`', '`name`', '`tags`'))
+            implode('`, `', array('`role`', '`name`', '`tags`'))
```
everzet: Same as this one

everzet:
You need to fix failing tests.

dantleech:
Updated

features/context.feature
```diff
@@ -310,7 +310,7 @@ Feature: Context consistency
     Then it should fail with:
       """
       [Behat\Testwork\Suite\Exception\SuiteConfigurationException]
-        `contexts` setting of the "default" suite is expected to be an array, string given.
+        `contexts` setting of the "default" suite is expected to be an array, `string` given.
```
stof: this looks wrong to me. string is the type here, not the value the user entered. So IMO, it should not be enclosed in backticks

dantleech: `gettype` could return `unknown type` or `NULL` but apparently it will [never return an empty string](http://php.net/manual/en/function.gettype.php). So I'm fine with reverting that.

dantleech:
Updated.

everzet:
Perfect! Thank you very much :)

PR #684, opened 2015-01-10
Add missing --no-colors

WouterJ:
Without this, I get `Failed asserting that "some binary string that has the correct text" contains "the expected text"` on Windows (not sure it happends on Linux).


everzet:
Thank you!

PR #693, opened 2015-01-18
Adding a prototype for the html formater

webit4me:
This is to correct the branch's misspell as well as tidying it up


webit4me:
@stof , @WouterJ , hope this one be better. And once more, thanks for the help :)

stof:
@webit4me not that your commit is not associated to your github account (not the same email address). You may want to alter your commit to use the proper email, or add it in your github account

src/Behat/Behat/Output/Node/Printer/Helper/HtmlPrinter.php
```diff
@@ -0,0 +1,312 @@
...
+use Behat\Gherkin\Node\FeatureNode;
+use Behat\Gherkin\Node\PyStringNode;
+use Behat\Gherkin\Node\ScenarioLikeInterface as Scenario;
+use Behat\Gherkin\Node\StepNode;
+use Behat\Testwork\Output\Printer\OutputPrinter;
+use Behat\Testwork\Tester\Result\TestResult;
+
+/**
+ *
+ */
```
stof: this empty docblock should be removed

src/Behat/Behat/Output/Node/Printer/Html/templates/footer.html
```diff
@@ -0,0 +1,186 @@
...
+                }
+                return skip;
+            }
+
+            return this.length && pat && pat.length ? this.each(function () {
+                innerHighlight(this, pat.toUpperCase());
+            }) : this;
+        };
+
+        jQuery.fn.removeHighlight = function () {
```
stof: defining these functions does not need to be done on domReady. You can do it earlier

src/Behat/Behat/Output/Node/Printer/Html/templates/footer.html
```diff
@@ -0,0 +1,186 @@
...
+         MIT license.
+
+         Johann Burkard
+         <http://johannburkard.de>
+         <mailto:jb@eaio.com>
+
+         */
+
+        jQuery.fn.highlight = function (pat) {
+            function innerHighlight(node, pat) {
```
stof: this function does not depend on the parent scope, so you don't need to redefine it for each highlighting call

src/Behat/Behat/Output/ServiceContainer/Formatter/HtmlFormatterFactory.php
```diff
@@ -0,0 +1,460 @@
...
+            'Prints the feature as is.',
+            array(
+                'timer'     => true,
+                'expand'    => false,
+                'paths'     => true,
+                'multiline' => true,
+            ),
+            $this->createOutputPrinterDefinition(),
+            new Definition('Behat\Testwork\Output\Node\EventListener\ChainEventListener', array(
+                    array(
```
stof: wrong indentation

src/Behat/Behat/Output/Node/Printer/Html/templates/navigator.html
```diff
@@ -0,0 +1,20 @@
+<nav class="navbar navbar-inverse navbar-fixed-top">
+    <div class="container">
+        <div class="navbar-header">
+            133 scenarios (133 passed, 0 Failed, 0 undefined, 0 pending, 0 Skipped)
+            <br>
+            679 steps (679 passed, 0 Failed, 0 undefined, 0 pending, 0 Skipped)
+            <br>
+            1m31.73s (memory)
```
stof: hardcoded stats look wrong to me

src/Behat/Behat/Output/Node/Printer/Html/AbstractHtmlPrinter.php
```diff
@@ -0,0 +1,48 @@
...
+
+use Behat\Behat\Output\Node\Printer\Helper\HtmlPrinter;
+use Behat\Testwork\Output\Printer\OutputPrinter;
+
+/**
+ * To retrieve (html printer) helper .
+ *
+ * @author Ali Bahman <abn@webit4.me>
+ */
+abstract class AbstractHtmlPrinter
```
stof: I would vote for removing this abstract class. It does not add much value IMO

stof:
You are missing HTML escaping in tons of places

src/Behat/Behat/Output/Node/Printer/Helper/HtmlPrinter.php
webit4me: Added a line of description

src/Behat/Behat/Output/Node/Printer/Html/templates/footer.html
webit4me: My bad, they are not in use at the moment anyway so i've removed them until i can find a solid solution

src/Behat/Behat/Output/Node/Printer/Html/templates/footer.html
webit4me: been removed since is not in use at the moment

src/Behat/Behat/Output/ServiceContainer/Formatter/HtmlFormatterFactory.php
webit4me: Tidied up the entire file :)

src/Behat/Behat/Output/Node/Printer/Html/templates/navigator.html
webit4me: don't know how lost the track on this :( , replaced values with their template's variables

src/Behat/Behat/Output/Node/Printer/Html/AbstractHtmlPrinter.php
webit4me: Agreed, will take care of it at some point today

webit4me: I took that abstraction out but i have to admit now there is too much duplication in such small module :) 
anyway all changes are in a single commit so we would be able to revert them if change our mind

webit4me:
@stof , escaping html in the html format sounds a bit tricky.
In one hand i would say some developers might prefer to use some html tags in their features to style up the result/report as well as in some cases we might have html tags as part of the step arguments e.g. passing html partial  inside a multiple lines steps (""") to be checked during a test.
any thought? :)

jakzal:
> some developers might prefer to use some html tags in their features to style up the result/report

Styling reports by modifying features sounds terribly wrong.

webit4me:
so what's the suggestion , shall we escape all html special characters? if so what will happened to  fixtures with some special character in them e.g. "<?php" in the [append_snippets.feature](https://github.com/Behat/Behat/blob/master/features/append_snippets.feature) ?

WouterJ:
When escaping, they will be shown as <?php.

Op za 24 jan. 2015 16:27 schreef Ali Bahman <notifications@github.com>:

webit4me:
so if thats the case i think i can easily walk through all variables passing to the getTemplate method and clear them up just before being passed to the template file.

webit4me:
Unfortunately it wasn't possible to escape all html tags by walking on the variables passing to he template. so i have done it in different places.
but still i would suggest to consider the opportunity of styling the report output by amending the fixture, i.e. like at the moment in the pretty formatter, one can put some more pressure on specific word in a scenario's title by wrapping it in a failed tag and make it bolder for the developer whom going to satisfy that test 
e.g. Scenario: Extension {+failed} should {-failed} be successfully loaded

PR #698, opened 2015-01-23
Fix infinite loop

stof:
See http://3v4l.org/K2nbU for the reason.

The fix already exists in the 3.0 branch: https://github.com/Behat/Behat/blob/v3.0.13/src/Behat/Behat/Context/Reader/AnnotatedContextReader.php#L86-90
However, the comment in this branch looks weird.

### DIFF
--- a/features/annotations/traits_inheritance.feature
+++ b/features/annotations/traits_inheritance.feature
@@ -1,0 +1,145 @@
+@php-version @php5.4
+Feature: Support php 5.4 traits with inheritance
+  In order to have much cleaner horizontal reusability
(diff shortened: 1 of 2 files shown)


stof:
note that a few scenarios are failing in the Behat 2.5 branch because of changes in the Symfony Console component about the way colors are handled. This is not related to this fix. But the new scenario added in the first commit is really failing without the fix.

stof:
OK, rebased as I fixed color tests in #699.

everzet:
@stof awesome! Thanks for doing it.

PR #706, opened 2015-02-01
Use high res badges

hanneskaeufler:


everzet:

PR #713, opened 2015-02-16
Scenario that closes #668

### LINKED ISSUE #668: `FeatureContext` is used independent of what `suites` config says
Independent of what config you provide inside `suites`, if `FeatureContext` class is autoloadable, `default` suite will be used.

sskorc:
Refs #668


everzet:

PR #717, opened 2015-03-05
Fixed deprecated call to setFallbackLocale triggered by behat.xml

sminnee:
This is the equivalent of d8b73d1bd4925f926d8707c2d1e4ab5bd661c4b3 but for the 2.5 branch

### DIFF
--- a/src/Behat/Behat/DependencyInjection/config/behat.xml
+++ b/src/Behat/Behat/DependencyInjection/config/behat.xml
@@ -305,8 +305,10 @@
                 <service class="%behat.translator.message_selector.class%" />
             </argument>
 
-            <call method="setFallbackLocale">
-                <argument>en</argument>
+            <call method="setFallbackLocales">
+                <argument type="collection">
+                    <argument>en</argument>
+                </argument>
             </call>
 
             <!-- Translation loaders -->


sminnee:
Hrm, this failed on Symfony 2.0 - 2.2, but was necessary to work in Symfony 2.3. Not sure what to do to fix, except maybe mark the module as not working with 2.3?

stof:
Symfony 2.0 to 2.3 are EOLed, so I think it is fine to bump the requirement to 2.3+ for ``symfony/translator``.

What do you think @everzet ?

everzet:
@stof I agree.

stof:
@sminnee please update the ``symfony/translator`` constraint in the composer.json and remove these versions from the Travis config then

danieledangeli:
What about this feature? I still have an error with behat/behat": "2.5.4"

stof:
Well, this PR needs to be updated

danieledangeli:
@stof I fixed it in my project by requiring a specific version for symfony/translator (2.1). I saw that in your composer your are requiring for ~2.1 (so between 2.1 and 3.0). Which is the proper fix for that?

stof:
@danieledangeli don't use 2.1 (which is not maintained anymore). Use 2.6 if you don't want the deprecation warnings (only symfony 2.7-dev is throwing deprecation warnings).
and the proper fix for the issue is the one in this PR, except that it is not ready to be merged as it breaks the testsuite. The PR author has not updated the PR based on the feedback received

danieledangeli:
@stof Nice! Thanks!

chillu:
Should be closed as duplicate of https://github.com/Behat/Behat/pull/681? This other PR already bumps the dependency versions.

stof:
Closing as I merged #681

PR #720, opened 2015-03-16
[POC] running scenarios in separate processes

gonzalovilaseca:
This is a quick POC for running each scenario as a different process. It will require some refactoring in order to make it work properly, but just wanted to know if the community wants to go this way before spending more time on it.

The reason for this is that our suite eats up memory fast. Our app is on top of Sylius and Symfony, we've tried to make sure the container was released, we looked for memory leaks ... etc etc, but the memory just grew after each scenario and wasn't released before the next one would run.

File inclussion in autoloaders are never released from memory until the php session is closed, so memory usage grows due to this fact too.

By running a scenario in a different process the memory is released after each one.

### DIFF
--- a/src/Behat/Testwork/Tester/Runtime/RuntimeSuiteTester.php
+++ b/src/Behat/Testwork/Tester/Runtime/RuntimeSuiteTester.php
@@ -20,6 +20,8 @@ use Behat\Testwork\Tester\Setup\SuccessfulSetup;
 use Behat\Testwork\Tester\Setup\SuccessfulTeardown;
 use Behat\Testwork\Tester\SpecificationTester;
 use Behat\Testwork\Tester\SuiteTester;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Process\Process;
 
 /**
  * Tester executing suite tests in the runtime.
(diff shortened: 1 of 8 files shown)


src/Behat/Testwork/Tester/Runtime/RuntimeSuiteTester.php
```diff
@@ -54,17 +56,29 @@ public function setUp(Environment $env, SpecificationIterator $iterator, $skip)
...
+
         $results = array();
         foreach ($iterator as $specification) {
-            $setup = $this->specTester->setUp($env, $specification, $skip);
-            $localSkip = !$setup->isSuccessful() || $skip;
-            $testResult = $this->specTester->test($env, $specification, $localSkip);
-            $teardown = $this->specTester->tearDown($env, $specification, $localSkip, $testResult);
+            if ($batch) {
+                foreach ($specification->getScenarios() as $scenario) {
+                    $cmd = 'bin/behat ' . $specification->getFile() . ":" . $scenario->getLine();
```
stof: this does not work because you assume a specific path to Behat, which is likely to be wrong in many cases

src/Behat/Testwork/Tester/Runtime/RuntimeSuiteTester.php
```diff
@@ -54,17 +56,29 @@ public function setUp(Environment $env, SpecificationIterator $iterator, $skip)
...
+            } else {
+                $setup      = $this->specTester->setUp($env, $specification, $skip);
+                $localSkip  = !$setup->isSuccessful() || $skip;
+                $testResult = $this->specTester->test($env, $specification, $localSkip);
+                $teardown   = $this->specTester->tearDown($env, $specification, $localSkip, $testResult);
 
-            $integerResult = new IntegerTestResult($testResult->getResultCode());
-            $results[] = new TestWithSetupResult($setup, $integerResult, $teardown);
+                $integerResult = new IntegerTestResult($testResult->getResultCode());
+                $results[]     = new TestWithSetupResult($setup, $integerResult, $teardown);
```
stof: you should not change coding standards when contributing. Assignment operators are not aligned

gonzalovilaseca:
@stof Please read the description, this is just a POC, by no means this is something that should be merged. 
It's just so the community get's the concept, and if it's worth spending more time in doing it properly.

stof:
I don't like this concept of ``$batch``. And your POC is broken: as soon as I specify a path, you consider it is the inner process, even though multiple scenarios are matched.

Anyway, I'm not a fan of this. Behat should not force using subprocesses (especially not subprocesses running Behat again). And the way you implemented it makes it quite impossible to write formatters properly (try to combine your work with the JUnit formatter PR for instance).

The fact that Symfony does not release memory properly on shutdown is an issue which should be solved in Symfony, not in Behat: https://github.com/symfony/symfony/issues/13448
And if you want to isolate scenarios, a better way would be to run the Symfony code inside a subprocess only (which is supported by the BrowserKit component, but MinkExtension does not allow to configure the client as insulated yet)

stof:
Btw, in the case of the Symfony2 driver type, MinkExtension currently gets the client configured in the Symfony2 kernel, so making it insulated would be a matter of manipulating the container of your project to add the needed method call.

gonzalovilaseca:
@stof This POC only works when you specify a suite: I'm not going to waste time covering all cases if the core idea is flawed.

Even if the app memory is released there will allways be the increase in memory usage due to included files, so in a very large app test suite, you can run out of memory even if there are no leaks.

stof:
@gonzalovilaseca if you run the Symfony request in a subprocess, your application will only be loaded in the subprocess

gonzalovilaseca:
@stof I know, will try that route then, thanks.

gonzalovilaseca:
Just in case someone else lands here, there is also the option of settind the insultate on a per scenario or step basis:
```
/**
 * @BeforeStep
 */
public function insulateProcess(BeforeStepScope $scope)
{
    $this->getSession()->getDriver()->getClient()->insulate(true);
}
```

stof:
@gonzalovilaseca this won't do it on a per-scenario basis if you don't rset it at the end (it will impact all next steps too)

gonzalovilaseca:
@stof Sure, you can do that in the After* steps, it all depends on the case.

PR #724, opened 2015-03-30
Allow no tests to be considered passing tests.

typhonius:
I've run into an issue where no tests is considered test failure when running with --strict. Some of our suites have to be excluded from tests but this then breaks phing which expects a 0 error code.

This change would mean that $result->isPassed() would return true if the return code was 0 or -100 (no tests).

In the event that this breaks some kind of backwards compatibility, I'd also be happy to look at creating a new flag for this.

### DIFF
--- a/src/Behat/Testwork/Tester/Result/TestResult.php
+++ b/src/Behat/Testwork/Tester/Result/TestResult.php
@@ -21,6 +21,7 @@ interface TestResult
     const SKIPPED = 10;
     const PENDING = 20;
     const FAILED = 99;
+    const NO_TESTS = -100;
 
     /**
      * Checks that test has passed.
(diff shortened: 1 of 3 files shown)


stof:
Well, if you want suites without tests to be considered as passing, don't run them in strict mode. The soft mode already allows such suites to be marked as passing

typhonius:
Thanks for your response. Because everything gets run in a big loop over a bunch of files, we don't know whether a test will have 0 suites or some until we run it.

I have considered writing a new assertion that just returns true. I could then use this to run everything with --strict and have at least one test suite run.

I suppose conceptually I am wondering whether running no tests should constitute a test failure.

ciaranmcnulty:
Like @stof says this feels like something that you can solve by not using `--strict`.

ciaranmcnulty:
I'm closing as there's a straightforward solution (don't use `--strict`)

PR #725, opened 2015-04-02
Environment configuration should override file configuration

nervo:
From my point of view, configuration defined by environment variable (BEHAT_PARAMS) should override values defined in behat.yml.
Given the order behat actually merge them, it's the opposite :)

### DIFF
--- a/src/Behat/Testwork/ServiceContainer/Configuration/ConfigurationLoader.php
+++ b/src/Behat/Testwork/ServiceContainer/Configuration/ConfigurationLoader.php
@@ -99,16 +99,16 @@ final class ConfigurationLoader
         $configs = array();
         $this->profileFound = false;
 
-        // first is ENV config
-        foreach ($this->loadEnvironmentConfiguration() as $config) {
-            $configs[] = $config;
-        }
-
-        // second is file configuration (if there is some)
+        // first is file configuration (if there is some)
         if ($this->configurationPath) {
             foreach ($this->loadFileConfiguration($this->configurationPath, $profile) as $config) {
                 $configs[] = $config;
             }
+        }
+
+        // second is ENV config
+        foreach ($this->loadEnvironmentConfiguration() as $config) {
+            $configs[] = $config;
         }
 
         // if specific profile has not been found


everzet:
More explicit setting should always override more implicit one. Anything else causes a lot of confusion.

nervo:
Hm, yes, for sure.
Once again, from my point of view, configuration given by command line is far more explicit than configuration file :)
Am i the only one thinking like that ?

everzet:
Configuration given by command line is the most explicit one. Flags that you use when you run `behat` override both configuration and environment variables. Here's the order of configuration loading:

1. Environment variable
2. Configuration file
3. Explicit command line flags

I find this order very intuitive and the one creating least confusion. The fact that you're the first person in couple of years questioning it kind of proves that point :)

nervo:
Maybe i'm just part of the 3 persons on earth who need to use environment variables with behat :)

I think it's time to describe my use case: on a project, we have to keep at least two environments for testing. One is "dev", for developpers, where base_url looks like "project.dev/app.php". Other is "test" where base_url is unpredictable (vms are launched on the fly).
Things must be keeped simple for dev (aka. config file shared across all developpers, and a simple command line without parameters), and very specific on the test environment (some parameters must be given at run time)

I could not figure i'm the only one in this case :)

diamondsea:
Nervo is correct, the proper order for settings priority should be:

1. Command Line Options (most specific, overrides all others below)
2. Environment Variables  
3. Configuration Files    
4. Program Defaults       (least specific, only used if not overridden by one of the above)

(If you still don't believe me, here's a link on AWS describing the same precedence: http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html#config-settings-and-precedence and in the Java Spring docs: http://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html)

We are running a complicated environment at www.DOT.gov (Drupal Multisite with multiple 
stage and test servers, where testing specific sites against specific browsers and tags in different combinations on our local dev machines as well as on our servers.  

Having a single YAML file that allows all the combinations of what we need to do doesn't work (I'm going to submit another issue with suggestions on that).  The only way we've been able to do what we need is using environment variables (set inside scripts, in our case) to control the settings, and to do that we have had to remove all those changing options from our behat.yml file because it overwrites our environment settings.  While this works, it then makes it impossible to run a behat test just from the normal command line because we've had to remove all these settings to get it to work.

Changing the precedence to the order shown above will provide a more flexible system that can be more easily tailored.

nervo:
Yep, dude !

benji07:
:+

woecifaun:
:+1:

ciaranmcnulty:
I agree that Environment is more specific than config file.

One comes from a VCS (typically), the other is local to the machine.

*However* making this change would be a BC break so could potentially cause issues if done in `3.x`

diamondsea:
How about this to preserve BC but allow the proper precedence in 3.x:

> Set an environment variable, BEHAT_ENV_PRECEDENCE=1, that will activate the new behavior?    We're going to be setting environment variables anyway, what's one more to make it work?

Admittedly, it's a hack, but it would allow for BC until the normal precedence (cmd, env, yml, code) becomes the default (in whatever version that ends up being in) and the BEHAT_ENV_PRECEDENCE variable is deprecated/removed, while allowing the additional flexibility of using the environment variables now for those who need it.

nervo:
'Sounds weird, but acceptable :)

everzet:
I'm ok to go with `BEHAT_ENV_PRECEDENCE`. Feel free to open a separate PR

nervo:
Ok, have a look on https://github.com/Behat/Behat/pull/816

PR #735, opened 2015-04-30
RFC : Auto-spin for 'Then' steps

Chekote:
Just to be clear: I'm not asking this to be merged, I just want to discuss the concept. The implementation here is a very quick and ugly hack to the Behat source.

I posted a question to the [Behat Gitter](https://gitter.im/Behat/Behat) channel, but it's dead there. So I'm hoping I can get some input here:

I'm trying something out and wanted some advice/feedback. I'm wanting to setup auto-spin for all of our "Then" steps, as by our convention these are always assertions. I tried to do it with Behat's standard hooks, by catching the failed step in an `@AfterStep` and then detecting if the step needs to be re-attempted. But the scope I receive seems to be read-only.

Is this approach going to be possible? I don't see a way to define my own StepTester class. Is it even possible to do what I'm trying without hacking Behat? This pull requests demonstrates what exactly I'm trying to do. Hopefully it'll clear up any questions.

I'd love to hear some feedback on if anyone things this is a good/terrible idea, and how I would go about implementing this "properly", either for my own projects in isolation, or as a feature for Behat itself.

### DIFF
--- a/src/Behat/Behat/Tester/Runtime/RuntimeStepTester.php
+++ b/src/Behat/Behat/Tester/Runtime/RuntimeStepTester.php
@@ -14,6 +14,7 @@ use Behat\Behat\Definition\Call\DefinitionCall;
 use Behat\Behat\Definition\DefinitionFinder;
 use Behat\Behat\Definition\Exception\SearchException;
 use Behat\Behat\Definition\SearchResult;
+use Behat\Behat\Tester\Exception\PendingException;
 use Behat\Behat\Tester\Result\ExecutedStepResult;
 use Behat\Behat\Tester\Result\FailedStepSearchResult;
 use Behat\Behat\Tester\Result\SkippedStepResult;
@@ -23,6 +24,7 @@ use Behat\Behat\Tester\StepTester;
 use Behat\Gherkin\Node\FeatureNode;
 use Behat\Gherkin\Node\StepNode;
 use Behat\Testwork\Call\CallCenter;
+use Behat\Testwork\Call\CallResult;
 use Behat\Testwork\Environment\Environment;
 use Behat\Testwork\Tester\Setup\SuccessfulSetup;
 use Behat\Testwork\Tester\Setup\SuccessfulTeardown;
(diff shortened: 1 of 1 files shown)


Chekote:
@stof Where would be the best place for me to discuss this concept? I'm not getting any response here or on [Behat Gitter](https://gitter.im/Behat/Behat).

src/Behat/Behat/Tester/Runtime/RuntimeStepTester.php
```diff
@@ -122,12 +137,48 @@ private function testDefinition(Environment $env, FeatureNode $feature, StepNode
...
-        $result = $this->callCenter->makeCall($call);
+
+        $lambda = function() use ($call) {
+          return $this->callCenter->makeCall($call);
+        };
+
+        // @todo We can only "spin" if we are interacting with a remote browser. If the browser is
+        // running in the same thread as this test (such as with Goutte or Zombie), then spinning
+        // will only prevent that process from continuing, and the test will either pass immediately,
+        // or not at all. We need to find out how to check what Driver we're using...
```
stof: anything talking about browser should be removed. Behat does not know about browsers

stof:
Replacing the StepTester should be totally possible from an extension.

I don't like this idea:

- using the keyword to determine this is wrong. Behat accepts any step for any keyword (the different keywords are there to make your steps more readable, but Behat does not care at all)
- limiting it to Then step is not the right behavior either. ``When I click on the button" may benefit from spinning in case it cannot find the button for instance
- spinning any failing step automatically assumes that any failed step can be retried safely. I can think of cases where this will break your suite. The spinning should probably be done based on the type of exceptions being thrown (for things coming from Mink, it may be based on ``ExpectationException`` or ``ElementNotFoundException`` while not retrying other failures like failures to load the page for instance)

Chekote:
Thank you so much for your feedback, I really appreciate it. I knew you wouldn't like the idea XD.

I agree with you completely that from the perspective of Behat core, using the keyword to determine the behavior is definitely wrong. But in our particular project, it isn't because by convention we only use `Then` for assertions.

Another convention that we use in our tests is to use assertions to ensure a valid state is achieved before  attempting to interact with the system. For example, given your example of "When I click on the button", a previous step would already have asserted that there is a button. But I do definitely see a benefit of having the "When I click on the button" spin without the assertion step being necessary, as it'll make the scenario less verbose.

You make a good point about the exception types. I'll definitely need to address that.

ciaranmcnulty:
I agree that this is not appropriate for core, but probably makes sense for your particular usage so would be a good subject for an Extension.

I'm not sure this should be left open?

PR #746, opened 2015-05-26
Remove an argument from StreamOutputPrinter::getOutputFactory() call

jakzal:
The getOutputFactory() method does not have an argument.


stof:
thanks

PR #752, opened 2015-06-01
Remove usage from expected output in scenarios verifying failed output

### LINKED ISSUE #751: The scenarios should not depend on the way Symfony Console display the command help
The format of the command usage changed in Symfony 2.7, making scenarios fail because they assert the output.
It would be better to write these scenarios in a way not depending on the usage output (which will also avoid breaking them all each time we add a new option)

jakzal:
This will make scenarios less sensitive to changes in the way Symfony displays usage details. Usage output is not relevant to the scenarios we expected it in anyway.

Fixes #751


stof:
thanks

PR #754, opened 2015-06-06
Support for negative numbers without surrounding quotes

ryancookdev:
When a definition parameter is not quoted, it is still interpreted as a parameter as long as it has a numeric value. However, the regex only included digit (\d), period (.), and comma (,) characters. I added a hyphen character (-), as this is the symbol commonly used to represent negative numbers when a user is typing on a computer keyboard.

Prior to this commit, the hyphen would get ejected in the generated snippet (`*@Given I have a negative number -:num*/`) and the value passed would be positive instead of negative. This behavior could be corrected by adding quotes in the Gherkin file ("-3" instead of just -3), but the behavior is still counter-intuitive. The lexer is already interpreting numeric sequences, so it makes sense to at least support negative numbers. More complex number notations (hexadecimal, scientific, etc) will still need to be quoted to work properly.

### DIFF
--- a/features/definitions_patterns.feature
+++ b/features/definitions_patterns.feature
@@ -535,3 +535,35 @@ Feature: Step Definition Pattern
       1 scenario (1 passed)
       1 step (1 passed)
       """
+
+  Scenario: Negative number parameters without quotes
+    Given a file named "features/bootstrap/FeatureContext.php" with:
+      """
+      <?php
+
+      use Behat\Behat\Context\Context;
+
+      class FeatureContext implements Context
+      {
+          /**
+           * @Given I have a negative number :num
+           */
+          public function multipleWrongNamedParameters($num) {
+          PHPUnit_Framework_Assert::assertEquals('-3', $num);
+          }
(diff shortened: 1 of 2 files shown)


features/definitions_patterns.feature
```diff
@@ -535,3 +535,35 @@ Feature: Step Definition Pattern
...
+
+      use Behat\Behat\Context\Context;
+
+      class FeatureContext implements Context
+      {
+          /**
+           * @Given I have a negative number :num
+           */
+          public function multipleWrongNamedParameters($num) {
+          PHPUnit_Framework_Assert::assertEquals('-3', $num);
```
ciaranmcnulty: indentation here seems wrong

ciaranmcnulty:
Do you need a scenario that covers this for snippet generation?

features/definitions_patterns.feature
everzet: It doesn't really matters that much in tests.

everzet:

PR #762, opened 2015-06-17
Bugfix: notice ob_end_clean()

tscheepers:
I'm getting a notice `Notice: ob_end_clean(): failed to delete buffer. No buffer to delete in vendor/behat/behat/src/Behat/Testwork/Call/Handler/RuntimeCallHandler.php line 137` which is really weird because the method is only called from `handleCall`, maybe its a PHP bug.
The notice causes the tests to fail. This patch prevents the notice if `ob_start` failed for some reason.

screen shot 2015-06-17 at 14 37 36


stof:
please add curly braces for the if

tscheepers:
@stof done :-)

ciaranmcnulty:
It's a shame the root cause for your error isn't known, but :+1: on adding the guard.

Issue #759, opened 2015-06-11
No error on missing config file (specified with -c filename)


diamondsea:
If behat is invoked with a `-c configfile.yml` parameter, and that file is not found, it does not display an error but continues running without it.

```
$ bin/behat -c missing_file.yml features/example.feature
# behat runs without displaying any error (presumably using default behat.yml file)
```

adwww:
Hey, I had this issue too and thought that looks like it's worth fixing (PR attached). However it broke a test around the use of environment variables - I'm not sure fully what the impact is if others could take a look please?

PR #763, opened 2015-06-17
Throw exception if requested config file (-c) does not exist

watermanio:
Relates to issue #759 which I thought was clearly a bug, not alerting users that they have specifically requested a config file that gets quietly ignored.

However, running the existing tests showed two failures from this change in `parameters.feature`. I've "fixed" those failures, but I'm not 100% sure if the `-c unexistent` part was desirable there, or even harmful? Hmm.

### DIFF
--- a/features/config.feature
+++ b/features/config.feature
@@ -8,7 +8,7 @@ Feature: Config
       """
       """
     And a file named "features/bootstrap/FeatureContext.php" with:
-    """
+      """
       <?php
 
       use Behat\Behat\Context\Context;
@@ -18,7 +18,7 @@ Feature: Config
       }
       """
     And a file named "features/config.feature" with:
-    """
+      """
       Feature:
         Scenario:
           When this scenario executes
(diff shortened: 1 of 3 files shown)


features/config.feature
```diff
@@ -35,3 +35,59 @@ Feature: Config
 
           When this scenario executes
       """
+
+  Scenario: Alternative configuration file
+    Given a file named "alternative-behat.yml" with:
+    """
```
ciaranmcnulty: Indentation of your opening `"""` is off in all scenarios (this is common when copy/pasting)

src/Behat/Testwork/Cli/Application.php
```diff
@@ -99,16 +100,22 @@ public function doRun(InputInterface $input, OutputInterface $output)
...
 
-        if (is_file($path = $input->getParameterOption(array('--config', '-c')))) {
-            $this->configurationLoader->setConfigurationFilePath($path);
+        // Ignore all other inputs if config-reference is set:
+        if ($input->hasParameterOption(array('--config-reference'))) {
+            $input = new ArrayInput(array('--config-reference' => true));
         }
 
-        $this->add($this->createCommand($input, $output));
+        // Check the requested config file exists
```
ciaranmcnulty: You should not commit these comments - if you think it needs explanation then maybe these pieces of logic can be extracted into private methods (or collaborators)

ciaranmcnulty:
This change makes sense to me

PR #765, opened 2015-06-23
Bug: --verbose does not support verbosity level > 1

pfrenssen:
The Symfony Console component allows to increase the verbosity level in different steps. Starting from level 2 exception backtraces are shown in addition to the exception messages. However this currently doesn't work in Behat:

```
$ ./vendor/bin/behat --verbose=2
  [RuntimeException]                               
  The "--verbose" option does not accept a value.
```

Proposed commit fixes this, and adds a line to the help explaining that this option can be used.

### DIFF
--- a/src/Behat/Testwork/Cli/Application.php
+++ b/src/Behat/Testwork/Cli/Application.php
@@ -65,7 +65,11 @@ final class Application extends BaseApplication
         return new InputDefinition(array(
             new InputOption('--profile', '-p', InputOption::VALUE_REQUIRED, 'Specify config profile to use.'),
             new InputOption('--config', '-c', InputOption::VALUE_REQUIRED, 'Specify config file to use.'),
-            new InputOption('--verbose', '-v', InputOption::VALUE_NONE, 'Increase verbosity of exceptions.'),
+            new InputOption(
+                '--verbose', '-v', InputOption::VALUE_OPTIONAL,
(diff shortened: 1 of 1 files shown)


nchagrass:
This functionality already exists by using: "-v -vv -vvv"

pfrenssen:
Yes but it is broken for --verbose, that is being fixed by this pull request.

nchagrass:
I don't think we need to replicate the Symfony/Console behaviour here but it is not for me to say.

pfrenssen:
I think this is simply a bug, somebody accidentally used the option `VALUE_NONE` instead of `VALUE_OPTIONAL`. I don't think it was ever the intention to have no feature parity between the full option and the shortcut. Currently `--verbose` is broken since it can only be used to set the first out of 4  possible error levels.

I stumbled upon this because in my current project we're using Phing to run the tests instead of running them directly on the command line. The Phing BehatTask only supports POSIX long options, not the shorthand ones.

ciaranmcnulty:
:+1: Looks good to me

ademarco:
It works fine for me too, thanks to this patch we can now run Phing BehatTask successfully! :+1:

gboddin:
+1 for POSIX long options compliance !

sandervd:
:thumbsup: thanks @pfrenssen!

PR #753, opened 2015-07-07
TeamCity and jUnit output

dizzy7:
Hello.

The Behat first creates empty xml files (containing empty testsuites tag only) and fills them after the test is passed. But the TeamCity does realtime monitoring, so it processes emty xml files when they appear and does not return to them later after data is filled in. So in fact the TeamCity gets no infomation about tests failed or passed. It would be great if behat could create filled files at once avoiding empty ones.

Thanks.

### DIFF
--- a/features/junit_format.feature
+++ b/features/junit_format.feature
@@ -527,67 +527,6 @@
       """
     And the file "junit/default.xml" should be a valid document according to "junit.xsd"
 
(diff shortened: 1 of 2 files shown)


ciaranmcnulty:
On the face of it this seems reasonable, but you should replace the scenario with one explaining the new behaviour.

everzet:

PR #774, opened 2015-07-31
Fix JUnit output for Jenkins

v-abramchuk:
The current implementation of JUnit output does not work for Jenkins, always causing it to show the steps as passed.

This PR fixes the output format.


ciaranmcnulty:
You will need to also update the test suite.

Are you sure this is invalid JUnit XML? I know a few teams already using the JUnit formatter.

everzet:

Issue #744, opened 2015-05-21
Behat 3.0 does not allow suite names that contain a hyphen


legovaer:
Well, the title describes the issue pretty much. When using a hyphen in the name of a suite, Behat does not detect that suite;

`./bin/behat --suite=my-test-suite`

throws:
```
  [Behat\Testwork\Suite\Exception\SuiteNotFoundException]
  `my-test-suite` suite is not found or has not been properly registered.
```
As soon as I remove the hyphens, the exception is gone.

This is not a blocking issue, but I thought it might be interesting to share this.

stof:
This is because of the config normalization replacing hyphens with underscores in the config keys. We should disable this normalization for keys of prototyped nodes (I hope we could make it the default behavior in Symfony 3.0 as this is the logical behavior but it is not the default on for legacy reasons).

With the current code, your suite is actually named ``my_test_suite``

legovaer:
Indeed. Executing `./bin/behat --suite=my_test_suite` works, but isn't the expected behavior of course. Maybe a remark needs to be added to the documentation?

stof:
I'd rather fix this issue (it is a one-line addition in the code) than documenting it.

PR #776, opened 2015-08-17
Disable normalization of suite key

WouterJ:
Fixes https://github.com/Behat/Behat/issues/744


stof:
Can you make the same for all other prototyped nodes ?

PR #779, opened 2015-08-21
Performance improvement to regex transformation

Sam-Burns:
This PR uses array caching in the transformation of Turnip step annotations to regular expressions.

It will not noticeably affect those users who have lots of suites, each with its own Context.  However, if the user's tests are organised using traits and a single Context, this yields a significant performance improvement for large suites.

The number of times the `TurnipPatternPolicy::transformPatternToRegex()` method is executed is `(steps executed)*(step definitions available)` - in our case `5k * 670 = 3.35M`.  Blackfire reports the inclusive CPU time of the `TurnipPatternPolicy::transformPatternToRegex()` method as ~26.3% of total execution time without this PR, and negligible after the change proposed.

Clearly, using traits, a single context and only one suite isn't an approach that scales well in Behat-land.  Nevertheless, I think simple caching here is going to help performance for unwieldy test suites.

### DIFF
--- a/src/Behat/Behat/Definition/Pattern/Policy/TurnipPatternPolicy.php
+++ b/src/Behat/Behat/Definition/Pattern/Policy/TurnipPatternPolicy.php
@@ -29,7 +29,7 @@ final class TurnipPatternPolicy implements PatternPolicy
     /**
      * @var string[]
      */
-    private $regexCache = [];
+    private $regexCache = array();
 
     /**
      * @var string[]
--- a/src/Behat/Behat/Definition/Pattern/Policy/TurnipPatternPolicy.php
+++ b/src/Behat/Behat/Definition/Pattern/Policy/TurnipPatternPolicy.php
@@ -29,6 +29,11 @@ final class TurnipPatternPolicy implements PatternPolicy
     /**
      * @var string[]
      */
+    private $regexCache = [];
+
+    /**
+     * @var string[]
+     */
     private static $placeholderPatterns = array(
         "/(?<!\w)\"[^\"]+\"(?!\w)/",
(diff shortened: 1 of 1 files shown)


ciaranmcnulty:
In a suite that large, what impact does this have on memory usage? That's the other side of the coin when doing this type of optimisation. I'd love to see some stats from your suite?

If it's not a major impact I'm :+1: on this change.

stof:
Can you share the blackfire comparison between before and after the change ?

stof:
> It will not noticeably affect those users who have lots of suites, each with its own Context. However, if the user's tests are organised using traits and a single Context

How does it depend on the number of suites/contexts ? the TurnipPatternPolicy is shared between all suites/contexts AFAIK

Sam-Burns:
@ciaranmcnulty The array is getting 670 strings in our use case (i.e. no. of step definitions).  The strings are regexs made out of Gerkin sentence fragments here, so memory usage went up 202kB in our case.

Sam-Burns:
@stof You’re right, the class is shared across suites/contexts, but `Behat\Behat\Definition\Search\RepositorySearchEngine::searchDefinition()` is only calling `TurnipPatternPolicy::transformPatternToRegex()` once per *available* step definition (for the given suite).  It’s doing this to decide which step definition to use.

Sam-Burns:
@stof Regarding profiling data, I can’t run it for the full suite, because Blackfire times out after about 15 minutes for CLI jobs.  (Seriously, this test suite… *sigh*)  The attached data came from running Behat over a single tag with ~300 steps executing.  Inclusive CPU time drops from about a minute and a half to about 6 seconds.  The method was executed ~180k times in this subset of our tests.

screen shot 2015-08-21 at 12 32 53

PR #792, opened 2015-10-09
Use proper scope names in ScenarioHook

Taluu:
The scopes names were using `feature.before` and `feature.after` in `ScenarioHook`, while it should be `scenario.before` and `scenario.after`.

Can't see which tests I should update, I'll take a look (it should be within `hooks.feature` I guess ?)

### DIFF
--- a/src/Behat/Behat/Hook/Scope/ScenarioScope.php
+++ b/src/Behat/Behat/Hook/Scope/ScenarioScope.php
@@ -21,8 +21,8 @@ use Behat\Testwork\Hook\Scope\HookScope;
  */
 interface ScenarioScope extends HookScope
 {
-    const BEFORE = 'feature.before';
-    const AFTER = 'feature.after';
+    const BEFORE = 'scenario.before';
+    const AFTER = 'scenario.after';
 
     /**
      * Returns scope feature.


ciaranmcnulty:
It's kinda worrying that this change hasn't caused any difference in the test suite

Taluu:
pretty much agree...

ciaranmcnulty:
Can you add a test that will fail if your change is reverted?

Taluu:
I am using scenario scope hooks without this fix and it seems to be working. I'll try to add a test that fails without this fix, but I'm not quite sure how. I'll need some time to check it up.

everzet:
@Taluu can you add a test so I can merge it?

Taluu:
Will do. Any hint on how shuold I test that (and make sure a failure is triggered without this fix, otherwise it could be quite pointless) ? Because there seems to be a bunch of scenarios with before / after scenario hooks and it didn't fail before...

everzet:
## TLDR;

There are already tests that cover this case. The reason why those weren't failing is because of defensive/type-guiard programming on my side.

## Longer version

Those constants are used during creation of hooks and scopes. They are then used by the [`HookRepository`](https://github.com/Taluu/Behat/blob/fix-scenario-scope/src/Behat/Testwork/Hook/HookRepository.php#L69) to fetch callbacks appropriate for a just fired hook. Now technically, there is a bug and in case of `ScenarioHook` this guard will be ignored for firing both `Feature` and `Scenario` hooks. So why doesn't it fail? Keep reading.

If hook and scope name check out (even wrongly), we pass forward to [filtering guard](https://github.com/Taluu/Behat/blob/fix-scenario-scope/src/Behat/Testwork/Hook/HookRepository.php#L73). Filtering guard checks if the kind of hook (all hooks in case of Behat) is the filterable hook and then asks hook to filter itself based on given `Scope`. If you drill down to [RuntimeScenarioHook](https://github.com/Taluu/Behat/blob/fix-scenario-scope/src/Behat/Behat/Hook/Call/RuntimeScenarioHook.php#L33-L35), you'll see that the first thing this guy does is type-checks the scope you pass.

Basically it means that hooks have two guards in play - name-based and type-based ones. This is redundancy and an obvious design flaw. I shouldn't have checked type inside `RuntimeScenarioHook` and should've followed tell-don't-ask querying pattern. This way already existing tests would've highlighted the name-guard bug.

## This PR and implications

With two guards in place and commitment to BC, there's not much we can do. I'll still merge this PR as it should potentially provide very minor performance optimisation by avoiding unnecessary type checks.

Taluu:
Thanks for the explanation !

everzet:
@Taluu thanks for spotting!

PR #798, opened 2015-10-19
added .DS_Store to .gitignore for mac users

achange-in-grovo:


Taluu:
-1.

This should be in your global `.gitignore` configuration. Otherwise we'd need such a directive for all and all project managers (mac file systems, eclipse, vim, ....)

achange-in-grovo:
can we add that to the readme? with a link to here to help people set up the global gitignore? https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

stof:
@achange-in-grovo the fact that ignore rules specific to your dev environment should be ignored at the environment level rather than at the project level does not really belong to the readme either (otherwise most projects should be setting this in the readme)

achange-in-grovo:
good point. but beginners need love too :)

ciaranmcnulty:
This is not appropriate for the repository to enforce, otherwise every project will have a huge `.gitignore`.

PR #801, opened 2015-11-04
Allow behat to work with Symfony3

benji07:
Hey, i tried to install behat with symfony3, and it failed, so I made this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,12 +23,11 @@ matrix:
     - php: 5.3.3
       env: DEPENDENCIES='low'
     - php: 5.6
-      env: DEPENDENCIES='dev' SYMFONY_VERSION=2.8.*@dev
+      env: DEPENDENCIES='dev'
     - php: 5.6
-      env: DEPENDENCIES='dev' SYMFONY_VERSION="3.0.x-dev as 2.8"
+      env: DEPENDENCIES='dev' SYMFONY_VERSION="3.0.*@dev"
   allow_failures:
-    - env: DEPENDENCIES='dev' SYMFONY_VERSION=2.8.*@dev
-    - env: DEPENDENCIES='dev' SYMFONY_VERSION="3.0.x-dev as 2.8"
+    - env: DEPENDENCIES='dev' SYMFONY_VERSION="3.0.*@dev"
 
 before_script:
   - if [ "$DEPENDENCIES" = "dev" ]; then perl -pi -e 's/^}$/,"minimum-stability":"dev"}/' composer.json; fi;
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,10 +22,17 @@ matrix:
       env: SYMFONY_VERSION='2.5.*'
     - php: 5.3.3
       env: DEPENDENCIES='low'
+    - php: 5.6
+      env: SYMFONY_VERSION=2.8.*@dev
+    - php: 5.6
+      env: SYMFONY_VERSION="3.0.x-dev as 2.8"
+  allow_failures:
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -46,6 +46,9 @@
         }
     },
 
+    "minimum-stability": "dev",
+    "prefer-stable": true,
```
stof: -1 for this. This means there is nothing ensuring that Behat can be installed in ``minimum-stability: stable``.

You should change the minimum-stability in Travis jobs needing it instead (see how I did it in Assetic)

.travis.yml
```diff
@@ -22,10 +22,17 @@ matrix:
       env: SYMFONY_VERSION='2.5.*'
     - php: 5.3.3
       env: DEPENDENCIES='low'
+    - php: 5.6
+      env: SYMFONY_VERSION=2.8.*@dev
+    - php: 5.6
+      env: SYMFONY_VERSION="3.0.x-dev as 2.8"
```
stof: do we actually need this aliasing as 2.8 ?

composer.json
benji07: @stof this line ?
```
- if [ "$DEPENDENCIES" = "dev" ]; then perl -pi -e 's/^}$/,"minimum-stability":"dev"}/' composer.json; fi;
```

ciaranmcnulty:
I agree that setting minimum-stability is inappropriate.

The change to "~2.1|~3.0" is probably right, assuming 3.0 works?

.travis.yml
benji07: i borrow that in https://github.com/sonata-project/SonataIntlBundle/blob/master/.travis.yml#L43

benji07:
@ciaranmcnulty everything is broken :), i can't install "behat/mink", "behat/mink-selenium2-driver", "behat/symfony2-extension", and "behat/mink-browserkit-driver" but i will do PR for those too

benji07:
@ciaranmcnulty my test suite run properly with symfony3

composer.json
```diff
@@ -18,13 +18,13 @@
...
-        "symfony/translation":           "~2.3",
-        "symfony/yaml":                  "~2.1",
-        "symfony/class-loader":          "~2.1"
+        "symfony/console":               "~2.1|~3.0",
+        "symfony/config":                "~2.3|~3.0",
+        "symfony/dependency-injection":  "~2.1|~3.0",
+        "symfony/event-dispatcher":      "~2.1|~3.0",
+        "symfony/translation":           "~2.3|~3.0",
+        "symfony/yaml":                  "~2.1|~3.0",
+        "symfony/class-loader":          "~2.1|~3.0"
```
stof: you should also update the dev dependency

.travis.yml
```diff
@@ -22,10 +22,17 @@ matrix:
       env: SYMFONY_VERSION='2.5.*'
     - php: 5.3.3
       env: DEPENDENCIES='low'
+    - php: 5.6
+      env: DEPENDENCIES='dev'
+    - php: 5.6
+      env: DEPENDENCIES='dev' SYMFONY_VERSION="3.0.*@dev"
```
stof: ``DEPENDENCIES='dev'`` already installs the 3.0 versions of Symfony as this is the highest compatible version (except for symfony/process because you forgot to change the requirement). So it is useless to add a second job

benji07:
@stof i update the travis file & the composer file

stof:
thanks

Soullivaneuh:
@stof Could we have a new patch tag for this? Thanks.

PR #803, opened 2015-11-18
Added romanian translation

Chriton:


Chriton:
Hello,
Is there something else that I should do for this to be merged?

Issue #812, opened 2015-12-27
Could not install on php7


rennis23:
Installation on php7 is not possible.
Required extension "ext-mbstring" not exists.

***composer output***
```
Your requirements could not be resolved to an installable set of packages.
  Problem 1
    - behat/behat dev-master requires ext-mbstring * -> the requested PHP extension mbstring is missing from your system.
    - behat/behat dev-master requires ext-mbstring * -> the requested PHP extension mbstring is missing from your system.
    - Installation request for behat/behat dev-master -> satisfiable by behat/behat[dev-master]
```

rennis23:
Pull request with fix https://github.com/Behat/Behat/pull/813

stof:
Well, mbstring exists for PHP 7 too. It looks like you installed PHP 7 without the extension.

rennis23:
You have right.
This is problem with my docker configuration.
If anybody used official [php7](https://hub.docker.com/_/php/) image from docker hub you must install mbstring by command.
```
docker-php-ext-install mbstring
```

PR #813, opened 2015-12-27
Fix installation on php 7

rennis23:
https://github.com/Behat/Behat/issues/812

Symfony polyfill component resolve problem with installation and dependency.
https://github.com/symfony/polyfill

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: php
 
-php: [5.3, 5.4, 5.5, 5.6, 5.3.3, hhvm]
+php: [5.3, 5.4, 5.5, 5.6, 7.0, 5.3.3, hhvm]
 
 sudo: false
 
@@ -23,6 +23,8 @@ matrix:
     - php: 5.3.3
       env: DEPENDENCIES='low'
     - php: 5.6
+      env: DEPENDENCIES='dev'
+    - php: 7.0
       env: DEPENDENCIES='dev'
 
 before_script:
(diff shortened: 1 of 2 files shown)


jakzal:
:+1:

composer.json
```diff
@@ -24,7 +23,8 @@
         "symfony/event-dispatcher":      "~2.1|~3.0",
         "symfony/translation":           "~2.3|~3.0",
         "symfony/yaml":                  "~2.1|~3.0",
-        "symfony/class-loader":          "~2.1|~3.0"
+        "symfony/class-loader":          "~2.1|~3.0",
+        "symfony/polyfill":              "^1.0"
```
stof: please don't depend on the whole polyfill byt only the deps we need

.travis.yml
```diff
@@ -24,6 +24,8 @@ matrix:
       env: DEPENDENCIES='low'
     - php: 5.6
       env: DEPENDENCIES='dev'
+    - php: 7.0
+      env: DEPENDENCIES='dev'
```
stof: no need to test dev dependencies on both 5.6 and 7.0 IMO. Please remove this extra job

rennis23:
This pull request is not necessary.
I closed [issue](https://github.com/Behat/Behat/issues/812), because I had wrong docker configuration (described in issue).

ciaranmcnulty:
The symfony polyfill may still be a good idea!

stof:
@ciaranmcnulty the right solution might be to make the Symfony polyfill provide ``ext-mbstring``, so that it can fulfill such requirement, instead of forcing all Behat users to install it. This would make it opt-in.

PR #816, opened 2016-01-03
Handle environment precedence variable

nervo:
Before adding some tests, i'd like to have your opinion on this implementation

### DIFF
--- a/src/Behat/Behat/ApplicationFactory.php
+++ b/src/Behat/Behat/ApplicationFactory.php
@@ -106,6 +106,14 @@ final class ApplicationFactory extends BaseFactory
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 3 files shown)


sroze:
What's your use case? I would argue that the environment variables shouldn't take precedence, as that would potentially cause a lot of troubles with potential differences between setups. If you want to have installation-only settings you can still use the `behat.yml.dist` (versioned) and the `behat.yml` file locally.

nervo:
You can have a look on the original pr and all its comments right there: https://github.com/Behat/Behat/pull/725

everzet:
This complicates handling of environment variables way too much. I'd like to keep already complicated logic as it is without overcomplicating, so I'm closing this.

Issue #768, opened 2015-07-07
Empty Background results in Skip of feature


jayS-de:
I'm using behat 3.0.15 and had by accident an empty background definition with the result that all scenarios of the feature have been skipped. As it can be fixed by deleting the empty background this is only a minor issue.

```
Feature: Awesome stuff
    Background:
    Scenario: do awesome stuff
        Given something
        Then magic happens
```

everzet:
Thank you for reporting it. Should be fixed in 3.1!

PR #817, opened 2016-01-04
Fix for empty background results in skip of feature

aaa2000:
Fix the bug #768


everzet:
Nice! Thank you very much for contribution!

Issue #821, opened 2016-01-11
Drop support for obsolete PHP versions


Sam-Burns:
I suggest dropping support for PHP 5.3 and 5.4.  Also, maybe update the Github and Packagist repository descriptions.

Also - reasonable people may differ on this point - but I think you should drop support for PHP 5.5 as well, as it is not subject to routine bugfixing anymore.  If you don't like that idea, maybe leave it till July when security support stops.

Sam-Burns:
Ref https://github.com/Behat/Behat/pull/822

stof:
Well, given that people are using Behat to test their libraries, dropping support for PHP means forbidding them to test their package on these versions too.

So I would not drop support just for the sake of dropping support. What is the actual benefit you will see here ?

Taluu:
This is quite different here. Waht is proposed here is just to stop supporting these versions (we've got to move on eventually, otherwise, it's gonna be php4 all over again...), as in "It may work with php 5.3, it may not work".

What are the actual benefits ? for starters, the features brought by php 5.4+ (such as the traits, the major speed improvments (huuum php7...), splat operators, ...) which could be  (I am saying *could be*) useful.

Also, it may also be used as a kind of an example of libraries promotong newer versions of php, and maybe encouraging to upgrade there. But this may require a version bump though (that's what Symfony did with Symfony 2, promoting php 5.3 instead of php 5.2 as it was designed for at the beginning).

stof:
speed improvements is already there, as you can already run PHP 7 in your projects. The fact that the min version is PHP 5.3 does not mean you cannot use PHP 7. And this will not be a benefit for Behat itself (it does not change the maintenance load to have it faster when using PHP 7).
Changing the composer requirement means that people running a ``composer update`` on 5.3 will get old releases instead of the latest. So this is not a "it may work". It is a "we don't let you try"

stof:
And regarding traits, I cannot thing of places in the Behat codebase which would use them, even if the min requirement was 5.4 (I don't use traits much to be honest).

Taluu:
Maybe, but still,, some features that could be used can't be actually used (Traits and all). I do not have any examples in mind in Behat source code, but I think this argument is as valid as any other. IMHO, Behat 3 should have at least been developped with 5.4 (or 5.5, as it was the stable at the time).

[...]

PR #822, opened 2016-01-11
Dropping support for PHP <5.5

Sam-Burns:
Ref https://github.com/Behat/Behat/issues/821

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: php
 
-php: [5.3, 5.4, 5.5, 5.6, 7.0, hhvm]
+php: [5.5, 5.6, 7.0, hhvm]
 
 sudo: false
 
@@ -16,7 +16,7 @@ branches:
 
 matrix:
   include:
-    - php: 5.3.3
+    - php: 5.6
       env: DEPENDENCIES='low'
     - php: 5.6
       env: SYMFONY_VERSION='2.3.*'
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     ],
 
     "require": {
-        "php":                           ">=5.3.3",
+        "php":                           ">=5.5.0",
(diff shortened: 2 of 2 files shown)


ciaranmcnulty:
I think this should wait for a major version

Sam-Burns:
@ciaranmcnulty Don't object to you bumping the major version number instead of the minor version, but I don't think that's really necessary, and I know that isn't how Sensiolabs like to do things.

To be responsible, support for PHP 5.3 should have been dropped in the summer of '14, so I wouldn't delay any longer.  Also, regarding advertising the tool on Packagist as 'Scenario-oriented BDD framework for PHP 5.3', I think more could be done to make it sound contemporary and relevant.

ciaranmcnulty:
I don't think SemVer requires a major version bump so it's up to @everzet

Taluu:
> I don't think SemVer requires a major version bump so it's up to @everzet

It actually doesn't :

> What should I do if I update my own dependencies without changing the public API?
>
> That would be considered compatible since it does not affect the public API. Software that explicitly depends on the same dependencies as your package should have their own dependency specifications and the author will notice any conflicts. Determining whether the change is a patch level or minor level modification depends on whether you updated your dependencies in order to fix a bug or introduce new functionality. I would usually expect additional code for the latter instance, in which case it's obviously a minor level increment.

But if the public api changes, it does need to bump a minor / major (Depending on how the break is made).

And :+1:

Sam-Burns:
Thanks @Taluu. Together, we can end PHP 5 in our lifetime ;-)

stof:
See my comment on the issue for the actual implication of this change

everzet:
Provided an argument in the related issue. Closing this PR as it is non-merge able anyways ATM. If the discussion in related issue will arrive at conclusion of progressing with this PR - feel free to rebase and reopen it.

PR #823, opened 2016-01-12
Allow the recursive directory iterator to iterate symbolic links

sherakama:
Directory iterator is not iterating through symbolic links as it did in behat 1.x and most of behat 2.x.
The new RecursiveDirectoryIterator needs to be explicitly told to follow symlinks or it will not.

### DIFF
--- a/src/Behat/Behat/Gherkin/Specification/Locator/FilesystemFeatureLocator.php
+++ b/src/Behat/Behat/Gherkin/Specification/Locator/FilesystemFeatureLocator.php
@@ -131,13 +131,19 @@ final class FilesystemFeatureLocator implements SpecificationLocator
             return array($absolutePath);
         }
 
-        $iterator = new RegexIterator(
-            new RecursiveIteratorIterator(
-                new RecursiveDirectoryIterator($absolutePath)
-            ), '/^.+\.feature$/i',
+        $RDI = new RecursiveDirectoryIterator(
+            $absolutePath,
+            RecursiveDirectoryIterator::FOLLOW_SYMLINKS |
+            RecursiveDirectoryIterator::SKIP_DOTS
+        );
+        $RII = new recursiveIteratorIterator($RDI);
+        $REX = new RegexIterator(
+            $RII,
+            '/^.+\.feature$/i',
             RegexIterator::MATCH
         );
-        $paths = array_map('strval', iterator_to_array($iterator));
+
+        $paths = array_map('strval', iterator_to_array($REX));
         uasort($paths, 'strnatcasecmp');
 
         return $paths;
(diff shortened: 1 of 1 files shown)


src/Behat/Behat/Gherkin/Specification/Locator/FilesystemFeatureLocator.php
```diff
@@ -131,13 +131,19 @@ private function findFeatureFiles($path)
             return array($absolutePath);
         }
 
-        $iterator = new RegexIterator(
-            new RecursiveIteratorIterator(
-                new RecursiveDirectoryIterator($absolutePath)
-            ), '/^.+\.feature$/i',
+        $RDI = new RecursiveDirectoryIterator(
```
sroze: I'm not fond of these short variable names.

src/Behat/Behat/Gherkin/Specification/Locator/FilesystemFeatureLocator.php
```diff
@@ -131,13 +131,19 @@ private function findFeatureFiles($path)
...
+            $absolutePath,
+            RecursiveDirectoryIterator::FOLLOW_SYMLINKS |
+            RecursiveDirectoryIterator::SKIP_DOTS
+        );
+        $RII = new recursiveIteratorIterator($RDI);
```
sroze: You could simply construct them without any other variable like it was before btw.

sherakama:
@sroze,

Thanks for reviewing this. I have restored the original formatting. It was easier to debug when I had them separated out in to separate variables. I agree, we don't need them.

src/Behat/Behat/Gherkin/Specification/Locator/FilesystemFeatureLocator.php
```diff
@@ -132,11 +132,17 @@ private function findFeatureFiles($path)
         }
 
         $iterator = new RegexIterator(
-            new RecursiveIteratorIterator(
-                new RecursiveDirectoryIterator($absolutePath)
-            ), '/^.+\.feature$/i',
+            new recursiveIteratorIterator(
```
sroze: You should use an uppercase `R` else that will fail on some systems.

src/Behat/Behat/Gherkin/Specification/Locator/FilesystemFeatureLocator.php
```diff
@@ -132,11 +132,17 @@ private function findFeatureFiles($path)
...
+                new RecursiveDirectoryIterator(
+                    $absolutePath,
+                    RecursiveDirectoryIterator::FOLLOW_SYMLINKS |
+                    RecursiveDirectoryIterator::SKIP_DOTS
```
stof: this should be on the previous line to be more readable. It is the same argument, not a different argument

sroze:
@everzet this PR sounds good and should be merged before the 3.1 release :+1:

sherakama:
Thanks!

PR #833, opened 2016-01-26
Catch PHP7 errors implementing Throwable

### LINKED ISSUE #831: Catch  `\Throwable` in PHP 7 so that fatal errors do not halt test execution (how?)
I would like to contribute a patch for this, but I have a few questions before I can do that.

Now PHP 7 is out, fatal errors no longer need to halt test execution. It looks like appropriate place in  `makeCall` method in  in the [`CallCenter`](https://github.com/Behat/Behat/blob/718159921b1d844b8805b5ab2e7d1be34ca323cb/src/Behat/Testwork/Call/CallCenter.php) class

```php
public function makeCall(Call $call)
{
    try {
        $filteredCall = $this->filterCall($call);
        $result = $this->handleCall($filteredCall);
        $filteredResult = $this->filterResult($result);
    } catch (Exception $e) {
        return new CallResult($call, null, $e, null);
    } catch (\Throwable $e) {
        // Cast \Throwable to \Exception?
        $e = throwableToException($e);
        return new CallResult($call, null, $e, null);
    }

    return $filteredResult;
}
```

I have something along these lines working however I am not sure how to handle the cast of `Throwable` to `Exception` in the context of Behat.  Symfony handles it via [ `FatalThrowableException`](https://github.com/symfony/symfony/blob/a729462ce1405497b9bec285d2e83f6e60b987d5/src/Symfony/Component/Debug/Exception/FatalThrowableError.php) but to be frank I don't fully understand the constructor of that class, why is `ErrorException` instantiated but not assigned for example?

I would appreciate any pointers to getting a PR open (or a PR would be great too)

peterjmit:
Resolves #831 

Questions:

- ~~How to correctly handle the version differences? JUnit formatter/feature might be problematic~~ 
- `FatalThrowableError` has been re-appropriated from Symfony, do I need to include licensing?

### DIFF
--- a/features/junit_format.feature
+++ b/features/junit_format.feature
@@ -538,6 +538,7 @@
       """
     And the file "junit/default.xml" should be a valid document according to "junit.xsd"
 
+  @php-version @php5.3 @php5.4
   Scenario: Aborting due to PHP error
     Given a file named "features/bootstrap/FeatureContext.php" with:
       """
--- a/src/Behat/Testwork/Call/Exception/FatalThrowableError.php
+++ b/src/Behat/Testwork/Call/Exception/FatalThrowableError.php
@@ -1,0 +1,55 @@
+<?php
+
+/*
+ * This file is part of the Behat Testwork.
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Behat\Testwork\Call\Exception;
+
+use ErrorException;
+use ParseError;
+use Throwable;
+use TypeError;
+
+/**
+ * Fatal Throwable Error.
+ *
+ * @author Nicolas Grekas <p@tchwork.com>
+ */
+class FatalThrowableError extends ErrorException
+{
+    public function __construct(Throwable $e)
+    {
+        if ($e instanceof ParseError) {
+            $message = 'Parse error: '.$e->getMessage();
+            $severity = E_PARSE;
+        } elseif ($e instanceof TypeError) {
+            $message = 'Type error: '.$e->getMessage();
+            $severity = E_RECOVERABLE_ERROR;
+        } else {
+            $message = 'Fatal error: '.$e->getMessage();
+            $severity = E_ERROR;
+        }
+
+        parent::__construct(
+            $message,
+            $e->getCode(),
+            $severity,
+            $e->getFile(),
+            $e->getLine()
+        );
+
+        $this->setTrace($e->getTrace());
+    }
+
+    private function setTrace($trace)
+    {
+        $traceReflector = new \ReflectionProperty('Exception', 'trace');
+        $traceReflector->setAccessible(true);
+        $traceReflector->setValue($this, $trace);
+    }
+}
--- a/features/catch_throwable.feature
+++ b/features/catch_throwable.feature
@@ -1,0 +1,63 @@
+@php-version @php7
+Feature: Support PHP 7 Throwable
+  In order for my test suite to continue running despite fatal errors in my code
+  As a feature developer
+  I need Behat to gracefully handle errors implementing the Throwable interface
+
+  Background:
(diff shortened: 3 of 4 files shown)


src/Behat/Testwork/Exception/FatalThrowableError.php
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+namespace Behat\Testwork\Exception;
+
+use ErrorException;
+use ParseError;
+use Throwable;
+use TypeError;
+
+class FatalThrowableError extends ErrorException
```
stof: Please keep the ``@author`` annotation with the original author (i.e. @nicolas-grekas IIRC)

src/Behat/Testwork/Exception/FatalThrowableError.php
```diff
@@ -0,0 +1,50 @@
...
+
+/*
+ * This file is part of the Behat Testwork.
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Behat\Testwork\Exception;
```
stof: This exception should be in ``Behat\Testwork\Call\Exception``, as it is only used by the Call subsystem

features/junit_format.feature
```diff
@@ -538,6 +538,7 @@
       """
     And the file "junit/default.xml" should be a valid document according to "junit.xsd"
 
+  @php-version @php5.3 @php5.4 @php5.5 @php5.6
```
stof: PHP 5.6 already runs the scenario tagged as ``@php5.4``, due to the way our Travis config works. It is meant to be a min version (and the code used in the Travis setup is currently broken for PHP 7+ when generating the list btw)

features/junit_format.feature
```diff
@@ -538,6 +538,7 @@
       """
     And the file "junit/default.xml" should be a valid document according to "junit.xsd"
 
+  @php-version @php5.3 @php5.4
```
peterjmit: @stof not sure if this is what you meant, perhaps just `php5.3` if it works as a "minimum version"

stof: no need to put anything there, as 5.3 is the min version supported by Behat instead. And using it to exclude PHP 7 is wrong though. It has this behavior only because the code used on Travis actually treats PHP 7.0 as a 5.0 release (so excluding any test with a min version tag)

peterjmit: @stof this test fails on php 7 locally (imo is expected to fail), so do I remove the tags or not?

src/Behat/Testwork/Call/Exception/FatalThrowableError.php
```diff
@@ -0,0 +1,55 @@
...
+
+        parent::__construct(
+            $message,
+            $e->getCode(),
+            $severity,
+            $e->getFile(),
+            $e->getLine()
+        );
+
+        $this->setTrace($e->getTrace());
```
Taluu: why do you set the trace here like this, when just using the `Throwable` exception as a previous exception (last parameter of every Exception object or almost, including `ErrorException`) should do the trick ?

Taluu: In @nicolas-grekas's implementation, it had sense because he is making his own stack trace, while here you just forcefully set the trace to be the one of the `Throwable`

stof: @Taluu The goal is to make the FatalThrowableError appear as being the original error instead of seeing a FatalThrowableError traced in CallCenter and then having to investigate linked exceptions

peterjmit:
@stof @everzet  Scrutinizer timed out and I don't have access to restart, can one of you do the honors?

kamazee:
@peterjmit you can simply amend a commit with nothing (just to change its hash) and push (with force), it should restart builds.

peterjmit:
This should be ready to merge unless there is more feedback/changes required for the PHP7 tag on the `junit` feature

PR #838, opened 2016-02-09
"imports: " should consoider .dist files as well

DrSchimke:
When using `imports: [foo.yml, bar.yml]` in my behat.yml, it would be cool, if there was a fallback to `foo.yml.dist` and `bar.yml.dist`, if the non-dist files don't exists.

So I could check in the dists into VCS and locally change only the non-versioned ymls.

(_And en passant changing the code `ConfigurationLoader::parseImport()` to avoid this a bit confusing `file_exists && !file_exists` stuff_ :)


everzet:
Thank you very much for sending pr. Sadly I cannot accept this one.

I think this is taking conventions too far. There is a fine line by crossing which conventions become both too powerful and too confusing. For the use case you provided I definitely believe explicit configuration based on `--config` switch is better.

PR #843, opened 2016-02-18
Use the new Composer syntax for OR

estheban:
https://getcomposer.org/doc/articles/versions.md#range


stof:
Technically, ``|`` also works (it is the older syntax). But I agree about switching to the newest syntax as it is the one documented by Composer.

however, please change your commit message, as it does not describe at all what the commit is doing.

PR #852, opened 2016-02-29
Fix #848: If output buffer is active then close it

### LINKED ISSUE #848: Behat 3 consuming lot of memory for results.
While transition from Behat 2.5.5 to behat 3.1.0-RC2 it was realised that memory used to save results is huge compared to what was in Behat 2.5.5

Following are the observations with https://github.com/rajeshtaneja/MinkExtension-example.git
- 12 steps : 17.78Mb
- 72 steps: 18.73 Mb (Multiple features)
- 752 Steps : 27.94Mb (Single big feature)

We have around 45000 steps and with above results it will use around 550Mb. which is quite significant.

cd MinkExtension-example
curl http://getcomposer.org/installer | php
php composer.phar install
#### Run Behat for 4 steps
git checkout smallfeature
bin/behat
##### Run behat with steps
git checkout bigfeature
bin/behat
#### Run behat with multiple features
git checkout multiplefeature
bin/behat

rajeshtaneja:
Close output buffer if opened successfully.

### DIFF
--- a/src/Behat/Testwork/Call/Handler/RuntimeCallHandler.php
+++ b/src/Behat/Testwork/Call/Handler/RuntimeCallHandler.php
@@ -27,6 +27,12 @@ final class RuntimeCallHandler implements CallHandler
      */
     private $errorReportingLevel;
 
+
+    /**
+     * @var bool
+     */
+    private $obStarted = false;
+
     /**
      * Initializes executor.
      *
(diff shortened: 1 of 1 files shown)


everzet:
@rajeshtaneja thank you for sticking around and finding a way to fix this one.

rajeshtaneja:
Thanks @everzet

PR #855, opened 2016-03-07
Added JUnitSetupPrinter

jakob-erdmann:
I noticed that errors from @BeforeStep and @AfterStep hooks are not listed in the JUnit export. This causes Jenkins to not recognize the test as failed and therefore it is not listed under failed Tests.

This change adds a SetupPrinter to JUnit export which will add all Step hook failures to the export and marks them as either ``setup`` or ``teardown`` type.


everzet:

PR #860, opened 2016-03-17
Fix #859: Include basepath in generateKey

### LINKED ISSUE #859: Rerun cache file is overriden if if more then 2 runs are executed
To test multiple branches, I created multiple setups and ran behat in parallel. While doing so, it was realised that rerun file was overridden by the last failing run.  
Reason being *RerunController::generatekey()* only consider profile, suite, role, paths which are common in both the runs.

Include base path can avoid this.

rajeshtaneja:
Executing multiple runs at one time over write
rerun file of previous run. Avoid doing so by
including basepath of each run to generate uniquekey


everzet:

PR #864, opened 2016-03-19
Use only one autoloader if possible

boekkooi:
Hi there,

So I finally had some time in my project to go on the hunt for some strange behaviour with 3.1.0RC2.
Because the project that I'm working on has a less then normal structure and behat is installed as a symlink, bin/behat managed to load 2 different and conflicting autoloaders.

Since #830 requires both autoloaders I added a simple check to see if the first/current working directory autoloader adds behat, if so when there is no need to load any extra autoloaders.

Let me know what you think.

Cheers,
Warnar

### DIFF
--- a/bin/behat
+++ b/bin/behat
@@ -15,17 +15,19 @@ if (is_file($autoload = getcwd() . '/vendor/autoload.php')) {
     require $autoload;
 }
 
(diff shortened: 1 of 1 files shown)


everzet:

boekkooi:
No problem thanks for the merge and your great work!

PR #870, opened 2016-03-30
[git] Add .gitattributes to remove unneeded files

Slamdunk:


ciaranmcnulty:
If this is to remove unnecessary tests for prod, it should probably also exclude the various CI tool configs, and probably the `features` folder

Slamdunk:
This should be all, isn't it?

everzet:

PR #871, opened 2016-04-04
Disable xdebug in travis tests

shrikeh:


ciaranmcnulty:
This appears to be much faster on the PHP engines.

However `phpenv` is killing the HHVM build, which is actually 12X slower most of the time anyhow :-/

pamil:
@ciaranmcnulty changing it to `phpenv config-rm xdebug.ini || true` will keep the HHVM build up and running :tada:

.travis.yml
```diff
@@ -28,6 +31,12 @@ matrix:
       env: SYMFONY_VERSION='3.0.*'
     - php: 7.0
       env: DEPENDENCIES='dev'
+    - php: hhvm
+      env: XDEBUG_DISABLE=''
+
+before_install:
+  # Remove xdebug as per https://getcomposer.org/doc/articles/troubleshooting.md#xdebug-impact-on-composer
+  - eval "$XDEBUG_DISABLE"
```
pamil: IMO `phpenv config-rm xdebug.ini || true` will be more readable than all the stuff happening with environmental variables

.travis.yml
```diff
@@ -1,3 +1,4 @@
+---
```
stof: why ?

.travis.yml
```diff
@@ -41,3 +55,4 @@ before_script:
 script:
   - phpunit
   - behat -fprogress --strict --tags '~@php-version,'`php php_version_tags.php`
+...
```
stof: this looks wrong

.travis.yml
```diff
@@ -18,16 +19,29 @@ matrix:
...
+        - phpenv config-rm xdebug.ini
     - php: 7.0
       env: SYMFONY_VERSION='3.0.*'
+      before_install:
+        - phpenv config-rm xdebug.ini
     - php: 7.0
       env: DEPENDENCIES='dev'
+      before_install:
+        - phpenv config-rm xdebug.ini
+    - php: hhvm
```
stof: this is duplicating the HHVM build

.travis.yml
```diff
@@ -18,16 +19,29 @@ matrix:
   include:
     - php: 5.3
       env: DEPENDENCIES='low'
+      before_install:
+        - phpenv config-rm xdebug.ini
```
stof: this will not be added for normal PHP builds, so this looks weird

stof:
I suggest doing this simple change:

```yaml
before_install:
    # Remove xdebug as per https://getcomposer.org/doc/articles/troubleshooting.md#xdebug-impact-on-composer
    - if ["$TRAVIS_PHP_VERSION" = 'hhvm']; then phpenv config-rm xdebug.ini; fi
```

.travis.yml
shrikeh: Tsk @stof, read your specs ;-)

.travis.yml
```diff
@@ -28,6 +29,11 @@ matrix:
       env: SYMFONY_VERSION='3.0.*'
     - php: 7.0
       env: DEPENDENCIES='dev'
+  fast_finish: true
```
pamil: It's useless as there are no jobs that are allowed to fail.

.travis.yml
shrikeh: http://www.yaml.org/spec/1.2/spec.html#id2800401

pamil: This file does not use directives, so we can get rid of this weird syntax.

.travis.yml
stof: This file does not use directives, so the document markers are useless here. So I'd rather avoid them. Thus, it makes the file less readable as many people don't know about them (and the Travis doc does not use them).

This is why I'm asking you why you use them. I know that they exist, but I don't see any benefit here

shrikeh: You should always start and end your document with markers, whether you use directives or not. It's a document, so it should have document markers.

.travis.yml
shrikeh: A question, on a side note: 5.3/5.4 are EOL, should they not be under an `allowed_failures` section?

.travis.yml
shrikeh: Please re-read the example in section 2.2 of the YAML spec.

.travis.yml
stof: Please read section ``9.1.3`` about bare documents (just after your section about document markers). this is what is used in this file.

shrikeh: > 9.1.4. Explicit Documents

> An explicit document begins with an explicit directives end marker line but no directives. Since the existence of the document is indicated by this marker, the document itself may be completely empty.

stof: And a bare document does not use the markers, which is totally valid. and in this case, using a bare document is more readable than using an explicit document

Issue #832, opened 2016-01-21
SIGINT handling doesn't appear to work with PHP 7.0


davedevelopment:
I think. I can definitely ctrl-c out of a simple suite with php 5.6.0, but the same suite under 7.0.(0,2,snapshot) just carries on running. I've done a bit of investigating, but I'm still not any closer to knowing what's going on :)

davedevelopment:
I should add that I confirmed the signal handler in `SigintController` is successfully registered, it just doesn't seem to fire.

everzet:
I did notice that too, but thought it was to do with the lack of readline module in my php7 environment. Did something change in regards how SIGINT capture works in php7?

Taluu:
I also have this problem. But, in phpunit, it works perfectly, so maybe there could be a hint in it to correctly handle this ?

I tried to detect the function used (as this is a simple if on a `function_exists`), and it does return true...

davedevelopment:
I do have readline compiled in. I only glanced through the changelog for 7.0.0 and couldn't see anything relevant. I've just had another play and managed to get a fairly minimal test case together without behat, will carry on playing with it and see where I get.

davedevelopment:
It's something to do with the positioning of `declare(ticks=1)`. Reading the php docs, I assumed moving it up to the `bin/behat` executable would at least get it working, but didn't. Moving in to the block of a step does make it work.

``` cucumber
Feature: test
    Scenario: test
        Given I wait for 5 seconds
        Given I wait for 5 seconds
        Given I wait for 5 seconds
        Given I wait for 5 seconds
        Given I wait for 5 seconds
        Given I wait for 5 seconds
```

``` php
    /**
     * @Given I wait for 5 seconds
     */
    public function iWait()
    {
        declare(ticks = 1); # this makes the magic happen, removing it forces a full run
        $start = time();
        while(time() < $start + 5) {
            $a = 0; # tickable statement, php 7 would run the loop without ticking otherwise
        }
    }

```

I don't have any more time to dig in to this right now, but I guess there must be somewhere suitable for the `declare` directive to go.

davedevelopment:
Filed a bug report, will see what they say https://bugs.php.net/bug.php?id=71448

danepowell:
What's the latest status on this? It just bit me today, pretty annoying.

It looks like @peterjmit might have a fix, but I don't see an upstream PR for it.

pies:
It's very annoying.

PR #878, opened 2016-04-14
Handle interupts on PHP7

peterjmit:
Fix for https://github.com/Behat/Behat/issues/832

It appears that the scoping of the declare statement prevents ticks from being emitted on PHP7. This PR decorates StepTester to declare ticks during step testing and therefore allow interrupts to occur mid-spec.

I initially looked at decorating a class in `Testwork` so that this code was closer to the `SigintController` but Testwork only goes as far as the suite level if I have understood things correctly.

On reading the documentation for `declare(ticks = 1)` it almost seems like the behaviour prior to PHP7 may have been a bug, and things are now behaving as expected (might explain the lack of response on the PHP bug filed?).

### DIFF
--- a/src/Behat/Behat/EventDispatcher/ServiceContainer/EventDispatcherExtension.php
+++ b/src/Behat/Behat/EventDispatcher/ServiceContainer/EventDispatcherExtension.php
@@ -40,6 +40,7 @@ class EventDispatcherExtension extends BaseExtension
         $this->loadEventDispatchingScenarioTester($container);
         $this->loadEventDispatchingExampleTester($container);
         $this->loadEventDispatchingStepTester($container);
+        $this->loadTickingStepTester($container);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Behat/Behat/EventDispatcher/ServiceContainer/EventDispatcherExtension.php
```diff
@@ -153,4 +154,23 @@ protected function loadEventDispatchingStepTester(ContainerBuilder $container)
...
+     */
+    protected function loadTickingStepTester(ContainerBuilder $container)
+    {
+        if (!function_exists('pcntl_signal')) {
+            return;
+        }
+
+        $definition = new Definition('Behat\Behat\EventDispatcher\Tester\TickingStepTester', array(
+            new Reference(TesterExtension::STEP_TESTER_ID),
+            new Reference(self::DISPATCHER_ID)
```
Taluu: the second parameters looks useless as you don't even acknowledge it in the `TickingStepTester` class ? (there is only one argument in there)

peterjmit: that was a bad copy paste 😳

LeoBenoist:
:+1:

pamil:
:+1:

pjedrzejewski:
👍

PR #882, opened 2016-05-10
[WIP] Allow to use wildcards in imports

sergeyfedotov:
This PR provides a feature that will allow to use wildcards (provided by the glob function) in the config imports.

For example:
```yaml
imports:
    - path/*/behat.yml
```

### DIFF
--- a/src/Behat/Testwork/ServiceContainer/Configuration/ConfigurationLoader.php
+++ b/src/Behat/Testwork/ServiceContainer/Configuration/ConfigurationLoader.php
@@ -221,8 +221,8 @@ final class ConfigurationLoader
     {
         $configs = array();
         foreach ($paths as $path) {
-            foreach (glob($path) as $matchedPath) {
-                foreach ($this->parseImport($basePath, $matchedPath, $profile) as $importConfig) {
+            foreach ($this->resolveImport($basePath, $path) as $resolvedPath) {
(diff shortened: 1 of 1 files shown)


stof:
This won't work, as you glob it based on the current working directory, while the path is expected to be related to the base path (which may be a different working directory).

and this cannot be accepted without a scenario covering the feature (we actually need multiple scenarios to cover the different cases btw).

and this is also broken in case ``glob`` does not find anything and returns false.

sergeyfedotov:
> This won't work, as you glob it based on the current working directory, while the path is expected to be related to the base path (which may be a different working directory).
Oh, sorry. I missed it.

> and this is also broken in case glob does not find anything and returns false.
glob returns an empty array if found nothing

stof:
> glob returns an empty array if found nothing

it cannot also return ``false`` in some cases: http://php.net/glob#refsect1-function.glob-returnvalues

sergeyfedotov:
Actually this feature introduce a BC break. In the current implementation, an exception should be thrown when the imported file doesn't exists, but glob returns an empty array in this case and it is valid. May be an alternative syntax can be used. 

For example:
```yaml
imports:
    - [ path/*/behat.yml ] # array should be processed by the glob
    - path/to/behat.yml # scalar should be processed in the current manner
```

everzet:

PR #891, opened 2016-06-20
Fixed line number

yosmanyga:


stof:
We should link to a tag rather than to the master branch to avoid such issues

stof:
thanks

PR #903, opened 2016-07-24
Multiline step definitions

### LINKED ISSUE #348: Allow step regexes to span multiple lines.
Currently, the regexes used for step definitions are not allowed to span multiple lines, which makes them very hard to work with when they are long.  It also throws CodeSniffer warnings in our context files due to the line-length checks we have in place.

Standard DocBlock parsing allows for tags to span multiple lines, and Behat should support this.  If the line following the @tag line starts with more than one space character, then it is considered to be a continuation.

(See the parseTags() function at https://github.com/phpDocumentor/ReflectionDocBlock/blob/master/src/phpDocumentor/Reflection/DocBlock.php for the phpDocumentor parsing algorithm.)

Given that this will be in the middle of a regex, we need to handle white-space in a sane and understandable way.  I propose that all leading/trailing white-space is stripped from the lines, so that there is no unexpected behaviour.  In situations where there is white-space at the point of the line-break, we allow the regex to be closed/opened so that this can be explicitly included.

This means that either of the following would be equivalent:

```
/**
 * @When /^I do something that has quite a long description and sp
 *         ans multiple lines such as "([^"]*)"$/
 */
```
```
/**
 * @When /^I do something that has quite a long description and /
 *       /spans multiple lines such as "([^"]*)"$/
 */
```
They would both result in the following regex:
```/^I do something that has quite a long description and spans multiple lines such as "([^"]*)"$/```

However, this is just a suggestion for implementation.  You may be able to come up with a better alternative to handling this.

The important point is to allow these regular expressions to be wrapped.

### PR DESCRIPTION
**Pull request by everzet**
If you put "\" at the end of your step definition, Behat will expect next line to continue that definition:

    * @Given my stepdef \
    *        continues \
    *        for couple of lines

is equivalent to:

    * @Given my stepdef continues for couple of lines

Closes #348

### DIFF
--- a/src/Behat/Behat/Context/Reader/AnnotatedContextReader.php
+++ b/src/Behat/Behat/Context/Reader/AnnotatedContextReader.php
@@ -111,6 +111,7 @@ final class AnnotatedContextReader implements ContextReader
     {
         $callees = array();
         $description = $this->readDescription($docBlock);
+        $docBlock = $this->mergeMultilines($docBlock);
 
         foreach (explode("\n", $docBlock) as $docLine) {
             $docLine = preg_replace(self::DOCLINE_TRIMMER_REGEX, '', $docLine);
@@ -129,6 +130,18 @@ final class AnnotatedContextReader implements ContextReader
         }
 
         return $callees;
+    }
+
+    /**
+     * Merges multiline strings (strings ending with "\")
+     *
+     * @param string $docBlock
+     *
+     * @return string
+     */
+    private function mergeMultilines($docBlock)
+    {
+        return preg_replace("#\\\\$\s*\*\s*([^\\\\$]+)#m", '$1', $docBlock);
     }
 
     /**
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/Behat/Behat/Context/Reader/AnnotatedContextReader.php [hunk 1]
```diff
@@ -132,6 +133,18 @@ private function readDocBlockCallees($class, ReflectionMethod $method, $docBlock
...
     /**
+     * Merges multiline strings (strings ending with "\")
+     *
+     * @param string $docBlock
+     *
+     * @return string
+     */
+    private function mergeMultilines($docBlock)
+    {
+        return preg_replace("#\\\\$\s*\*\s*([^\\\\$]+)#m", '$1', $docBlock);
```
stof: you should replace ``\s*`` by ``\s*+`` to make the quantifiers possessive. This will avoid lots of useless backtracking.

and I think the ``[^\\\\$]+`` quantifier could also be possessive.

stof: btw, backtracking is even worse on PHP 7: some kind of backtracking can make PCRE reach the JIT stack limit much sooner than reaching the backtrack limit (set to 1000000 in PHP), so regexes working (inefficiently) on PHP 5.6 could break on PHP 7 for the same input

everzet:
Good catch. Will do this evening!

PR #905, opened 2016-07-25
Refactor transformations

everzet:
This PR extracts individual transformations from `RepositoryArgumentTransformer` into individual classes under `Behat\Behat\Transformation\Transformation` namespace. This should make adding new transformations and fixing old ones a breeze.

This PR is basis for upcoming transformation improvements in 3.2.

### DIFF
--- a/src/Behat/Behat/Transformation/RegexGenerator.php
+++ b/src/Behat/Behat/Transformation/RegexGenerator.php
@@ -1,0 +1,30 @@
+<?php
+
+/*
+ * This file is part of the Behat.
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Behat\Behat\Transformation;
+
+/**
+ * Regular expression generator.
+ *
+ * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ */
+interface RegexGenerator
+{
+    /**
+     * Generates regular expression using provided parameters.
+     *
+     * @param string $suiteName
+     * @param string $pattern
+     * @param string$language
+     *
+     * @return string
+     */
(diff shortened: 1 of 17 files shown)


src/Behat/Behat/Transformation/RegexGenerator.php
```diff
@@ -0,0 +1,30 @@
...
+ * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ */
+interface RegexGenerator
+{
+    /**
+     * Generates regular expression using provided parameters.
+     *
+     * @param string $suiteName
+     * @param string $pattern
+     * @param string$language
```
pamil: Missing space.

src/Behat/Behat/Transformation/Transformation/RuntimeTransformation.php
```diff
@@ -8,14 +8,16 @@
...
-namespace Behat\Behat\Transformation\Call;
+namespace Behat\Behat\Transformation\Transformation;
 
 use Behat\Behat\Transformation\Transformation;
 use Behat\Testwork\Call\RuntimeCallee;
 
 /**
  * Transformation that is created and executed in the runtime.
  *
+ * @deprecated Will be removed in 4.0. Use specific transformations instead
```
stof: I suggest using the ``@trigger_error('...', E_USER_DEPRECATED)`` convention used by Symfony (which reminds me that someone wrote a Behat extension to collect deprecations, which I said I would review and I forgot to do it)

stof: btw, if you rename the class, deprecating it is useless: it is already a BC break. you would need to keep it in the ``Call`` namespace

everzet:
I'll move `RuntimeTransformation` back. I don't mind with triggering deprecation errors yet, it's not as current of occurrence as in Symfony ATM.

Issue #863, opened 2016-03-19
Transform based on type of argument


ciaranmcnulty:
Currently we associate Transforms with placeholders based on name, when they're really more about types.

An example might be I have a `:sender` and `:recipient` but they're both Emails.

We should support type-based transform matching:

```php
/**
 * @Transform
 */
function toEmail($string) : Email // use @return for PHP5
{
    return Email::fromString();
}

/**
 * @Then an email should be sent from :sender to :recipient
 */
function checkEmailWasSent(Email $sender, Email $recipient)
{
    // ...
}
```

sroze:
Can't agree more on the need. Why not simply improving the existing syntax by optionally prefixing the `:` by the class name (with or without namespace)?

```php
/**
 * @Transform Email:
 */
function toEmail($string)
{
    return Email::fromString($string);
}

/**
 * @Then an email should be sent from :sender to :recipient
 */
function checkEmailWasSent(Email $sender, Email $recipient)
{
    // ...
}
```

That way, we can also mix the both, by having a transformer for a given type *and* argument name:
```php
/**
 * @Transform Email:secret
 */
function toSecretEmail($secret)
{
    return Email::fromSecret($secret);
}
```

ciaranmcnulty:
That's a nice idea Samuel. 

I have a concern that the trailing comma is not obvious enough at first glance? Maybe something like `Email:*`?

sroze:
True, that's not obvious at all. Why not simply having the colon (`:`) optional? It won't conflict with the table or regex transformers.

```php
use FQCN\Email;

/**
 * @Transform Email
 */
function toEmail($string)
{
    return Email::fromString($string);
}
```

pamil:
What do you think about `@Transform Email::class`?

stof:
@pamil what would be your syntax corresponding to the ``@Transform Email:secret`` case ?

stof:
@sroze the issue with your ``<classname>:<token>`` proposal is that it conflicts with the table transformations

sroze:
@stof when the class name is `Table`, this is true. That's an extremely rare case I would say. We can handle this specific case by throwing an exception and ask to use the FQCN of the `Table` class.

stof:
This is not the only prefix used (we also have ``row`` and ``rowtable``). And using this pattern for class names would mean that we can never use this pattern anymore in the future for other kinds of builtin transformations.

And we cannot throw an exception, as this would prevent using the existing feature.

[...]

PR #907, opened 2016-07-26
From-string object transformer

everzet:
## TLDR;

This PR introduces automatic ValueObject transformer into Behat. It is first half of #863. It has following attributes:

1. Is automatically executed for all your class type-hints in your step definitions
2. Attempts to automatically transform such arguments calling typehint's `fromString()` factory method
3. Has the lowest precedence of all transformers. Meaning that if you provide any other kind of transformation that matches said argument, from-string object transformer wouldn't get executed

### Feature

Given simple feature:

```gherkin
Feature:
  Scenario:
    Given I am "everzet"
```

### Stepdef

And a step definition, using type-hint:

```php
class User {
    public function __construct($name) { $this->name = $name; }
}

{
    /** @Given I am :user */
    public function iAm(User $user) {
        // TODO: do something with $user
    }
}
```

### Test

When you attempt to execute this, you'll get exception like this:

```bash
...> Method fromString does not exist
```

There are now two possible routes out of this situation...

#### Route 1: Write custom transformation

As before, you can write your own custom transformation and it will work as it always did:

```php
class User {
    public function __construct($name) { $this->name = $name; }
}

{
    /** @Transform :user */
    public function transformUser($user) {
        return new User($user);
    }

    /** @Given I am :user */
    public function iAm(User $user) {
        // TODO: do something with $user
    }
}
```

#### Route 2: Add `fromString`

Or, you can add `fromString` factory method to your `User` class and that will also work exactly as you might expect:

```php
class User {
    private function __construct($name) { $this->name = $name; }
    static public function fromString($name) { return new static($name); }
}

{
    /** @Given I am :user */
    public function iAm(User $user) {
        // TODO: do something with $user
    }
}
```

## Implications

There are no known BC breaks or implications on existing codebases. Please, highlight if you are aware of any.

### DIFF
--- a/src/Behat/Behat/Transformation/Call/Filter/DefinitionArgumentsTransformer.php
+++ b/src/Behat/Behat/Transformation/Call/Filter/DefinitionArgumentsTransformer.php
@@ -100,7 +100,11 @@ final class DefinitionArgumentsTransformer implements CallFilter
                 continue;
             }
 
-            return $transformer->transformArgument($definitionCall, $index, $value);
+            $transformedValue = $transformer->transformArgument($definitionCall, $index, $value);
+
+            if ($value !== $transformedValue) {
+                return $transformedValue;
+            }
         }
 
         return $value;
--- a/src/Behat/Behat/Transformation/Exception/FactoryMethodNotFound.php
+++ b/src/Behat/Behat/Transformation/Exception/FactoryMethodNotFound.php
@@ -1,0 +1,64 @@
+<?php
+
+/*
+ * This file is part of the Behat.
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Behat\Behat\Transformation\Exception;
+
+use InvalidArgumentException;
+
+/**
+ * Represents an exception caused by usage of type-hinted step definition argument without factory method.
+ *
+ * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ */
+final class FactoryMethodNotFound extends InvalidArgumentException implements TransformationException
+{
+    /**
+     * @var string
+     */
+    private $class;
+    /**
+     * @var string
+     */
+    private $method;
+
+    /**
+     * Initializes exception.
+     *
+     * @param string $message
+     * @param string   $class
+     */
+    public function __construct($message, $class, $method)
+    {
+        parent::__construct($message);
+
+        $this->class = $class;
+        $this->method = $method;
+    }
+
+    /**
+     * Returns class name.
+     *
+     * @return string
+     */
+    public function getClass()
+    {
+        return $this->class;
+    }
+
+    /**
+     * Return factory method name.
+     *
+     * @return string
+     */
+    public function getFactoryMethod()
+    {
+        return $this->method;
+    }
+}
--- a/src/Behat/Behat/Transformation/Exception/FactoryMethodNotFound.php
+++ b/src/Behat/Behat/Transformation/Exception/FactoryMethodNotFound.php
@@ -32,14 +32,15 @@ final class FactoryMethodNotFound extends InvalidArgumentException implements Tr
      * Initializes exception.
      *
      * @param string $message
-     * @param string   $class
+     * @param string $class
+     * @param string $factoryMethod
      */
-    public function __construct($message, $class, $method)
+    public function __construct($message, $class, $factoryMethod)
     {
         parent::__construct($message);
 
         $this->class = $class;
-        $this->method = $method;
+        $this->method = $factoryMethod;
     }
 
     /**
--- a/src/Behat/Behat/Transformation/Transformer/FromStringObjectTransformer.php
+++ b/src/Behat/Behat/Transformation/Transformer/FromStringObjectTransformer.php
@@ -11,8 +11,9 @@
 namespace Behat\Behat\Transformation\Transformer;
 
 use Behat\Behat\Definition\Call\DefinitionCall;
+use Behat\Behat\Transformation\Exception\FactoryMethodNotFound;
 use Closure;
-use ReflectionMethod;
+use ReflectionClass;
 use ReflectionParameter;
 
 /**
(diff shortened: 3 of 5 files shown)


ciaranmcnulty:
While I would use this as an extension, I don't think this should be core.

Thus far Behat has not introduced any conventions to how users design or write code, and I don't think it's appropriate for the core app to do so.

If you want to go ahead, the error message needs to be far more explicit about why the hell it things `fromString` should exist or why it's looking for it.

everzet:
> Thus far Behat has not introduced any conventions to how users design or write code, and I don't think it's appropriate for the core app to do so.
Why?

src/Behat/Behat/Transformation/Call/Filter/DefinitionArgumentsTransformer.php
```diff
@@ -100,7 +100,11 @@ private function transformArgument(DefinitionCall $definitionCall, $index, $valu
                 continue;
             }
 
-            return $transformer->transformArgument($definitionCall, $index, $value);
+            $transformedValue = $transformer->transformArgument($definitionCall, $index, $value);
+
+            if ($value !== $transformedValue) {
```
ciaranmcnulty: This condition is not needed. If they're equal you can still return `$transformedValue` (as `$value` is returned below);

Additionally the `continue` above doesn't add much to comprehension, you could flip that `if` and nest the `return` inside it.

src/Behat/Behat/Transformation/Exception/FactoryMethodNotFound.php
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * @var string
+     */
+    private $method;
+
+    /**
+     * Initializes exception.
+     *
+     * @param string $message
+     * @param string   $class
```
ciaranmcnulty: Missing a  docblock for`$method`

src/Behat/Behat/Transformation/Exception/FactoryMethodNotFound.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        return $this->class;
+    }
+
+    /**
+     * Return factory method name.
+     *
+     * @return string
+     */
+    public function getFactoryMethod()
```
ciaranmcnulty: If it's a factoryMethod then the constructor param should be named the same

src/Behat/Behat/Transformation/Transformer/FromStringObjectTransformer.php
```diff
@@ -0,0 +1,145 @@
...
+use ReflectionParameter;
+
+/**
+ * Transforms typehinted parameters using their `fromString` factory method (if exists).
+ *
+ * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ */
+final class FromStringObjectTransformer implements ArgumentTransformer
+{
+    const FACTORY_METHOD = 'fromString';
```
ciaranmcnulty: Could this be passed into the constructor instead? It could come from configuration but have a default?

src/Behat/Behat/Transformation/Transformer/FromStringObjectTransformer.php
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function transformArgument(DefinitionCall $definitionCall, $argumentIndex, $argumentValue)
+    {
+        $class = $this->getParameterClassByIndex($definitionCall, $argumentIndex);
+
+        if (!$class->hasMethod(self::FACTORY_METHOD)) {
```
ciaranmcnulty: Probably also need to check the method is compatible i.e. Unary, can take a string as it's arguments (compatible typehints)

ciaranmcnulty:
New exception message is a big improvement

src/Behat/Behat/Transformation/Call/Filter/DefinitionArgumentsTransformer.php
everzet: I don't want successful transformations to stack.

src/Behat/Behat/Transformation/Transformer/FromStringObjectTransformer.php
everzet: That's why I don't want it to be configurable and anything but `fromString`. If you want to configure it - you should use custom transformation.

src/Behat/Behat/Transformation/Transformer/FromStringObjectTransformer.php
everzet: If you want to configure it - you should use custom transformation.

src/Behat/Behat/Transformation/Exception/FactoryMethodNotFound.php
everzet: Ta. Fixed.

jakzal:
It doesn't feel right to introduce the `fromString` as a convention. It will encourage adding this kind of method to all value objects (since transformers will become so easy). In general I share @ciaranmcnulty's concerns expressed in #863. 

I really like option 3 from https://github.com/Behat/Behat/issues/863#issuecomment-235400924. Any reason why we couldn't support a variation of that in PHP 5.x?

pamil:
Totally agree with @jakzal and @ciaranmcnulty, especially that with the second or third option from https://github.com/Behat/Behat/issues/863#issuecomment-235400924 implemented, it will be a no-brainer (and ~6 lines of code) to implement own type-based transformations.

Support for PHP 5.x will be very appreciated, we use a lot of transformers that can be type-based in [Sylius](https://github.com/Sylius/Sylius/tree/master/src/Sylius/Behat/Context/Transform) and can't bump up dependencies to `^7.0`.

stof:
> When you attempt to execute this, you'll get exception like this:
> 
> ...> Method fromString does not exist
> 
> There are now two possible routes out of this situation...

I think that this message is not clear at all. I think it would be better if the transformation was checking the existence of ``fromString`` to decide whether it supports the argument or no, which would result in the argument not being transformed if it cannot.
Thus, it would remove the requirement to keep this transformer in last position forever (the current implementation requires it because if partial support check)

everzet:
Ok, it seems we're overwhelmingly against this feature in its current form. I then see two options for this PR:

1. Close it like it never happened
2. Do what @stof suggested. Remove the exception throwing by silently checking for `fromString` existence before transformation. If method exists - automatically use it for transformation, if not - silently skip the transformation

Thoughts?

everzet:
@jakzal, @pamil I'd keep discussion about wider object transformations improvement outside of the scope of this PR and keep it in the #863.

PR #910, opened 2016-07-31
Return type transformations

everzet:
Second half of #863 This PR introduces two new transformations:

## 1. Pure type-based based transformation for php7

Pure typed transformation, that matches all type-hints:

```php
/**
 * @Transform
 */
public function toUser($user) : User
{
    return User::fromString($user);
}
```

## 2. Both named and typed transformation for php7

In case you want to handle specific case differently, you'll be able to use both variable name AND type. Transformation will get applied ONLY if both match the argument:

```php
/**
 * @Transform :name
 */
public function toNamedUser($name) : User
{
    return User::named($name);
}
```

## Order of application

These will be applied in following order: "name & return type", "just return type", "just name". The first found wins.

### DIFF
--- a/features/definitions_transformations.feature
+++ b/features/definitions_transformations.feature
@@ -316,3 +316,70 @@ Feature: Step Arguments Transformations
       3 scenarios (3 passed)
       6 steps (6 passed)
       """
+
+  Scenario: By-type object transformations
+    Given a file named "features/my.feature" with:
+      """
+      Feature:
+        Scenario:
+          Given I am "everzet"
+          And he is "sroze"
+          Then I should be a user named "everzet"
+          And he should be a user named "sroze"
+      """
+    And a file named "features/bootstrap/FeatureContext.php" with:
+      """
+      <?php
+      class User {
+          private function __construct($name) { $this->name = $name; }
+          static public function named($name) { return new static($name); }
+      }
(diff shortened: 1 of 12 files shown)


everzet:
Just realised I've made a BC break in the `Transformation` interface. Will need to further refactor...

sroze:
So we don't care about introducing a BC break in the `SimpleArgumentTransformation` interface? It can probably be avoided quite easily by introducing a new one I think, wouldn't it be better? Or are you planning to introduce many BC breaks that would force you to release a new major version soon?

everzet:
@sroze good catch. That totally wasn't been intentional.

everzet:
`SimpleArgumentTransformation` interface didn't exist until this work. So we're all good :)

pamil:
Is it planned to add support for PhpDoc `@return` tag to the core if transformation does not have a return type? If so, I'd love to add it.

everzet:
@pamil I'm not planning on adding it, but if you want to - please go ahead.

pamil:
All right then, I'll try to send a PR this week.

nifr:
@pamil Did you start with the PR to add `@return` phpdoc? There's no reference in here and I can't find it in open PRs. I think this feature would be a great addition to behat.

PR #911, opened 2016-08-01
Fix context isolation for Scenario Outlines

everzet:
This PR fixes context isolation for Scenario Outlines. It achieves that without BC breaks by implementing additional tester decorator - `IsolatingScenarioTester`. It's a bit clunky, but totally working solution until the next major, when we can break BC.


jakzal:
:+1:

PR #916, opened 2016-08-06
 Default jUnit XML files Match Feature Basename

### LINKED ISSUE #915: Default jUnit XML files Match Feature Basename
Problem:
The jUnit XML file output by Behat creates a single file name 'default.xml' which is problematic with multiple behat feature files. 

Solution: 
Have the default XML files match the original feature file name

generalconsensus:
fixes #915

### DIFF
--- a/src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitOutlineStoreListener.php
+++ b/src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitOutlineStoreListener.php
@@ -56,7 +56,7 @@ final class JUnitOutlineStoreListener implements EventListener
     {
         $this->captureOutlineOnBeforeOutlineEvent($event);
         if ($event instanceof SuiteTested) {
-            setFormatterFileName($event, $formatter);
+            $formatter->fileName = $this->getFormatterFileName($event);
         }
 
         $this->printHeaderOnBeforeSuiteTestedEvent($formatter, $event);
@@ -64,17 +64,16 @@ final class JUnitOutlineStoreListener implements EventListener
     }
 
     /**
-     * Get FeatureNode from within Event
+     * Get FeatureNode from within Event and return the feature path name.
      *
      * @param Event $event
      */
-    private function setFormatterFileName(Event $event, Formatter $formatter) {
+    private function getFormatterFileName(Event $event) 
+    {
         foreach ($event->getSpecificationIterator() as $node) {
             if ($node instanceof FeatureNode) {
                 $path = pathinfo($node->getFile(), PATHINFO_FILENAME);
-                if (!empty($path)) {
-                    $formatter->fileName = $path;
-                }
+                return !empty($path) ? $path : NULL;
             }
         }
     }
--- a/src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitOutlineStoreListener.php
+++ b/src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitOutlineStoreListener.php
@@ -19,6 +19,7 @@ use Behat\Testwork\EventDispatcher\Event\BeforeSuiteTested;
 use Behat\Testwork\Output\Formatter;
 use Behat\Testwork\Output\Node\EventListener\EventListener;
 use Symfony\Component\EventDispatcher\Event;
+use Behat\Behat\Gherkin\Specification\LazyFeatureIterator;
 
 /**
  * Listens for Outline events store the current one
@@ -53,7 +54,21 @@ final class JUnitOutlineStoreListener implements EventListener
      */
     public function listenEvent(Formatter $formatter, Event $event, $eventName)
     {
-        $this->captureOutlineOnBeforeOutlineEvent($event);
+      $this->captureOutlineOnBeforeOutlineEvent($event);
+      if (method_exists($event, 'getSpecificationIterator')) {
+        $iterators = $event->getSpecificationIterator();
+        foreach ($iterators as $iterator) {
+          if ($iterator instanceof FeatureNode){
+            $file = $iterator->getFile();
+            if (!empty($file)) {
+              $path_parts = pathinfo($file);
+              if (!empty($path_parts['filename'])) {
+                $formatter->fileName = $path_parts['filename'];
+              }
+            }
+          }
+        }
+      }
 
         $this->printHeaderOnBeforeSuiteTestedEvent($formatter, $event);
         $this->printFooterOnAfterSuiteTestedEvent($formatter, $event);
--- a/src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitOutlineStoreListener.php
+++ b/src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitOutlineStoreListener.php
@@ -19,7 +19,7 @@ use Behat\Testwork\EventDispatcher\Event\BeforeSuiteTested;
 use Behat\Testwork\Output\Formatter;
 use Behat\Testwork\Output\Node\EventListener\EventListener;
 use Symfony\Component\EventDispatcher\Event;
-use Behat\Behat\Gherkin\Specification\LazyFeatureIterator;
+use Behat\Gherkin\Node\FeatureNode;
 
 /**
  * Listens for Outline events store the current one
@@ -55,13 +55,13 @@ final class JUnitOutlineStoreListener implements EventListener
     public function listenEvent(Formatter $formatter, Event $event, $eventName)
     {
       $this->captureOutlineOnBeforeOutlineEvent($event);
-      if (method_exists($event, 'getSpecificationIterator')) {
-        $class = 'Behat\Gherkin\Node\FeatureNode';
-        foreach ($event->getSpecificationIterator() as $iterator) {
-          if ($iterator instanceof $class) {
-            $file = $iterator->getFile();
-            if (($path_parts = pathinfo($file)) && !empty($path_parts['filename'])) {
-              $formatter->fileName = $path_parts['filename'];
+      if ($event instanceof SuiteTested) {
+        foreach ($event->getSpecificationIterator() as $node) {
+          if ($node instanceof FeatureNode) {
+            $file = $node->getFile();
+            $path = pathinfo($file, PATHINFO_FILENAME);
+            if (!empty($path)) {
+              $formatter->fileName = $path;
             }
           }
         }
--- a/src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitOutlineStoreListener.php
+++ b/src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitOutlineStoreListener.php
@@ -56,15 +56,12 @@ final class JUnitOutlineStoreListener implements EventListener
     {
       $this->captureOutlineOnBeforeOutlineEvent($event);
       if (method_exists($event, 'getSpecificationIterator')) {
-        $iterators = $event->getSpecificationIterator();
-        foreach ($iterators as $iterator) {
-          if ($iterator instanceof FeatureNode){
+        $class = 'Behat\Gherkin\Node\FeatureNode';
+        foreach ($event->getSpecificationIterator() as $iterator) {
+          if ($iterator instanceof $class) {
             $file = $iterator->getFile();
-            if (!empty($file)) {
-              $path_parts = pathinfo($file);
-              if (!empty($path_parts['filename'])) {
-                $formatter->fileName = $path_parts['filename'];
-              }
+            if (($path_parts = pathinfo($file)) && !empty($path_parts['filename'])) {
+              $formatter->fileName = $path_parts['filename'];
             }
           }
         }
--- a/src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitOutlineStoreListener.php
+++ b/src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitOutlineStoreListener.php
@@ -54,21 +54,29 @@ final class JUnitOutlineStoreListener implements EventListener
      */
     public function listenEvent(Formatter $formatter, Event $event, $eventName)
     {
-      $this->captureOutlineOnBeforeOutlineEvent($event);
-      if ($event instanceof SuiteTested) {
-        foreach ($event->getSpecificationIterator() as $node) {
-          if ($node instanceof FeatureNode) {
-            $file = $node->getFile();
-            $path = pathinfo($file, PATHINFO_FILENAME);
-            if (!empty($path)) {
-              $formatter->fileName = $path;
-            }
-          }
+        $this->captureOutlineOnBeforeOutlineEvent($event);
+        if ($event instanceof SuiteTested) {
+            setFormatterFileName($event, $formatter);
         }
-      }
 
         $this->printHeaderOnBeforeSuiteTestedEvent($formatter, $event);
         $this->printFooterOnAfterSuiteTestedEvent($formatter, $event);
+    }
+
+    /**
+     * Get FeatureNode from within Event
+     *
+     * @param Event $event
+     */
+    private function setFormatterFileName(Event $event, Formatter $formatter) {
+        foreach ($event->getSpecificationIterator() as $node) {
+            if ($node instanceof FeatureNode) {
+                $path = pathinfo($node->getFile(), PATHINFO_FILENAME);
+                if (!empty($path)) {
+                    $formatter->fileName = $path;
+                }
+            }
+        }
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitOutlineStoreListener.php
```diff
@@ -53,7 +54,21 @@ public function __construct(SuitePrinter $suitePrinter)
...
+            $file = $iterator->getFile();
+            if (!empty($file)) {
+              $path_parts = pathinfo($file);
+              if (!empty($path_parts['filename'])) {
+                $formatter->fileName = $path_parts['filename'];
+              }
+            }
+          }
+        }
+      }
```
everzet: Coding style is broken. Please follow PSR-0

everzet: Also, refactoring drastically needed. Too many nested ifs

sroze:
It would be good to add some tests too.

generalconsensus:
Not quite clear on how to follow PSR-0, but I did refactor the if statements. Please advise

src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitOutlineStoreListener.php
```diff
@@ -53,7 +54,18 @@ public function __construct(SuitePrinter $suitePrinter)
...
     public function listenEvent(Formatter $formatter, Event $event, $eventName)
     {
-        $this->captureOutlineOnBeforeOutlineEvent($event);
+      $this->captureOutlineOnBeforeOutlineEvent($event);
+      if (method_exists($event, 'getSpecificationIterator')) {
+        $class = 'Behat\Gherkin\Node\FeatureNode';
+        foreach ($event->getSpecificationIterator() as $iterator) {
+          if ($iterator instanceof $class) {
+            $file = $iterator->getFile();
+            if (($path_parts = pathinfo($file)) && !empty($path_parts['filename'])) {
```
sroze: You can get the file name directly with `pathinfo($file, PATHINFO_FILENAME)`

src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitOutlineStoreListener.php
```diff
@@ -53,7 +54,18 @@ public function __construct(SuitePrinter $suitePrinter)
      */
     public function listenEvent(Formatter $formatter, Event $event, $eventName)
     {
-        $this->captureOutlineOnBeforeOutlineEvent($event);
+      $this->captureOutlineOnBeforeOutlineEvent($event);
+      if (method_exists($event, 'getSpecificationIterator')) {
+        $class = 'Behat\Gherkin\Node\FeatureNode';
+        foreach ($event->getSpecificationIterator() as $iterator) {
```
sroze: I would name the variable `$iterator` as `$node` instead as this iterator return `Node`s.

src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitOutlineStoreListener.php
```diff
@@ -53,7 +54,18 @@ public function __construct(SuitePrinter $suitePrinter)
      */
     public function listenEvent(Formatter $formatter, Event $event, $eventName)
     {
-        $this->captureOutlineOnBeforeOutlineEvent($event);
+      $this->captureOutlineOnBeforeOutlineEvent($event);
+      if (method_exists($event, 'getSpecificationIterator')) {
+        $class = 'Behat\Gherkin\Node\FeatureNode';
```
sroze: No sure why you need this class in a variable...?

src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitOutlineStoreListener.php
```diff
@@ -19,6 +19,7 @@
 use Behat\Testwork\Output\Formatter;
 use Behat\Testwork\Output\Node\EventListener\EventListener;
 use Symfony\Component\EventDispatcher\Event;
+use Behat\Behat\Gherkin\Specification\LazyFeatureIterator;
```
sroze: This is a useless `use` as you don't use the class.

src/Behat/Behat/Output/Node/Printer/JUnit/JUnitSuitePrinter.php
```diff
@@ -44,7 +44,11 @@ public function printHeader(Formatter $formatter, Suite $suite)
 
         /** @var JUnitOutputPrinter $outputPrinter */
         $outputPrinter = $formatter->getOutputPrinter();
-        $outputPrinter->createNewFile($suite->getName());
+        if (!empty($formatter->fileName)) {
+          $outputPrinter->createNewFile($formatter->fileName);
```
sroze: Could you use only one `createNewFile` call, such as:
```php
$outputPrinter->createNewFile($formatter->fileName ?: $suite->getName());
```
(or put the filename in a variable is you prefer.

src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitOutlineStoreListener.php
```diff
@@ -53,7 +54,18 @@ public function __construct(SuitePrinter $suitePrinter)
      */
     public function listenEvent(Formatter $formatter, Event $event, $eventName)
     {
-        $this->captureOutlineOnBeforeOutlineEvent($event);
+      $this->captureOutlineOnBeforeOutlineEvent($event);
+      if (method_exists($event, 'getSpecificationIterator')) {
```
sroze: Instead of this method check, I would check if the event is an instance of the [`SuiteTested` class](https://github.com/Behat/Behat/blob/fb3241979bd080ac17f157244e9d6fc84a22bf9d/src/Behat/Testwork/EventDispatcher/Event/SuiteTested.php).

sroze: Instead of this method check, I would check if the event is an instance of the [`SuiteTested` class](https://github.com/Behat/Behat/blob/fb3241979bd080ac17f157244e9d6fc84a22bf9d/src/Behat/Testwork/EventDispatcher/Event/SuiteTested.php), that way you are sure that the return value is the expected type.

Even better (but not mandatory here I would say - @everzet to confirm) would be to introduce an interface for the events that contains this method, such as the [hook `SuiteScope` interface](https://github.com/Behat/Behat/blob/fb3241979bd080ac17f157244e9d6fc84a22bf9d/src/Behat/Testwork/Hook/Scope/SuiteScope.php).

generalconsensus:
@sroze First time contributor to Behat. Points well taken! Let me know what you think of the changes?

sroze:
Now you need to follow the PSR-2, I believe that @everzet's mention of PSR-0 was a typo. Basically, at least the 4 spaces indentation :wink:

src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitOutlineStoreListener.php
```diff
@@ -53,7 +54,18 @@ public function __construct(SuitePrinter $suitePrinter)
      */
     public function listenEvent(Formatter $formatter, Event $event, $eventName)
     {
-        $this->captureOutlineOnBeforeOutlineEvent($event);
+      $this->captureOutlineOnBeforeOutlineEvent($event);
+      if ($event instanceof SuiteTested) {
+        foreach ($event->getSpecificationIterator() as $node) {
```
sroze: Can you isolate the `foreach` loop into a dedicated method on the class? An explicit name (as you can see on other methods of this class) is required tho

src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitOutlineStoreListener.php
```diff
@@ -53,7 +54,18 @@ public function __construct(SuitePrinter $suitePrinter)
...
+          if ($node instanceof FeatureNode) {
```
sroze: I personally prefer to `continue` if not `instanceof`. That helps readability by reducing the depth of indentation.

src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitOutlineStoreListener.php
```diff
@@ -53,7 +54,18 @@ public function __construct(SuitePrinter $suitePrinter)
...
+            $file = $node->getFile();
```
sroze: Not sure that you need this variable now as you use it only once

src/Behat/Behat/Output/Node/Printer/JUnit/JUnitSuitePrinter.php
```diff
@@ -44,7 +44,7 @@ public function printHeader(Formatter $formatter, Suite $suite)
 
         /** @var JUnitOutputPrinter $outputPrinter */
         $outputPrinter = $formatter->getOutputPrinter();
-        $outputPrinter->createNewFile($suite->getName());
+        $outputPrinter->createNewFile((!empty($formatter->fileName) && $formatter->fileName) ? $formatter->fileName : $suite->getName());
```
sroze: You probably only need the `empty` test here

src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitOutlineStoreListener.php
```diff
@@ -54,12 +55,31 @@ public function __construct(SuitePrinter $suitePrinter)
     public function listenEvent(Formatter $formatter, Event $event, $eventName)
     {
         $this->captureOutlineOnBeforeOutlineEvent($event);
+        if ($event instanceof SuiteTested) {
+            setFormatterFileName($event, $formatter);
```
sroze: Please write a test for the feature you introduced/changed. You'll see that it is not working and that you are missing an obvious `$this->` here ;-)

src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitOutlineStoreListener.php
```diff
@@ -54,12 +55,31 @@ public function __construct(SuitePrinter $suitePrinter)
...
         $this->printHeaderOnBeforeSuiteTestedEvent($formatter, $event);
         $this->printFooterOnAfterSuiteTestedEvent($formatter, $event);
     }
 
     /**
+     * Get FeatureNode from within Event
+     *
+     * @param Event $event
+     */
+    private function setFormatterFileName(Event $event, Formatter $formatter) {
```
sroze: You need to add a new line before `{`

src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitOutlineStoreListener.php
sroze: I would also prefer having a method that returns the formater fileName and you simply assign the returned value of this method to `$formatter->fileName` (so you don't have to pass the `Formatter` object to your method)

src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitOutlineStoreListener.php
```diff
@@ -56,25 +56,24 @@ public function listenEvent(Formatter $formatter, Event $event, $eventName)
...
-    private function setFormatterFileName(Event $event, Formatter $formatter) {
+    private function getFormatterFileName(Event $event) 
+    {
         foreach ($event->getSpecificationIterator() as $node) {
             if ($node instanceof FeatureNode) {
                 $path = pathinfo($node->getFile(), PATHINFO_FILENAME);
-                if (!empty($path)) {
-                    $formatter->fileName = $path;
-                }
+                return !empty($path) ? $path : NULL;
```
sroze: If would `return` only if the path is not empty so it would try for the next `FeatureNode` in the iterator. And if not not empty path is found then it will return `null` by default.

sroze:
I will ask you again to add some tests to prove that your feature is working and to prevent anybody to break it in the future. No tests on a testing framework sounds bad ;-P

everzet:
@generalconsensus thanks for contributing for the first time!

There are couple of design issues with your PR. I would be happy to either fix them myself for you or give you indication and explanation on how/why to fix them (as a coaching exercise). Either way is fine with me :)

generalconsensus:
@everzet Thank you! Yes to coaching exercise! Let me know how I can make this PR work for Behat!

src/Behat/Behat/Output/Node/Printer/JUnit/JUnitSuitePrinter.php
```diff
@@ -44,7 +44,7 @@ public function printHeader(Formatter $formatter, Suite $suite)
 
         /** @var JUnitOutputPrinter $outputPrinter */
         $outputPrinter = $formatter->getOutputPrinter();
-        $outputPrinter->createNewFile($suite->getName());
+        $outputPrinter->createNewFile(!empty($formatter->fileName) ? $formatter->fileName : $suite->getName());
```
everzet: Here. Use delegation instead of control (also known as "Tell, don't ask" principle).

Can you refactor this piece of code to avoid employing creating and reading public properties from the formatter?

generalconsensus:
@everzet I believe this is what you are looking for, please advise

src/Behat/Behat/Output/Node/Printer/JUnit/JUnitSuitePrinter.php
```diff
@@ -44,7 +44,7 @@ public function printHeader(Formatter $formatter, Suite $suite)
 
         /** @var JUnitOutputPrinter $outputPrinter */
         $outputPrinter = $formatter->getOutputPrinter();
-        $outputPrinter->createNewFile(!empty($formatter->fileName) ? $formatter->fileName : $suite->getName());
+        $outputPrinter->createNewFile(!empty($name = $formatter->getName()) ? $name : $suite->getName());
```
everzet: 1. You are asking `$formatter` to give you the name
2. You are also asking `$suite` to give you the name
3. You then create a file name on the spot from these two pieces of information

You basically just asked a person to hand you over their phone just to check the current time. Instead of simply asking the person what time is it. Imagine next time person forgets his phone at his desk, but still does wear a watch on his hand - your time-asking routine will completely fail, even though person *can* technically tell you what time is it. If only you asked...

You have `$formatter`, you have `$suite`, you have a behaviour of figuring out the filename based on the information these two have. Give that behaviour a name (e.g. `generateFilename()`) and assign it to one of these two objects (e.g. `$formatter->generateFilenameFor($suite)`).

Tell objects to tell you the time, don't ask for their phone to check the time...

generalconsensus:
Please let me know if this is a better design pattern

src/Behat/Testwork/Output/Formatter.php
```diff
@@ -30,6 +30,13 @@
     public function getName();
 
     /**
+     * Returns file name.
+     * @param string $suite
+     * @return string
+     */
+    public function generateFilenameFor($suite);
```
everzet: Adding new methods to existing interfaces would be a BC break. Besides that, `generateFilenameFor()` seems to be a very-specific behaviour for a particular formatter, not all formatters in general.

src/Behat/Testwork/Output/NodeEventListeningFormatter.php
```diff
@@ -95,6 +99,14 @@ public function getName()
     /**
      * {@inheritdoc}
      */
+    public function generateFilenameFor($suite)
```
everzet: Please typehint the `$suite` variable.

src/Behat/Testwork/Output/Formatter.php
generalconsensus: I'm not quite clear on the best way forward with this -- I see that the PassthroughFormatter is another Formatter that I should be concerned with. Should I call the generateFilenameFor() method only when I know the $formatter is the NodeEventListeningFromatter class?

everzet:
@generalconsensus I went through your solution thoroughly and am afraid it wouldn't work as you probably expect it to.

First of all, you're still interacting with `SuiteTested`. That event is fired *after* all features in a suite were tested.

You also change the behaviour of `JUnitSuitePrinter::printHeader()` function to use the feature filename instead of suite name. `JUnitSuitePrinter::printHeader()` is called *once* before the suite execution. That filename you generate will be shared across all features in the suite.

This all means that if you have 1 suite with 3 features, you will still get 1 xml file at the end. All you did is made that 1 xml file to be named with a first feature filename instead of suite name.

If you want to actually proceed with your idea, I'd suggest you to extend the test case and focus on describing how the new functionality will work with suites, features and files. Use https://github.com/Behat/Behat/blob/master/features/junit_format.feature as a starting point.

everzet:
I'm closing this PR as it goes in the wrong direction. @generalconsensus please feel free to open another PR when you've written examples on how your change will work in the following cases:

* 1 suite with 1 feature
* 1 suite with 0 features
* 1 suite with 3 features
* 2 suites, 1 with 3 features, 1 with 2

We would be happy to help from that point. But until then you need to do some initial thinking on your own :)

PR #922, opened 2016-08-21
Append snippets revamp

everzet:
## TLDR;

`SnippetAcceptingContext` and `CustomSnippetAcceptingContext` were both deprecated and together with their support will be removed in 4.0.

Instead of them two new CLI options were introduced:

* `--snippets-for=CONTEXT_CLASS` - Specifies which context class Behat should generate snippets for.
* `--snippets-type=[regex|turnip]` - Specifies which snippet type Behat should generate.

`=CONTEXT_CLASS` is optional in `--snippets-for=CONTEXT_CLASS` if you provide `--snippets-for` as a last option in CLI (`behat features/my.feature --snippets-for`). If no context class provided, Behat will run in interactive mode.

## LV;

Using interfaces as indicators for where to generate snippets was a pretty dumb idea on my part. I constantly stumble into team commits that include "removing stupid `SnippetAcceptingContext`" comments. Problem is, in Behat 3 your suite has more than one context and depending on which context you generate snippets to, Behat needs to generate different method names (to avoid collisions). You do need to tell Behat which context you're expecting snippets to go to, but using interfaces for that purposes wasn't the right way to go about it.

This PR ditches aforementioned interfaces in favour of two new explicit CLI options - `--snippets-for` and `--snippets-type`.

Both new options work perfectly fine with existing `--append-snippets` and `--no-snippets` options. So from now on, you would use one of the following combinations for your snippet-generating routine:

* `bin/behat --snippets-for=FeatureContext` - Will print snippets in terminal.
* `bin/behat --append-snippets --snippets-for=FeatureContext` - Will append `turnip` snippets into `FeatureContext`.
* `bin/behat --append-snippets --snippets-for=FeatureContext --snippets-type=regex` - Will append `regex` snippets into `FeatureContext`.

To help newcomers and experienced users alike, unhelpful message:

screen shot 2016-08-21 at 14 09 55

was replaced with:

screen shot 2016-08-21 at 14 10 43

If typing is not your cup of tea, providing `--snippets-for` as a last option without value given will run interactive mode:

screen shot 2016-08-28 at 12 26 44

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
     },
 
     "require-dev": {
-        "symfony/process": "~2.1|~3.0",
+        "symfony/process": "~2.5|~3.0",
         "phpunit/phpunit": "~4.5",
         "herrera-io/box": "~1.6.1"
     },
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         "ext-mbstring":                  "*",
         "behat/gherkin":                 "~4.4",
         "behat/transliterator":          "~1.0",
-        "symfony/console":               "~2.1||~3.0",
+        "symfony/console":               "~2.5||~3.0",
         "symfony/config":                "~2.3||~3.0",
         "symfony/dependency-injection":  "~2.1||~3.0",
         "symfony/event-dispatcher":      "~2.1||~3.0",
--- a/features/append_snippets.feature
+++ b/features/append_snippets.feature
@@ -8,17 +8,15 @@ Feature: Append snippets option
       """
       <?php
 
-      use Behat\Behat\Context\CustomSnippetAcceptingContext,
+      use Behat\Behat\Context\Context,
           Behat\Behat\Tester\Exception\PendingException;
       use Behat\Gherkin\Node\PyStringNode,
           Behat\Gherkin\Node\TableNode;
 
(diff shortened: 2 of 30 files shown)


jakzal:
Could `--snippets-for` also accept a path to the context class? I often put my context classes in a namespace, which would require me to type a lot. A path can be autocompleted by shell, and since we're in a command line context while calling the command it feels more natural.

stof:
:+1: for this revamping

everzet:
@ciaranmcnulty @jakzal 

screen shot 2016-08-28 at 12 26 44

ciaranmcnulty:
Fantastic! Will this work with --append-snippets too?

I think in the case of only one context existing you could default to that one?

everzet:
> Fantastic! Will this work with --append-snippets too?
yes it will

> I think in the case of only one context existing you could default to that one?
first context is the default choice - just pressing ↩ will result in that context being selected.

ciaranmcnulty:
I mean in the --append-snippets case where selecting None would not make sense

everzet:
@ciaranmcnulty it would. Because `--snippets-for` without value will ask you the same question for each suite. This question provides a fine-grained control not only for which contexts to append snippets to, but which suites to ignore for now ;)

Issue #924, opened 2016-08-24
Sharing context between steps of a scenario


sroze:
In order to simply and have more human-readable scenarios, it would be nice to ease sharing state between steps. This idea is already explored by the [`ScenarioStateBehatExtension` extension](https://github.com/gorghoa/ScenarioStateBehatExtension) but we can have a lot of benefits and an easier syntax by adding it to the core of Behat.

# Current situation

As an example, the following scenario:
```feature
Scenario: Add a token to a bucket
   Given there is a bucket named "ContinuousPipe"
   When I add a token "1234" in the bucket "ContinuousPipe"
```

In the step definition, I can use a repository or just a property of the context to share the created `Bucket` object in this example:

```php
    /**
     * @Given there is a bucket named :name
     */
    public function iCreateABucketNamed($name)
    {
        $this->bucketRepository->save(Bucket::named($name));
    }

    /**
     * When I add a token :token in the bucket :bucketName
     */
    public function iAddATokenInTheBucket($token, $bucketName)
    {
        $bucket = $this->bucketRepository->findByName($bucketName);
        $bucket->addToken($token);
    }
```

In my point of view, the drawback here is to have to mention again the bucket name in the second step. An option would be to simply have a `bucket` property in the context but that requires every other step that wants to use this variable to have access to this property.

# Ideal situation

```feature
Scenario: Add a token to a bucket
   Given there is a bucket named "ContinuousPipe"
   When I add a token "1234" to this bucket
```

```php
    /**
     * @Given there is a bucket named :name
     */
    public function iCreateABucketNamed($name) : Bucket
    {
        return Bucket::named($name);
    }

    /**
     * @When I add a token :token to this bucket
     */
    public function iAddATokenToThisBucket(Bucket $bucket, $token)
    {
        $bucket->addToken($token);
    }
```

The value returned by the previous steps (if any) is then able to be inject in the following steps. Such a feature allow for instance to have these 2 definitions in different context files without any problem, or even to have another step that create a bucket a completely different way without any change on the second step.

I'm happy to send a PR before the 3.2.0 release if that makes sense to the community :)

spolischook:
what about context property, that can be set in one step and read in another? Or you mean something else?

sroze:
@spolischook just updated the issue body, I pressed "Create" too fast. Is it clearer now?

avant1:
I like the idea of saving domain objects and avoiding calling them by name (if it is not important for scenario) in favor of something like "this bucket".
[...]

PR #926, opened 2016-08-25
Allow injection of returned values into next steps' arguments 

sroze:
This PR aims to implement a solution for the #924. At the moment, it is a 5 minute prototype that breaks loads of interfaces, but it is supposed to show how simple such feature can be added in Behat.


everzet:
Sorry for the thinking and work on this one, but I'm closing it. See the reasoning and explanation in #924

PR #927, opened 2016-08-28
Remove support for Symfony prior to 2.7

everzet:
2.3 maintenance period effectively ended. The only actively maintained
Symfony 2.x versions are 2.7 and 2.8. I'm bumping minimum required Symfony
version to 2.7.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,6 @@ matrix:
     - php: 5.3
       env: DEPENDENCIES='low'
     - php: 5.6
-      env: SYMFONY_VERSION='2.3.*'
-    - php: 5.6
       env: SYMFONY_VERSION='2.7.*'
     - php: 5.6
       env: SYMFONY_VERSION='2.8.*'
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,7 @@
   * Transformations refactoring ([PR #905](https://github.com/Behat/Behat/pull/905))
   * `*SnippetAcceptingContext` interfaces deprecation ([PR #922](https://github.com/Behat/Behat/pull/922))
   * Exception handlers extension point added ([PR #931](https://github.com/Behat/Behat/pull/931))
+  * Bumped minimum dependencies version to Symfony 2.7 ([PR #927](https://github.com/Behat/Behat/pull/927)) 
 
 ### Bug fixes
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,6 +27,13 @@ where BC break is introduced as a measure to fix a serious issue.
 
 You can read detailed guidance on what BC means in [Symfony2 BC guide](http://symfony.com/doc/current/contributing/code/bc.html).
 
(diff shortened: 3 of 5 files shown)


ciaranmcnulty:
Do we have a versions policy anywhere?

This is a good move but brace yourself for the backlash :-)

everzet:
> Do we have a versions policy anywhere?

Good shout.

andytson:
Does "remove support" have to mean "don't allow it to work"?

patkar:
@andytson This PR means newer Behat releases after it got merged (and of course dev-master) are not installable with Symfony <2.7 anymore. Composer will only install the latest release before this PR.

ciaranmcnulty:
@andytson it would be irresponsible to claim support for a version in composer.json but not actually test against it

andytson:
maybe, yep, if the tests start to fail, but before the failure?

andytson:
anyway, I have no uses of Symfony 2.6- so no stake in this, just thought I'd be devils advocate

everzet:
@ciaranmcnulty added versioning policy

Taluu:
Why 2.7 and not 3.1 instead ? Or, if it's because there is no LTS version of 3.x versions (if I remember correctly), 2.8, as it is a LTS ?

patkar:
@Taluu 2.7 and 2.8 are still both LTS versions. And you're right, there is no 3.x LTS version. And according to the version policy it's about dropping unsupported versions of 3rd party deps, and not dropping supported versions ;)

Taluu:
Maybe, but there are things that are in 2.8 (and thus in 3.0, as 2.8 is 3.0 with deprecated things) that aren't in 2.7. That's limiting the scope for future improvements, especially as 2.7 and 2.8 have almost the same eol dates (with a few monts separating these two)

patkar:
So you want to annoy all Symfony 2.7 users whom are using behat? :scream:

everzet:
Just to be crystal clear - this PR is only about dropping support for 2.0-2.6. 2.7 support will remain no matter the outcome of this PR - it is a LTR and killing support for it for no reason is like shooting ourselves in the foot.

everzet:
After thorough thinking on this one I decided to postpone the bump until the 4.0 release. Here's my thinking:

1. There's no technical need to bump entire Symfony dependency to 2.7. Some dependencies that did need to be bumped were already bumped in their own PRs
2. If we remove support for unmaintained Symfony versions, it would make sense to also remove support for unsupported PHP versions. And we already agreed that it is too big of a change
3. We will likely need to bump minimum Symfony version to 3.x again in couple of months with 4.0 release. That seems excessive
4. Bumping versions required explicit change of policy mid-major. And I'm not quite comfortable with what it means

Overall, it's the point number 1 that I think swayed me over. If there's no technical need for a bump, there's no need for a bump.

Only meaningful reason left I see to do it is to nudge people to upgrade to latest Symfony. Behat is opinionated in many way, but which version of Symfony you use is not in that list.

Issue #932, opened 2016-08-30
--rerun should execute features in suites and not with all suites.


rajeshtaneja:
rerun option executes all failed features with all suites and doesn't respect if feature is part of the suite or not.

## Steps to reproduce:
> git clone https://github.com/rajeshtaneja/MinkExtension-example.git
cd MinkExtension-example
curl http://getcomposer.org/installer | php
php composer.phar install

#### Run behat
>git checkout multiplesuites
bin/behat

* Only one feature file is executed with 3 scenarios.
* Check /tmp/behat_rerun_cache/xxxx.scenarios and you will see only 1 scenario listed

##### Re-run behat
>bin/behat --rerun

You will observe features are executed twice, as temp context doesn't have search.feature it should only be executed once.

PR #933, opened 2016-08-30
#932 Save failed runs with suite information

rajeshtaneja:
Save failed runs with suite information, so
rerun run failed scenarios of suite

### DIFF
--- a/features/rerun_with_multiple_suite.feature
+++ b/features/rerun_with_multiple_suite.feature
@@ -1,0 +1,296 @@
+Feature: Rerun with multiple suite
+  In order to test only failed scenarios in different suite
+  As a feature developer
+  I need to have an ability to rerun failed previously scenarios
+
+  Background:
+    Given a file named "behat.yml" with:
+      """
+      default:
+        suites:
+          default:
+            paths:
+              - features/apples.feature
+              - features/bananas.feature
+            contexts:
+              - FeatureContext: [ { param1: val1, param2: val1 } ]
+          suite1:
+            paths: {}
+          suite2:
+            paths:
+              - features/bananas.feature
+      """
+    And a file named "features/bootstrap/FeatureContext.php" with:
+      """
+      <?php
+
+      use Behat\Behat\Context\Context;
+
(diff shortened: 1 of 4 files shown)


everzet:
Can you provide a test case? Simple version of "Steps to reproduce" section from your issue would work nicely!

rajeshtaneja:
Information has been added https://github.com/Behat/Behat/issues/932

## Steps to reproduce:
> git clone https://github.com/rajeshtaneja/MinkExtension-example.git
cd MinkExtension-example
curl http://getcomposer.org/installer | php
php composer.phar install

#### Run behat
>git checkout multiplesuites
bin/behat

* Only one feature file is executed with 3 scenarios.
* Check /tmp/behat_rerun_cache/xxxx.scenarios and you will see only 1 scenario listed

##### Re-run behat
>bin/behat --rerun

You will observe features are executed twice, as temp context doesn't have search.feature it should only be executed once.

everzet:
It is not a cost effective test case if it is not automated :) Can you write a representative feature file under `features/` folder?

rajeshtaneja:
Our features are distributed in different modules, so if plugin is installed it will have it's own features and context.

Sorry I didn't get the cost factor problem involved. Can you please explain. Other then the current solution  , I can override scenario list locator to check if it's a json file or just scenarios and return feature accordingly.

everzet:
@rajeshtaneja write an automated test case (living under `features/` folder) as part of this PR.

rajeshtaneja:
Thanks @everzet ,

Added testcase.

everzet:
@rajeshtaneja that's awesome! Thank you.

rajeshtaneja:
Thank you for keeping up the awesome work. 

Sorry, I didn't get your comment earlier about the testcase. I was reading between the lines....

Can you please let us know due date for 3.2.0

everzet:
@rajeshtaneja you can expect the RC1 in the next couple of days.

rajeshtaneja:
Thanks @everzet

PR #941, opened 2016-09-08
Transformation should be allowed if printable chars are used

rajeshtaneja:
Table row can have symbols/words/space
All printable chars should be allowed
while doing check for supported pattern


rajeshtaneja:
The failure seems to be happening because of https://github.com/Behat/Gherkin/commit/db8be8a11b49926fc52e3b4c53134b711b7ed487

everzet:
Yup

everzet:

Issue #951, opened 2016-09-20
Stop on failure with strict mode


estevejm:
Given that
- `--strict` option will make the tests fail if there's a missing step
- `--stop-on-failure` option will stop test execution as soon as a step fails

If we run Behat with both options I would expect it to stop test execution as soon as it detects a missing step. Unfortunately, this is not how this is working currently. 

Do you think it makes sense to change this behaviour? I would be happy to help...

BTW, I'm using PHP 5.6 with Behat 3.1

everzet:
I agree with this

everzet:
Closed in 954

PR #954, opened 2016-09-24
Strict stop on failure

estevejm:
Given that
- `--strict` option will make the tests fail if there's a missing step
- `--stop-on-failure` option will stop test execution as soon as a step fails

If we run Behat with both options it will stop test execution as soon as it detects a missing step.

This PR should close issue #951


everzet:
That is perfect! Thank you very much for contribution!

PR #955, opened 2016-09-24
Further improve snippet generation

### LINKED ISSUE #952: Snippets-for should be default behaviour
I've been teaching a workshop today.

Previous workflow:
* Run `behat --init`
* Write feature
* Run `behat --append-snippets`

New workflow:
* Run `behat --init`
* Write feature
* Run `behat --append-snippets --snippets-for`

It's not obvious to new users (from observation) that `--snippets-for` can be given as a boolean flag, so they are confused what value to give it.

This behaviour should be on by default in an interactive prompt, or perhaps be implied by `--append-snippets`, otherwise 3.2 requires more complex commands than 3.1 so is damaging to new users.

everzet:
Currently, when not using `SnippetAcceptingContext` interface (as it is deprecated), people are forced to use the `--snippets-for` flag in CLI. However, it is now obvious that for newcomers it is not transparent that `--snippets-for` has an interactive mode when called without argument. That causes unnecessary confusion.

This PR refactors slightly both pattern and context class identifiers to make both more flexible. As a result, context identifiers as of this commit operate in the following way:

1. If suite has any context implementing `SnippetAcceptingContext`, that class will be used for snippet generation.
2. If above is not true, but user explicitly provided a context class via `--snippets-for`, the provided class will be used
   instead.
3. If none of the above is  true, but the current CLI environment supports interactive mode - interactive mode will kick in automatically.
4. If none of the above is true, the help message asking user to use `--snippets-for` will be provided.

Closes #952

### DIFF
--- a/src/Behat/Behat/Context/Cli/InteractiveContextIdentifier.php
+++ b/src/Behat/Behat/Context/Cli/InteractiveContextIdentifier.php
@@ -57,7 +57,7 @@ final class InteractiveContextIdentifier implements TargetContextIdentifier
      */
     public function guessTargetContextClass(ContextEnvironment $environment)
     {
-        if (!$this->input->isInteractive()) {
+        if ($this->input->hasParameterOption('--no-interaction')) {
             return null;
         }
 
--- a/src/Behat/Behat/Context/Cli/InteractiveContextIdentifier.php
+++ b/src/Behat/Behat/Context/Cli/InteractiveContextIdentifier.php
@@ -57,6 +57,10 @@ final class InteractiveContextIdentifier implements TargetContextIdentifier
      */
     public function guessTargetContextClass(ContextEnvironment $environment)
     {
+        if (!$this->input->isInteractive()) {
+            return null;
+        }
+
         $suiteName = $environment->getSuite()->getName();
         $contextClasses = $environment->getContextClasses();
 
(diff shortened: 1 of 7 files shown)


src/Behat/Behat/Context/Cli/InteractiveContextIdentifier.php
```diff
@@ -57,6 +57,10 @@ public function __construct(TranslatorInterface $translator, InputInterface $inp
      */
     public function guessTargetContextClass(ContextEnvironment $environment)
     {
+        if ($this->input->hasParameterOption('--no-interaction')) {
```
stof: you should use ``$input->isInteractive()`` instead. The input can be non-interactive without having this option, thanks to autodetection of whether the command is run in a TTY or no.

everzet: I did [initially](https://github.com/Behat/Behat/pull/955/commits/283586a1ae521e5be0ef77c8042b18ec141caad6#diff-cc5e27e0868da5db12e6f72502938043R58). It breaks apart in the `Symfony\Console` prior to 2.7

everzet: Good catch, though. I made the tradeoff more explicit: https://github.com/Behat/Behat/commit/c04e67f3a1cfad72b7e79e192e537fb42268af26

PR #958, opened 2016-10-05
Simplify SuiteController::execute method logic

spolischook:


src/Behat/Testwork/Suite/Cli/SuiteController.php
```diff
@@ -60,24 +60,22 @@ public function configure(Command $command)
      * {@inheritdoc}
      */
     public function execute(InputInterface $input, OutputInterface $output)
-    {
-        $exerciseSuiteName = $input->getOption('suite');
+    {
```
stof: please don't add trailing whitespaces

src/Behat/Testwork/Suite/Cli/SuiteController.php
```diff
@@ -60,24 +60,22 @@ public function configure(Command $command)
      * {@inheritdoc}
      */
     public function execute(InputInterface $input, OutputInterface $output)
-    {
-        $exerciseSuiteName = $input->getOption('suite');
```
stof: please keep the variable definition outside the ``if``. It is more readable

src/Behat/Testwork/Suite/Cli/SuiteController.php
spolischook: fixed

src/Behat/Testwork/Suite/Cli/SuiteController.php
spolischook: fixed

stof:
This is breaking everything. Previously, when the suite name was not provided, all suites were registered. Now, you register none of them

PR #969, opened 2016-11-28
test suite filter with dash in name

mathroc:
this a test to illustrate #968 (I'm too lazy to run the tests on my machine so I'm editing from github and relying on travis here). 

I'll have a look later at how to fix this


everzet:
@mathroc thank you very much for effort, but can you add new scenario instead of changing the old one? I'd like scenarios to focus on one edge case at a time.

mathroc:
of course, there you go !

PR #971, opened 2016-12-06
Suite name with hyphen fix

### LINKED ISSUE #968: cli suite filter argument does not works for suite with "-" in names
> $ behat --suite=my-suite
[Behat\Testwork\Suite\Exception\SuiteNotFoundException]
  `my-suite` suite is not found or has not been properly registered.

but it works if I rename the suite to "my_suite" or "mysuite"

avant1:
This allows to use suites with hyphen in name in `--suite` CLI option.

Fixes #968.
I've included work from #969, but moved scenario to another feature file.

BTW this is a potential BC break for users who had hyphens in suite names and who used underscored version of names in `--suite` filters. 
With this PR `--suite=some_suite` will not work if suite has name `some-suite`.

However, this is a bugfix, and using underscored names was not documentated anywhere.

### DIFF
--- a/features/rerun_with_multiple_suite.feature
+++ b/features/rerun_with_multiple_suite.feature
@@ -19,8 +19,6 @@ Feature: Rerun with multiple suite
           suite2:
             paths:
               - features/bananas.feature
-          suite-3:
-            paths: {}
       """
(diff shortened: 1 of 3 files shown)


everzet:

PR #974, opened 2016-12-20
Helper containers

everzet:
## Narrative

Since I introduced support for multiple contexts in 3.0, the most asked question among the Behat newcomers goes along the lines of "how do I share state between multiple contexts?". I was reluctant to provide a systematic, built-in answer to this question for two reasons:

1. Behat is not your dependency management framework. Wiring applications together is usually a responsibility of application frameworks. I considered that it is better left to Symfony2Extension and friends to provide servicing configuration.
2. I generally consider sharing state between contexts to be an anti-pattern. It often highlights wrongly drawn context boundaries and trades off organizational simplicity for a short-term convenience.

Late push for container interoperability made me believe I finally have an answer to sharing state without giving Behat responsibility it ultimately wasn't built for. On top of that, recently conducted by me coaching, pairing and conversations with people inside and across the communities (thanks @tooky) made me realize that state/behaviour sharing still has very legitimate use-cases in many situations. Especially for Modelling by Example and similar workflows.

I believe this PR to be an elegant answer to the need without breaking very clear responsibility context Behat has. This PR is me finally trying to answer the question most community members keep asking.

## General principles

Before starting with the spec, I decided on two balancing principles that the solution must follow:

1. **Convenience limited to simple cases**. It should be simple enough to never get to the point where Behat gets confused with the application framework. Service sharing should work seamlessly for simple cases, but nudge people towards proper dependency management approaches in more complicated ones. If all you need is to instantiate an isolated environment with 1-2 independent, but shared services, you should be able to bootstrap in seconds.
2. **Complex cases served via interoperability**. In case of complex cases Behat must favour reuse. Whatever artefact Behat users end up producing as part of their service sharing procedure should be fairly decoupled from Behat and reusable with other, unaware of Behat frameworks. If you spend a lot of time wiring up your complex dependency tree for Behat, that work shouldn't be lost on your tests alone.

## Enter helper containers

Helper containers are service containers that you can configure to provide shared services for your contexts. And those services indeed be shared between your contexts, with all state and behaviour. Here's simple rules I followed in designing them:

- A single optional container is allowed per suite
- Having container enables you to use its services as context arguments via `@name` syntax
- Container is rebuilt and is isolated between scenarios
- Container is configured via suite's `services` option
- Container is a class implementing `Interop\Container\ContainerInterface`
- There is a built-in container if you need a very simple service-sharing, configurable through the same `services` setting
- There is an extension point that allows Behat extensions provide their own containers for end-users via `@name` syntax

## How to use a helper container

The rest of this PR will go in more general details on how to use the feature. For more detailed info and examples, [please read the feature file](https://github.com/Behat/Behat/blob/91d0b721ec5c48f9970c0d6a8154d5d2017283dd/features/helper_containers.feature).

### Built-in container

The simplest way of using helper container is to define a service (or services) that would be injected (and shared) between your contexts in the scope of a single scenario:

```yml
default:
  suites:
    default:
      contexts:
        - FirstContext:
          - "@shared_service"
        - SecondContext:
          - "@shared_service"

      services:
        shared_service: "SharedService"
```

This definition will make Behat automatically instantiate `SharedService` before each scenario and pass it as a first argument (via `@...` notation) to both `FirstContext` and `SecondContext` - exact same instance.

There's even a bit more verbose syntax that allows you to provide custom arguments and even factory method for each class:

```yml
      services:
        shared_service:
          class: "SharedService"
          factory_method: "fromNothing"
          arguments: []
```

**_Please note that built-in container does not allow you to use defined services as other service arguments. This is to both keep the built-in container away from becoming a full-blown service container and to nudge you towards external container in cases where you do need a deeply nested, highly-wired dependency tree._**

### External container class

Alternatively, you can use a custom container class that implements [`Interop\Container\ContainerInterface`](https://github.com/container-interop/container-interop/blob/master/src/Interop/Container/ContainerInterface.php):

```yml
default:
  suites:
    default:
      contexts:
        - FirstContext:
          - "@shared_service"
        - SecondContext:
          - "@shared_service"

      services: "MyContainer"
```

Don't forget the class itself. You can even place it near you context classes (in `features/bootstrap`):

```php
<?php

use Interop\Container\ContainerInterface;

final class MyContainer implements ContainerInterface
{
    // ...
}
```

And yes, you can utilise custom factory methods in those too:

```yml
default:
  suites:
    default:
      contexts:
        - FirstContext:
          - "@shared_service"
        - SecondContext:
          - "@shared_service"

      services: "MyContainer::fromEnvironment"
```

### Container from extension

The third and the last way is to use containers provided by your favourite extensions. Like that:

```yml
default:
  suites:
    default:
      contexts:
        - FirstContext:
          - "@shared_service"
        - SecondContext:
          - "@shared_service"

      services: "@symfony_extension.container"
```

The example above doesn't yet work, because it requires extension authors to actually provide that container explicitly from withing their config. But I hope they'll soon follow. Feel free to [help them](https://github.com/FriendsOfBehat/SymfonyExtension/pulls) :).

## Further details

For further details, as usual, I advise you to dive into [the feature file](https://github.com/Behat/Behat/blob/95f7b6d17538d740b999e63e2e300f78e57a7d54/features/helper_containers.feature) itself. It provides ultimate information, reasoining and examples behind this feature. Oh, yeah, it also is a stellar automated test to make sure I implemented feature correctly in the first place :)

As usual, feedback is greatly appreciated.


composer.json
```diff
@@ -14,17 +14,18 @@
...
+        "ext-mbstring":                        "*",
+        "behat/gherkin":                       "^4.4.4",
+        "behat/transliterator":                "~1.0",
+        "symfony/console":                     "~2.5||~3.0",
+        "symfony/config":                      "~2.3||~3.0",
+        "symfony/dependency-injection":        "~2.1||~3.0",
+        "symfony/event-dispatcher":            "~2.1||~3.0",
+        "symfony/translation":                 "~2.3||~3.0",
+        "symfony/yaml":                        "~2.1||~3.0",
+        "symfony/class-loader":                "~2.1||~3.0",
```
stof: I would actually stop aligning constraint to avoid such kind of diffs

src/Behat/Behat/HelperContainer/ServiceContainer/HelperContainerExtension.php
```diff
@@ -0,0 +1,118 @@
...
+     *
+     * @return bool
+     *
+     * @deprecated remove after upgrading to Symfony 2.8+
+     */
+    private function isDefinitionShared(Definition $definition)
+    {
+        if (method_exists($definition, 'isShared')) {
+            return $definition->isShared();
+        } else if (method_exists($definition, 'getScope')) {
```
stof: ``if`` is enough thanks to the return. And I'm not even sure we need this second check

src/Behat/Behat/HelperContainer/ServiceContainer/HelperContainerExtension.php
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /**
+     * Checks if provided definition is shared.
+     *
+     * @param Definition $definition
+     *
+     * @return bool
+     *
+     * @deprecated remove after upgrading to Symfony 2.8+
```
stof: you should not mark it as deprecated, as you should not call deprecated APIs internally. And this is a private function anyway.
This looks more like a TODO than a deprecation here btw

src/Behat/Behat/HelperContainer/ServiceContainer/HelperContainerExtension.php
```diff
@@ -0,0 +1,118 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        $references = $this->processor->findAndSortTaggedServices($container, self::HELPER_CONTAINER_TAG);
+
+        foreach ($references as $reference) {
+            if ($this->isDefinitionShared($container->getDefinition($reference))) {
```
stof: should be ``(string) $reference`` as ``getDefinition`` expects a string (the id), not a Reference object

src/Behat/Behat/HelperContainer/ServiceContainer/HelperContainerExtension.php
```diff
@@ -0,0 +1,118 @@
...
+                throw new WrongServicesConfigurationException(sprintf(
+                    'Container services must not be configured as shared, but `@%s` is.', $reference
```
stof: same here about using the string

composer.json
everzet: Agreed. I grown to dislike this in recent years too.

src/Behat/Behat/HelperContainer/ServiceContainer/HelperContainerExtension.php
everzet: Good shout. I kinda use these interchangeably.

src/Behat/Behat/HelperContainer/ServiceContainer/HelperContainerExtension.php
everzet: To stop static analysis tools (like Scrutinizer) from complaining about non-existing method :(

src/Behat/Behat/HelperContainer/ServiceContainer.php
```diff
@@ -0,0 +1,24 @@
...
+use Interop\Container\ContainerInterface;
+
+/**
+ * Stores services available for contexts and injected via argument resolvers.
+ *
+ * @see ServicesResolver
+ *
+ * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ */
+interface ServiceContainer extends ContainerInterface
```
everzet: I was trying to give people who don't care about `Interop\Container\ContainerInterface` a clear, transparent entry.

everzet: I'm actually not super-attached to it.

src/Behat/Behat/HelperContainer/ServiceContainer/HelperContainerExtension.php
everzet: Implicit conversion VS explicit conversion. I'm easier with this case than the one you mentioned next.

src/Behat/Behat/HelperContainer/ServiceContainer/HelperContainerExtension.php
everzet: This again can be considered implicit VS explicit. Except in this case I obviously tell `sprintf` I need a string (`@%s`). Hence, I'd argue conversion is already explicit here and additional conversion would only confuse things.

That is unless you have a different tech/clarity reason I haven't thought of yet :)

src/Behat/Behat/HelperContainer/ServiceContainer.php
everzet: Actually, having separate interface breaks the second of feature principles. So I'd go ahead and drop `Behat\Behat\HelperContainer\ServiceContainer`.

src/Behat/Behat/HelperContainer/ServiceContainer/HelperContainerExtension.php
stof: Thus Symfony never explicitly mentions that ``getDefinition`` supports castable objects. If it works today, it is purely because of an implementation detail (probably because of using ``lowercase`` which supports it). So it means that passing a castable object is *not* covered by the BC policy (as it is a case not respecting the documented contract). So making it explicit is necessary to be future-proof

src/Behat/Behat/HelperContainer/ServiceContainer/HelperContainerExtension.php
stof: yeah, I was not sure about it being necessary here.

everzet: Cool. I'll introduce casting only to the `getDefinition(...)` case then.

everzet:
@stof thanks for your feedback, as always. I fixed everything you highlighted.

features/helper_containers.feature
```diff
@@ -0,0 +1,307 @@
+Feature: Per-suite helper containers
+  In order to share state and behaviour between contexts
+  developers need to have a way to create or register shared service container
+
+  Rules:
+    - A single optional container is allowed per suite
+    - Having container enables you to use its services as context arguments via `@name` syntax
```
ciaranmcnulty: 'Having a container"

src/Behat/Behat/Context/Argument/AggregateFactory.php
```diff
@@ -0,0 +1,52 @@
...
+use Behat\Testwork\Suite\Suite;
+
+/**
+ * Aggregate factory. Delegates to other (registered) factories to do the job.
+ *
+ * @see ContextEnvironmentHandler
+ *
+ * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ */
+final class AggregateFactory implements SuiteScopedResolverFactory
```
ciaranmcnulty: I'd rather this was called Composite, after the design pattern. Aggregate risks confusion with the DDD concept, or other usages where an aggregate is a collection

src/Behat/Behat/Context/Argument/NullFactory.php
```diff
@@ -0,0 +1,31 @@
...
+use Behat\Testwork\Suite\Suite;
+
+/**
+ * NoOp factory. Always returns zero resolvers.
+ *
+ * @see ContextEnvironmentHandler
+ *
+ * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ */
+final class NullFactory implements SuiteScopedResolverFactory
```
ciaranmcnulty: There might be better naming - `noop` from the comment makes sense, or maybe `empty`?

src/Behat/Behat/HelperContainer/Argument/ServicesResolverFactory.php
```diff
@@ -0,0 +1,169 @@
...
+    /**
+     * Creates custom container using class/constructor given.
+     *
+     * @param string $settings
+     *
+     * @return mixed
+     */
+    private function createContainerFromString($settings)
+    {
+        if ('@' === mb_substr($settings, 0, 1)) {
```
ciaranmcnulty: A similar check takes place in `ServicesResolverFactory` - can these be done in one place? Perhaps a value object?

ciaranmcnulty:
If I have services option and the symfony2extension is there fallback, or are all @-params interpreted as non-Symfony services?

src/Behat/Behat/Context/Argument/NullFactory.php
everzet: If we follow the pattern naming from the previous comment, we keep following it for null objects :)

src/Behat/Behat/HelperContainer/Argument/ServicesResolverFactory.php
everzet: Potentially. But that's too small of an op to worry about reuse at this point.

everzet:
@ciaranmcnulty service-based resolvers come first as they are more specific (as in local to suites). I'd avoid using `services` with `Symfony2Extension` until it is updated to support new functionality in proper way.

Issue #991, opened 2017-01-26
Context constructor initializer removes null arguments


Taluu:
When trying to instanciate a context with a "nullable" service (e.g, having a `?MyService` in php 7.1), if the value is indeed `null`, the arguments seems to be filtered when passed to the instanciator.

Thus, we have the following error `UnknownParameterException: Can not find a matching value for ...`. If I pisted the thing correctly, it seems that somewhere along the way, the `null` is stripped from the arguments list. And as this is not an `optionnal`, but a `nullable` (php 7.1) parameter, the count of non optionnal parameters and given arguments is not exact and thus the exception is thrown.

In my case, after pondering it, I think that it is fine if the service is optionnal (having `null` as a default value) so I changed it, but it is not always the case...

I am not really sure how to fix this though.

spolischook:
You can set null as dependency and then replace it in your Extension
```yml
my_own_service:
  class: Namespace\To\Class
  arguments:
    - ~
```
In this way you should have null in constructor.
In Extension you can replace this argument according to some conditions:
```php
public function process(ContainerBuilder $container)
{
$container
  ->getDefinition('my_own_service')
  ->replaceArgument(0, new Reference('dependecy_service'));
}
```

or just use setter injection

Taluu:
Actually, that is not the point. Yes, using the container in the extension (or whatever) does the trick, but when you are using the helper container, as follows :

```yaml
suites:
    defaults:
        contexts:
            FeatureContext:
                - @my_service

        services: @my.container
```

The container being :

```php
<?php
use Interop\Container\ContainerInterface;

class MyContainer implements ContainerInterface
{
    // ...
    public function get($id) 
    {
        if ('my_service' === $id) {
            return null;
         }

         // .... other cases
     }

    // ....
}
```

And in your context, the following :

```php
<?php

// ....
public function __construct(?MyService $foo)
{
    // ...
}
```

Behat will then complain with thrown error, that it doesn't have the right number of args ; one way or another, the `null` is stripped of the arguments stack, and thus complains that there is no match for `$foo`. Hence the bug.

PR #992, opened 2017-01-27
Do not treat first argument as numbered argument if typehinted

### LINKED ISSUE #990: MixedArgumentOrganizer misinterpret the first argument
When looping over arguments in `MixedArgumentOrganiser::splitArguments`, if an argument matches the typehint of the first argument (index 0), it is evaluated as `false` and thus not registered as a `typeHintedArgument`, but a `numberedArgument`.

https://github.com/Behat/Behat/blob/master/src/Behat/Testwork/Argument/MixedArgumentOrganiser.php#L91

Maybe we should check if the `num` is a digit or something ?

Taluu:
Fixes #990 : If the first argument (index 0) is typehinted, as `0 == false`, it was never registered as a typehinted argument but as a numbered argument.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 ### Fixed
   * [#993](https://github.com/Behat/Behat/pull/993) Fix mixed arguments organizer not marking typehinted arguments as "defined"
+  * [#992](https://github.com/Behat/Behat/pull/993) Do not misinterpret first argument as a numbered argument if it is in fact typehinted
 
 ## [3.3.0] - 2016-12-25
 ### Added
--- a/features/helper_containers.feature
+++ b/features/helper_containers.feature
@@ -314,7 +314,6 @@ Feature: Per-suite helper containers
           default:
             contexts:
               - FirstContext:
-                - foo
                 - "@typehinted_service"
                 - bar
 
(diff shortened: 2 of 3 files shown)


stof:
Please add a test covering this

Taluu:
Any idea how ? The class is full of privates, and the return is correct with or without this patch. It is just that during the execution of privates methods, the split is wrong (saw it with a debugger while trying to resolve #991)...

stof:
Well, is there a bug or no ?

Taluu:
Let's say it's more a problem *during execution*, that somewhat resolves itself by the end. I think some screens of my debugger would be more talkative ;

Here is the constructor of my context :

`function __construct(HttpClient $client, StreamFactory $streamFactory, MessageFactory $messageFactory, LastHistory $history, ?Twig_Environment $twig)`

image

The first argument (here is is `HttpClient`) should be in here actually, as the context is typehinted to receive this parameter :

image

But then once we go to the validate part, the args are correctly set ;

image

 So this is not a "real" bug per se, but still, some surprises could still happen afterwards IMO

My patch fixes that, see :

image

And it is not really testable as we can't hook into the private methods, as this is really implementation details...

stof:
OK then

everzet:
@Taluu can you update the changelog?

Taluu:
@everzet Rebased and updated

everzet:
@Taluu thank you!

Issue #989, opened 2017-01-26
Supporting Puli for extensions


Taluu:
Hello,

Currently, when developping an extension, when configuring it in the behat.yml config file, we need to refer it from its FQCN rather than the defined config key (`My\BehatExtension` instead of `my_extension`). This is probably because of the fact that the "extension" is not loaded when we just require it, and have no way to tell behat "i've got a new extension".

So I was wondering if supporting Puli would be (or not..) an idea to help Behat discover the extensions, and then allow to use them directly.

For an exemple, you could look at how the discovery is done in the php-http package : 
- https://github.com/php-http/discovery/blob/master/src/Strategy/PuliBetaStrategy.php
- https://github.com/php-http/curl-client/blob/master/puli.json (example of a puli package)

If the idea is validated, I could try a PR

everzet:
Might be interesting. As long as it doesn't introduce complexity into the core, colour me interested.

Taluu:
I'll try to see how it could be done then, but IMO shouldn't be really complicated...

everzet:
Two constraints I'd put on it are:

1. It needs to be completely optional
2. It mustn't introduce extra complexity to the core. Ideally, it should reduce existing complexity if possible
3. Prefer more general, less tool-specific (Puli) implementation

PR #995, opened 2017-02-03
Add support for Puli

Taluu:
this implements what is described in #989, although it is right now pretty specific for puli (as there is no "standards" currently).

I am not sure how to test this, but maybe registering the built-in extensions in the puli json file, and then remove the `extensions` parameter from the constructor could lead to something ? And then test that a built-in extension loads itself correctly ?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,9 @@
     "require-dev": {
         "symfony/process": "~2.5|~3.0",
         "phpunit/phpunit": "~4.5",
-        "herrera-io/box": "~1.6.1"
+        "herrera-io/box": "~1.6.1",
+        "puli/composer-plugin": "^1.0@beta",
+        "puli/cli": "^1.0"
     },
 
     "suggest": {
@@ -52,6 +54,9 @@
             "dev-master": "3.2.x-dev"
         }
     },
+
+    "prefer-stable": true,
+    "minimum-stability": "dev",
 
     "bin": ["bin/behat"]
 }
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         "symfony/process": "~2.5|~3.0",
         "phpunit/phpunit": "~4.5",
         "herrera-io/box": "~1.6.1",
-        "puli/composer-plugin": "^1.0@beta",
+        "puli/composer-plugin": "dev-master",
         "puli/cli": "^1.0"
     },
 
--- a/src/Behat/Testwork/ServiceContainer/ExtensionManager.php
+++ b/src/Behat/Testwork/ServiceContainer/ExtensionManager.php
@@ -20,10 +20,6 @@ use Behat\Testwork\ServiceContainer\Exception\ExtensionInitializationException;
 final class ExtensionManager
 {
     /**
-     * @var string
-     */
-    private $extensionsPath;
-    /**
      * @var Extension[]
      */
     private $extensions = array();
@@ -34,30 +30,24 @@ final class ExtensionManager
     private $debugInformation = array(
         'extensions_list' => array()
     );
+    /**
+     * @var ExtensionInstantiator[]
+     */
+    private $instantiators = array();
 
     /**
      * Initializes manager.
      *
-     * @param Extension[] $extensions     List of default extensions
-     * @param null|string $extensionsPath Base path where to search custom extension files
+     * @param Extension[]             $extensions    List of default extensions
+     * @param ExtensionInstantiator[] $instantiators Extension instantiators to use
      */
-    public function __construct(array $extensions, $extensionsPath = null)
+    public function __construct(array $extensions, array $instantiators)
     {
         foreach ($extensions as $extension) {
             $this->extensions[$extension->getConfigKey()] = $extension;
         }
 
-        $this->extensionsPath = $extensionsPath;
-    }
-
-    /**
-     * Sets path to directory in which manager will try to find extension files.
-     *
-     * @param null|string $path
-     */
-    public function setExtensionsPath($path)
-    {
-        $this->extensionsPath = $path;
+        $this->instantiators = $instantiators;
     }
 
     /**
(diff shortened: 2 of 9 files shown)


Taluu:
Something on what could be improved on the "extension detection" would be to have a real system of extension locators (one for class locator, one for file locator, one for puli locator, ... and so on, the first "supports" win), but I am not too sure if this wouldn't be too overkill ?

src/Behat/Testwork/ServiceContainer/ExtensionManager.php
```diff
@@ -228,4 +234,29 @@ private function validateExtensionInstance($extension, $locator)
...
         }
     }
+
+    private function discoverExtensionWithPuli($locator)
+    {
+        static $extensions = null;
+
+        // feed extension on first call
+        if (null === $extensions) {
+            $extensions = [];
```
Taluu: I have to change this into `array()`, I know.

everzet: Sorry.

everzet:
I'd rather not tie `ExtensionManager` to `Puli` directly. Sounds like abstraction is due.

stof:
> I am not sure how to test this, but maybe registering the built-in extensions in the puli json file, and then remove the extensions parameter from the constructor could lead to something ? And then test that a built-in extension loads itself correctly ?

Test this by writing a Behat scenario doing it with a custom extension.

Taluu:
> Test this by writing a Behat scenario doing it with a custom extension.

Yup, but I need to be sure that puli is available, that's the main pain IMO. Maybe add a tag or something ? I don't really want to add it in the `require-dev` section...

everzet:
> I don't really want to add it in the require-dev section...

@Taluu why not?

src/Behat/Testwork/ServiceContainer/ExtensionManager.php
```diff
@@ -228,4 +234,29 @@ private function validateExtensionInstance($extension, $locator)
             ), $locator);
         }
     }
+
+    private function discoverExtensionWithPuli($locator)
```
everzet: I'd rather this method and all the other mentions of Puli to be hidden behind an abstraction.

Taluu:
@everzet Because I feel like it should be more as a suggestion or optionnal dependency rather than required dev deps, but if you say it's okay, then I don't really mind

src/Behat/Testwork/ServiceContainer/ExtensionManager.php
Taluu: You mean abstracting external tools or abstracting the loaders (one for the file loader, one for class locator, one for puli, ... and so on) ?

stof:
@Taluu dev requirements are things needed to run **our** tests. They have **no** impact on our users. So we can put what we need in them.
Forcing contributors to install puli separately to be able to run tests would make things harder for them, for no benefit

composer.json
```diff
@@ -53,5 +54,8 @@
         }
     },
 
+    "prefer-stable": true,
+    "minimum-stability": "dev",
```
Taluu: Had to it because `puli/composer-plugin` is still in beta (composer doesn't accept to install it then, as it is "not stable" yet)

src/Behat/Testwork/ServiceContainer/ExtensionManager.php
```diff
@@ -34,30 +30,24 @@
...
 
-        $this->extensionsPath = $extensionsPath;
-    }
-
-    /**
-     * Sets path to directory in which manager will try to find extension files.
-     *
-     * @param null|string $path
-     */
-    public function setExtensionsPath($path)
```
Taluu: this wasn't used, but as it is public, I'm wondering if this could be considered a BC break ; but I didn't find a way on how to keep compatibility though...

Taluu:
@everzet PR updated with an abstraction on the initializers. The tests are not passing on php 5.3 for reasons that are eluding me (I can't manage to find the "line 446" on the puli plugin...).

Tests are missing but I'll add these later.

everzet:
I just reran the test and it still fails. Something is broken

composer.json
everzet: I think this is the bit that breaks the build on 5.3 :/

Taluu:
Mmh, on travis it seems it is installing the `beta-6` of the puli plugin (which faults at https://github.com/puli/composer-plugin/blob/1.0.0-beta6/src/PuliPlugin.php#L446), while there is the `beta-10` available. Maybe it's due to that ? I'll try to update the plugin and put another version then instead ?

Taluu:
BTW, wasn't 5.3 support dropped ? Why is it still on travis ?

composer.json
everzet: I feel uneasy about doing this with stable core release. If that's the only way of doing the feature, my preference shifts to just providing an extension point in the core, but Puli being an extension (maybe even official).

Taluu: So just splitting the extension loader ? Can do

everzet: Yeah, well, the only challenge you will how is that now you will need to introduce an extension point for extension instantiators, which is definitely a head-scratcher.

Taluu: Well, first things first...

Or what can be done, as it is done in the travis configuration, it's to enable these only in travis (or in dev through a global configuration ?). So it would mean to remove these two configuration options. Don't forget that it's only for dev-mode...

Taluu: But it doesn't solve the problem that if Puli is provided in an extension (official or not), it would indeed be hard to solve the problem of extending these (you would actually need to load the extension through a normal way ?)

Or what could be done, instead of relying on these, would be to support something like a composer package type, which would be loaded automatically ?

Taluu:
Closing this one, as this could be in a separate (official or not) extension (but before we need to solve the ["head-scratching" issue](https://github.com/Behat/Behat/pull/995#discussion_r110240969) mentioned by @everzet in favor of #1018

PR #996, opened 2017-02-10
[Feature] Accept multiple path arguments

yuloh:
This PR adds the ability to pass multiple path arguments:

```
behat features/apples.feature features/pears.feature
```

We use CircleCI to run a large suite of Behat tests across multiple docker containers.  CircleCI can [automatically balance](https://circleci.com/blog/announcing-automatic-test-balancing/) tests across the containers to keep the run time as low as possible.  The problem is, Circle assumes the test runner will accept multiple file arguments.

I'm working around this at the moment with a bash script and a Behat extension so I can invoke behat like `behat --feature apples.feature --feature pears.feature`.  I can't make an extension that directly changes the `paths` argument to accept an array because other extensions expect a string.

This is a breaking change, so it will need to wait for Behat 4.0.0.

Backwards (In)compatibility:

* You can't add an optional argument after an array argument, so any extension adding an optional argument will break.
* Any extension using `$input->getArgument('paths')` or `$input->setArgument('paths')` and expecting a string will break.

What do you think?


gquemener:
I personnally think this would be an awesome improvement, very usefull on CircleCI indeed to parallelize execution.

src/Behat/Testwork/Tester/Exception/WrongPathsException.php
```diff
@@ -20,30 +20,30 @@
...
     }
 
     /**
      * Returns path that caused exception.
      *
-     * @return string
+     * @return string[]
      */
-    public function getPath()
+    public function getPaths()
```
everzet: Please also add `getPath()` with `@deprecated` tag for BC. It can return the first path given.

src/Behat/Testwork/Tester/Exception/WrongPathsException.php
```diff
@@ -20,30 +20,30 @@
...
+     * @param string   $message
+     * @param string[] $paths
      */
-    public function __construct($message, $path)
+    public function __construct($message, $paths)
     {
         parent::__construct($message);
 
-        $this->path = $path;
+        $this->paths = $paths;
```
everzet: For BC please ensure that `$paths` argument can be either a string or an array. Both must be supported for now.

```
$this->paths = (array) $paths;
```

everzet:
> Any extension using $input->getArgument('paths') or $input->setArgument('paths') and expecting a string will break.

Argh. Good catch :/ I'm sure there is a way though...

stof:
@everzet but would it work for the circleCI features ?

Btw, this would be a nightmare to implement, because you can have spaces inside path, so we would have to provide our own escaping inside this string. And require to escape spaces in the path to avoid considering the argument as a path list is a BC break for *callers* of Behat (and makes it much harder to use Behat by default), which is worse than breaking extensions IMO (especially given that extensions should generally not mess with arguments implemented by another one IMO)

stof:
thus, if an extension relies on the argument directly, changing the content of the string from a path to a space-delimited list of escaped path would *also* break BC for the extension, as it would have to change the way it uses the string.

everzet:
:/

gquemener:
Here's Circle CI behaviour : https://circleci.com/docs/parallel-manual-setup/#using-configuration-files-modifier

> By default, the file arguments will be appended to the end of the command.

I'm not sure how it'll react if behat expects it to be quoted, but I wouldn't recommend so.

Well, that's too bad that the argument is already named `paths` and not `path` (any historical reason for that?).

What about bumping to 4.0.0 ?

yuloh:
@everzet I rebased on master and made WrongPathsException backwards compatible.

It sounds like there isn't a good way to prevent breaking BC.  I'm personally ok with waiting until 4.0, even if it takes a few years :)

pscheit-lillydoo:
voice-over-sponge-bob-voice: 

3 years later

adrienbrault:
Hey, is anything else needed for this besides a rebase?

What could I do to help?

ciaranmcnulty:
Can you please target this to the 4.x branch?

PR #997, opened 2017-02-15
Allow to use `*.yaml` configuration files

slavcodev:
This PR adds the ability to use `*.yaml` extension for the configuration, since this extension is considered official.

To keep backward compatibility but avoid extra file operation, checking with `is_file` was replaced with `glob`.

### DIFF
--- a/src/Behat/Behat/ApplicationFactory.php
+++ b/src/Behat/Behat/ApplicationFactory.php
@@ -111,11 +111,22 @@ final class ApplicationFactory extends BaseFactory
     protected function getConfigPath()
     {
         $cwd = rtrim(getcwd(), DIRECTORY_SEPARATOR) . DIRECTORY_SEPARATOR;
-        $configDir = 'config' . DIRECTORY_SEPARATOR;
-        $paths = glob("{$cwd}{,{$configDir}}behat.y{a,}ml{,.dist}", GLOB_BRACE);
-
-        if (count($paths)) {
-            return current($paths);
+        $configDir = $cwd . 'config' . DIRECTORY_SEPARATOR;
+        $paths = array(
+            $cwd . 'behat.yaml',
+            $cwd . 'behat.yml',
+            $cwd . 'behat.yaml.dist',
+            $cwd . 'behat.yml.dist',
+            $configDir . 'behat.yaml',
+            $configDir . 'behat.yml',
+            $configDir . 'behat.yaml.dist',
+            $configDir . 'behat.yml.dist',
+        );
+        
+        foreach ($paths as $path) {
+            if (is_file($path)) {
+                return $path;
+            }
         }
 
         return null;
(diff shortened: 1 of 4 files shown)


src/Behat/Behat/ApplicationFactory.php
```diff
@@ -110,16 +110,9 @@ protected function getEnvironmentVariableName()
...
-                $cwd . DIRECTORY_SEPARATOR . 'behat.yml',
-                $cwd . DIRECTORY_SEPARATOR . 'behat.yml.dist',
-                $cwd . DIRECTORY_SEPARATOR . 'config' . DIRECTORY_SEPARATOR . 'behat.yml',
-                $cwd . DIRECTORY_SEPARATOR . 'config' . DIRECTORY_SEPARATOR . 'behat.yml.dist',
-            ),
-            'is_file'
-        );
+        $cwd = rtrim(getcwd(), DIRECTORY_SEPARATOR) . DIRECTORY_SEPARATOR;
+        $configDir = 'config' . DIRECTORY_SEPARATOR;
+        $paths = glob("{$cwd}{,{$configDir}}behat.y{a,}ml{,.dist}", GLOB_BRACE);
```
stof: having ``{`` both for PHP interpolation and the string content is hardly readable. Please use ``sprintf`` instead

src/Behat/Behat/ApplicationFactory.php
```diff
@@ -110,16 +110,9 @@ protected function getEnvironmentVariableName()
...
 
         if (count($paths)) {
             return current($paths);
```
stof: this is now broken, as there is no guarantee that the glob will return the files in the order of precedence defined by Behat when there are several such files (for instance, when having both a committed ``behat.yml.dist`` and overwriting it with a git-ignored ``behat.yml``.

To limit the number of ``is_file`` calls, the solution could be to use a loop of the paths, do ``is_file`` inside the loop and then returning early as soon as a file is found (as we don't care about others anyway) instead of filtering the whole list

src/Behat/Behat/ApplicationFactory.php
stof: ``GLOB_BRACE`` is not defined on all systems. Some systems don't support it, which means you would drop support for them here.

stof:
Some scenario must be written to cover this case.
And the order of precedence must be defined clearly.

src/Behat/Behat/ApplicationFactory.php
slavcodev: The glob pattern take a care about order, there is order of files

~~~
behat.yaml
behat.yml
behat.yaml.dist
behat.yml.dist
config/behat.yaml
config/behat.yml
config/behat.yaml.dist
config/behat.yml.dist
~~~

slavcodev:
Thanks for review, I've updated PR.

stof:
This still need scenarios covering this new feature

everzet:
@slavcodev can you update the `CHANGELOG`?

slavcodev:
Sure, shall I mentioned that `*.yml` is considered deprecated?

stof:
I don't think we should consider it as deprecated IMO. ``yml`` is also very common for YAML files

slavcodev:
You're right it is common, but doe sit make sense to support both? I mean i the next major releases?

everzet:

Issue #976, opened 2016-12-23
Apostrophes causing unexpected capitals in snippets


WillGibson:
When we have steps like...

```
Given the Client's Account is Suspended
````

...the suggested snippets some out with a capital S in the method names where we would expect a lower one...

```
public function theClientSAccountIsSuspended()
```

...instead of this, which we would expect...

```
public function theClientsAccountIsSuspended()
```

Not a big deal really, but it is triggering some anal retentive tremors in myself and a colleague, so I will probably try to find time to make a PR to change this behaviour in the new year.

**Comment by WillGibson** [comment]
Hi @everzet,

I just started to prepare for this and figured a good first step would be to check all the Behat tests pass before I do anything so I know for sure if I broke stuff :-)

Mostly they do but I'm getting some fails like on...

```
LANG=C bin/behat features/extensions.feature:127
```

...which is failing because of some extra output at my end not being what the test is expecting...

```
Failed asserting that 'FF

      --- Failed steps:

      001 Scenario:                  # features/extensions.feature:2
            Given non-existent class # features/extensions.feature:3
              Fatal error: Class 'Non\Existent\Cls' not found (Behat\Testwork\Call\Exception\FatalThrowableError)

      002 Scenario:                   # features/extensions.feature:4
            Given non-existent method # features/extensions.feature:5
              Fatal error: Call to undefined method FeatureContext::getName() (Behat\Testwork\Call\Exception\FatalThrowableError)

      2 scenarios (2 failed)
      2 steps (2 failed)
      /private/var/folders/k1/n7wy5pjj069dvry_2tspxcxh0000gq/T/behat/3a3a1475b52d2e7f1f1542f8518417c5/custom_extension.php:12:
      string(16) "Non\Existent\Cls"
      /private/var/folders/k1/n7wy5pjj069dvry_2tspxcxh0000gq/T/behat/3a3a1475b52d2e7f1f1542f8518417c5/custom_extension.php:17:
      array(2) {
        [0] =>
        string(14) "FeatureContext"
        [1] =>
        string(7) "getName"
      }' contains "FF

      --- Failed steps:

      001 Scenario:                  # features/extensions.feature:2
            Given non-existent class # features/extensions.feature:3
              Fatal error: Class 'Non\Existent\Cls' not found (Behat\Testwork\Call\Exception\FatalThrowableError)

      002 Scenario:                   # features/extensions.feature:4
            Given non-existent method # features/extensions.feature:5
[...]

PR #1001, opened 2017-03-03
Feature/976 apostrophes vs capitals in snippets

WillGibson:
Addresses "Apostrophes causing unexpected capitals in snippets #976" with another PR on Behat/Transliterator (https://github.com/Behat/Transliterator/pull/18).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         "php": ">=5.3.3",
         "ext-mbstring": "*",
         "behat/gherkin": "^4.4.4",
-        "behat/transliterator": "~1.0",
+        "behat/transliterator": "^1.2",
         "symfony/console": "~2.5||~3.0",
         "symfony/config": "~2.3||~3.0",
         "symfony/dependency-injection": "~2.1||~3.0",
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -11,6 +11,8 @@ cache:
 branches:
   only:
     - master
+    # ToDo: Remove this line once all done
+    - feature/976-apostrophes-vs-capitals-in-snippets
```
Taluu: you actually don't need that line, PRs are automatically tested

WillGibson: Ah, yes, it wasn't a PR before. I'll remove that.

composer.json
```diff
@@ -13,11 +13,18 @@
...
+            "type": "git",
+            "url": "https://github.com/BiffBangPow/Transliterator"
+        }
+    ],
+
     "require": {
         "php": ">=5.3.3",
         "ext-mbstring": "*",
         "behat/gherkin": "^4.4.4",
-        "behat/transliterator": "~1.0",
```
WillGibson: This will need sorting out before merging.

Taluu:
Actually I am unsure of the interest of this PR, as it concerns only the pr made in the translit (this shouldn't be tested here, it works without it)

WillGibson:
@Taluu I think there would need to be a PR here because I added a scenario into features/snippets.feature because it is already testing the area of snippet generation in here, so it seemed write to test the new behaviour in the same way.

Taluu:
The generation is already tested actually, so it doesn't make sense to test every single variant on the main repo IMO. If it is tested in the translit repo, this should be enough

WillGibson:
What do you think here @stof? Do you want these tests here? Or do you agree with @Taluu?

composer.json
everzet: @WillGibson are you ok to bump it?

WillGibson: Yup, give me a few minutes...

WillGibson: @everzet That's done now, and I squashed it all into a single commit. Just waiting on the build.

Issue #999, opened 2017-02-20
PSR-11 has been accepted


Taluu:
PSR-11 has been officially accepted (cf https://github.com/php-fig/fig-standards/commit/a13a2f700d7c6b4de871fddde50b785b543e043e), so instead of having to use the [container-interop interfaces](https://github.com/container-interop/container-interop), we should use the [proper interfaces](https://github.com/php-fig/container). There is a compatibility though between the two sets of interfaces (container-interop now extends php-fig's container interfaces)

This may be considered a BC break though, so I guess it would only be for Behat 4 ?

stof:
I'm quite sure we don't need to wait until Behat 4, if we keep implement the container-interop interface on our own container for BC reasons.

We can change the check for the expected ContainerInterface to expect a PSR one. This would not be a BC break, as container-interop now extends PSR-11 (we would have to bump the min version to 1.2 of course).

Removing container-interop entirely may be harder regarding BC indeed, in case someone expects the Behat container itself being an interop one. But this is not necessary for the first step of PSR-11 support.

everzet:
Merged on master

PR #1014, opened 2017-04-03
Use psr/container

thewilkybarkid:
This replaces the now-deprecated `container-interop/container-interop` with `psr/container` in what-should-be-a BC way.

(The exceptions are now a different type, is that acceptable?)

### DIFF
--- a/src/Behat/Behat/HelperContainer/BuiltInServiceContainer.php
+++ b/src/Behat/Behat/HelperContainer/BuiltInServiceContainer.php
@@ -12,7 +12,7 @@ namespace Behat\Behat\HelperContainer;
 
 use Behat\Behat\HelperContainer\Exception\ServiceNotFoundException;
 use Behat\Behat\HelperContainer\Exception\WrongServicesConfigurationException;
-use Interop\Container\ContainerInterface;
+use Psr\Container\ContainerInterface;
 use ReflectionClass;
 use ReflectionMethod;
 
--- a/src/Behat/Behat/HelperContainer/Exception/HelperContainerException.php
+++ b/src/Behat/Behat/HelperContainer/Exception/HelperContainerException.php
@@ -11,13 +11,13 @@
 namespace Behat\Behat\HelperContainer\Exception;
 
 use Behat\Testwork\Environment\Exception\EnvironmentException;
-use Interop\Container\Exception\ContainerException;
+use Psr\Container\ContainerExceptionInterface;
 
 /**
  * All HelperContainer exceptions implement this interface.
  *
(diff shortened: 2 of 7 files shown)


src/Behat/Behat/HelperContainer/Exception/HelperContainerException.php
```diff
@@ -11,13 +11,13 @@
...
-use Interop\Container\Exception\ContainerException;
+use Psr\Container\ContainerExceptionInterface;
 
 /**
  * All HelperContainer exceptions implement this interface.
  *
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
  */
-interface HelperContainerException extends ContainerException, EnvironmentException
+interface HelperContainerException extends ContainerExceptionInterface, EnvironmentException
```
stof: to be BC, we should keep extending the interop interface for now (which extends the PSR one). Otherwise, existing try/catch may not work anymore (if they were catching the interop one)

src/Behat/Behat/HelperContainer/BuiltInServiceContainer.php
```diff
@@ -12,7 +12,7 @@
 
 use Behat\Behat\HelperContainer\Exception\ServiceNotFoundException;
 use Behat\Behat\HelperContainer\Exception\WrongServicesConfigurationException;
-use Interop\Container\ContainerInterface;
+use Psr\Container\ContainerInterface;
```
stof: our class must keep implementing the interop interface to be BC (which will make it implement the PSR one)

Taluu: Basically we need to require `container-interop` from `1.2`, which will extend the thing. Then we indeed need to check if the psr is implemented

Taluu:
see #999 and more specifically @stof's [comment](https://github.com/Behat/Behat/issues/999#issuecomment-281030020)

stof:
Here is what we need to end up with:

- we need to require interop 1.2+ rather than 1.1+ (to have the version using PSR)
- any place consuming a container should accept the PSR interface
- any place defining container-related things (the container itself and the exceptions) should use the interop class (which extends PSR) to keep backward compatibility

everzet:
Postponed till FIG actually accepts PSR-11. Currently it is in review:
http://www.php-fig.org/psr/

Taluu:
The website doesn't seem to be up to date, see https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-11-container.md

It is categorized as "accepted"

PR #1018, opened 2017-04-07
Extension Instantiators abstraction

Taluu:
See https://github.com/Behat/Behat/pull/995#discussion_r110239510 https://github.com/Behat/Behat/pull/995#issuecomment-280620803 This is basically #995 without Puli stuff

The instantiators are now separated and loaded in new array parameter `$instantiators` in the `ExtensionManager` class.

What could be done would be the following (to ease adding other instantiators) :
- Use a type of plugin loaded through composer (Thus binding even more Behat to Composer ?) ?
- Use the container to load instantiators (it is currently not the case) through a tag of some sort

### DIFF
--- a/src/Behat/Testwork/ServiceContainer/ExtensionManager.php
+++ b/src/Behat/Testwork/ServiceContainer/ExtensionManager.php
@@ -20,10 +20,6 @@ use Behat\Testwork\ServiceContainer\Exception\ExtensionInitializationException;
 final class ExtensionManager
 {
     /**
-     * @var string
-     */
-    private $extensionsPath;
-    /**
      * @var Extension[]
      */
     private $extensions = array();
@@ -34,30 +30,24 @@ final class ExtensionManager
     private $debugInformation = array(
         'extensions_list' => array()
     );
+    /**
+     * @var ExtensionInstantiator[]
+     */
+    private $instantiators = array();
 
     /**
      * Initializes manager.
      *
-     * @param Extension[] $extensions     List of default extensions
-     * @param null|string $extensionsPath Base path where to search custom extension files
+     * @param Extension[]             $extensions    List of default extensions
+     * @param ExtensionInstantiator[] $instantiators Extension instantiators to use
      */
-    public function __construct(array $extensions, $extensionsPath = null)
+    public function __construct(array $extensions, array $instantiators)
     {
         foreach ($extensions as $extension) {
             $this->extensions[$extension->getConfigKey()] = $extension;
         }
 
-        $this->extensionsPath = $extensionsPath;
-    }
-
-    /**
-     * Sets path to directory in which manager will try to find extension files.
-     *
-     * @param null|string $path
-     */
-    public function setExtensionsPath($path)
-    {
-        $this->extensionsPath = $path;
+        $this->instantiators = $instantiators;
     }
 
     /**
@@ -129,21 +119,6 @@ final class ExtensionManager
     }
 
     /**
-     * Attempts to guess full extension class from relative.
-     *
-     * @param string $locator
-     *
-     * @return string
-     */
-    private function getFullExtensionClass($locator)
-    {
-        $parts = explode('\\', $locator);
-        $name = preg_replace('/Extension$/', '', end($parts)) . 'Extension';
-
-        return $locator . '\\ServiceContainer\\' . $name;
-    }
-
-    /**
      * Initializes extension by id.
      *
      * @param string $locator
@@ -175,24 +150,16 @@ final class ExtensionManager
      */
     private function instantiateExtension($locator)
     {
-        if (class_exists($class = $locator)) {
-            return new $class;
-        }
-
-        if (class_exists($class = $this->getFullExtensionClass($locator))) {
-            return new $class;
-        }
-
-        if (file_exists($locator)) {
-            return require($locator);
-        }
-
-        if (file_exists($path = $this->extensionsPath . DIRECTORY_SEPARATOR . $locator)) {
-            return require($path);
+        foreach ($this->instantiators as $instantiator) {
+            try {
+                return $instantiator->instantiate($locator);
+            } catch (ExtensionInitializationException $e) {
+                // ignored if the instantiator does not support the locator
+            }
         }
 
         throw new ExtensionInitializationException(sprintf(
-            '`%s` extension file or class could not be located.',
+            '`%s` extension could not be located.',
             $locator
         ), $locator);
     }
--- a/src/Behat/Testwork/ServiceContainer/ExtensionManager.php
+++ b/src/Behat/Testwork/ServiceContainer/ExtensionManager.php
@@ -40,14 +40,16 @@ final class ExtensionManager
     /**
      * Initializes manager.
      *
-     * @param Extension[]                         $extensions      List of default extensions
-     * @param ExtensionInstantiator[]|string|null $instantiators   Extension instantiators to use
+     * @param Extension[]                         $extensions              List of default extensions
+     * @param ExtensionInstantiator[]|string|null $extensionsInstantiators Extension instantiators to use
      */
-    public function __construct(array $extensions, $instantiators = null)
+    public function __construct(array $extensions, $extensionsInstantiators = null)
     {
         foreach ($extensions as $extension) {
             $this->extensions[$extension->getConfigKey()] = $extension;
         }
+
+        $instantiators = $extensionsInstantiators;
 
         // BC Layer, using default instantiators
         // to be removed in 4.0
--- a/src/Behat/Testwork/ServiceContainer/ExtensionManager.php
+++ b/src/Behat/Testwork/ServiceContainer/ExtensionManager.php
@@ -10,6 +10,8 @@
 
 namespace Behat\Testwork\ServiceContainer;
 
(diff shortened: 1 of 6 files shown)


Taluu:
The failure on scrutinizer is due to him going haywire on a bc compat feature (if the instantiators are empty, defining some new ones)

src/Behat/Testwork/ServiceContainer/ExtensionManager.php
```diff
@@ -34,30 +32,50 @@
...
 
-        $this->extensionsPath = $extensionsPath;
+        // BC Layer for `extensionsPath`
+        // to be removed in 4.0
+        if (empty($instantiators)) {
+            $instantiators = array(
+                new ExtensionClassInstantiator(),
+                new ExtensionFileInstantiator($extensionsPath),
+            );
+        }
```
Taluu: could be written as :

```php
$this->instantiators = empty($instantiators)
    ? array(new ExtensionClassInstantiator(), new ExtensionFileInstantiator($extensionsPath))
    : $instantiators;
```

But IMO this is not elegant (that's pretty subjective), even though it would comply with scrutinizer...

src/Behat/Testwork/ServiceContainer/ExtensionManager.php
```diff
@@ -34,30 +32,50 @@
...
      *
      * @param null|string $path
+     * @deprecated since 3.4, to be removed in 4.0
+     *  (Set the extensions path through the proper instantiator instead)
      */
     public function setExtensionsPath($path)
     {
-        $this->extensionsPath = $path;
+        foreach ($this->instantiators as $instantiator) {
+            if ($instantiator instanceof ExtensionFileInstantiator) {
```
Taluu: instead of testing if the instantiator is the `ExtensionFileInstantiator`, maybe testing if the instantiator has a method `setExtensionsPath` would be better ?

src/Behat/Testwork/ServiceContainer/Instantiator/ExtensionClassInstantiator.php
```diff
@@ -0,0 +1,51 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Behat\Testwork\ServiceContainer\Instantiator;
+
+use Behat\Testwork\ServiceContainer\ExtensionInstantiator;
+use Behat\Testwork\ServiceContainer\Exception\ExtensionInitializationException;
+
+final class ExtensionClassInstantiator implements ExtensionInstantiator
```
everzet: Missing description. Focus on differences

src/Behat/Testwork/ServiceContainer/Instantiator/ExtensionFileInstantiator.php
```diff
@@ -0,0 +1,56 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Behat\Testwork\ServiceContainer\Instantiator;
+
+use Behat\Testwork\ServiceContainer\ExtensionInstantiator;
+use Behat\Testwork\ServiceContainer\Exception\ExtensionInitializationException;
+
+final class ExtensionFileInstantiator implements ExtensionInstantiator
```
everzet: Missing description. Focus on differences

Taluu:
Docblocks added @everzet

everzet:
So how are you planning to use this extension point? Create your own custom binstub?

everzet:
Ah. I missed the options you outlined in the description.

Taluu:
Using a custom binstub could be a first solution before having a real solution (it wouldn't really change what is currently done, as there are no extension points at all). But meh, I'm not a fan of this method...

stof:
Using the container to load instantiators for extensions looks weird to me, as extensions are precisely responsible for building the container, so the container is not usable at the time extensions are instantiated

src/Behat/Testwork/ServiceContainer/ExtensionManager.php
```diff
@@ -34,30 +32,46 @@
...
+            ? array(new ExtensionClassInstantiator(), new ExtensionFileInstantiator($extensionsPath))
+            : $instantiators
+        ;
     }
 
     /**
      * Sets path to directory in which manager will try to find extension files.
      *
      * @param null|string $path
+     * @deprecated since 3.4, to be removed in 4.0
```
stof: I suggest adding deprecation warnings using the Symfony convention (i.e. `@trigger_error('...', E_USER_DEPRECATED);`) in deprecated APIs.

Behat does not provide reporting of usage of deprecated APIs yet (I think there is a third-party extension doing it though), but it may still help people using  Behat programmatically (which are the only ones caring about it, as the built-in usage of Behat should not trigger deprecated APIs for the ExtensionManager)

src/Behat/Testwork/ServiceContainer/ExtensionManager.php
stof: as it is a deprecated API anyway, this is fine

src/Behat/Testwork/ServiceContainer/ExtensionManager.php
```diff
@@ -175,20 +174,12 @@ private function initialize($locator)
...
-            return require($path);
+        foreach ($this->instantiators as $instantiator) {
+            try {
+                return $instantiator->instantiate($locator);
+            } catch (ExtensionInitializationException $e) {
+                // ignored if the instantiator does not support the locator
+            }
         }
 
         throw new ExtensionInitializationException(sprintf(
```
stof: The final exception should be changed, as it would be wrong for custom instantiator.

You may want to build the message based on exception messages provided by each instantiator.

src/Behat/Testwork/ServiceContainer/ExtensionManager.php
```diff
@@ -34,30 +32,46 @@
...
      * Initializes manager.
      *
-     * @param Extension[] $extensions     List of default extensions
-     * @param null|string $extensionsPath Base path where to search custom extension files
+     * @param Extension[]             $extensions      List of default extensions
+     * @param string|null             $extensionsPath  Base path where to search custom extension files (deprecated)
+     * @param ExtensionInstantiator[] $instantiators   Extension instantiators to use
      */
-    public function __construct(array $extensions, $extensionsPath = null)
+    public function __construct(array $extensions, $extensionsPath = null, array $instantiators = array())
```
stof: the `extensionsPath` argument should be deprecated.

So instantiators should be passed as second argument in the non-deprecated signature. The BC layer would enter into action when it is not an array.

your current logic would force keeping the `$extensionsPath` argument in 4.0 when removing the BC layer.

src/Behat/Testwork/ServiceContainer/ExtensionManager.php
```diff
@@ -34,30 +32,46 @@
...
     {
         foreach ($extensions as $extension) {
             $this->extensions[$extension->getConfigKey()] = $extension;
         }
 
-        $this->extensionsPath = $extensionsPath;
+        // BC Layer, using default instantiators
```
stof: We need a deprecation warning when reaching the Bc layer IMO.

src/Behat/Testwork/ServiceContainer/ExtensionManager.php
Taluu: Not sure if building the last exception from the other exceptions is relevant, as these are just "failures". Removing the `class or file` part looks like a good idea, to stay generic.

What could be done would be to list which instantiators were tried ?

src/Behat/Testwork/ServiceContainer/ExtensionManager.php
Taluu: done

src/Behat/Testwork/ServiceContainer/ExtensionManager.php
Taluu: done

src/Behat/Testwork/ServiceContainer/ExtensionManager.php
```diff
@@ -34,30 +32,60 @@
...
+     * @deprecated since 3.4, to be removed in 4.0
+     *  (Set the extensions path through the proper instantiator instead)
      */
     public function setExtensionsPath($path)
     {
-        $this->extensionsPath = $path;
+        @trigger_error(
+            sprintf(
+                '%s is deprecated since 3.4, and will be removed in 4.0. Use the proper instantiator instead.',
+                E_USER_DEPRECATED
```
stof: this must not be inside the sprintf. Your code is broken. what should be inside the sprintf is `__FUNCTION__`.

Taluu: my bad, fixed.

ciaranmcnulty:
@Taluu I can't remember what this is about; are you still keen on it being merged?

Taluu:
IIRC, the goal would be to have a sort of "extensions loader" which doesn't need to hack behat. But to do that, we need to be able to make the extension loaders abstract instead of being hardcoded in Behat's core.

I'd be willing to contribute to it if you think the idea is still relevant.

ciaranmcnulty:
I'm not even sure I really wanted this feature; I'm going to close for now but once the project's tidied up please feel free to re-propose

Issue #1008, opened 2017-03-20
Argument resolver doesn't find matching arguments


elvetemedve:
# The bug
Let's  suppose there are `Api` and `FakeApi` **classes**, where  `FakeApi extends Api`.
And I want Behat to create my context class with these constructor arguments:
```php
class MyContext implements Context
{
    public __construct(Api $api, FakeApi $fakeApi) {}
}
```
In this case the following error will be reported:

>  [Behat\Testwork\Argument\Exception\UnknownParameterValueException]                                                        
>   Can not find a matching value for an argument `$fakeApi` of the method `MyContext::__construct()`.

## Explanation
That happens because `MixedArgumentOrganiser` try to find which parameter to assign by [checking that the parameter is an instance of the argument](https://github.com/Behat/Behat/blob/master/src/Behat/Testwork/Argument/MixedArgumentOrganiser.php#L145-L150). Since both the `Api` instance and `FakeApi` instance are instances of `Api`, argument resolver will associate these parameters with the `$api` argument.

## Workaround
As a possible solution, I removed type hinting to make argument resolver work by argument numbers.

everzet:
Thanks for reporting it.

glennmcewan:
Using the `Symfony2Extension`, the easiest way would be to name the parameters you're passing to the `Context`.

The issue in this case would be an easy* fix, by checking which one has a more direct relationship to each argument (for example, using `get_class` to check if `FakeApi` is equal, where `Api` would fail, or by climbing the inheritance chain to find which service is the closest relative).

\* I say easy, because it then becomes tricky when you have two arguments which ask for the same interface -- implemented by two different classes -- to be injected. How on earth do you know which you want in that scenario? It is also tricky when you have two arguments which request different interfaces, and each service candidate satisfies both interfaces.

I've had a toy around with this and in this case where you're asking for `FakeApi` which is a child of `Api`, it can be solved by adding a new method to `MixedArgumentOrganiser` like `prepareTypehintedArguments` similar to the other methods, and then iterating each argument and matching the best fitting candidate. I'll make a PR as a POC shortly as I've walked in to the same issue as this…

glennmcewan:
@everzet @elvetemedve delay in making the PR due extenuating circumstances, but I have created it here: behat/behat#1028

[...]

PR #1028, opened 2017-05-04
Fix Parent / Child class argument ambiguity issue with MixedArgumentResolver

glennmcewan:
This PR updates the logic of the `MixedArgumentResolver@prepareTypehintedArguments` method to resolve an issue where a constructor asks for two typehinted arguments where one of the arguments is interchangeable with the other. A parent / child class relationship will cause this error.

See the original issue: behat/behat#1008

~~I have not added in any tests, as the unit tests were failing locally when I ran them on a fresh clone of master.~~
I have now added in a test case for the bug that I have fixed.

### DIFF
--- a/src/Behat/Testwork/Argument/MixedArgumentOrganiser.php
+++ b/src/Behat/Testwork/Argument/MixedArgumentOrganiser.php
@@ -199,13 +199,23 @@ final class MixedArgumentOrganiser implements ArgumentOrganiser
 
         $candidates = $typehintedArguments;
 
-        $this->applyPredicateToTypehintedArguments($parameters, $candidates, $arguments, 'classMatchingPredicateForTypehintedArguments');
+        $this->applyPredicateToTypehintedArguments(
+            $parameters,
+            $candidates,
+            $arguments,
+            [$this, 'classMatchingPredicateForTypehintedArguments']
+        );
 
         // This iteration maps up everything else, providing the argument is an instanceof the parameter.
-        $this->applyPredicateToTypehintedArguments($parameters, $candidates, $arguments);
+        $this->applyPredicateToTypehintedArguments(
+            $parameters,
+            $candidates,
+            $arguments,
+            [$this, 'isInstancePredicateForTypehintedArguments']
+        );
 
         return $arguments;
-        }
+    }
 
     /**
      * Applies a predicate for each candidate when matching up typehinted arguments.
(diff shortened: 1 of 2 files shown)


src/Behat/Testwork/Argument/MixedArgumentOrganiser.php
```diff
@@ -112,26 +112,26 @@ private function isStringKeyAndExistsInParameters($argumentKey, $parameterNames)
...
-     * @param ReflectionParameter[] $parameters
-     * @param mixed                 $value
+     * @param  ReflectionParameter[] $parameters
+     * @param  mixed                 $value
      *
-     * @return null|integer
+     * @return Boolean
      */
-    private function getParameterNumberWithTypehintingValue(array $parameters, $value)
+    public function isParameterTypehintedInArgumentList(array $parameters, $value)
```
elvetemedve: Why did you make this method public? I think it should remain private.

src/Behat/Testwork/Argument/MixedArgumentOrganiser.php
```diff
@@ -174,27 +174,78 @@ private function prepareNamedArguments(array $parameters, array $namedArguments)
...
             $this->markArgumentDefined($num);
+
+                    unset($candidates[$candidateIndex]);
+
+                    break 1;
         }
+            }
+        }
+
+        foreach ($parameters as $num => $parameter) {
```
elvetemedve: This foreach block looks duplicated to me, the only difference is the `if ($reflectionClass->getName() === get_class($candidate))` statement. It would be nice if you could eliminate this repetition.

glennmcewan: I noticed this as Scrutinizer was complaining about the repetition, but I don't know which pattern would be good to eliminate this repetitiveness without adding in a lot of superfluous code (visitor pattern or something).

src/Behat/Testwork/Argument/MixedArgumentOrganiser.php
glennmcewan: Ah, it was a new method that I added in, automatically went as public. I had forgotten to check the visibility of other methods. I have updated this one.

src/Behat/Testwork/Argument/MixedArgumentOrganiser.php
glennmcewan: I've refactored it to remove repetition, but now Scrutinizer thinks the method is rated `D`, and after quite a bit more, it is now `B`.

glennmcewan:
The  Scrutinizer rating is now stuck at `B`. It would be silly to even "simplify" this method, as there is already quite a superfluous chain of simple methods.

elvetemedve:
@glennmcewan I think the refactoring you made is good enough. Applying a complex design pattern like the visitor would make it over-engineered in my opinion.

The class is rated to **B**, because one of the method is **B**, particularly `applyPredicateToTypehintedArguments()`. If you [take a look at the report](https://scrutinizer-ci.com/g/Behat/Behat/inspections/b9764823-9872-4813-8f9b-6566f686c752/code-structure/operation/Behat%5CTestwork%5CArgument%5CMixedArgumentOrganiser%3A%3AapplyPredicateToTypehintedArguments), you can see that **Total Lines** is a bit high. Try to remove all the empty lines and see if Scrutinizer likes it.

glennmcewan:
@elvetemedve I agree, implementing the visitor pattern would be an unnecessary way of doing it, but I still feel like the refactoring I have done now is a bit excessive.

I don't know what Scrutinizer's rating algorithm is, but total lines of 26 in a low complexity method shouldn't mark it down to a **B** _imo_. I just had a quick test, and it didn't make a difference. I think removing the inner loop did the trick as that removes paths / complexity. Overkill imo, but if that makes the builds pass...

Anyway, Scrutinizer is now happy, Behat tests are written, I think this is ready :)

elvetemedve:
@glennmcewan I agree that number of lines should be okay as well as a nested `foreach`. Scrutinizer is too strict in this case. But it's not my call to decide. :)

everzet:
Thank you very much for a great work! I did some very small refactoring and merged into master.

Issue #1043, opened 2017-05-25
Behat fails to parse individual scenarios when relative paths are used


deviantintegral:
If you try to run an individual scenario, it only works if you are running form the root of your project. Otherwise, you can pass a directory or feature file only. If you convert the path to the features to be an absolute path, it works fine. For example:

```
$ ../../vendor/bin/behat -c tests/src/Behat/behat.local.yml $(pwd)/tests/src/Behat/features/testing.feature:3
```

works fine, but if I remove `$(pwd)/` behat fails with:

```
$ ../../vendor/bin/behat -c tests/src/Behat/behat.local.yml tests/src/Behat/features/testing.feature:3
  [Behat\Testwork\Tester\Exception\WrongPathsException]
  No specifications found at path(s) `tests/src/Behat/features/testing.feature:3`.
```

spolischook:
I guess behat try to find your feature from the path of your config. Try:
```bash
bin/behat -c tests/src/Behat/behat.local.yml features/testing.feature:3
```
I suggest use composer bin path, see https://getcomposer.org/doc/articles/vendor-binaries.md

deviantintegral:
That works, but only if I move features up to the root of my project. This is for a Drupal module, where the tests are owned by the module (as there isn't a "site" to test against).

Is there any reason why Behat couldn't convert the relative paths to absolute paths when referencing them internally?

spolischook:
Is this issue same for feature running? or it happens only for scenario run?

glennunipro:
This is a bug which I have found to be caused by the `GherkinExtension` class; it does not give the basePath to each of its loaders, so the loaders are not aware of what the base path is.

This works when you're already in the root, because `is_file` will return true (because the relative path is equal to the absolute path, so it doesn't need to go down the route of looking in the base path).

So, I've created a PR [here](1046) which resolves this, feel free to give it a try and let me know how that works.

glennunipro:
Turns out that in my previous comment, the solution was completely wrong. It did fix the case, when running Behat in this scenario, but it broke the unit tests spectacularly (due to the way the Behat self tests work, writing feature files to a temporary folder, outside of the base path).

The solution instead is to "sanitise" the given path in the FilesystemFeatureLocator, so that it strips out the line ranges from the path before searching for the file in the filesystem.

I've updated my original PR, builds now passing (except one code style recommendation). If you check out that branch into your project, it should resolve your issue @deviantintegral.

PR #1046, opened 2017-05-30
Allow Behat to find feature files with specified line / range when not in the root folder

glennunipro:
This is to resolve the bug raised in #1043.

In the scenario where you are running a specific Behat scenario whilst not in the project root (where `behat.yml` lives), Behat trips up as it cannot find the specified file, even if it has been passed properly.

To replicate the bug on master, change in to a sub directory, and call a specific scenario in a given feature file, example:
```
cd foo
../vendor/bin/behat -c ../behat.yml features/acme.feature:9
```

When running this, Behat will only look for `features/acme` in the `foo` directory, it will not try and look in the base path because the Gherkin loaders have not been told what the base path is, which causes a failure in the `behat\gherkin` repository (the `Behat\Gherkin\Loader\AbstractFileLoader@findAbsolutePath` method).

One line change to resolve it.

### DIFF
--- a/src/Behat/Behat/Gherkin/ServiceContainer/GherkinExtension.php
+++ b/src/Behat/Behat/Gherkin/ServiceContainer/GherkinExtension.php
@@ -225,6 +225,7 @@ final class GherkinExtension implements Extension
         }
 
         $definition->addMethodCall('setCache', array($cacheDefinition));
+        $definition->addMethodCall('setBasePath', array('%paths.base%'));
         $definition->addTag(self::LOADER_TAG, array('priority' => 50));
         $container->setDefinition('gherkin.loader.gherkin_file', $definition);
     }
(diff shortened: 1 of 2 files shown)


deviantintegral:
Looks like this isn't working for the case where features are in a subdirectory:

Works:

```
$ ../../vendor/bin/behat -c tests/src/Behat/behat.local.yml tests/src/Behat/features
```

and this one:

```
$ ../../vendor/bin/behat -c tests/src/Behat/behat.local.yml tests/src/Behat/features/media-settings.feature
```

but not with a single scenario (even though the prior command confirmed the line number):

```
$ ../../vendor/bin/behat -c tests/src/Behat/behat.local.yml tests/src/Behat/features/media-settings.feature:33

 [Behat\Testwork\Tester\Exception\WrongPathsException]
  No specifications found at path(s) `tests/src/Behat/features/media-settings.feature:33`.

```

glennunipro:
@deviantintegral I'm going to close this; I have had a similar issue when running tests for Drupal modules and the fix for it seems to be different to your case.

I'm unable to replicate your scenario, but if you make a public repo replicating the issue, I can take a look.

Providing that the path to the feature is relative to the **config** file, this should work nicely.

deviantintegral:
Unfortunately, I get the same exception with this patch even when I use paths relative to the feature file, like `features/media-settings.feature`.

This particular example is in private code, but I think we'll be working on behat tests for the Facebook Instant Articles module this week.

PR #1053, opened 2017-06-26
Fix to inability to use LineFilter and LineRangefilter Filters

generalconsensus:
This PR fixes an issue where the user provides a line for a specific scenario(s). The AbstractFileLoader class doesn't find the basePath because FilesystemFeatureLocator doesn't set the $basePath via the _construct. It only sets the $basePath for the FilesystemFeatureLocator. Not sure if this is the best place for this. Without it any command for scenarios by line number fail due to the Gherkin class not having a basePath set.
````
bin/behat --suite=default features/donate.feature:61

### DIFF
--- a/src/Behat/Behat/Gherkin/Specification/LazyFeatureIterator.php
+++ b/src/Behat/Behat/Gherkin/Specification/LazyFeatureIterator.php
@@ -64,13 +64,15 @@ final class LazyFeatureIterator implements SpecificationIterator
      * @param Gherkin           $gherkin
      * @param string[]          $paths
      * @param FilterInterface[] $filters
+     * @param string            $basePath
      */
-    public function __construct(Suite $suite, Gherkin $gherkin, array $paths, array $filters = array())
+    public function __construct(Suite $suite, Gherkin $gherkin, array $paths, array $filters = array(), $basePath = '')
     {
         $this->suite = $suite;
         $this->gherkin = $gherkin;
         $this->paths = array_values($paths);
         $this->filters = array_merge($this->getSuiteFilters($suite), $filters);
+        $gherkin->setBasePath($basePath);
     }
 
     /**
--- a/src/Behat/Behat/Gherkin/Specification/Locator/FilesystemFeatureLocator.php
+++ b/src/Behat/Behat/Gherkin/Specification/Locator/FilesystemFeatureLocator.php
@@ -77,7 +77,7 @@ final class FilesystemFeatureLocator implements SpecificationLocator
         if ($locator) {
             $filters = array(new PathsFilter($suiteLocators));
 
-            return new LazyFeatureIterator($suite, $this->gherkin, $this->findFeatureFiles($locator), $filters);
+            return new LazyFeatureIterator($suite, $this->gherkin, $this->findFeatureFiles($locator), $filters, $this->basePath);
         }
 
         $featurePaths = array();
@@ -85,7 +85,7 @@ final class FilesystemFeatureLocator implements SpecificationLocator
             $featurePaths = array_merge($featurePaths, $this->findFeatureFiles($suiteLocator));
         }
 
-        return new LazyFeatureIterator($suite, $this->gherkin, $featurePaths);
+        return new LazyFeatureIterator($suite, $this->gherkin, $featurePaths, [], $this->basePath);
     }
 
     /**

### CODE REVIEW

**Comment by everzet** [comment]
Can you provide a test case?

**Comment by stof** [comment]
I'm -1 for having the locator altering the state of the Gherkin object. This is likely to create weird edge cases. The right path should be set when instantiating the Gherkin object instead.

**Comment by glennunipro** [comment]
@generalconsensus you may want to look at PR #1046 (and look through the individual commits, to see my approach and how it changed as I learnt more); I tried resolving this previously, and yes I tried the base path but this causes the Behat self-tests to fail, because of the way they search for features.

There are a multitude of ways to resolve this -- one of which is quite sizeable, and potentially a BC break; it will require having a `%paths.features%` parameter, so that you can change the _feature_ path, but leave the *base* path alone. I feel that is a good feature to add, as the features path doesn't need to be in the root.

@stof see commit here: https://github.com/Behat/Behat/pull/1046/commits/fb06d564c18ee3bc1723fe0935bd8406b7c16ee0
That's how I set the base path on to Gherkin, which is a bit more correct, but it ultimately runs in to the same issue.

This will need some more thought / investigation, unless I am wildly over-complicating this.

@everzet this typically happens when you run `behat` on a specific line, when you are *not* in the base path. Unsure how to write a test case for this...the Behat self-tests break when changing the Gherkin base path (which should be the feature path). Perhaps the logic in https://github.com/Behat/Behat/blob/master/features/bootstrap/FeatureContext.php needs reviewing...I had a look but was unsure about it.

**Comment by generalconsensus** [comment]
Ok I've left a pr on the Gherkin repo: https://github.com/Behat/Gherkin/pull/125. That fixes the bug that fails to find the relative file.

File: src/Behat/Behat/Gherkin/Specification/Locator/FilesystemFeatureLocator.php [hunk 1]
```diff
@@ -77,15 +77,15 @@ public function locateSpecifications(Suite $suite, $locator)
         if ($locator) {
             $filters = array(new PathsFilter($suiteLocators));
 
-            return new LazyFeatureIterator($suite, $this->gherkin, $this->findFeatureFiles($locator), $filters);
+            return new LazyFeatureIterator($suite, $this->gherkin, $this->findFeatureFiles($locator), $filters, $this->basePath);
```
glennunipro: Note, you're passing in only **4** arguments here. The constructor expects **5**, you'll want to include the `$filters` parameter here.

src/Behat/Behat/Gherkin/Specification/Locator/FilesystemFeatureLocator.php
```diff
@@ -77,15 +77,15 @@ public function locateSpecifications(Suite $suite, $locator)
...
         }
 
         $featurePaths = array();
         foreach ($suiteLocators as $suiteLocator) {
             $featurePaths = array_merge($featurePaths, $this->findFeatureFiles($suiteLocator));
         }
 
-        return new LazyFeatureIterator($suite, $this->gherkin, $featurePaths);
+        return new LazyFeatureIterator($suite, $this->gherkin, $featurePaths, $this->basePath);
```
glennunipro: Note, you're passing in only *4* arguments here. The constructor expects *5*, you'll want to include the filters parameter here.

glennunipro:
I would recommend setting Gherkin's base path when it is registered in to the container via the Extension. Feels a bit more 'expected' that way, imo. Below is a commit where I have done it, on a closed PR which was not merged:

https://github.com/Behat/Behat/commit/fb06d564c18ee3bc1723fe0935bd8406b7c16ee0

Also left a review on your latest commits; that's why your builds are failing. Try run the PHPUnit tests locally if possible; saves Travis having to run them etc.

generalconsensus:
@glennunipro Sir, thank you for your help. I will admit my intention is just to fix not to extend or change functionality. These locator examples below worked back with Behat 2.x as far as I remember. Having just built a fresh Behat 3.3.1 instance I was disappointed to find that this functionality no longer working.  I'm only patching to get things working. I don't know about setting base path through extension. I leave that to core contributors to decide on how best to execute. 

```
            "a scenario at specific line <comment>(*.feature:10)</comment>.",
            "all scenarios at or after a specific line <comment>(*.feature:10-*)</comment>.",
            "all scenarios at a line within a specific range <comment>(*.feature:10-20)</comment>."
```

glennunipro:
@generalconsensus no worries, I have ran in to the same problem previously, which is why I have the examples to hand :)

I think that setting the base path is the correct way to resolve it, however this causes a failure with Behat's self tests. I couldn't quite figure it out, but it is down to the way the `FeatureContext` sets the feature path (random temporary location, for each scenario). I have a feeling that the Extension is not receiving the new base path each time, but was unable to prove that.

glennunipro:
Narrowed it down a little; need to amend the logic in `Behat\Gherkin\Node\FeatureNode`, because `Behat\Gherkin\Filter\PathsFilter` is given an array of paths, which the FeatureNode's file path should reside within (filters out any given features where the `FeatureNode@getFile` method does not equal the _start_ of any string paths that it is given to filter against). `Behat\Gherkin\Filter\PathsFilterTest` specifically tests that it will not match for partial paths, so that tells me `FeatureNode` **should** always be given an absolute path for the file property.

To do this, I think the solution is to update the way the Gherkin Loaders create the `FeatureNode`s in the `@load` method so that they are given an absolute path, rather than a filename (which is somewhat inaccurate, as if the base path is not set, the 'filename' will be `features/xyz.feature` (not a filename, that's a relative path).

I'll have a look at options with this, would be handy to get some feedback from @everzet / @stof to make sure this is not taking the wrong approach.

generalconsensus:
@glennunipro Thanks for taking so much time on this. I will see if I can help by this afternoon.

glennunipro:
No worries; it's an issue I have ran in to before, dropped it as it seemed to take an unnecessary amount of effort. A fresh look over it revealed a few new ideas.

glennunipro:
Got it working locally, PR raised #1056. Failing CI as it cannot pull in my fork of Gherkin (which has another change, and a PR raised).

May need review as to whether or not the change is correct in the Gherkin repo.

src/Behat/Behat/Gherkin/Specification/LazyFeatureIterator.php
```diff
@@ -64,13 +64,15 @@
...
+     * @param string            $basePath
      */
-    public function __construct(Suite $suite, Gherkin $gherkin, array $paths, array $filters = array())
+    public function __construct(Suite $suite, Gherkin $gherkin, array $paths, array $filters = array(), $basePath = '')
     {
         $this->suite = $suite;
         $this->gherkin = $gherkin;
         $this->paths = array_values($paths);
         $this->filters = array_merge($this->getSuiteFilters($suite), $filters);
+        $gherkin->setBasePath($basePath);
```
stof: this should not alter the Gherkin object it received as argument. Change the service definition of the gherkin service instead.

glennunipro:
@stof I'm not sure if this PR is relevant now that #1056 has been merged in, which was my attempt at resolving the same issue.

@generalconsensus did #1056 fix your case too?

stof:
#1056 is doing exactly what I was asking there. I forgot about it when reviewing this PR yesterday. I'm quite sure this PR is unnecessary now.

everzet:
Closing then. @generalconsensus feel free to reopen if you still experience the issue.

PR #1054, opened 2017-06-30
Explicit PSR-11 support

everzet:


Taluu:
Hum, I think like #1014, there are several bc breaks. See the comments @stof made (It was on thing implementing / extending the interop interfaces, now implementing the psr's ones ; one should expect that these classes / exceptions accept the typehinting from the interop, but they don't).

Only the typehints should be changed to be compatible with psr 11.

Unless I'm reading and understanding it wrongly...

everzet:
@Taluu @stof `container-interop` was updated with a minor to migrate to PSR-11. In 1.2 they basically [extend](https://github.com/container-interop/container-interop/blob/master/src/Interop/Container/ContainerInterface.php#L13) all the base PSR-11 interfaces.

There's also no public API change, so for users upgrade from Interop 1.1 to Interop 1.2 (which extends PSR-11) is non-breaking. This all means that for people depending on Interop container interfaces the change should be transparent. Unless they are locking particular minor version of Interop in their `composer.json`, which there's no reason to.

everzet:
@Taluu @stof I specified explicit conflict with `container-interop` prior to `composer.json`, which would explicitly force Composer to update `container-interop` to 1.2. Unless users locked their `container-interop` version, in which case they would be forced to unlock it.

Taluu:
The mentionned BC break is for example on your exceptions. Before the PSR-11 implementation, even though it is using `interop > = 1.2`, one would for example try to catch a `Interop\NotFoundException`, while now only a `Psr\Container\NotFoundException` (the names are redacted, I'm too lazy to llok for the real names), which would now be incorrect.

@stof's point was that the thing in behat that extends / implements something from interop AND/OR PSR11 should still accept / implement the interop interfaces, so that nothing breaks.

everzet:
@Taluu you are right in regards to exceptions. Switched all internal container classes back to extend from `Interop-Container`.

Issue #1035, opened 2017-05-16
PHP 5.3 Short Array Syntax In MixedArgumentOrganiser.php


alexfinnarn:
I see that you changed some code to use short array syntax and then set that back in commits like https://github.com/Behat/Behat/commit/a02da149a4a4bcf2de279429e762b110b6b21cee. 

I just ran into a case where https://github.com/Behat/Behat/blob/d7cc1409e17824d3041fab6cd89bb623f5d9caf8/src/Behat/Testwork/Argument/MixedArgumentOrganiser.php#L239 ended up causing a test run to error. 

If you are in the mood to revert back to PHP 5.3 syntax there, then go nuts. Otherwise, you can ignore or close this issue.

everzet:
Missed this one :/

PR #1068, opened 2017-08-31
Add PHP 5.3 to travis

caciobanu:
This will catch issues like #1035.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: php
 
-php: [5.4, 5.5, 5.6, 7.0, 7.1, hhvm]
+php: [5.3, 5.4, 5.5, 5.6, 7.0, 7.1, hhvm]
 
 sudo: false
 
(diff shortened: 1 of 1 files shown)


caciobanu:
Build is failing due to short array syntax. Do you want me to fix those also, in this PR ?

Taluu:
It was intentionnally removed from the test matrix in https://github.com/Behat/Behat/commit/f3e1481b6fca205ec9e0d1242be3d49af538624d and https://github.com/Behat/Behat/commit/cf47261ea5a2e7376dafa307b3783b8768795fca

Taluu:
And the ticket you are mentionning is not related (at all) to php 5.3. The short array syntax is not used because of consistency across all usages in the project.

caciobanu:
So 5.3 is not supported anymore? If yes maybe composer.json should be updated?

everzet:
We didn't explicitly drop the support, but we stopped bothering with it as much. Version was dropped from the matrix because newer Travis VMs don't even have 5.3.

Bottom line: Behat should still work on PHP 5.3, but it might get broken from version to version (unintentionally). Next major will kill support altogether.

PR #1069, opened 2017-08-31
Rework argument validators

everzet:
This is primarily a refactoring extracting argument validation routine from inside the argument organiser. But it has a side-benefit where we now will have proper argument validation for transformations and step definitions.

This refactoring, as always, is a pre-work for a bigger feature to come.

### DIFF
--- a/src/Behat/Behat/Context/ContextFactory.php
+++ b/src/Behat/Behat/Context/ContextFactory.php
@@ -102,19 +102,21 @@ final class ContextFactory
      */
     private function resolveArguments(ReflectionClass $reflection, array $arguments, array $resolvers)
     {
+        $newArguments = $arguments;
+
         foreach ($resolvers as $resolver) {
-            $arguments = $resolver->resolveArguments($reflection, $arguments);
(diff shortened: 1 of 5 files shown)


src/Behat/Behat/Context/ContextFactory.php
```diff
@@ -43,6 +48,7 @@
     public function __construct(ArgumentOrganiser $argumentOrganiser)
     {
         $this->argumentOrganiser = $argumentOrganiser;
+        $this->validator = new Validator();
```
Taluu: shouldn't you inject this as constructor argument ?

everzet: No. I don't envision any other implementation of Validator at the moment. I also don't see it becoming dependent on any external service.

In this regard there's no real difference between

```php
public function __construct(...)
{
    $this->validator = new Validator();
}
```

and

```php
public function __construct(..., Validator $validator = null)
{
    $this->validator = $validator ?: new Validator();
}
```

I'd rather not expose any extra dependencies without good reason yet.

PR #1070, opened 2017-09-01
Hold helper container on instances of Environment

everzet:
This is a second refactoring to enable us a bigger feature in upcoming release. However, despite aiming at supporting a bigger feature this refactoring brings value on its own - it exposes your preconfigured service-container via `Environment::getServiceContainer()`. Extension creators can use that to their own benefit.

### DIFF
--- a/src/Behat/Behat/Context/Environment/Handler/ContextEnvironmentHandler.php
+++ b/src/Behat/Behat/Context/Environment/Handler/ContextEnvironmentHandler.php
@@ -10,10 +10,12 @@
 
 namespace Behat\Behat\Context\Environment\Handler;
 
+use Behat\Behat\Context\Argument\ArgumentResolver;
 use Behat\Behat\Context\Argument\SuiteScopedResolverFactory;
 use Behat\Behat\Context\Argument\NullFactory;
 use Behat\Behat\Context\ContextClass\ClassResolver;
 use Behat\Behat\Context\ContextFactory;
+use Behat\Behat\Context\Environment\ContextEnvironment;
 use Behat\Behat\Context\Environment\InitializedContextEnvironment;
 use Behat\Behat\Context\Environment\UninitializedContextEnvironment;
 use Behat\Testwork\Environment\Environment;
@@ -36,7 +38,7 @@ final class ContextEnvironmentHandler implements EnvironmentHandler
      */
     private $contextFactory;
     /**
-     * @var SuiteScopedResolverFactory
+     * @var ArgumentResolver|SuiteScopedResolverFactory
      */
     private $resolverFactory;
     /**
@@ -47,10 +49,10 @@ final class ContextEnvironmentHandler implements EnvironmentHandler
     /**
      * Initializes handler.
      *
-     * @param ContextFactory             $factory
-     * @param SuiteScopedResolverFactory $resolverFactory
+     * @param ContextFactory                              $factory
+     * @param ArgumentResolver|SuiteScopedResolverFactory $resolverFactory
      */
-    public function __construct(ContextFactory $factory, SuiteScopedResolverFactory $resolverFactory = null)
+    public function __construct(ContextFactory $factory, $resolverFactory = null)
     {
         $this->contextFactory = $factory;
         $this->resolverFactory = $resolverFactory ?: new NullFactory();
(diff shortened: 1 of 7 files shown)


src/Behat/Behat/Context/Environment/Handler/ContextEnvironmentHandler.php
```diff
@@ -184,4 +186,22 @@ private function resolveClass($class)
...
+     *
+     * @return ArgumentResolver[]
+     */
+    private function createArgumentResolvers(ContextEnvironment $environment)
+    {
+        if ($this->resolverFactory instanceof ArgumentResolver) {
+            return $this->resolverFactory->createArgumentResolvers($environment);
+        } elseif ($this->resolverFactory instanceof SuiteScopedResolverFactory) {
+            return $this->resolverFactory->generateArgumentResolvers($environment->getSuite());
+        }
```
sroze: To prevent a less-likely developer WTF, I'd throw an exception here: as you don't have the type-hint anymore, somebody could inject so weird thing, that would be ignored and would end up to be no argument resolver. Throwing here if `$this->resolverFactory !== null` would highlight the issue.

src/Behat/Behat/Context/Argument/SuiteScopedResolverFactory.php
```diff
@@ -18,6 +18,8 @@
  * @see ContextEnvironmentHandler
  *
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * @deprecated since 3.4. Use `ArgumentResolverFactory` instead
```
Taluu: You should add a `trigger_error` with `E_DEPRECATED` on the file, as @stof suggested in https://github.com/Behat/Behat/pull/1018#discussion_r113420353

src/Behat/Behat/Context/Environment/Handler/ContextEnvironmentHandler.php
```diff
@@ -184,4 +186,22 @@ private function resolveClass($class)
...
+    /**
+     * Creates argument resolvers for a given environment.
+     *
+     * @param ContextEnvironment $environment
+     *
+     * @return ArgumentResolver[]
+     */
+    private function createArgumentResolvers(ContextEnvironment $environment)
+    {
+        if ($this->resolverFactory instanceof ArgumentResolver) {
```
stof: wrong check. It must be `ArgumentResolverFactory`

src/Behat/Behat/Context/Environment/Handler/ContextEnvironmentHandler.php
stof: Another option is to create an internal class implementing `ArgumentResolverFactory` by wrapping a `SuiteScopedResolverFactory` (which is easy to do).

This class would then wrap the SuiteScopedResolverFactory into this internal adapter implementation in case the provided argument does not implement the new interface but only the old one. And then, the private property can be guaranteed to always contain a `ArgumentResolverFactory`

src/Behat/Behat/Context/Environment/Handler/ContextEnvironmentHandler.php
```diff
@@ -47,10 +49,10 @@
     /**
      * Initializes handler.
      *
-     * @param ContextFactory             $factory
-     * @param SuiteScopedResolverFactory $resolverFactory
+     * @param ContextFactory                              $factory
+     * @param ArgumentResolver|SuiteScopedResolverFactory $resolverFactory
```
stof: wrong phpdoc. Should be `ArgumentResolverFactory`

src/Behat/Behat/Context/Argument/SuiteScopedResolverFactory.php
stof: ỳou cannot, as the implementation still implements it for BC

src/Behat/Behat/HelperContainer/Argument/ServicesResolverFactory.php
```diff
@@ -45,6 +48,8 @@ public function __construct(TaggedContainerInterface $container)
 
     /**
      * {@inheritdoc}
+     *
+     * @deprecated
```
stof: The deprecation should have a message giving the version it was deprecated in, and the alternative.

And I would also add a `@trigger_error('...', E_USER_DEPRECATED);` in the method.

src/Behat/Behat/Context/Environment/Handler/ContextEnvironmentHandler.php
everzet: @stof I like the way you think :)

src/Behat/Behat/HelperContainer/Argument/ServicesResolverFactory.php
everzet: Good shout.

PR #1071, opened 2017-09-01
Services autowiring

everzet:
# Services autowiring feature

This PR introduces new minor feature into Behat 3.x (planned for 3.4).

## Narrative

HelperContainers were introduced in 3.3 to help resolve particular case of sharing behaviour between contexts. But even before the problem of sharing behaviour there was always a problem of growing context classes. Everyone once stumbled into situation where you have tens of properties on your context class, most of which are used only once, in a single transformation or a step definition.

This is a very similar problem to what MVC users experience with their actions and controllers. For years my advice was similar to theirs - split your contexts down by lines of responsibility. This works well, but both adds additional weight to every decision at the beginning of the project, because splitting stateful classes later is harder than at the start.

Looking at MVC frameworks like Laravel and Symfony 4, they both now have an alternative solution to this problem, built on top of their service containers - auto-wiring. With auto-wiring framework can automatically inject the service your controller or action needs, just by reflecting upon the argument types you requested. From the user point of view this makes introducing dependency into your action as simple as just adding type-hinted argument. And actions created that way are easily transferrable across controllers with a simple cut-and-paste.

Imagine if you can do the same with your step definitions, transformations and event context constructors? With this PR merged, you can.

## Autowiring

Before starting with the spec and implementation, I set out on a list of basic rules that the implementation will follow:

- Autowiring only works with helper containers
- Autowiring is off by default
- Autowiring is enabled/disabled by a suite-level `autowire` flag
- It works for context constructor arguments
- It works for step definition arguments
- It works for transformation arguments
- It only wires arguments that weren't otherwise set
- Service arguments must be last in step definitions
- Service arguments must be last in transformations

## How to use service autowiring

The rest of this PR explains how to use autowiring. For more detailed info and examples, please read the feature attached to this PR.

### Enabling autowiring

Service auto-wiring only works with [helper containers](https://github.com/Behat/Behat/pull/974). So you need to enable them for your suite to benefit from the feature:

```yaml
default:
  suites:
    default:
      services:
        SharedService1: "SharedService1"
        SharedService2: "SharedService2"
```

Note that I use service class names as their names - without that auto-wiring wouldn't work. There's also a simpler notation now:

```yaml
default:
  suites:
    default:
      services:
        SharedService1: ~
        SharedService2: ~
```

After that is done, you'll need to explicitly enable auto-wiring (it is disable by default), again, at the suite level:

```yaml
default:
  suites:
    default:
      # ....
      autowire: true
```

With those two done, you're all set!

### Autowiring constructor arguments

As the name suggests, services can be wired into your context constructors. All you need to do is type-hint them:

```php
class MyContext implements Context {
    public function __construct(SharedService1 $s1, SharedService2 $s2) {
        // ...
    }
}
```

You might not expect it to, but auto-wiring works with mixing manual and automatic arguments:

```yaml
default:
  suites:
    default:
      contexts:
        - MyContext:
            name: "Konstantin"
      # ....
      autowire: true
```

```php
class MyContext implements Context {
    public function __construct(string $name, SharedService1 $s1, SharedService2 $s2) {
        // ...
    }
}
```

### Autowiring step definition arguments

Probably the most useful application is that you can use auto-wiring in step definitions. Yes, that means you don't need to inject your service into context constructor and store it on the context properties - you can just ask for it when and if you need it, straight inside the stepdef:

```php
    /** @When I set the state to :value */
    public function setState($value, Service2 $s2) {
        $s2->state = $value;
    }
```

**Note that service arguments should always go after all the other arguments in the stepdef.**

### Autowiring transformation arguments

Perhaps even more useful is the fact that you can do the same with transformations:

```php
    /** @Transform */
    public function usernameToUser(string $username, Users $repository): User {
        return $repository->findByUsername($username);
    }
```

**Note that service arguments should always go after all the other arguments in the transformations.**

## Further details

For further details, as usual, I advise you to dive into the feature file itself. It provides ultimate information, reasoining and examples behind this feature. Oh, yeah, it also is a stellar automated test to make sure I implemented feature correctly in the first place :)

As usual, feedback is greatly appreciated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 ### Added
+  * [#1071](https://github.com/Behat/Behat/pull/1071): Services auto-wiring
   * [#1054](https://github.com/Behat/Behat/pull/1054): [PSR-11](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-11-container.md)
     support for helper containers.
   * Support for modern PHPUnit.
--- a/features/autowire.feature
+++ b/features/autowire.feature
@@ -1,0 +1,162 @@
+Feature: Helper services autowire
+  In order to speed up the development process at early stages
+  developers need to have a convenient way of requesting services without going through the explicit configuration layer
+
+  Rules:
+    - Autowiring only works with helper containers
+    - Autowiring is off by default
+    - Autowiring is enabled/disabled by a suite-level `autowire` flag
+    - It works for context constructor arguments
+    - It works for step definition arguments
+    - It works for transformation arguments
+    - It only wires arguments that weren't otherwise set
+
+  Background:
+    Given a file named "behat.yaml" with:
+      """
+      default:
+        suites:
+          default:
+            services: ServiceContainer
+            autowire: true
+      """
+    And a file named "features/bootstrap/ServiceContainer.php" with:
+      """
+      <?php use Psr\Container\ContainerInterface;
+
+      class Service1 {public $state;}
+      class Service2 {public $state;}
+      class Service3 {public $state;}
+
+      class ServiceContainer implements ContainerInterface {
+          private $services = array();
+
+          public function has($class) {
+              return in_array($class, array('Service1', 'Service2', 'Service3'));
+          }
+
+          public function get($class) {
+              return isset($this->services[$class])
+                   ? $this->services[$class]
+                   : $this->services[$class] = new $class;
+          }
+      }
+      """
+
+  Scenario: Constructor arguments
+    Given a file named "features/autowire.feature" with:
+      """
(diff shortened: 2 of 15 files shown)


cherifGsoul:
Im confused a bit! where this should be used? In a Acceptance tests (ATDD) or Modelling By example (BDD)? 
What kind of services should be wired here? I guess Application Services.
I suggest to write documentations for use cases of this feature.

everzet:
@cherifGsoul It's supposed to be used anywhere it makes your life easier. If you don't know when to use it, then please don't use it! This feature isn't built for best practice (it is not a silver bullet), it is build for good practice (it is a solution to a problem).

tikolakin:
Not sure, but it seems the second time the new feature is introduced in such way. I like it.
it's a good addition to helpers. Behat becomes a better ecosystem to develop in comparison to rest of bdd frameworks 🥇

PR #1073, opened 2017-09-10
fix some issues reported by PhpStorm inspections

Haehnchen:
This pr fixes some PhpStorm inspections together with _Php Inspections (EA Extended)_. Just fixed some of them that were clear to me.

### DIFF
--- a/src/Behat/Behat/Context/Environment/Handler/ContextEnvironmentHandler.php
+++ b/src/Behat/Behat/Context/Environment/Handler/ContextEnvironmentHandler.php
@@ -140,7 +140,8 @@ final class ContextEnvironmentHandler implements EnvironmentHandler
                 $arguments = array();
 
                 if (is_array($context)) {
-                    $class = current(array_keys($context));
+                    $keys = array_keys($context);
+                    $class = current($keys);
                     $arguments = $context[$class];
                 }
 
--- a/src/Behat/Behat/Context/Reader/AnnotatedContextReader.php
+++ b/src/Behat/Behat/Context/Reader/AnnotatedContextReader.php
@@ -196,7 +196,7 @@ final class AnnotatedContextReader implements ContextReader
      */
     private function isNotAnnotation($docLine)
     {
-        return '@' !== substr($docLine, 0, 1);
+        return '@' !== $docLine[0];
     }
 
     /**
--- a/src/Behat/Behat/Output/Node/Printer/Helper/StepTextPainter.php
+++ b/src/Behat/Behat/Output/Node/Printer/Helper/StepTextPainter.php
@@ -58,7 +58,7 @@ final class StepTextPainter
         $paramStyle = $style . '_param';
 
         // If it's just a string - skip
-        if ('/' !== substr($regex, 0, 1)) {
+        if ('/' !== $regex[0]) {
             return $text;
         }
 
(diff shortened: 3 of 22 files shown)


everzet:
You broke tests.

src/Behat/Behat/Context/Reader/AnnotatedContextReader.php
```diff
@@ -196,7 +196,7 @@ private function isEmpty($docLine)
      */
     private function isNotAnnotation($docLine)
     {
-        return '@' !== substr($docLine, 0, 1);
+        return '@' !== $docLine[0];
```
pfrenssen: PHPStorm suggests this as an improvement? This makes it less clear to me, I know this works and that you can retrieve individual characters in a string by accessing them using the array key syntax, but the original code is much easier to understand for a human being.

src/Behat/Testwork/Argument/PregMatchArgumentOrganiser.php
```diff
@@ -62,7 +62,8 @@ private function cleanupMatchDuplicates(array $match)
         $arguments = array();
 
         $keys = array_keys($cleanMatch);
-        for ($keyIndex = 0; $keyIndex < count($keys); $keyIndex++) {
+        $keysCount = count($keys);
+        for ($keyIndex = 0; $keyIndex < $keysCount; $keyIndex++) {
```
pfrenssen: How about `foreach()`?

src/Behat/Behat/Context/Reader/AnnotatedContextReader.php
stof: and this code breaks for an empty string

src/Behat/Testwork/Argument/PregMatchArgumentOrganiser.php
stof: they keyIndex is sometimes incremented manually inside the loop, so a foreach cannot be used.

src/Behat/Behat/Output/Node/Printer/Helper/StepTextPainter.php
```diff
@@ -58,7 +58,7 @@ public function paintText($text, Definition $definition, TestResult $result)
         $paramStyle = $style . '_param';
 
         // If it's just a string - skip
-        if ('/' !== substr($regex, 0, 1)) {
+        if ('/' !== $regex[0]) {
```
stof: this breaks in case of empty string

src/Behat/Behat/Context/Reader/AnnotatedContextReader.php
kalessil: Covered here:
```php
           if ($this->isEmpty($docLine)) {
                continue;
            }
            if ($this->isNotAnnotation($docLine)) {
                continue;
            }
```
I'd suggest even to simplify `if ($this->isEmpty($docLine) || $this->isNotAnnotation($docLine)) ...`

src/Behat/Behat/Context/Environment/Handler/ContextEnvironmentHandler.php
```diff
@@ -140,7 +140,8 @@ function ($context) {
                 $arguments = array();
 
                 if (is_array($context)) {
-                    $class = current(array_keys($context));
+                    $keys = array_keys($context);
+                    $class = current($keys);
```
kalessil: PhpStrom stubs bug: https://github.com/JetBrains/phpstorm-stubs/commit/2aa970a6cad16b775bc958400114711184fed6c8, I'd suggest reverting.

src/Behat/Testwork/Argument/PregMatchArgumentOrganiser.php
kalessil: `for ($keyIndex = 0, $keysCount = count($keys); ...)` would be more semantically correct IMO

PR #1075, opened 2017-09-13
Fix PHP 5.3 compatibility

smoorhouse11:
Removed short array syntax to maintain php 5.3 style. I believe this is a bug since composer.json still has a PHP requirement value of >= 5.3.3.


everzet:
Thank you for spot and fix :)

Issue #1080, opened 2017-09-24
Autowiring not existing classes


Taluu:
Hello,

Just trying the new autowire feature. The thing is, I have a nullable constructor argument with an object typehint, which *can* be an not existing class.

```php
<?php
function __construct(NotExistingClass $class = null)
```

If used with autowire, we have a `ReflectionException` saying that `NotExistingClass` does not exist.

I think that if this argument is nullable (has a `null` default value or has a nullable typehint since 7.1), if the class does not exist, it should use `null` instead.

This case can happen on extension proposing contexts. For example, my extension Behapi *can* use twig but does *not* require it. If it is not available, it is set to `null` and then silently ignored.

Taluu:
As #1081 was merged in 3.5, it's safe to close. :}

PR #1081, opened 2017-09-24
Fix autowiring null

Taluu:
Ref #1080, but not sure it really fixes it, as we still have a problem if the class is not loadable (because it doesn't exist). But IMO this is not the concern of the autowirer, but more the responsability of the organizers...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Fixed
+  * [#1081](https://github.com/Behat/Behat/pull/1081): Do not use isset to be
+    able to check for intentional `null` values when autowiring contexts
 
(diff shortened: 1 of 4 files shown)


Taluu:
Nevermind, solved the problem with the last commit, a `try... catch ReflectionException` was all it needed. :}

everzet:
@Taluu you used slightly wrong assertion class. Can you update and I'll merge?

Taluu:
You mean using the namespaced one instead ? Sure, I'll rebase while I'm at it then.

Taluu:
Rebased and fixed the correct class. :}

everzet:
@Taluu top one, man :)

Taluu:
You mean in the Autowirer ? Looks good to me, It should not autowire if the key / argument exists in the index or the arguments or if the class can't be loaded ?

Unless I'm reading it wrong.

Taluu:
A shot in the dark : Ping ? @ciaranmcnulty @stof @everzet

ciaranmcnulty:
Thanks, @Taluu

PR #1095, opened 2017-10-24
Update dependencies to allow Symfony 4.x-components.

dbrumann:

PR #1096, opened 2017-10-27
Allow to use latest PHPUnit but keep the compatibility with legacy versions

jakzal:
Latest 4.8.36 comes with compatibility layer, so we can start using PHPUnit namespaces while continue using the 4.8 branch on legacy PHP versions. This will work even on PHP 5.3.

Note that [currently phpunit is not run on Travis](https://travis-ci.org/Behat/Behat/jobs/282180374#L519) with 7.x. phpunit can't find any tests.

screen shot 2017-10-27 at 11 00 48

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
 
     "require-dev": {
         "symfony/process": "~2.5|~3.0",
-        "phpunit/phpunit": "~4.5",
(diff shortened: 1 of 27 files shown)


jakzal:
This should also help with #1095

spolischook:
:+1:

Issue #994, opened 2017-02-02
Symfony's class loader will be deprecated in 3.3+


Taluu:
cf symfony/symfony#21353 : Instead of using Symfony's autoloader in the Autoloaders, we should consider using Composer's. See the mentionned PR on symfony for more details.

stof:
I'd rather deprecate the behat autoloading config entirely, in favor of the ``autoload-dev`` feature of composer (which did not exist at the time it was added in Behat)

ciaranmcnulty:
That's a good choice for Symfony projects, and similar 'modern' frameworks but not all projects use composer - consider a user running Wordpress and using behat.phar

I do agree we should encourage Composer usage, and we could do so in the current version of Behat by updating the documentation to include composer by default, but we need to consider how it will work in the use case of Behat as a stand-alone tool.

everzet:
@ciaranmcnulty I think the key message here is that Symfony autoloader we use for this feature really goes away. Which means that if we are to support this feature in the future, we should either copy the component into our codebase or reimplement it.

So our choice is:

1. Drop the feature
2. Copy-paste Autoloader component into Behat core
3. Reimplement Autoloader in Behat core
4. Switch to using Composer autoloader

Dropping the feature does make it tricky to use `behat.phar`. But I'd rather bundle `Composer` into PHAR stub than drag dead component over.

stof:
We have the Composer ClassLoader class inside the phar, as Behat itself uses composer.

everzet:
Yup. I meant exactly reappropriating that thing to autoload `features/bootstrap`

stof:
Btw, using it should be a no-brainer, as the API of ``Symfony\Component\ClassLoader\ClassLoader`` is similar to the one of ``Composer\Autoload\ClassLoader``

everzet:
I do think autoloading feature should be turned off in non-phar version of 4.x

stof:
and probably even in phar versions when we can load the project-level composer autoloading file (we should warn users about using the composer autoload-dev for such cases even in 3.x)

tristanbes:
What's the status of this since 3.3 is almost due date 📦  ?

MisterGlass:
I just ran an upgrade and am now running into the deprecation error every time I run tests. Is there any update/timeline on this?

Taluu:
[...]

PR #1107, opened 2017-11-28
symfony/class-loader 4.0 doesn't exist

dunglas:
The Symfony's ClassLoader component has been deprecated in favor of Composer and doesn't exist in version 4.
This PR removes the invalid version, but some work need to be done to refactor classes using this component.


Taluu:
See #994. What you are proposing here is a temporary solution, as removing it would break BC.

dunglas:
It’s just a quick improvement, because classloader v4 doesn’t exist and will not.

Taluu:
Yep, it seems I forgot to add a :+1: :}

PR #1109, opened 2017-12-08
Update .travis.yml

DonCallisto:


DonCallisto:
In a project of mine, `--prefer-lowest` makes behat fail. Let's test here and see what's no good.

DonCallisto:
Ok, I've noticed that behat 2 is not maintained anymore. So I'm gonna raise this into my `composer.json` file.

PR #1111, opened 2017-12-11
Feature: Enable to change Suite after initialization object

spolischook:
This is needed for change Suite after it's creation.

### DIFF
--- a/src/Behat/Testwork/Suite/Suite.php
+++ b/src/Behat/Testwork/Suite/Suite.php
@@ -48,12 +48,4 @@ interface Suite
      * @return mixed
      */
     public function getSetting($key);
-
-    /**
-     * Set key value pair to the settings
-     *
-     * @param string $key
-     * @param mixed $value
-     */
-    public function setSetting($key, $value);
 }
--- a/src/Behat/Testwork/Suite/Suite.php
+++ b/src/Behat/Testwork/Suite/Suite.php
@@ -48,4 +48,12 @@ interface Suite
      * @return mixed
      */
     public function getSetting($key);
+
+    /**
+     * Set key value pair to the settings
+     *
+     * @param string $key
+     * @param mixed $value
+     */
+    public function setSetting($key, $value);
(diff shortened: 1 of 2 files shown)


src/Behat/Testwork/Suite/Suite.php
```diff
@@ -48,4 +48,12 @@ public function hasSetting($key);
...
      */
     public function getSetting($key);
+
+    /**
+     * Set key value pair to the settings
+     *
+     * @param string $key
+     * @param mixed $value
+     */
+    public function setSetting($key, $value);
```
Taluu: That's a BC break. Do not add methods to the interface (the class is OK though)

spolischook: @Taluu I know that this is Bc break, but it's more convenient to relay on interface

spolischook:
@Taluu In our case we are redistribute suites after it's configurations according to statistic of suite duration. So we are make auto generated suites that will execute approx same time, see [OroBehatStatisticExtension](https://github.com/oroinc/platform/tree/master/src/Oro/Bundle/TestFrameworkBundle/BehatStatisticExtension)

src/Behat/Testwork/Suite/Suite.php
Taluu: It may be more convenient but it's still a BC break, so you'll have to comment out this one.

everzet:
@spolischook I know I'm late with comment, but as @Taluu said, we need to respect BC.

spolischook:
BC break removed.

ciaranmcnulty:
@spolischook Can you add a new interface that extends Suite, and have GenericSuite implement it? This would mean you can rely on an interface still

Just adding it to the concrete class risks us breaking your code in future by not respecting the BC

pamil:
@spolischook have you seen if https://github.com/FriendsOfBehat/SuiteSettingsExtension addresses your issue?

PR #1116, opened 2018-01-20
Uniq filename for xml file

tikolakin:
Hi there,
I encountered issue with reports file names. As far as junit report create a xml file with the name of a suite it makes difficulties to manage those files if we run tests in parallel 
So far we solved it with separate folder for each report file
Now we are adding another xml reporter and managing files became complicated

Any suggestion for a filename?
Thanks


spolischook:
For parallel execution we are using generating divided suites with auto generated names, like User_1 User_2

tikolakin:
we could do it, but we do not generate suites.
our solution is very simple - we parse `./features` then behat executes a single feature
based on behat config - all report files are placed in the same folder - and it's a think we want to improve

spolischook:
Use uniqid is not a perfect deal, because it will litter the report directory. Use the build number should make a deal.

spolischook:
Also you can change the directory for output for each run and this is should resolve your issue

tikolakin:
so, at the time being, we do create a separate folder for each xml file
to do this we have to manage it with BEHAT_PARAMS, where we provide a report path 
I would like to not do it as it's now more complicated with another xml reporter
Build number doesn't solve it, as single build has xml report per feature file

> it will litter the report directory
people can clean report folder programmatically if they want (mostly issue with local execution)
but ok on CI

tikolakin:
well, when you run test on local you usually do not need xml reporter))
it's only valuable on CI when we have separate work directory for each build 
report folder should not be trash)

tikolakin:
hi, @everzet @stof 
are you guys available to give a feedback on the above? 
Thanks :1st_place_medal:

chernecov:
Any updates?

pamil:
It is a BC break to change that filename. Have you seen #915?

tikolakin:
hi @pamil, thanks for the reference
I've just had a quick look at that feature request and I don't think it will solve all the cases
example where it won't work is
feature file has 5 scenarios (it also might include scenario outline)
and that scenarios are executed in parallel 
that will still provide 5 reports file with same file name overriding each other 

while that feature request has a value - I'm still sure that provided unique file name is a more general solution that will work in most of the cases.

tikolakin:
@ciaranmcnulty 
> I'd be ok with the behat instance getting in ID at the CLI though? like --runner-id= maybe?
it still doesn't resolve the issue when you have run Behat in parallel 
By this time we used another workaround ...
I don't it's actually since such a long time :)

PR #1118, opened 2018-01-24
Catch PHP 7 Errors and rethrow them as Exception

### LINKED ISSUE #1117: Catch errors/exceptions in hook and return non-0 status code
I'm using https://github.com/BehatExtension/DoctrineDataFixturesExtension to load fixtures before running my tests. But I made a small mistake that caused a TypeError to be raised. The output looked like this:
```
$ vendor/bin/behat

In LoadData.php line 56:

  Type error: Too few arguments to function App\DataFixtures\ORM\Scenarios\LoadData::__construct(), 0 passed in /var/www/symfony/vendor/doctrine/data-fixtures/lib/Doctrine/Common/DataFixtures/Loader.php on line 210 and exactly 1 expected


behat [-s|--suite SUITE] [-f|--format FORMAT] [-o|--out OUT] [--format-settings FORMAT-SETTINGS] [--init] [--lang LANG] [--name NAME] [--tags TAGS] [--role ROLE] [--story-syntax] [-d|--definitions DEFINITIONS] [--snippets-for [SNIPPETS-FOR]] [--snippets-type SNIPPETS-TYPE] [--append-snippets] [--no-snippets] [--strict] [--order ORDER] [--rerun] [--stop-on-failure] [--dry-run] [--] [<paths>]

$ echo $?
0
```

It seems that Behat just thinks everything is fine. It doesn't continue with the tests, but it reports a `0` as return code. The issue is that since the return code is `0` our CI thinks that everything is 👍

Should this be handled better by Behat? Or should the DoctrineDataFixturesExtension capture this somehow?

ruudk:
This will make sure that the Symfony Console application returns a non-zero exit code

Fixes #1117

### DIFF
--- a/src/Behat/Testwork/Cli/Application.php
+++ b/src/Behat/Testwork/Cli/Application.php
@@ -10,6 +10,7 @@
 
 namespace Behat\Testwork\Cli;
 
+use Behat\Testwork\Call\Exception\FatalThrowableError;
 use Behat\Testwork\ServiceContainer\Configuration\ConfigurationLoader;
 use Behat\Testwork\ServiceContainer\ContainerLoader;
 use Behat\Testwork\ServiceContainer\Exception\ConfigurationLoadingException;
@@ -121,7 +122,14 @@ final class Application extends BaseApplication
 
         $this->add($this->createCommand($input, $output));
 
-        return parent::doRun($input, $output);
+        try {
+            return parent::doRun($input, $output);
+        } catch(\Error $error) {
+            // Catch PHP 7 Error's and rethrow them as FatalThrowableError exception
+            // This will make sure that the Symfony Console application returns a non-zero exit code
+
+            throw new FatalThrowableError($error);
+        }
     }
 
     protected function getDefaultCommands()


ruudk:
Not sure if tests are failing because of my change or that Master is just broken... @everzet Do you know?

ruudk:
@stof What do you think? I really think this can help to make sure failures are correctly handled on CI.

composer.json
```diff
@@ -20,6 +20,7 @@
         "behat/transliterator": "^1.2",
         "symfony/console": "~2.5||~3.0||~4.0",
         "symfony/config": "~2.3||~3.0||~4.0",
+        "symfony/debug": "~2.1||~3.0||~4.0",
```
Taluu: ... why ? as you're not using anything from this component. BTW, at least use supported versions such as `2.8`, `3.3`, ... and so on.

src/Behat/Testwork/Cli/Application.php
```diff
@@ -121,7 +122,14 @@ public function doRun(InputInterface $input, OutputInterface $output)
 
         $this->add($this->createCommand($input, $output));
 
-        return parent::doRun($input, $output);
+        try {
+            return parent::doRun($input, $output);
+        } catch(\Error $exception) {
```
OskarStark: should be `$error`

composer.json
ruudk: Removed, not sure why I added that.

src/Behat/Testwork/Cli/Application.php
ruudk: Good one, fixed!

ruudk:
Scrutinizer is complaing because it doesn't understand `\Error`

stof:
Could we have a Behat scenario covering this case to prevent regressions ?

ruudk:
@stof I'd love to, but can you point me in the right direction?

stof:
Well, write a scenario in https://github.com/Behat/Behat/tree/master/features reproducing the case triggering an Error at a point where your patch is needed to get the proper exit code.
I even suggest you to try running your scenario *without* the patch as well, to see it failing (otherwise it means that either your patch is useless or your scenario does not cover it).

ruudk:
I cannot even reproduce it anymore, so it could be that it fixed itself. 🙈

stof:
Well, maybe it was fixed in the Symfony Console component in the meantime. I don't remember which changes we made in the component in this timeframe, but I know that some work on this topic was done at some point.

ruudk:
@stof Its amazing how you can keep track of every single repository on the internet 🙌 Thanks for your help and keep it up 💪

PR #1120, opened 2018-01-26
Fix the test suite by working around a Symfony problem

jakzal:
There is a Symfony BC break that broke the test suite: https://github.com/symfony/symfony/issues/25825

`-sbig_brother` is treated as a chain of options, and `h` makes the command display a help message. 

This change will at least make the build green again, before the issue is fixed in Symfony.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,7 +6,7 @@ clone_folder: c:\projects\behat
 
 environment:
     matrix:
-        - PHP_DOWNLOAD_FILE: php-7.1.9-nts-Win32-VC14-x86.zip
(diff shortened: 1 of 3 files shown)


jakzal:
AppVeyor issue seems to be caused by something else.

jakzal:
All green 😃

stof:
@jakzal when working on the fix on the Symfony side, please make sure to add a test covering this case though, so that a similar BC break is catched by the testsuite in the future.

jakzal:
Thanks :+1:

For now, the BC breaking change was reverted in Symfony. We're not sure if/how the issue it was trying to fix is going to be taken care of.

PR #1130, opened 2018-02-22
Fix quoteless definition arguments matching with unicode characters

ilovemistakes:
This allows us to use unicode characters in definition arguments without quotes. This is very convenient when using ordinals like "1ый", "2ая", etc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Fixed
+  * [#1129](https://github.com/Behat/Behat/issues/1129): Fix quoteless
+    definition arguments matching with unicode characters
 
 ## [3.4.3] - 2017-11-27
 ### Fixed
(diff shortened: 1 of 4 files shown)


stof:
putting the regex in unicode mode would means that any non-unicode input would not be supported anymore as it will make the regex fail. Are we sure that we won't break existing projects ?

ilovemistakes:
Do you have any case in mind where non-unicode input is not supported in unicode mode?
Gherkin already has a [requirement](https://github.com/Behat/Gherkin/blob/master/src/Behat/Gherkin/Lexer.php#L63) for feature files to be in UTF-8 encoding.
According to [PHP docs](http://php.net/manual/en/reference.pcre.pattern.modifiers.php) the PCRE_UTF8 "turns on additional functionality of PCRE" and "Pattern and subject strings are treated as UTF-8". And they are in UTF-8. So, in our case, non-UTF-8 input (english characters) is a subset of UTF-8 input. Any english word will match in both unicode and non-unicode modes. I've written a [test](https://github.com/Behat/Behat/pull/1130/files#diff-89340ed26b1d2fcb84f043a335e04d05R622) to make sure that english words still match.
The only case of breaking existing projects I can think of is when somebody believes that quoteless words, containing non-english characters, must not be treated as definition arguments. But i see no reasons for this belief, as [documentation](http://behat.org/en/latest/user_guide/context/definitions.html) states, that 
> All token values of the pattern (e.g. :methodArgument) will be captured and passed to the method argument

There are no mentions about english and non-english characters.

ciaranmcnulty:
Thanks @ilovemistakes 

We'll need to keep an eye open for any new issues from people using unicode languages

PR #1146, opened 2018-05-10
Use the Symfony hosted version of PHP to fix the appveyor build

jakzal:

PR #1147, opened 2018-05-10
Use latest version of php for appveyor

Taluu:
The build fails because it goes on a 404.


Taluu:
looks like @jakzal's already on it in #1146 . Closing.

PR #1148, opened 2018-05-10
Rework appveyor

Taluu:
And try to test it on multiple php versions too. I also added php 7.2 to travis while I was at it.

Mostly inspired from @Majkl578 and @photodude work on [doctrine dbal's appveyor config](https://github.com/doctrine/dbal/blob/master/.appveyor.yml)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: php
 
-php: [5.6, 7.0, 7.1]
+php: [5.6, 7.0, 7.1, 7.2]
 
 sudo: false
 
@@ -20,8 +20,8 @@ matrix:
       env: SYMFONY_VERSION='2.7.*'
     - php: 5.6
       env: SYMFONY_VERSION='2.8.*'
-    - php: 7.1
-      env: SYMFONY_VERSION='4.0.*@dev'
+    - php: 7.2
+      env: SYMFONY_VERSION='^3.4'
 
 before_install:
   - phpenv config-rm xdebug.ini
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,7 +6,7 @@ clone_folder: c:\projects\behat
 
 environment:
     matrix:
-        - PHP_DOWNLOAD_FILE: php-7.1.3-Win32-VC14-x86.zip
+        - php: 7.2
 
 branches:
   only:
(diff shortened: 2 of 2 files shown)


appveyor.yml
```diff
@@ -16,31 +19,49 @@ skip_commits:
...
+
+            if (!(Test-Path c:\tools\composer\composer.phar)) {
+                appveyor-retry appveyor DownloadFile https://getcomposer.org/composer.phar -Filename C:\tools\composer\composer.phar
+                Set-Content -path 'C:\tools\composer\composer.bat' -Value ('@php C:\tools\composer\composer.phar %*')
+            }
+        }
+
     - cd c:\projects\behat
-    - composer install --no-progress --ansi
+    - rm composer.lock
```
Majkl578: You don't need this one.

Taluu: yeah wasn't sure if some cache would not kick in. Removed.

Majkl578:
Don't trust metrics here, cache is never persisted for PRs. If you want to benchmark (it will always vary due to load etc.), you will need to test it in the repository itself (branch is enough), not through PR. :)

Majkl578:
You can also check Doctrine DBAL build times here: https://ci.appveyor.com/project/doctrine/dbal
(The config is referenced above and is a bit more more complicated than yours, also collects and sends coverage report.)

Taluu:
> However, this will extend the build time from ~2.5min to ~17.5min. Is it worth it? We already run tests against all supported php versions on travis. Appveyor is ther just so we confirm there's no incompatibility issues with windows. Latest PHP version should be sufficient imo.

Yeah, been wondering that. Thought some cache could kick in, but then we could indeed at least test only the latest version... or maybe 5.6 and 7.2 ?

Majkl578:
@Taluu Btw you will probably need different directories for different PHP versions in order to make cache work properly.

appveyor.yml
```diff
@@ -16,31 +19,51 @@ skip_commits:
     message: /\[ci skip\]/
 
 cache:
-    - c:\php -> appveyor.yml
+    - C:\ProgramData\chocolatey\bin -> appveyor.yml
+    - C:\ProgramData\chocolatey\lib -> appveyor.yml
+    - C:\tools\php5.6 -> appveyor.yml
```
Majkl578: If you make the version a subdirectory of "C:\tools\php", this and other occurrences may not turn that ugly. :)

Taluu:
I think I'm gonna test only on latest php version as @jakzal suggested, as it's a mess with the env vars. It seems that chocolatey doesn't properly install when env vars are involved...

jakzal:
Looks good to me 👍

.travis.yml
```diff
@@ -20,7 +20,7 @@ matrix:
       env: SYMFONY_VERSION='2.7.*'
     - php: 5.6
       env: SYMFONY_VERSION='2.8.*'
-    - php: 7.1
+    - php: 7.2
       env: SYMFONY_VERSION='4.0.*@dev'
```
stof: the one which should be tested explicitly here is the 3.4 LTS, like done for 2.7 and 2.8. The `4.0.*@dev` is useless

appveyor.yml
```diff
@@ -16,31 +16,48 @@ skip_commits:
     message: /\[ci skip\]/
 
 cache:
-    - c:\php -> appveyor.yml
+    - C:\ProgramData\chocolatey\bin -> appveyor.yml
+    - C:\ProgramData\chocolatey\lib -> appveyor.yml
+    - C:\tools\php -> appveyor.yml
+    - C:\tools\composer -> appveyor.yml
+    - '%LOCALAPPDATA%\Composer\files -> composer.json'
```
stof: no need to expire them when the composer.json is changing. Even if we change one dependency, the others could still benefit from the cache.

Taluu:
Changes done, as @stof suggested them. I rebase against the master too while I was at it.

PR #1153, opened 2018-05-29
Add local cache for transformPatternToRegex

jdeniau:
This little cache helps us save 1,5 second on one file containing 2200 lines and "saved" nearly one billion call to `preg_match`

### DIFF
--- a/src/Behat/Behat/Definition/Pattern/PatternTransformer.php
+++ b/src/Behat/Behat/Definition/Pattern/PatternTransformer.php
@@ -26,6 +26,8 @@ final class PatternTransformer
      */
     private $policies = array();
 
+    private $patternToRegexpCache = array();
+
     /**
      * Registers pattern policy.
      *
@@ -34,6 +36,7 @@ final class PatternTransformer
     public function registerPatternPolicy(PatternPolicy $policy)
     {
         $this->policies[] = $policy;
+        $this->patternToRegexpCache = array();
     }
 
     /**
@@ -68,11 +71,16 @@ final class PatternTransformer
      */
     public function transformPatternToRegex($pattern)
     {
-        foreach ($this->policies as $policy) {
-            if ($policy->supportsPattern($pattern)) {
-                return $policy->transformPatternToRegex($pattern);
+        if (!isset($this->patternToRegexpCache[$pattern])) {
+            foreach ($this->policies as $policy) {
+                if ($policy->supportsPattern($pattern)) {
+                    $this->patternToRegexpCache[$pattern] = $policy->transformPatternToRegex($pattern);
+                    break;
+                }
             }
         }
+        return $this->patternToRegexpCache[$pattern];
+
 
         throw new UnknownPatternException(sprintf('Can not find policy for a pattern `%s`.', $pattern), $pattern);
     }
--- a/src/Behat/Behat/Definition/Pattern/PatternTransformer.php
+++ b/src/Behat/Behat/Definition/Pattern/PatternTransformer.php
@@ -79,9 +79,11 @@ final class PatternTransformer
                 }
             }
         }
+
+        if (!isset($this->patternToRegexpCache[$pattern])) {
+            throw new UnknownPatternException(sprintf('Can not find policy for a pattern `%s`.', $pattern), $pattern);
+        }
+
         return $this->patternToRegexpCache[$pattern];
-
-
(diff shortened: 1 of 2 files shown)


jdeniau:
@ciaranmcnulty changed

ciaranmcnulty:
Looks like a cool optimisation - my only worry would be if there's an adverse affect on memory usage for suites which have a large number of _unique_ step defs - any ideas how to profile it?

jdeniau:
In the test I mention, I profiled with blackfire, I could only watch `supportsDefinitionAndArgument` wich call `transformPatternToRegexp` (because in the optimized version, the call disapear, I think blackfire does not display "small" calls).

Well, the memory usage for `supportsDefinitionAndArgument` "bumped" from 45k to 76k. For the record the global memory usage for this test is 1080M in both case (0.004% → 0.007%), so I think it's pretty anectdotic.

As a matter of fact, I think we do more reuse regexp more than we create new ones, especially with tools like `behatch/context`

src/Behat/Behat/Definition/Pattern/PatternTransformer.php
```diff
@@ -68,11 +71,16 @@ public function generatePattern($type, $stepText)
...
-                return $policy->transformPatternToRegex($pattern);
+        if (!isset($this->patternToRegexpCache[$pattern])) {
+            foreach ($this->policies as $policy) {
+                if ($policy->supportsPattern($pattern)) {
+                    $this->patternToRegexpCache[$pattern] = $policy->transformPatternToRegex($pattern);
+                    break;
+                }
             }
         }
+        return $this->patternToRegexpCache[$pattern];
```
ciaranmcnulty: you should only return it if it's set (the loop maybe got to the end and didn't find anything)

ciaranmcnulty: Note the unreachable exception after this - you could move the exception up and put a condition around it

ciaranmcnulty: The fact the tests didn't catch this means there's one missing :-)

jdeniau: 🤦‍ Of course, it's fixed, and I added a test about it

jdeniau:
@ciaranmcnulty I do not really know how to fix the scrutinizer test without increasing the overall complexity.

I could add some private functions but it seems kind of really complex for that case, WDYT ?

jakzal:
This would normally be solved by decoration, but we don't have an interface here ;)

I'd indeed extract some private methods to make the public method more concise and readable, i.e.:

```php
     public function transformPatternToRegex($pattern)
    {
        if (!isset($this->patternToRegexpCache[$pattern])) {
            $this->patternToRegexpCache[$pattern] = $this->transformPatternToRegexWithSupportedPolicy($pattern);
        }

        return $this->patternToRegexpCache[$pattern];
    }

    private function transformPatternToRegexWithSupportedPolicy(string $pattern): string
    {
        foreach ($this->policies as $policy) {
            if ($policy->supportsPattern($pattern)) {
                return $policy->transformPatternToRegex($pattern);
            }
        }

        throw new UnknownPatternException(sprintf('Can not find policy for a pattern `%s`.', $pattern), $pattern);
    }
```

jakzal:
Sorry... I forgot that Behat still supports php 5, and my example used php 7 constructs. Also, I haven't tested the "rafactoring". Did it in my head ;)

tests/Behat/Tests/Definition/Pattern/PatternTransformerTest.php
```diff
@@ -0,0 +1,86 @@
...
+
+use Behat\Behat\Definition\Pattern\PatternTransformer;
+use Behat\Behat\Definition\Pattern\Policy\PatternPolicy;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Class PatternTransformerTest
+ * @author Julien Deniau <julien.deniau@mapado.com>
+ */
+class PatternTransformerTest extends TestCase
```
everzet: Bonus points for having integration test!

src/Behat/Behat/Definition/Pattern/PatternTransformer.php
```diff
@@ -26,6 +26,8 @@
      */
     private $policies = array();
 
+    private $patternToRegexpCache = array();
```
everzet: Ideally needs a typehint.

ciaranmcnulty:
Thanks, @jdeniau!

PR #1138, opened 2018-04-06
Add behat-zf-extension to suggestion in composer.json

RageZBla:

PR #1155, opened 2018-06-02
Remove suggestions from composer.json

everzet:
As suggested in #1138, I propose to remove suggestions from the package.json. The utility of this was always questionable in the context information provide.

People who want Behat to integrate with Symfony2 OR Mink in most cases already know about the extensions they need to install. Even if they weren't there are better ways to instruct them.


ciaranmcnulty:
It's possible mink could stay...

ciaranmcnulty:
Looks good but Travis is borked

PR #1160, opened 2018-07-03
Register cli input/output services as synthetic

### LINKED ISSUE #1159: Symfony/DIC upstream break
I've reported an issue to the symfony project that stops behat working entirely
https://github.com/symfony/symfony/issues/27802

There may be a workaround we can implement (or it may be triggered by how we're using the container) in case that Symfony change makes it into a stable release.

jakzal:
> Services that are set at runtime are called synthetic services. This service has to be configured so the container knows the service exists during compilation (otherwise, services depending on kernel will get a "service does not exist" error).

source: https://symfony.com/doc/current/service_container/synthetic_services.html

Fixes #1159

### DIFF
--- a/src/Behat/Testwork/Cli/ServiceContainer/CliExtension.php
+++ b/src/Behat/Testwork/Cli/ServiceContainer/CliExtension.php
@@ -81,6 +81,7 @@ final class CliExtension implements Extension
     public function load(ContainerBuilder $container, array $config)
     {
         $this->loadCommand($container);
+        $this->loadSyntheticServices($container);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Behat/Testwork/Cli/ServiceContainer/CliExtension.php
```diff
@@ -103,6 +104,12 @@ protected function loadCommand(ContainerBuilder $container)
         $container->setDefinition(self::COMMAND_ID, $definition);
     }
 
+    protected function loadSyntheticServices(ContainerBuilder $container)
+    {
+        $container->register(self::INPUT_ID)->setSynthetic(true);
```
stof: I think we should also make them public

jakzal: Updated.

ciaranmcnulty:
Thanks @jakzal

Issue #1108, opened 2017-12-01
Serialization of 'Closure' is not allowed


respinoza:
Hello,

We upgraded behat and it pulled some 4.0 symfony components and we get the following error

```php
In RemoveUnusedDefinitionsPass.php line 76:
                                             
  [Exception]                                
  Serialization of 'Closure' is not allowed  
                                             

Exception trace:
 serialize() at /web/vendor/symfony/dependency-injection/Compiler/RemoveUnusedDefinitionsPass.php:76
 Symfony\Component\DependencyInjection\Compiler\RemoveUnusedDefinitionsPass->process() at /web/vendor/symfony/dependen
cy-injection/Compiler/RepeatedPass.php:57
 Symfony\Component\DependencyInjection\Compiler\RepeatedPass->process() at /web/vendor/symfony/dependency-injection/Co
mpiler/Compiler.php:95
 Symfony\Component\DependencyInjection\Compiler\Compiler->compile() at /web/vendor/symfony/dependency-injection/Contai
nerBuilder.php:699
 Symfony\Component\DependencyInjection\ContainerBuilder->compile() at /web/vendor/behat/behat/src/Behat/Testwork/Cli/A
pplication.php:187
 Behat\Testwork\Cli\Application->createContainer() at /web/vendor/behat/behat/src/Behat/Testwork/Cli/Application.php:1
61
 Behat\Testwork\Cli\Application->createCommand() at /web/vendor/behat/behat/src/Behat/Testwork/Cli/Application.php:122
 Behat\Testwork\Cli\Application->doRun() at /web/vendor/symfony/console/Application.php:129
 Symfony\Component\Console\Application->run() at /web/vendor/behat/behat/bin/behat:34
```

If we force to use symfony dependency injection to be ^3.0, everything works as normal (tested with 3.4)

t1k3hu:
`"symfony/dependency-injection": "3.3.13"`

valorin:
Looks like this was introduced here: https://github.com/Behat/Behat/issues/1108

drbyte:
Experiencing the same problem here.

drbyte:
Had to manually add a `dev` dependency for `symfony/dependency-injection:"3.*"` in my `composer.json`:

`composer require --dev symfony/dependency-injection:"3.*"`
and then 
`composer update`

Not a fan of having to force this. 
Been fence-sitting about converting all my tests to something other than behat. This may be the shove I needed to go and do it. Sigh.

[...]

PR #1163, opened 2018-07-12
Fix for #1108: Serialization of 'Closure' is not allowed

philtweir:
Allows for changes in Symfony serialization in 4.x, where `json_encode` has been replaced by `serialize` and an error is thrown if a closure is found. This replaces the service container object as a Definition argument with a Symfony DI Reference that can be serialized.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Fixed
+  * Serializable definitions for Symfony 4.x
 
 ## [3.4.3] - 2017-11-27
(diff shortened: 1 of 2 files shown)


philtweir:
Having had a look through the features, I'm not too sure what the best approach for a test for this would be, but happy to take some direction.

philtweir:
For context, in case anyone is having similar issues: Behat-Laravel-Extension#78 was what led to this PR - in the meantime, following the instructions there should (as a stop-gap) get Behat working for Laravel >=5.6.

titopixelfusion:
It would be great to merge this PR

ciaranmcnulty:
Yep looks good, thank you

ciaranmcnulty:
Thanks @philtweir!

philtweir:
Thanks @ciaranmcnulty !

PR #1171, opened 2018-08-28
Remove the dependency on Symfony's ClassLoader

felds:
It solves the issue #994 and allows installing Behat inside a Symfony 4 project.


stof:
Note that I don't think `SuiteWithContextsSetup` works fine when using composer's `autoload-dev` instead of the Behat autoloading config

Meyfarth:
Hello, when will it be released? We're using `dev-master` in order to avoid this bug, but I would like to use a stable release of behat.

Thanks

mikadamczyk:
Hi, is any chance to release in the nearest future? Some packages do not register Symfony services when Class Loader Component is present.

PR #1172, opened 2018-08-29
Remove herrera-io/box as a dependency

felds:
This package was not being used directly in the code.

It cuts down on some packages, including two abandoned ones: `phine/exception` and `phine/path`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,8 +31,7 @@
 
     "require-dev": {
         "symfony/process": "~2.5|~3.0|~4.0",
-        "phpunit/phpunit": "^4.8.36|^6.3",
-        "herrera-io/box": "~1.6.1"
+        "phpunit/phpunit": "^4.8.36|^6.3"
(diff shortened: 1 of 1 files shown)


Taluu:
I think it was used to generate the phar. So not directly used by behat (in its code), but for releases.

felds:
It looks to be the case. But anyways, I don't think it should  carry 

--

stof:
This is a dev requirement, not a normal requirement. This relates to the tooling used by Behat.

felds:
@stof is there any other way to keep tooling working without carrying it over on every installation? Since Behat is a dev library, putting it under `require-dev` doesn't make much of a difference…

Maybe moving it to the `suggest` section?

ciaranmcnulty:
@felds that's not how require-dev works; transitive dependencies aren't installed by anyone requiring behat/behat

PR #1184, opened 2018-11-02
Fix phar deployment on Travis

### LINKED ISSUE #1078: phar files are not provided on github
phar bundles are a wonderful way to provide a software such as Behat.
Sadly, recent releases are still not available through GitHub downloads.

michalpipa:
Phar files were not built on Travis, because builds were enabled only on master branch, not on tags. You can see requests log here: [Requests](https://travis-ci.org/Behat/Behat/requests). So I have enabled builds on tags.

Enabling builds revealed another problem. In old Travis config (before my changes) deploy stage was executed on every of four PHP 5.6 jobs. This means that there could be zero to four deployments depending of how many jobs have failed. The result was unpredictable, because each upload was overwriting previous one. Final deploy was the result of the longest running job.

To solve those problems I've used Travis staged builds. There are two stages: test and deploy. Test stage consists of eight jobs running in parallel (the same jobs as before). Deploy stage has only one job and runs only on tagged commits and after all jobs from test stage have succeeded. 

This PR fixes #1078.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,20 +8,7 @@ cache:
   directories:
     - $HOME/.composer/cache
 
-branches:
-  only:
-    - master
-
-matrix:
-  include:
-    - php: 5.6
-      env: DEPENDENCIES='low'
-    - php: 5.6
-      env: SYMFONY_VERSION='2.7.*'
-    - php: 5.6
-      env: SYMFONY_VERSION='2.8.*'
-    - php: 7.2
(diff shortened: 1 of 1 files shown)


ciaranmcnulty:
Fixed with move to GHA as far as I can see?

ciaranmcnulty:
Thanks for the efforts @michalpipa I'm sorry It wasn't merged at the time

PR #1185, opened 2018-11-05
Get contexts via polymorphism

Chekote:
## Scenario:
Given I have a context class MyContext
And I have a context class MyExtendingContext that extends MyContext
When I register MyExtendingContext in the Behat environment
And I try to use InitializedContextEnvironment::getContext to get MyContext

## Desired Behavior:
Then I should receive MyExtendingContext as it "is-a" MyContext

## Current Behavior:
Then I do not receive MyExtendingContext as it is not a MyContext

## Implementation:
I modified the hasContextClass and getContext methods to check for extending or implementation of the specified class or interface.

### DIFF
--- a/src/Behat/Behat/Context/Environment/InitializedContextEnvironment.php
+++ b/src/Behat/Behat/Context/Environment/InitializedContextEnvironment.php
@@ -97,7 +97,30 @@ final class InitializedContextEnvironment implements ContextEnvironment, Service
      */
     public function hasContextClass($class)
     {
-        return isset($this->contexts[$class]);
+        return (bool) $this->resolveContextClass($class);
+    }
+
+    /**
+     * Resolves the specified class to a registered context.
+     *
+     * Will return an instance of the specified class if it is registered directly,
+     * or any registered context that extends/implements the specified class/interface.
+     *
+     * @param  string       $class the fully qualified class name.
+     * @return Context|null The context class, or null if no matching context was found.
+     */
+    public function resolveContextClass($class) {
+        if (isset($this->contexts[$class])) {
+            return $this->contexts[$class];
+        }
+
+        foreach ($this->contexts as $context) {
+            if ($context instanceof $class) {
+                return $context;
+            }
+        }
+
+        return null;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Taluu:
I'm -1 on this, because I don't think it should be behat's problem to resolve the context class. Especially if the context is extended by several other contexts, then what ?

Anyway, you shouldn't really extend contexts IMO, but use several of them instead (through composition).

Chekote:
Yes, I definitely agree that ideally you should compose rather than extend, but I have a specific use case where I’m refactoring a framework to do just that, and not being able to resolve via polymorphism is making that tough. Of course if this is not accepted, I can just use a temporary fork.

Regarding the problem of multiple extending contexts, it wouldn’t be a problem. I’d be asking for a MyContext. If there were multiple contexts that were a MyContext, any of them would suffice.

Chekote:
@Taluu So I have a concrete example of why I'm doing this, and I'm wondering if there's another solution since this approach looks like it won't be approved:

We're using MinkContext but we need to enhance the behavior of some of the Behat steps. Our current approach is to extend MinkContext, override the methods, and then register our extending context instead of the base MinkContext class.

The problem arises from the fact that we need to modify an existing Behat step. If I was to write a context class that does not extend MinkContext but attempts to implement a step with the same pattern, Behat is going to complain that there is ambiguity when both contexts are registered.

The only other approach I can think of is to NOT register MinkContext, and instead re-implement all of the method signatures in my own context class, and have it call an internal instance of MinkContext that is not registered with the Behat environment. But that just feels wrong.

Do you have any suggestions on how to effectively override a small subset of the MinkContext methods if we shouldn't use inheritance?

Taluu:
As you said, the best way would be imo to extend the `Minkcontext` and register only _your_ context.

Who cares if it's called `MinkContext` ? Unless other contexts are calling `MinkContext` specifically (would be strange imo) ?

Chekote:
Yes, that's exactly the scenario. The extending context that fixes the visibility checks of mink is in a separate composer package. The projects that leverage the functionality are coded against Mink. To utilize the fixes with this PR in place, the projects just require the package and swap out the MinkContext for the extending context in the behat.yml. All of the existing code that tries to get access to MinkContext still works because the extending context "is-a" MinkContext.

It is possible to utilize the package without this PR, but then you need to tightly couple your other contexts against the extending context. You could argue that that's inappropriate because if you're coding against MinkContext, then you're coding against what it considers visible, and if you want some other behavior, you should code against that context instead.

Chekote:
I'm going to close this in favor of refactoring the inheritance to use composition as you suggested. Thanks so much for helping me reason this out, I really appreciate your assistance.

PR #1190, opened 2018-11-17
Nullable arguments

### LINKED ISSUE #1189: If a step argument is nullable, it should use `null` as value if no match
With the following step :

```php
/** 
 * @Then foo :bar 
 * @Then bar
 */
public function foo(?string $bar): void
{
}
```

If the second version of the step is used, it crashes saying that it can't find a value for `$bar` argument. IMO, if the parameter allows `null` value, it should validate and use `null` as a value if none is given.

I can try to work on a patch for that. WDYT ?

Taluu:
Fix #1189 

If a parameter is nullable but doesn't have a `null` default value, it should IMO use `null` as a value. Not sure though if it's not better to leave it as it is currently, with nothing more than a default value detection.

### DIFF
--- a/features/arguments.feature
+++ b/features/arguments.feature
@@ -181,3 +181,41 @@ Feature: Step Arguments
       1 scenario (1 passed)
       1 step (1 passed)
       """
+
+  @php-version @php7.1
+  Scenario: Nullable arguments
+    Given a file named "features/bootstrap/FeatureContext.php" with:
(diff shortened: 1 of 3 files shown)


Taluu:
After some thoughts, I really think it should not use `null` as a value if no default value but nullable. Closing, feel free to reopen and merge if you still think it's worth it.

PR #1198, opened 2018-12-23
Added Korean language

getsolaris:
Added Korean language


ciaranmcnulty:
Is there a way we can get someone else Korean to validate the translations?

stof:
I think locales are sorted in the file currently.

getsolaris:
Thank you

Issue #1194, opened 2018-12-10
RerunController should not attempt write cache if --rerun was not specified


oleg-andreyev:
Noticed that `RerunController` is trying to write cache, even if `--rerun` was not specified.

https://github.com/Behat/Behat/blob/master/src/Behat/Behat/Tester/Cli/RerunController.php#L89-L91

ciaranmcnulty:
It's only a 'rerun' the 2nd time around

The idea is you:

 1. Run Behat normally
 2. Get a fail, fix the issue
 3. Run it with `--rerun`

This only works if step 1 creates the log

I'd be open to `--no-rerun` or something?

Issue #714, opened 2015-02-19
--rerun behaving unexpectedly


WouterJ:
**First run (without `-rerun`):**
````
$ "bin/behat" -fprogress features/inschrijfproces_portal/1_aanvrager.feature
...................................................................... 70
.................................................................F..F. 140
F-

34 scenarios (31 passed, 3 failed)
142 steps (138 passed, 3 failed, 1 skipped)
2m15.74s (14.12Mb)
````

**Second run:**
````
$ "bin/behat" -fprogress features/inschrijfproces_portal/1_aanvrager.feature --rerun
..F-

1 scenario (1 failed)
4 steps (2 passed, 1 failed, 1 skipped)
0m10.02s (13.44Mb)
````
Expected: 3 scenarios
Got: 1 scenario

**Third run:**
````
$ "bin/behat" -fprogress features/inschrijfproces_portal/1_aanvrager.feature --rerun
No scenarios
No steps
0m1.36s (10.13Mb)

  [Behat\Testwork\Tester\Exception\WrongPathsException]
  No specifications found at path(s) `...\behat_rerun_cache\6f8b5d46103670016ba60fbe9c6ee5dc.scenarios`.
````
Expected: 3 scenarios (or at least 1, based on the previous - wrong - result)
Got: 0 scenarios

stof:
Which version of Behat are you using ?

WouterJ:
3.0.14 (latest stable)

everzet:
Not enough context. Could you debug?

everzet:
I'm closing this issue as it is seems to be very old and might as well have been fixed. Please reopen with updated stats if that's not the case!

PR #1199, opened 2018-12-24
Rerun should not write cache if it was explicitly set by user

oleg-andreyev:
By merging this PR `rerun` won't be enabled by default, which is minor behavior change.
To enable `rerun` user must explicitly define `--rerun` argument.

Issue #1194

### DIFF
--- a/features/rerun.feature
+++ b/features/rerun.feature
@@ -118,7 +118,7 @@ Feature: Rerun
       """
 
   Scenario: Rerun only failed scenarios
-    Given I run "behat --no-colors -f progress features/apples.feature"
+    Given I run "behat --no-colors -f progress features/apples.feature --rerun"
     When I run "behat --no-colors -f progress features/apples.feature --rerun"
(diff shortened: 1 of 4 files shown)


ciaranmcnulty:
@oleg-andreyev What's the reason for this change? I'm nervous of the BC implication

oleg-andreyev:
@ciaranmcnulty it was quite a while, I need to remember the use case.

oleg-andreyev:
@ciaranmcnulty the reason was #714 #1194, behat process is trying to write some tmp file for rerun without any notice and if /tmp is closed it's failing, I think rerun is good feature, but you should explicitly enable it

ciaranmcnulty:
Maybe it should be a softer failure? We could write if available and not if we can't

ciaranmcnulty:
Rather than change the behaviour can we have a `--no-rerun` for cases where `/tmp` isn't writeable?

PR #1204, opened 2019-01-26
Fix Symfony/Translator Deprecation Warnings on "transChoice"

mab05k:
This should fix deprecation warnings when calling the Translator in the `CounterPrinter` class. I don't think we can update to the non-deprecated interface because the library requires PHP >7.1.3.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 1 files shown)


mab05k:
I think I am going to need some help/guidance with getting the tests passing. The builds are failing with only pushing up changes to the CHANGELOG.md. Is `master` broken?

ciaranmcnulty:
No, working off master is correct 

It looks like the newer release of behat/gherkin is causing an issue

srosato:
What can we do to fix it?

PR #1208, opened 2019-02-27
Allow multiple argumentTransformers to process the same argument

das-peter:
Currently `DefinitionArgumentsTransformer::transformArgument()` bails after the first supported argumentTransformer processed the variable but in certain cases e.g. the argument is a TableNode you might have argumentTransformers that process different parts of an argument.
The made change allows all supported argumentTransformers to run on an argument.


ciaranmcnulty:
Closing for inactivity

PR #1209, opened 2019-02-27
Symfony 4.2 Compatibility .env.local

YoannChabert:
With the new system with .env.local in Sf4.2, we go through the file /src/.bootstrap.php

### DIFF
--- a/bin/behat
+++ b/bin/behat
@@ -11,7 +11,9 @@
 
 define('BEHAT_BIN_PATH',     __FILE__);
 
-if (is_file($autoload = getcwd() . '/vendor/autoload.php')) {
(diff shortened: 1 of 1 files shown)


stof:
Hardcoding such Symfony-specific thing in the Behat runner is not the good way to handle that.

and anyway, loading the .env files does not need to be done before loading the composer autoloader (and loading `vendor/autoload.php` can be done as many times as you want without any issue).

So -1 for that change.

Support for the Symfony Flex `.bootstrap.php` file is something to be handled by the Symfony extension.

YoannChabert:
Oh yeah. I understand. In fact, I have to update the boostrap.php file in my feature directory, which point on the env file i need.

Sorry for this PR :)

Hope be better next time !

PR #1212, opened 2019-03-07
Update composer package description

tkotosz:
Since the tool is not specifically for PHP 5.3 I think it make sense to remove this from the description.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
     "name": "behat/behat",
-    "description": "Scenario-oriented BDD framework for PHP 5.3",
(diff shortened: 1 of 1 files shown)


Shine-neko:
Duplicate ? https://github.com/Behat/Behat/pull/1299

tkotosz:
ye this was just to update the title in case the php requirement will still not be updated... but since now the plan is that the tool will go to php7 I am happy to close this PR and leave the composer desc change to the other PR :)

pamil:
Thank you, Tibor! :1st_place_medal:

PR #1214, opened 2019-03-11
Skipped tests should not be marked as fails

andrewnicols:
By throwing an exception in any setUp hook (beforeSuite, beforeFeature, beforeScenario) a feature is marked as Skipped. This behaviour works as anticipated if you only consider the CLI output of the run.

However, the return status for skipped tests is incorrectly reported as being failed, regardless of whether strict mode is used.

According to the documentation for strict mode:
```
      --strict                           Passes only if all tests are explicitly passing.
```
This suggests that in soft mode, only _failing_ tests should be considered a fail, and that Pending and Skipped tests should not be considered to be a failure.

This has a knock-on effect that if using the `--rerun` option, if you have no failures, but one or more _skipped_ tests then the rerun file is empty, and the exit code is non-zero, and CI systems using the return status of the Behat CLI command will automatically attempt to rerun, causing the entire suite to be re-run.

This pull request changes the resultCode to be returned as SKIPPED for any test which was not successful in the _setup_ phase. This means that soft mode only regards FAILED tests as being Failed as intended. This change does not affect the result when strict mode is used.

This change is accompanied by a four new test which ensure that both soft, and strict mode give the correct return code and result (including exit code) for both a skipped Feature, and a skipped Scenario. Since it does not make sense to skip an entire suite I have not added a test to cover this case.

### DIFF
--- a/src/Behat/Behat/Definition/Exception/BeforeStepException.php
+++ b/src/Behat/Behat/Definition/Exception/BeforeStepException.php
@@ -1,0 +1,37 @@
+<?php
+
+/*
+ * This file is part of the Behat.
+ * (c) Andrew Nicols <andrew@nicols.co.uk>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Behat\Behat\Definition\Exception;
+
+use Behat\Behat\Definition\Definition;
+use RuntimeException;
+
+/**
+ * Represents an exception caused by a BeforeStepException.
+ *
+ * If multiple step definitions in the boundaries of the same suite use same regular expression, behat is not able
+ * to determine which one is better and thus this exception is thrown and test suite is stopped.
+ *
+ * @author Andrew Nicols <andrew@nicols.co.uk>
+ */
+final class BeforeStepException extends RuntimeException implements SearchException
+{
+    /**
+     * Initializes BeforeStepException.
+     */
+    public function __construct()
+    {
+        $message = sprintf(
+            "Step skipped due to failed BeforeStep setup"
+        );
+
+        parent::__construct($message);
+    }
+}
--- a/src/Behat/Behat/Definition/Exception/SkippedStepException.php
+++ b/src/Behat/Behat/Definition/Exception/SkippedStepException.php
@@ -1,0 +1,41 @@
+<?php
+
+/*
+ * This file is part of the Behat.
+ * (c) Andrew Nicols <andrew@nicols.co.uk>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Behat\Behat\Definition\Exception;
+
+use Behat\Behat\Definition\Definition;
+use RuntimeException;
+
+/**
+ * Represents an exception caused by a redundant step definition.
+ *
+ * If multiple step definitions in the boundaries of the same suite use same regular expression, behat is not able
+ * to determine which one is better and thus this exception is thrown and test suite is stopped.
+ *
+ * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ */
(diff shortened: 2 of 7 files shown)


andrewnicols:
Okay so it seems that `@BeforeStep` hooks are also treated as test setUp.

Whilst it makes sense to consider things which happen before the test considered as setup, it feels wrong to consider actions which occur midway through the test to be setup.

andrewnicols:
I believe I've successfully converted @BeforeStep exceptions to a fail.
The remaining failures are existing failures currently seen on master (verified by Travis).

The second patch in my changeset changes the step executor to treat the first failure resulting from a Step Setup, or Step Test stage as a fail. After that point, any subsequent skip in that Scenario is treated as a Skip.
If the Feature, or Scenario caused the skip, then the test is still marked as Skipped.

MalikOvaiz:
I have relevant issue, can you please answer about this issue https://stackoverflow.com/questions/71894719/how-to-pass-test-cases-having-scenarios-with-skipped-tags-in-behat

src/Behat/Behat/Definition/Exception/BeforeStepException.php
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Represents an exception caused by a BeforeStepException.
+ *
+ * If multiple step definitions in the boundaries of the same suite use same regular expression, behat is not able
+ * to determine which one is better and thus this exception is thrown and test suite is stopped.
+ *
+ * @author Andrew Nicols <andrew@nicols.co.uk>
+ */
+final class BeforeStepException extends RuntimeException implements SearchException
```
ciaranmcnulty, line 24: This looks like a copy of RedundantStepException and isn't used anywhere? Can it be removed?

src/Behat/Behat/Definition/Exception/SkippedStepException.php
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Represents an exception caused by a redundant step definition.
+ *
+ * If multiple step definitions in the boundaries of the same suite use same regular expression, behat is not able
+ * to determine which one is better and thus this exception is thrown and test suite is stopped.
+ *
+ * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ */
+final class SkippedStepException extends RuntimeException implements SearchException
```
ciaranmcnulty, line 24: This looks like a copy of RedundantStepException and isn't used anywhere? Can it be removed?

ciaranmcnulty (commented):
I'm not sure I buy into the functionality being designed - if a setup step fails (e.g. the database is not available) why would we not want to detect it?

ciaranmcnulty:
Closing for inactivity, but please feel free to re-open

andrewnicols:
Apologies - I missed your questions in amongst the rest of my notificaitons.

I'll sort out those RedundantStepException duplicates - they were from a different branch and got mixed in by mistake.

> I'm not sure I buy into the functionality being designed - if a setup step fails (e.g. the database is not available) why would we not want to detect it?

In our case we support a wide range of:
- databases (mariadb, mysql, postgres, oracle, MS SQL server)
- cache stores (redis, memcached, mongodb)
- other optional features

On our CI testing infrastructure we cover most of those combinations, but not necessarily all at the same time. For example we only test one database server at a time and there may be a page which we want to test for Postgres which shows information about the Postgres database configuration which is obviously irrelevant to other database drivers.

Likewise we have some tests for our various caches, BigBlueButton, and a variety of other items, which only run if the external entity is configured (we usually use a range of Docker images for this). Again, we don't run these on every iteration and it gets very convoluted to have a tags configuration to "run these tags, but exclude these tags" when we could easily detect these and just skip the scenario as proposed here.

andrewnicols:
Here's relevant documentation for PHPUnit: https://phpunit.readthedocs.io/en/9.5/incomplete-and-skipped-tests.html#skipping-tests

PR #1225, opened 2019-05-23
Change the order in which configuration is loaded

### LINKED ISSUE #1224: Merging of 'default' configuration is inconsistent with merging of profile configuration in general
The configuration merged behaves differently depending on the profile you load.

`behat.basic.yml`
```yaml
default:
  extensions:
    Behat\MinkExtension:
      browser_stack:
        capabilities:
          browserstack-tunnel: true
```

`behat.production.yml`:
```yaml
imports:
  - behat.basic.yml
default:
  extensions:
    Behat\MinkExtension:
      browser_stack:
        capabilities:
          browserstack-tunnel: false
chrome:
  extensions:
    Behat\MinkExtension:
      browser_stack:
        capabilities:
          browserstack-debug: true
```

Running `behat --profile=default`
Expected:
```yaml
browserstack-tunnel: false
```

Is:
```yaml
browserstack-tunnel: false
```

Running `behat --profile=chrome`

Expected:
```yaml
browserstack-debug: true
browserstack-tunnel: false
```

Is:
```yaml
browserstack-debug: true
browserstack-tunnel: true // <- this should be FALSE
```

I find this an inconsistency in the configuration merging. I thought this would have been addressed in #230, but it seems that the fix was only for the calling of the specific profile.

func0der:
To prevent defaults from being overwritten by imports, if we load a profile other than 'default', we change the point at which the including file's 'default' definition is inserted in the config array.

Should fix #1224

### DIFF
--- a/src/Behat/Testwork/ServiceContainer/Configuration/ConfigurationLoader.php
+++ b/src/Behat/Testwork/ServiceContainer/Configuration/ConfigurationLoader.php
@@ -213,14 +213,14 @@ final class ConfigurationLoader
     {
         $configs = array();
 
-        // first load default profile from current config, but only if custom profile requested
-        if ('default' !== $profile && isset($config['default'])) {
-            $configs[] = $config['default'];
-        }
-
-        // then recursively load profiles from imports
+        // recursively load profiles from imports first
         if (isset($config['imports']) && is_array($config['imports'])) {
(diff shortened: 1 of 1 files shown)


func0der:
I provided a simple fix. I do not know, why the tests are failing. The output seems to be cut off.

func0der:
Can anyone help me out here?

PR #1230, opened 2019-06-24
PHP 7.3 support

snapshotpl:


.travis.yml
```diff
@@ -22,6 +22,8 @@ matrix:
       env: SYMFONY_VERSION='2.8.*'
     - php: 7.2
       env: SYMFONY_VERSION='^3.4'
+    - php: 7.3
+      env: SYMFONY_VERSION='^3.4'
```
Taluu: Should test `^4.2` imo.

Taluu:
We should also remove the line on symfony 2.7 as it's now dead.

ciaranmcnulty:
Thanks @snapshotpl

PR #1231, opened 2019-06-24
Lowercase `Boolean`

snapshotpl:
`Bool` type should be written by lowercase, because some tools (like psalm) trying to recognize this name as class name.


pamil:
Going further, it could be changed to `bool` to be the same as the scalar typehint.

ciaranmcnulty:
Wasn't this some legacy of Symfony coding standards?

stof:
@ciaranmcnulty yes indeed. But Symfony changed that a while ago already.

PR #1234, opened 2019-08-13
Fix tests with the latest version of gherkin

jakzal:


ciaranmcnulty:
@stof @everzet The scrutinizer check looks like it's worked, but there's a web hook or something missing

I can't force the merge as `master` is a protected branch

PR #1236, opened 2019-09-24
Add support for php 7.4

snapshotpl:


.travis.yml
```diff
@@ -26,9 +26,10 @@ matrix:
       env: SYMFONY_VERSION='^3.4'
     - php: 7.3
       env: SYMFONY_VERSION='^4.2'
-
-before_install:
-  - phpenv config-rm xdebug.ini
+    - php: 7.4snapshot
+      env: SYMFONY_VERSION='^3.4'
```
stof: no need for these 2 jobs testing 7.4 against specific SF versions. A single job running PHP 7.4 is enough (and we only need a single job forcing specifc SF LTS versions)

.travis.yml
```diff
@@ -26,9 +26,10 @@ matrix:
       env: SYMFONY_VERSION='^3.4'
     - php: 7.3
       env: SYMFONY_VERSION='^4.2'
-
-before_install:
-  - phpenv config-rm xdebug.ini
```
stof: this should be kept for versions having xdebug. Change it to `phpenv config-rm xdebug.ini || true` to avoid making it fail when xdebug is not available.

snapshotpl:
Looks like Transliterator's dependecy is not ready for 7.4

stof:
@snapshotpl any chance to contribute there first then ?

snapshotpl:
@stof Done https://github.com/Behat/Transliterator/pull/26

snapshotpl:
Waiting for v1.3 of behat/transliterator appear on packagist https://packagist.org/packages/behat/transliterator

ciaranmcnulty:
@snapshotpl The transliterator tag is released now on packagist

ciaranmcnulty:
(we don't need 'snapshot' now I think)

snapshotpl:
One test in php 7.4 doesn't pass. @ciaranmcnulty can you help me to resolve this https://travis-ci.org/Behat/Behat/jobs/637319029#L316 ?

ciaranmcnulty:
Hm, something must have changed around error handling...

Sam-Burns:
@ciaranmcnulty @snapshotpl we really need to delete the offending test at https://github.com/snapshotpl/Behat/blob/ea89bd18f290a26c0c7fc0e8aa3205242bf7e5b8/features/junit_format.feature#L529-L587 .

It's tagged only to run in PHP 5.3 and 5.4. The system by which these tags are used in Travis is a bit of a mess because:
- Detection of whether we're in PHP 5.4 is done using the minor version _only_, hence problems in PHP 7.4.
- This problem didn't come up with PHP 7.3 because of a missing decimal point, meaning it was trying to run tests tagged as `php53` but not `php5.3`.

I'll do a separate pull request cleaning up the whole fiasco at https://github.com/Behat/Behat/blob/7aa8161be7499ecc8b73ccc6dcd18f4062e6bf8e/.travis.yml#L40 , but deleting this test is definitely necessary as it is tagged only to run in PHP versions we don't even seem to be testing in anymore.

The rest of your PHP looks great, thanks so much for doing this.

snapshotpl:
@Sam-Burns nice catch! So I'm waiting for PR to clean this an then I will rebase your fix here

Sam-Burns:
Regarding the Appveyor and Travis builds: I think if we merge (or you rebase on top of) https://github.com/Behat/Behat/pull/1273 and https://github.com/Behat/Behat/pull/1270, this PR should be ready to merge.

snapshotpl:
@Sam-Burns awesome. Now waiting for fixed builds. @ciaranmcnulty ?

snapshotpl:
@ciaranmcnulty ready to merge!

pamil:
Thanks, Witold! :tada:

PR #1238, opened 2019-09-24
The PHP DOM extension is required for JUnit

pfrenssen:
I'm getting the following error when running a test scenario on a minimal Ubuntu Bionic image:

```
Fatal error: Class 'DOMDocument' not found (Behat\Testwork\Call\Exception\FatalThrowableError)
```

It is thrown in [`Behat\Testwork\Output\Printer\JUnitOutputPrinter`](https://github.com/Behat/Behat/blob/master/src/Behat/Testwork/Output/Printer/JUnitOutputPrinter.php#L62).

### DIFF
--- a/src/Behat/Testwork/Output/Printer/JUnitOutputPrinter.php
+++ b/src/Behat/Testwork/Output/Printer/JUnitOutputPrinter.php
@@ -10,6 +10,7 @@
 
 namespace Behat\Testwork\Output\Printer;
 
+use Behat\Testwork\Output\Exception\MissingExtensionException;
 use Behat\Testwork\Output\Printer\Factory\FilesystemOutputFactory;
 use Symfony\Component\Console\Output\OutputInterface;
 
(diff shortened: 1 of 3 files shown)


jakzal:
It's only required by the junit output printer and not technically needed for other Behat features to work.

Would it make more sense if we checked if the extension is present in the JUnitOutputPrinter class, and throw an exception if it's not?

pfrenssen:
Yes that is also an option, then we can move it to `suggest`.

src/Behat/Testwork/Output/Printer/JUnitOutputPrinter.php
```diff
@@ -57,6 +57,10 @@ public function __construct(FilesystemOutputFactory $outputFactory)
      */
     public function createNewFile($name, array $testsuitesAttributes = array())
     {
+        // This requires the DOM extension to be enabled.
+        if (!extension_loaded('dom')) {
+            throw new \Exception('The PHP DOM extension is required to generate JUnit reports.');
```
jakzal: I'd throw something more specific, like `RuntimeException`.

jakzal: There are also some more Printer specific exceptions in this package: https://github.com/Behat/Behat/tree/master/src/Behat/Testwork/Output/Exception

I think the intention might've been for Printer implementations to throw exceptions implementing the PrinterException.

pfrenssen:
Sorry for the delay, this fell off my radar after I moved on to other tasks. I have updated the PR with a dedicated exception which implements `PrinterException`.

Issue #1076, opened 2017-09-14
3.4.0 breaks the ability to run individual tests


pfrenssen:
### Update + quick fix

This is a long standing issue which is affecting many users. It has been solved but the PR is not yet accepted. If you are affected by this you can solve it by adding the following lines to your `composer.json`:

```
"require-dev": {
    "behat/gherkin": "^4.6.0",
    "cweagans/composer-patches": "^1.6.7",
},
"patches": {
    "behat/gherkin": {
        "Correctly filter features, now that the base path is correctly set": "https://github.com/Behat/Gherkin/compare/v4.6.0...a7d84eb6ba245ea268a304312d14660788bf6c29.patch"
    }
}
```

Note that this problem is not caused directly by Behat but it is due to a confusion between absolute and relative paths in the `Behat/Gherkin` library. We are keeping this issue open in the Behat queue so that it is easier to find for people who encounter this problem when using Behat.

If possible please vote for + review + approve the fix at https://github.com/Behat/Gherkin/pull/129 - hopefully once it is approved it can be merged.

### Original summary

This was originally reported at https://github.com/Behat/Gherkin/issues/128 but it appears to be a bug in Behat and not in Gherkin.

In https://github.com/Behat/Behat/pull/1056 a change was introduced to make it possible to run individual tests from inside a different folder than where the tests are located. Unfortunately this breaks the more common use case of running the tests from inside the actual folder where they are located.

To replicate this, create a folder structure in which the tests are located in a subfolder of the root of the project, e.g in the `tests/` folder:

```
$ composer require behat/behat-3.4.0
$ mkdir tests && cd tests/
$ ../vendor/bin/behat --init
# Create a test
$ vim features/homepage.feature
# Make sure the gherkin cache is cleared, this might "hide" the problem
$ rm -rf /tmp/behat_*
# Try to run a single test
$ ../vendor/bin/behat features/homepage.feature -vvv
```

Result:

```
$ ../vendor/bin/behat features/homepage.feature -vvv


No scenarios
No steps
0m0.73s (9.22Mb)

                                                                                                                                            
  [Behat\Testwork\Tester\Exception\WrongPathsException]                                                                                     
  No specifications found at path(s) `features/homepage.feature`. This might be because of incorrect paths configuration in your `suites`.  
                                                                                                                                            

Exception trace:
 () at /home/pieter/v/joinup-dev/vendor/behat/behat/src/Behat/Testwork/Tester/Cli/ExerciseController.php:111
[...]

PR #1239, opened 2019-09-25
Fix specifications locator to find feature files

milkovsky:
This is my fix for the https://github.com/Behat/Behat/issues/1076 issue.
Looks like feature files were not discovered properly. I have extended the locator and iterator for this.

Before:
```
./bin/behat features/migrate/articles.feature
...
No specifications found at path(s) `features/migrate/articles.feature`. This might be because of incorrect paths configuration in your `suites`.
```

After:
```
./bin/behat features/migrate/articles.feature
...
1 scenario (1 passed)
2 steps (2 passed)
```

P.S. Tests\Behat\Gherkin\Loader\GherkinFileLoaderTest shows no errors.

### DIFF
--- a/src/Behat/Behat/Gherkin/Specification/LazyFeatureIterator.php
+++ b/src/Behat/Behat/Gherkin/Specification/LazyFeatureIterator.php
@@ -64,12 +64,14 @@ final class LazyFeatureIterator implements SpecificationIterator
      * @param Gherkin           $gherkin
      * @param string[]          $paths
      * @param FilterInterface[] $filters
+     * @param FeatureNode[]     $features
      */
-    public function __construct(Suite $suite, Gherkin $gherkin, array $paths, array $filters = array())
+    public function __construct(Suite $suite, Gherkin $gherkin, array $paths, array $filters = array(), $features = array())
     {
         $this->suite = $suite;
         $this->gherkin = $gherkin;
         $this->paths = array_values($paths);
+        $this->features = $features;
         $this->filters = array_merge($this->getSuiteFilters($suite), $filters);
     }
 
--- a/src/Behat/Behat/Gherkin/Specification/Locator/FilesystemFeatureLocator.php
+++ b/src/Behat/Behat/Gherkin/Specification/Locator/FilesystemFeatureLocator.php
@@ -76,8 +76,13 @@ final class FilesystemFeatureLocator implements SpecificationLocator
 
         if ($locator) {
             $filters = array(new PathsFilter($suiteLocators));
+            $features = array();
+            foreach ($this->findFeatureFiles($locator) as $featureFile) {
+                // Collect all the feature specification files.
+                $features = array_merge($features, $this->gherkin->load($featureFile, $filters));
+            }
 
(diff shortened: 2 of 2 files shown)


milkovsky:
Patch for your composers:
[behat-run-single-feature.patch.txt](https://github.com/Behat/Behat/files/3652984/behat-run-single-feature.patch.txt)

milkovsky:
image
Yeeeeiii

src/Behat/Behat/Gherkin/Specification/LazyFeatureIterator.php
```diff
@@ -64,12 +64,14 @@ final class LazyFeatureIterator implements SpecificationIterator
      * @param Gherkin           $gherkin
      * @param string[]          $paths
      * @param FilterInterface[] $filters
+     * @param FeatureNode[]     $features
      */
-    public function __construct(Suite $suite, Gherkin $gherkin, array $paths, array $filters = array())
+    public function __construct(Suite $suite, Gherkin $gherkin, array $paths, array $filters = array(), $features = array())
```
joshlopes: missing the type `array`

src/Behat/Behat/Gherkin/Specification/Locator/FilesystemFeatureLocator.php
```diff
@@ -76,8 +76,13 @@ public function locateSpecifications(Suite $suite, $locator)
 
         if ($locator) {
             $filters = array(new PathsFilter($suiteLocators));
+            $features = array();
+            foreach ($this->findFeatureFiles($locator) as $featureFile) {
+                // Collect all the feature specification files.
+                $features = array_merge($features, $this->gherkin->load($featureFile, $filters));
```
joshlopes: i imagine this to be quite slow for big projects. wondering if we can improve it somehow.

jakzal:
Tests are failing. This patch seems to be duplicating features in some cases.

jakzal:
I don't think this is the right approach. It should be possible to fix this bug without changing the existing interfaces.

Before https://github.com/Behat/Behat/commit/08802b7f2d736b88df339e5411a1c722e92adbe3 it all worked fine, so I expect the issue is related to normalising the path, rather than changing the logic of finding features.

src/Behat/Behat/Gherkin/Specification/Locator/FilesystemFeatureLocator.php
milkovsky: Not sure. In my case the $locator is an exact feature file path. In this case foreach will loop only 1 time.
In addiition I guess it will be as "slow" as the next foreach block in this file below.

milkovsky:
>  It should be possible to fix this bug without changing the existing interfaces.
Thank you for the feedback. I simply did not find another way to set the `features` property. I guess we can just add setFeatures() method or smth similar to the FeatureIterator.php.

> Before 08802b7 it all worked fine, 
Good to know. I've just installed behat a few days ago, I am not familiar with the previous versions. That was a quick workaround from my side to get the tests working. Feel free to close this PR if you preffer a different approach.

pfrenssen:
https://github.com/Behat/Gherkin/pull/129 is the fix for this, already since 2017

PR #1241, opened 2019-09-26
add php 7.4

andreybolonin:


jakzal:
To progress with the build update the line that disables xdebug:

`phpenv config-rm xdebug.ini || echo "XDebug is not enabled"`

stof:
I'm -1 for this PR, in favor of https://github.com/Behat/Behat/pull/1236/files, because these PR is adding the 2 jobs that I rejected during the review of the other one, and does not add the job running unmodified deps.

ciaranmcnulty:
Agree with @stof

PR #1244, opened 2019-10-02
Show only user-land trace for exceptions and errors in very verbose mode

wouterj:
When some of my tests error, I almost always rerun with `-vvv` to get the trace of the error. However, most of this trace is taken by Behat's function call's. They never are interesting (unless we're debugging some Behat bug).

That's why I propose to only show user-land trace in very verbose (`-vv`) mode and only the full trace in debug verbose (`-vvv`) mode.

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -348,6 +348,13 @@ EOL;
                     return str_replace('/', DIRECTORY_SEPARATOR, $matches[0]);
                 }, $text
             );
+
+            // error stacktrace
+            $text = preg_replace_callback(
+                '/#\d+ [^:]+:/', function ($matches) {
+                    return str_replace('/', DIRECTORY_SEPARATOR, $matches[0]);
+                }, $text
+            );
         }
 
         return $text;
@@ -402,10 +409,13 @@ EOL;
     {
         $output = $this->process->getErrorOutput() . $this->process->getOutput();
 
-        // Normalize the line endings in the output
+        // Normalize the line endings and directory separators in the output
         if ("\n" !== PHP_EOL) {
             $output = str_replace(PHP_EOL, "\n", $output);
         }
+
+        // Remove location of the project
+        $output = str_replace(realpath(dirname(dirname(__DIR__))).DIRECTORY_SEPARATOR, '', $output);
 
         // Replace wrong warning message of HHVM
         $output = str_replace('Notice: Undefined index: ', 'Notice: Undefined offset: ', $output);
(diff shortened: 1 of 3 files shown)


src/Behat/Testwork/Exception/ExceptionPresenter.php
```diff
@@ -93,13 +94,45 @@ public function presentException(Exception $exception, $verbosity = null)
             }
         }
 
-        if (OutputPrinter::VERBOSITY_VERY_VERBOSE <= $verbosity) {
-            return $this->relativizePaths(trim($exception));
+        if (OutputInterface::VERBOSITY_VERY_VERBOSE <= $verbosity) {
+            if (OutputInterface::VERBOSITY_DEBUG > $verbosity) {
```
wouterj: the old `OutputPrinter` verbosity constants were still used. These are deprecated and the ones from Symfony's `OutputInterface` are used throughout Behat. As their values are not equal, we should change them here as well.

src/Behat/Testwork/Exception/ExceptionPresenter.php
```diff
@@ -93,13 +94,45 @@ public function presentException(Exception $exception, $verbosity = null)
             }
         }
 
-        if (OutputPrinter::VERBOSITY_VERY_VERBOSE <= $verbosity) {
-            return $this->relativizePaths(trim($exception));
```
stof: trimming got lost

features/bootstrap/FeatureContext.php
```diff
@@ -402,14 +402,17 @@ private function getOutput()
     {
         $output = $this->process->getErrorOutput() . $this->process->getOutput();
 
-        // Normalize the line endings in the output
+        // Normalize the line endings and directory separators in the output
         if ("\n" !== PHP_EOL) {
-            $output = str_replace(PHP_EOL, "\n", $output);
+            $output = str_replace(array(PHP_EOL, DIRECTORY_SEPARATOR), array("\n", '/'), $output);
```
stof: this seems to be breaking tests on Windows actually.

features/error_reporting.feature
```diff
@@ -56,6 +56,13 @@ Feature: Error Reporting
               PHPUnit\Framework\Assert::assertEquals($arg1, $this->result);
           }
 
+          /**
+           * @When an exception is thrown
+           */
+          public function anExceptionIsThrown()
+          {
+              throw \Exception();
```
stof: missing `new`

wouterj:
Thanks for the review. I've fixed the windows tests. Unfortunately, the string representation of exceptions were changed from PHP 5.x to PHP 7.x. So I've mark the tests 7.x only: they do not depend on something special that is different in 5.x and 7.x, so we can be sure that if a 7.x test is green, it would be green in 5.x as well.

PR #1247, opened 2019-11-06
addition of support for symfony 5

flug:

PR #1249, opened 2019-11-08
removal of old versions for the update to sf5

flug:
pre-pull-request for the removal of versions not maintained by symfony and the removal of old versions of php that are no longer maintained. This will facilitate the migration of behat code to the latest symfony 5 components. 

pr ref : https://github.com/Behat/Behat/pull/1247

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: php
 
-php: [5.6, 7.0, 7.1, 7.2, 7.3]
+php: [7.1, 7.2, 7.3]
 
 sudo: false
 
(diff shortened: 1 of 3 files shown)


Taluu:
That would be awesome to stop supporting php 5... But alas, it's not planned until at least 4.0. Unless it has changed since then (Let a man dream 😂 )

flug:
@Taluu it might just be the beginning of version 4  :stuck_out_tongue_winking_eye:

ciaranmcnulty:
What @Taluu said, basically

flug:
@ciaranmcnulty is the removal of php 5.* critical for you or is it really necessary to remove it? to know if my other pull request should be modified or not?

rpkamp:
So when is 4.x planned?

pamil:
Do we really need to wait for 4.x to remove PHP 5 support? It is not a BC break: https://www.doctrine-project.org/2017/07/25/php-7.1-requirement-and-composer.html

Taluu:
It's not on the sense of semver but that's what is applied on Behat (and most projects from symfony ecosystem)... :{

ciaranmcnulty:
As I said, we're not planning to drop support yet

PR #1259, opened 2019-11-30
Initial stab at creating scoped PHARs

rpkamp:
This PR can be used to build a scoped PHAR where all classes in vendor namespaces are prefixed with `Behat\`. So for example, `Symfony\Component\Console\Command\Command` becomes `Behat\Symfony\Component\Console\Command\Command`. That way, if people have Symfony installed in their project as well, it won't clash with the version that is bundled with Behat.

This works fine for the base use case, but falls apart at the moment Behat extensions come into the mix, because behat extensions are classes that need to use parts of Symfony (such as `Symfony\Component\DependencyInjection\ContainerBuilder`), which is not possible inside the PHAR, because those classes are prefixed in the PHAR (so `Behat\Symfony\Component\DependencyInjection\ContainerBuilder` exists in the PHAR, but `Symfony\Component\DependencyInjection\ContainerBuilder` does _not_).

Output when trying to run the scoped PHAR with my [behat mailhog extension](https://github.com/rpkamp/mailhog-behat-extension):

```
PHP Fatal error:  Declaration of rpkamp\Behat\MailhogExtension\ServiceContainer\MailhogExtension::configure(Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition $builder): void must be compatible with Behat\Testwork\ServiceContainer\Extension::configure(Behat\Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition $builder) in /home/remon/code/mailhog/mailhog-behat-extension/src/ServiceContainer/MailhogExtension.php on line 31
```

I'm not really sure how to solve this in a nice way.

Creating an abstraction layer between Behat and Symfony might work (basically a bridge between Behat and the Symfony version Behat works with) but that would force all current extensions to be rewritten in order to be compatible with that abstraction layer, which would be a lot of work.

Another solution would be for extensions to be written differently (using the scoped namespace) when used in the PHAR than when used outside the PHAR, which sounds brittle and puts quite a strain on extension maintainers.

But maybe there are other solutions that I'm missing?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,11 +42,11 @@ script:
   - behat -fprogress --strict --tags '~@php-version,'`php php_version_tags.php`
 
 before_deploy:
-  - curl -LSs https://box-project.github.io/box2/installer.php | php
-  - export PATH=.:$PATH
+  - wget https://github.com/humbug/box/releases/download/3.8.3/box.phar
+  - chmod +x box.phar
   - rm -Rf ./vendor
   - composer install --no-dev -o
-  - box.phar build
+  - ./box.phar compile -vvv
 
 deploy:
   provider: releases
--- a/box.json
+++ b/box.json
@@ -1,9 +1,9 @@
 {
-    "chmod": "0755",
     "directories": ["src"],
     "files": [
         "LICENSE",
-        "i18n.php"
+        "i18n.php",
+        "vendor/symfony/translation/Resources/data/parents.json"
     ],
     "finder": [
         {
@@ -12,8 +12,10 @@
             "in": "vendor"
         }
     ],
-    "compactors": "Herrera\\Box\\Compactor\\Php",
-    "main": "bin/behat",
+    "compactors": [
+        "KevinGH\\Box\\Compactor\\Php",
+        "KevinGH\\Box\\Compactor\\PhpScoper"
+    ],
     "output": "behat.phar",
-    "stub": true
+    "stub": "build/stub.php"
 }
--- a/build/stub.php
+++ b/build/stub.php
@@ -1,0 +1,38 @@
+#!/usr/bin/env php
+<?php
+
+Phar::mapPhar('behat');
+
+// Load the autoloader specific to the project Behat is being used in so
+// any installed behat extensions and feature contexts can be autoloaded
+if (is_file($projectAutoloader = behatPhar_getProjectDir() . '/vendor/autoload.php')) {
+    var_dump('Require '.$projectAutoloader);
+    require $projectAutoloader;
+}
+
+require 'phar://behat/vendor/autoload.php';
+
+define('BEHAT_BIN_PATH', Phar::running(false));
+
+/**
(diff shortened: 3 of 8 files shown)


Taluu:
Have you tried to have a look at how phpunit does it ? It supports extensions pretty well. And I didn't see anywhere on phpunit if it declared (and thus maintained...) a custom autoloadder.

If memory serves right, it depends on the "local" (where the tool is executed) composer, from a bootstrap file.

rpkamp:
> Have you tried to have a look at how phpunit does it ?

I have. This PR is largely based on how they do it.

> It supports extensions pretty well.

It has full support for extensions, which is easy for them, because their extensions don't rely on third party code. Behat extensions on the other hand _do_ rely on third party code (i.e. Symfony). Entirely different situation.

> And I didn't see anywhere on phpunit if it declared (and thus maintained...) a custom autoloadder.

Correct, because PHPUnit doesn't autoload the test cases. IIRC they just `require` all `*Test.php` files in your `tests/` directory by default (can be overriden of course). 
Behat however instructs the composer autoloader how to load feature contexts from the `features/` directory.

> If memory serves right, it depends on the "local" (where the tool is executed) composer, from a bootstrap file.

Correct, but Behat can't do that, because it won't be able to find the FeatureContexts. Unless we tell people they need to add them to their `composer.json`'s `autoload` section. In that case Behat could drop autoloading completely. But I'm not sure if that's such a great idea...

rpkamp:
Also, copying composer's autoloader was just a first attempt to make it work. We could replace it with a custom / simpler implementation if it's decided this is the way to go, since about half of the methods aren't being used by Behat.

This was just get it to work. Make it work, make it pretty, make it fast :wink:

rpkamp:
The more I think about this, the more I think it would make sense to have the loading of FeatureContext done by composer in the project itself instead of taking it on inside Behat, that would make creating a scoped phar much easier.

This would mean a BC break however, and people having to set up autoloading for it in their `composer.json` somehow:

```
"autoload-dev": {
    "psr4": {
        "": "features/bootstrap/",
    }
}
```

ciaranmcnulty:
I recommend you get it working with humbug/box (and auto deploying*) as a first step, without worrying about scoping and then layer on the scoping in a later PR

I'm up for removing the Behat autoloading in $some_version - it's only there because not every project used composer in the old days. It'd remove some complexity (but be a different PR).

The general approach of scoping the PHAR is +1 from me if it helps us move more rapidly on upgrading the non-PHP dependencies, especially symfony. I guess that's what causes the issues with extensions - I'm not particularly aware of any other projects using PHARs and also symphony's extension mechanism :/

* This seems cleaner with GitHub actions

rdoering:
We run into a case where needed this feature and hope that it will be merge.

rpkamp:
>  I'm not particularly aware of any other projects using PHARs and also symphony's extension mechanism :/

Yes, that is the main issue for sure. The easiest way around this would be to create a custom interface inside Behat for extensions. Internally that could still use Symfony, but it shouldn't be exposed to the client at all.

PR #1264, opened 2019-12-17
ExtensionManager#getExtension() can also return null

Lctrs:


Taluu:
Why ? It doesn't mke sense...

Lctrs:
Look at the body of the function. It does return `null` if there is no extension defined with the given `$key`.

pamil:
Thank you, @Lctrs! :1st_place_medal:

Issue #1255, opened 2019-11-22
Remove (abandonded) `container-interop/container-interop` dependency


bobdenotter:
Hi, 

Now that https://github.com/container-interop/container-interop/issues/94 has been done, composer update gives this mildly annoying message: 

```
Package container-interop/container-interop is abandoned, you should avoid using it. Use psr/container instead.
```

It's trivial in the greater scheme of things, but it'd be nice if this could be remedied.

danepowell:
There's already an open issue for this from quite some time ago: https://github.com/Behat/Behat/issues/1110

It's unfortunate that this deprecation was known about for two years and nothing was done about it...

Taluu:
As I already said on linked issue, this is considered to be a bc break, even though it's an acceptable one.

Ping @ciaranmcnulty, @stof and @jakzal

pamil:
Fixed in #1281.

Issue #1110, opened 2017-12-09
Drop `container-interop/container-interop` dependency


mdrost:
As `container-interop/container-interop` [is going to be abandoned](https://github.com/container-interop/container-interop/issues/94) I encourage you to drop that dependency and rely only on `psr/container`.

Taluu:
Would be a BC Break (but imo would be an acceptable one)

dimacros:
The package is deprecated. You must remove the dependency.

danepowell:
To be clear, apart from the stability and security concerns related to relying on deprecated packages, all users of Behat will now get a scary yellow deprecation warning like this from Composer:
> Package container-interop/container-interop is abandoned, you should avoid using it. Use psr/container instead.

pamil:
Fixed by #1281.

tedw:
FYI `container-interop/container-interop` was added back to `composer.json` in Sept. 2021 https://github.com/Behat/Behat/pull/1281#issuecomment-1111383428

PR #1267, opened 2020-01-13
Drop container-interop/container-interop from required packages

gjuric:
As noted in #1110 and #1255 this package has been abandoned and
replaced with psr/container.

I understand that this is a BC break, but once the changes for Symfony 5 get merged in maybe a new version could be released.


gjuric:
It looks like it's more complex then just removing it from the list of dependencies.

Will reopen if I get the time to clean up everything.

PR #1274, opened 2020-01-29
ReturnTypeTransformation: Add support for php 7.4

stephanedelprat:
> Fatal error: 8192: Function ReflectionType::__toString() is deprecated in vendor/behat/behat/src/Behat/Behat/Transformation/Transformation/ReturnTypeTransformation.php line 133

Since getName() is not documented, I am not sure from which PHP version it is available.

Behat 3.5 is for PHP 5.6+, is that correct ?

In case `getName()` is missing in some PHP version it would require a : 

```
return method_exists($type, 'getName') ? $type->getName(); : (string) $type;
```

### DIFF
--- a/src/Behat/Behat/Transformation/Transformation/ReturnTypeTransformation.php
+++ b/src/Behat/Behat/Transformation/Transformation/ReturnTypeTransformation.php
@@ -130,7 +130,7 @@ final class ReturnTypeTransformation extends RuntimeCallee implements SimpleArgu
             return null;
         }
 
-        return $type->getName();
+        return method_exists($type, 'getName') ? $type->getName(); : (string) $type;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Behat/Behat/Transformation/Transformation/ReturnTypeTransformation.php
```diff
@@ -130,7 +130,7 @@ static private function getReturnClass(ReflectionFunctionAbstract $reflection)
             return null;
         }
 
-        return (string) $type;
+        return $type->getName();
```
Daniel-Marynicz: This will brake compatibility with php 5.3

stephanedelprat: Ok, I changed it

PR #1275, opened 2020-01-31
fix php 7.1 deprecation for ReflectionType::__toString

Daniel-Marynicz:


rpkamp:
Not that this is a duplicate of https://github.com/Behat/Behat/pull/1274

I personally like this solution better (less magic).

Daniel-Marynicz:
i also added  https://github.com/Behat/Behat/issues/1276

weaverryan:
Do we know that getName() was added in 7.1? If so, this looks good to me. If not, what version was it added in? If it has always existed, we could *just* use it always.

Daniel-Marynicz:
> Do we know that getName() was added in 7.1? If so, this looks good to me. If not, what version was it added in? If it has always existed, we could _just_ use it always.
It is 
https://www.php.net/manual/en/class.reflectionnamedtype.php  
And and in this doc we have info (PHP 7 >= 7.1.0)

Daniel-Marynicz:
> > Do we know that getName() was added in 7.1? If so, this looks good to me. If not, what version was it added in? If it has always existed, we could _just_ use it always.
> > It is
> > https://www.php.net/manual/en/class.reflectionnamedtype.php
> > And and in this doc we have info (PHP 7 >= 7.1.0)

And probably i need also check `instanceof \ReflectionNamedType`

weaverryan:
Or you could *just* do the instanceof check and no version check?

Daniel-Marynicz:
> Or you could _just_ do the instanceof check and no version check?

ok changed this to only instanceof  check.

ciaranmcnulty:
Thanks!

PR #1256, opened 2019-11-23
Update dependencies to allow Symfony 5.x components

japicoder:
Two days ago the lastest version of the Symfony framework was finally release, including a lot of new and interesting improvements.

For that reason, at the Symfony Hackday we are doing an effort to have all libraries ready for Symfony 5.0.

#SymfonyHackday

PR #1278, opened 2020-02-05
Fix fatal due to wrong symfony/event-dispatcher version detection

### LINKED ISSUE #1277: Fatal error on 3.6.0 if installed alongside older symfony/event-dispatcher (e.g. due to guzzle)
The 3.6.0 release is producing a fatal error on two of our projects, with this trace:

```
Fatal error: Declaration of TestworkEventDispatcherPhp72Trait::dispatch($event, ?string $eventName = NULL): object must be compatible with Symfony\Component\EventDispatcher\EventDispatcher::dispatch($eventName, ?Symfony\Component\EventDispatcher\Event $event = NULL) in /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/behat/behat/src/Behat/Testwork/EventDispatcher/TestworkEventDispatcher.php on line 25

Call Stack:
    0.0002     407808   1. {main}() /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/behat/behat/bin/behat:0
    0.0223    3670008   2. Behat\Testwork\Cli\Application->run() /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/behat/behat/bin/behat:34
    0.0299    4066528   3. Behat\Testwork\Cli\Application->doRun() /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/symfony/console/Application.php:145
    0.0299    4066560   4. Behat\Testwork\Cli\Application->createCommand() /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/behat/behat/src/Behat/Testwork/Cli/Application.php:122
    0.0299    4066560   5. Behat\Testwork\Cli\Application->createContainer() /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/behat/behat/src/Behat/Testwork/Cli/Application.php:161
    0.0478    5164888   6. Behat\Testwork\ServiceContainer\ContainerLoader->load() /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/behat/behat/src/Behat/Testwork/Cli/Application.php:185
    0.0775    6846376   7. Behat\Testwork\ServiceContainer\ContainerLoader->loadExtensions() /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/behat/behat/src/Behat/Testwork/ServiceContainer/ContainerLoader.php:67
    0.0881    7695040   8. Behat\Testwork\ServiceContainer\ContainerLoader->loadExtension() /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/behat/behat/src/Behat/Testwork/ServiceContainer/ContainerLoader.php:130
    0.0884    7700128   9. Behat\Testwork\Output\ServiceContainer\OutputExtension->load() /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/behat/behat/src/Behat/Testwork/ServiceContainer/ContainerLoader.php:159
    0.0885    7702208  10. Behat\Testwork\Output\ServiceContainer\OutputExtension->loadFormatters() /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/behat/behat/src/Behat/Testwork/Output/ServiceContainer/OutputExtension.php:125
    0.0885    7702208  11. Behat\Behat\Output\ServiceContainer\Formatter\PrettyFormatterFactory->buildFormatter() /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/behat/behat/src/Behat/Testwork/Output/ServiceContainer/OutputExtension.php:186
    0.0885    7702208  12. Behat\Behat\Output\ServiceContainer\Formatter\PrettyFormatterFactory->loadRootNodeListener() /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/behat/behat/src/Behat/Behat/Output/ServiceContainer/Formatter/PrettyFormatterFactory.php:64
    0.0885    7705672  13. spl_autoload_call() /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/behat/behat/src/Behat/Behat/Output/ServiceContainer/Formatter/PrettyFormatterFactory.php:101
    0.0885    7705752  14. Composer\Autoload\ClassLoader->loadClass() /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/behat/behat/src/Behat/Behat/Output/ServiceContainer/Formatter/PrettyFormatterFactory.php:101
    0.0885    7705752  15. Composer\Autoload\includeFile() /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/composer/ClassLoader.php:322
    0.0886    7709208  16. include('/var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/behat/behat/src/Behat/Behat/EventDispatcher/Event/BackgroundTested.php') /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/composer/ClassLoader.php:444
    0.0886    7709208  17. spl_autoload_call() /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/behat/behat/src/Behat/Behat/EventDispatcher/Event/BackgroundTested.php:22
    0.0886    7709288  18. Composer\Autoload\ClassLoader->loadClass() /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/behat/behat/src/Behat/Behat/EventDispatcher/Event/BackgroundTested.php:22
    0.0886    7709288  19. Composer\Autoload\includeFile() /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/composer/ClassLoader.php:322
    0.0887    7713152  20. include('/var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/behat/behat/src/Behat/Testwork/EventDispatcher/Event/LifecycleEvent.php') /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/composer/ClassLoader.php:444
    0.0887    7713152  21. spl_autoload_call() /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/behat/behat/src/Behat/Testwork/EventDispatcher/Event/LifecycleEvent.php:22
    0.0887    7713208  22. Composer\Autoload\ClassLoader->loadClass() /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/behat/behat/src/Behat/Testwork/EventDispatcher/Event/LifecycleEvent.php:22
    0.0887    7713208  23. Composer\Autoload\includeFile() /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/composer/ClassLoader.php:322
    0.0888    7715136  24. include('/var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/behat/behat/src/Behat/Testwork/Event/Event.php') /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/composer/ClassLoader.php:444
    0.0888    7715136  25. spl_autoload_call() /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/behat/behat/src/Behat/Testwork/Event/Event.php:7
    0.0888    7715216  26. Composer\Autoload\ClassLoader->loadClass() /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/behat/behat/src/Behat/Testwork/Event/Event.php:7
    0.0888    7715216  27. Composer\Autoload\includeFile() /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/composer/ClassLoader.php:322
    0.0889    7726768  28. include('/var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/behat/behat/src/Behat/Testwork/EventDispatcher/TestworkEventDispatcher.php') /var/www/www.edbookfest/releases/e3b6c9d6c8b804e4c0bd8b2293ef854bd2a152f6/vendor/composer/ClassLoader.php:444
```

On investigating, these projects are both using guzzle 3.9.3. 

That requires `symfony/event-dispatcher: ~2.1` with the result that the event-dispatcher package is pinned at 2.8.52.

However, they are also both using symfony/console 4.2 - which requires symfony/contracts (at 1.1.8)

This dependency setup is valid as far as composer/symfony/our app at runtime is concerned. 

However, behat [uses the presence of symfony/contracts to decide which version of symfony is in use](https://github.com/Behat/Behat/blob/master/src/Behat/Testwork/EventDispatcher/TestworkEventDispatcher.php#L21). Based on this, it switches between two different definitions of TestworkEventDispatcher to be compatible with both older and newer versions of symfony.

Unfortunately as can be seen this test is not sufficiently robust.

We have a couple of other projects that use guzzle 3, and the old event-dispatcher, but don't use symfony/console, and these are passing fine. So the two versions of the TestworkEventDispatcher seem to be OK, it's just that it's not safe to rely on checking for the symfony/contracts class to decide which to use.

acoulton:
The version detection added in #1256 to support symfony 5 relies
only on checking for the presence of a class provided by the
(separate) symfony/event-dispatcher-contracts package.

However, symfony/console 4.2 also requires symfony/contracts
but can coexist with older versions of symfony/event-dispatcher.

Therefore for example if a project has a constraint on updating
symfony/event-dispatcher (e.g. because of using Guzzle 3,
which requires event-dispatcher ~2.1) composer will install:

* symfony/console @ 4.2.12 (latest that doesn't advertise a
  conflict with newer event-dispatcher)
* symfony/contracts @ 1.1.8 (latest that satisfies the console
  requirement)
* symfony/event-dispatcher @ 2.8.52 (latest that satisfies
  guzzle)

In this setup the presence of symfony/contracts incorrectly
signals to behat that we are running symfony 5 and causes a
fatal error when it loads the new implementation of
TestworkEventDispatcher.

The only change between the two versions is that the method
signature for dispatch has changed order and types.

Instead of looking at packages, it is more robust just to use
reflection to detect which version of the dispatch() method
is defined, and then load the extension class that matches
that signature.

Fixes #1277


src/Behat/Testwork/EventDispatcher/TestworkEventDispatcher.php
```diff
@@ -17,8 +17,27 @@
...
+    $reflection  = new \ReflectionClass(\Symfony\Component\EventDispatcher\EventDispatcher::class);
+    $dispatch    = $reflection->getMethod('dispatch');
+    $first_param = $dispatch->getParameters()[0];
+    switch ($first_param->getName()) {
+        case 'event':
+            // This is the new Symfony 5 event dispatcher interface
+            // This test retained from existing code. In reality it's unclear that it could ever be false, since
+            // symfony/event-dispatcher 5.x requires php: ^7.2.5 in their composer, and before that the class has the
+            // old signature.
+            return PHP_VERSION_ID > 70200;
```
acoulton: I have left this PHP > 7.2 check in for now, as I'm not clear quite why it was added originally. It's not possible to have symfony 5 on an older php version so I think it was always redundant.

acoulton:
The appveyor build appeared to fail due to not being able to find php, I'm not sure what that means but I don't think it can be related to my changes.

src/Behat/Testwork/EventDispatcher/TestworkEventDispatcher.php
```diff
@@ -17,8 +17,27 @@
...
-if (class_exists(\Symfony\Contracts\EventDispatcher\Event::class) && PHP_VERSION_ID > 70200) {
+        case 'eventName':
+            return false;
+
+        default:
+            throw new \UnexpectedValueException('Could not identify which version of symfony/event-dispatcher is in use, could not define a compatible TestworkEventDispatcher');
+    }
+};
+
+if ($isSymfony5AndPhpGreater72()) {
```
lchrusciel: Instead of checking method existance, what do you think about using by default old implementation? 
```suggestion
if (class_exists(\Symfony\Component\EventDispatcher\EventDispatcher::class) && PHP_VERSION_ID > 70200) {
```

src/Behat/Testwork/EventDispatcher/TestworkEventDispatcher.php
ciaranmcnulty: You're correct, this could be removed

src/Behat/Testwork/EventDispatcher/TestworkEventDispatcher.php
```diff
@@ -17,8 +17,27 @@
  *
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
  */
+$isSymfony5AndPhpGreater72 = function () {
```
ciaranmcnulty: as we're in global scope here I'd rather not create a variable - can you make this self-calling?

```
if ((function(){ ... })()) {
```

src/Behat/Testwork/EventDispatcher/TestworkEventDispatcher.php
acoulton: I'm not sure how that would help? `\Symfony\Component\EventDispatcher\EventDispatcher` exists in all versions of symfony - the class name hasn't changed, just the signature of `dispatch`

ciaranmcnulty: Which is an annoying 'backward compatible' change

Taking a step back, why do we need two implementations of the class? can't we extend in a way that copes with either?

lchrusciel: My bad, wrong class checked, as I mistaken Event and EventDispatcher.

src/Behat/Testwork/EventDispatcher/TestworkEventDispatcher.php
acoulton: @ciaranmcnulty me either but I don't think it can be self-calling on php5 which I think you still want to support? Best I could do would probably be to unset the variable immediately after the if block.

Or I could unset it sooner if I moved the logic to a script beside the class definition so this file becomes `if (require(__DIR__.'/_identify_symfony_eventdispatcher_version.php))` and that script cleans up all its variables before it returns.

src/Behat/Testwork/EventDispatcher/TestworkEventDispatcher.php
acoulton: Removed the 7.2 check, can squash it all up once we have a solution to the global lambda

src/Behat/Testwork/EventDispatcher/TestworkEventDispatcher.php
ciaranmcnulty: yeah don't worry - I forgot we were supporting such a wide range

src/Behat/Testwork/EventDispatcher/TestworkEventDispatcher.php
acoulton: @ciaranmcnulty I don't think so for two reasons:

a) symfony < and > 5 have hard typehints on dispatch that are not compatible, and you extend and replace the inherited dispatch method (I've not looked into why that is). There's no way to define your own `dispatch` with a signature that's compatible with both symfony versions.

b) even if you could somehow do that, symfony5 uses php 7.2 hints (`string` and ` : object`) that will generate parse errors if running behat on php5 which I assume is why there's the odd magic of defining that method in a trait to prevent the file being loaded in an older php environment.

Worth noting the symfony change isn't intended to be backward compatible, the method signature changed in 5.0. The issue is behat wants to attempt to support symfony 2.x - 5.x all from a single release of the behat codebase. 

It would probably have been better to make symfony5 support a breaking change and optional upgrade (and php7-only at the same time) on behat - there's a very large amount of magic/variation/duplication going on in #1256 to support both versions at once.

ciaranmcnulty: Thanks for the explanation, it makes a lot of sense

OK, to make this cleaner how about we do this?

- Move the 2 implementations into separate autoloadable files (maybe remove the trait if we can) called something like `-Legacy` and `-Symfony5`
- Make this file do the `if` and then a `class_alias` to whichever implementation should be loaded
- have a dummy implementation inside an `if(false)` here to fool the composer autoloader

acoulton:
@ciaranmcnulty ok so.... it turns out the php72 check is required *is* required because as of symfony/event-dispatcher 4.3 [they have been doing magic themselves in the dispatch method](https://github.com/symfony/event-dispatcher/blob/4.3/EventDispatcher.php) to make it accept both old and new arguments so:

* On php < 7.1.3 the absolute latest event-dispatcher you can get is 3.4 which has an official signature of `public function dispatch($eventName, Event $event = null)` [here](https://github.com/symfony/event-dispatcher/blob/3.4/EventDispatcher.php#L37)
* On php >= 7.1.3 with event-dispatcher 4.4 the signature is `public function dispatch($event/*, string $eventName = null*/)` note the second arg is actually commented out [here](https://github.com/symfony/event-dispatcher/blob/4.4/EventDispatcher.php#L51) which means the Testwork class can add it as an extra optional param without php complaining about incorrect inheritance
* On php >= 7.2.5 with event-dispatcher 5.0 the signature is `public function dispatch(object $event, string $eventName = null): object`

And note those are cumulative - e.g. on php7.2 you could be running any of (2.x/3.x/4.0/4.1/4.2 - all equivalent, 4.3/4.4 or 5.0.

So I still think a PHP_VERSION constraint is useless / misleading albeit putting it back in has made the specific combinations on your travis go green again.

But I think if you actually want to support all of these, we need to sniff for all three variants of the method signature.

I'm also not sure whether it's safe to use the legacy extension class on 4.3/4.4 since it doesn't do any of the symfony magic to guess which order the arguments are in. If it's only called internally within behat it's probably fine, if anything from symfony core is dispatching events then there may be other unpredictable results not yet encountered.

acoulton:
If this was my package, I'd be tempted to revert all the symfony5 support and release a 3.7, then work on a new 4.0 branch requiring symfony5 and php7. Appreciate you may not want to go that way.

ciaranmcnulty:
> If this was my package, I'd be tempted to revert all the symfony5 support and release a 3.7, then work on a new 4.0 branch requiring symfony5 and php7.

Sadly that won't remove the tag of 3.6.0 :/

ciaranmcnulty:
I agree it sounds like sniffing the three signatures (FFS) is the right approach - I'd like to do some sort of class_alias to separate out the different implementations from the decision logic

acoulton:
@ciaranmcnulty 

> Sadly that won't remove the tag of 3.6.0 :/

true but we could at least upgrade past it (and I suspect a lot of people don't have 3.6.0 yet so would go straight to 3.7.0 or higher...)

I think I have a robust sniff for the signatures, if travis passes this time. I agree splitting into class_alias is probably the way to go, I need to take a break just now but will try to look later on.

src/Behat/Testwork/EventDispatcher/TestworkEventDispatcher.php
```diff
@@ -17,28 +17,37 @@
  *
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
  */
-$isSymfony5AndPhpGreater72 = function () {
+$isSymfony5EventDispatcher = function () {
     $reflection  = new \ReflectionClass(\Symfony\Component\EventDispatcher\EventDispatcher::class);
     $dispatch    = $reflection->getMethod('dispatch');
+
+    if ($dispatch->getNumberOfParameters() === 1) {
+        // This is the 4.3 / 4.4 version, which has `public function dispatch($event/*, string $eventName = null*/)` and
```
ciaranmcnulty: Great that you've written these comments 👍

acoulton: Always nice to know what on earth I was thinking later on :)

src/Behat/Testwork/EventDispatcher/TestworkEventDispatcher.php
acoulton: This is dealt with in the new class_alias version, the var is unset as soon as we've used it so nothing leaks to global state.

acoulton:
@ciaranmcnulty that's the fix rewritten to use a class_alias based approach as requested.

ciaranmcnulty:
This looks great, thank you

PR #1279, opened 2020-02-05
Drop container-interop/container-interop from required packages

slavcodev:
The package has been abandoned and replaced with psr/container.

Related issues:  #1110, #1255

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,8 +24,7 @@
         "symfony/event-dispatcher": "^2.7.51 || ^3.0 || ^4.0 || ^5.0",
         "symfony/translation": "^2.7.51 || ^3.0 || ^4.0 || ^5.0",
         "symfony/yaml": "^2.7.51 || ^3.0 || ^4.0 || ^5.0",
-        "psr/container": "^1.0",
-        "container-interop/container-interop": "^1.2"
+        "psr/container": "^1.0"
     },
 
     "require-dev": {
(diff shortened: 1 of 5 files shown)


ciaranmcnulty:
How can we do this without the BC issues?

Taluu:
Well, the package is abandoned but it extends the right one (`psr/container`)... So maybe we could switch the asked interface for the real one, make a conflict for the version of this package and then extensions would have to require the deprecated package if they want to continue to work otherwise ?

I'm aware it's a tiny BC break, but it's best that way IMO.

ciaranmcnulty:
I guess we could ship our own implementations?

ciaranmcnulty:
Or maybe we can support the container if used but not `require` it in composer?

slavcodev:
How does this break BC? Interlop interfaces extend PSR, so all end-user imlementation will still work.

Taluu:
Extensions rely on behat, but can be transitively relying on the interlop  (big mistake, transitive dependency, I'm aware).

But then again, I think removing the require and using the psr interfaces is fine.

slavcodev:
> can be transitively relying on the interlop

Well, these are their mistakes, those who do not properly manage dependencies.

Taluu:
I'm one but TBH I don't really care, plan to use the proper interface if Behat uses it in the next release. :joy:

upyx:
Hello!
Guys, it is the really annoying warning in all projects which use Behat. Why not just bump the version? If someone had been so retrograde to rely on an abandoned package, let him use the old version of Behat.

upyx:
I've created another PR without BC break. Let's get it over with.

PR #1281, opened 2020-02-11
Make container-interop/container-interop optional dependency

upyx:
Alternative way to #1279 without BC break.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,8 @@ matrix:
     - php: 5.6
       env: DEPENDENCIES='low'
     - php: 5.6
+      env: SKIP_INTEROP='~@interop&&'
+    - php: 5.6
       env: SYMFONY_VERSION='2.7.*'
     - php: 5.6
       env: SYMFONY_VERSION='2.8.*'
@@ -34,13 +36,14 @@ before_install:
 
 before_script:
   - if [ "$SYMFONY_VERSION" != "" ]; then composer require --no-update "symfony/symfony:${SYMFONY_VERSION}"; fi;
+  - if [ "$SKIP_INTEROP" != "" ]; then composer remove --dev --no-update container-interop/container-interop; fi;
   - if [ "$DEPENDENCIES" != "low" ]; then composer update; fi;
   - if [ "$DEPENDENCIES" = "low" ]; then composer update --prefer-lowest; fi;
   - export PATH=./bin:$PATH
 
 script:
   - ./vendor/bin/phpunit
-  - behat -fprogress --strict --tags '~@php-version,@php'`php -r 'echo PHP_MAJOR_VERSION;'`
+  - behat -fprogress --strict --tags "$SKIP_INTEROP"'~@php-version,@php'`php -r 'echo PHP_MAJOR_VERSION;'`
 
 before_deploy:
   - curl -LSs https://box-project.github.io/box2/installer.php | php
--- a/src/Behat/Behat/HelperContainer/Exception/NotFoundException.php
+++ b/src/Behat/Behat/HelperContainer/Exception/NotFoundException.php
@@ -1,0 +1,27 @@
+<?php
+
+/*
+ * This file is part of the Behat.
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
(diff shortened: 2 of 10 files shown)


upyx:
@Taluu what is needed to merge PR?

Taluu:
Add tests to show nothing breaks ?

E.g having a container that extends the psr interface and passes, and another one that uses the old interop interface and works ?

And also, the 3rd party extensions will need to require the psr or the interop, but I guess it should be done anyway.

That is why it is still a minor BC break, but an acceptable one IMO.

src/Behat/Behat/HelperContainer/Exception/NotFoundException.php
```diff
@@ -0,0 +1,24 @@
...
+
+class_alias(
+    interface_exists('Interop\\Container\\Exception\\NotFoundException')
+        ? 'Interop\\Container\\Exception\\NotFoundException'
+        : 'Psr\\Container\\NotFoundExceptionInterface',
+    'Behat\\Behat\\HelperContainer\\Exception\\NotFoundException'
+);
+
+if (false) {
+    interface NotFoundException
```
stof: I suggest adding a phpdoc on these fake interface to tag them as `@internal`. We don't want external packages to use them.

upyx:
> Add tests to show nothing breaks ?
> 
Done :wink:

upyx:
Should I rebase and squash it?

.travis.yml
```diff
@@ -36,12 +38,13 @@ before_script:
   - if [ "$SYMFONY_VERSION" != "" ]; then composer require --no-update "symfony/symfony:${SYMFONY_VERSION}"; fi;
   - if [ "$DEPENDENCIES" != "low" ]; then composer update; fi;
   - if [ "$DEPENDENCIES" = "low" ]; then composer update --prefer-lowest; fi;
+  - if [ "$SKIP_INTEROP" != "" ]; then composer remove --dev container-interop/container-interop; fi;
```
stof: this should be before the `composer update` commands (with a `composer remove --no-update`. Otherwise, we might loose the `--prefer-lowest` effect.

.travis.yml
```diff
@@ -16,6 +16,8 @@ matrix:
   include:
     - php: 5.6
       env: DEPENDENCIES='low'
+    - php: 5.6
+      env: DEPENDENCIES='low' SKIP_INTEROP='~@interop&&'
```
stof: IMO, that should not be a `DEPENDENCIES='low'` job.

upyx: Why do you think so? `container-interop` is an abandon package, so that is test backward compatibility with the oldest supported versions.

stof: well, this job is not about testing `container-interop`

upyx: Job? I don't understand. It seem like ... um ... environment variable. What do you suggest?

.travis.yml
upyx: Thank you!

.travis.yml
upyx: I've found your description about that in other PR. I agree with you.

mvorisek:
Can a release with this be made?

pamil:
@mvorisek I'll try to release it in the nearest week or so.

tedw:
FYI `container-interop/container-interop` was added back to `composer.json` in this commit for some reason:
https://github.com/Behat/Behat/commit/706b5a549077fe04fb22808b0d19d95b628455f4#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R34

I’m still getting this warning:
>Package container-interop/container-interop is abandoned, you should avoid using it. Use psr/container instead.” warning.

stof:
@tedw no. This commit only added a comma at the end of the line

stof:
And the line in the commit you mentioned is in the `require-dev` section, which will not have any effect when you use `behat/behat` in your own projects. Run `composer why container-interop/container-interop` to see which package depends on it

tedw:
@stof Thanks for the quick reply! Apologies for the confusion on my part. I just started working on a preexisting site and was still seeing the warning but it turned out to be some dependency that was loading an older version of behat.

Thanks again for the reply and your work on behat!

PR #1282, opened 2020-02-13
Updated the year on Changelog dates

choult:
Updated the year on Changelog dates for v3.6.0 and v3.6.1 to match actual release dates


ciaranmcnulty:
Oops, thanks

Issue #1280, opened 2020-02-06
packagist requires php


michield:
https://packagist.org/packages/behat/behat#v3.6.1

says requires PHP 5.3.3 and up, but it is now 5.5 and up.
It won't work with PHP 5.4

stof:
What do you mean by "it is now 5.5. and up` ?

michield:
I can't remember exactly, but I linted the code with php5.4 and it gives errors, due to some parts having syntax that is [not 5.4 compatible](https://www.php.net/manual/en/migration55.new-features.php). IIRC it was the Classname::method notation.

pamil:
Will be fixed by #1284.

PR #1270, opened 2020-01-16
Fix issues with PHP version handling in build

Sam-Burns:
Builds are failing in PHP7.4, because of an issue with the build scripts.

Tests tagged as`php5.4` are running in 7.4, because only the minor version is checked.

This pull request fixes these issues, and simplifies the handling of language version numbers generally.

PR #1284, opened 2020-02-15
Restore PHP 5.3/5.4 compat

dvdoug:
Behat 3.6 has broken support for PHP <5.5 even though `composer.json` is still claiming to support PHP >= 5.3.3.

This PR hopefully fixes that and adds PHP 5.3, 5.4 and 5.5 to CI as it currently doesn't seem to test these older versions 🤔 

Fully behind you guys dropping support for older PHP in the future, just please update `composer.json` first 😉

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,15 @@ branches:
 
 matrix:
   include:
+    - php: 5.3
+      dist: precise
+      env: DEPENDENCIES='low'
+    - php: 5.4
+      dist: trusty
+      env: DEPENDENCIES='low'
+    - php: 5.5
+      dist: trusty
+      env: DEPENDENCIES='low'
     - php: 5.6
       env: DEPENDENCIES='low'
     - php: 5.6
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,13 +16,10 @@ matrix:
   include:
     - php: 5.3
       dist: precise
-      env: DEPENDENCIES='low'
     - php: 5.4
       dist: trusty
-      env: DEPENDENCIES='low'
     - php: 5.5
       dist: trusty
-      env: DEPENDENCIES='low'
     - php: 5.6
       env: DEPENDENCIES='low'
     - php: 5.6
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,20 +8,20 @@ cache:
   directories:
     - $HOME/.composer/cache
 
-branches:
-  only:
-    - master
-
 matrix:
   include:
     - php: 5.3
       dist: precise
+      env: PHP_VERSION=53
     - php: 5.4
       dist: trusty
+      env: PHP_VERSION=54+
     - php: 5.5
       dist: trusty
+      env: PHP_VERSION=54+
     - php: 5.6
       env: DEPENDENCIES='low'
+      env: PHP_VERSION=54+
     - php: 5.6
       env: SKIP_INTEROP='~@interop&&'
     - php: 5.6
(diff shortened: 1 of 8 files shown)


dvdoug:
Scrutinizer failure is a false-positive, there's nothing I can do about that. I've reported it, and they say there's already a ticket their side for it.

This should be good to go

upyx:
May be is it the time to drop support of old PHP versions?

dvdoug:
> 
> 
> May be is it the time to drop support of old PHP versions?

Need to fix it first though, otherwise running `composer update` breaks projects (I discovered this through a broken build)

upyx:
> Scrutinizer failure is a false-positive, there's nothing I can do about that.

It failures because of the first '\\' symbol. It should be skipped in fully-qualified class names:
```php
interface_exists('Symfony\\Component\\Translation\\TranslatorInterface')
```

dvdoug:
The leading slash is equivalent to not having one, see rule #1 on https://www.php.net/manual/en/language.namespaces.rules.php. Happy to remove though.

The failure on line 10 is because the interface exists on the version of Symfony that isn't installed on Scrutinzer and the static analysis isn't picking up on the fact that it's guarded by `interface_exists` on line 9.

.travis.yml
```diff
@@ -14,6 +14,15 @@ branches:
 
 matrix:
   include:
+    - php: 5.3
+      dist: precise
+      env: DEPENDENCIES='low'
+    - php: 5.4
+      dist: trusty
+      env: DEPENDENCIES='low'
```
stof: I suggest removing these `DEPENDENCIES='low'`. We should rather test on higher (compatible) versions than on lowest ones, to match what our users will get (we need one job using `DEPENDENCIES='low'` in our matrix, but if you look at it, you will see that it also has other normal jobs running on that PHP version)

dvdoug: Done

ciaranmcnulty:
I updated from master and the 5.3 build doesn't seem to be working

dvdoug:
@ciaranmcnulty Traits are 5.4 and https://github.com/Behat/Behat/pull/1270/files seems to have removed that restriction from the test suite

dvdoug:
I'm not sure there's an elegant way of excluding that test file from the 5.3 build using the revised tag system, I'm thinking it might be easier to simply `rm` it inside Travis on the 5.3 build? How would maintainers feel about that?

Another idea would be adding `Given my PHP version is >= x.y` as step definition to the test suite but that seems like a lot of extra code for the purpose of skipping 1 test.

Or...could temporarily revert #1270, release a final 5.3 supporting version and then put #1270 back on master alongside a bumped minimum PHP version?

Sam-Burns:
@dvdoug This approach seems to be passing on Travis, if it is of any interest https://github.com/Sam-Burns/Behat/compare/master...Sam-Burns:restore_php54_compat

Sam-Burns:
Would be good to get your fix included, and do a patch release with working PHP 5.3 support. After speaking to @ciaranmcnulty, it sounds like it may well be the last release to support EOL PHP versions 🍾

dvdoug:
> 
> 
> Would be good to get your fix included, and do a patch release with working PHP 5.3 support. After speaking to @ciaranmcnulty, it sounds like it may well be the last release to support EOL PHP versions 🍾

Awesome @Sam-Burns, I've cherry-picked that one into my branch and rebased on master.

Sam-Burns:
Scrutinizer is complaining about a reference to a nonexistent interface, although frankly that is the entire point of the relevant file. I'd suggest ignoring it, if it's possible to merge the PR anyway.

dvdoug:
> 
> 
> Scrutinizer is complaining about a reference to a nonexistent interface, although frankly that is the entire point of the relevant file. I'd suggest ignoring it, if it's possible to merge the PR anyway.

Yep, it's a false positive dating back to before this PR. Any chance this could get merged/released @ciaranmcnulty?

pamil:
Fixes #1280.

pamil:
I've rebased it with master, but it's failing on one of the builds connected to the container interop. @dvdoug, could you please take a look at it?

pamil:
Thanks, Doug! :tada:

PR #1299, opened 2020-06-03
Bump requirements to PHP ^7.2 and Symfony ^4.4|^5.0

### LINKED ISSUE #1258: Say farewell to unsupported Symfony versions
This library does an amazing job supporting Symfony 2.7, 2.8 and 3.3. But some day we need to decide that it is time to stop supporting versions that has reached EOL. Keeping support for these versions will put too much maintenance work. 
(I know [this PR](https://github.com/Behat/Behat/pull/1256) is struggling to make sure we support 4 major versions of symfony)

I suggest to only support Symfony 3.4.3 and greater.

If we not are going to decide to drop the old versions now, when is a better time?

pamil:
Fixes #1258.

This PR does not remove any dead code, just bumps up requirements and adjusts CI configs.

What to do later?
 - Remove dead code that brings compatibility for Symfony <4.4
 - Remove dead code that brings compatibility for PHP <7.2
 - Setup some coding standard tool in order to maintain a consistent code style (and replace those `array()` with `[]`)
 - Merge some cool stuff 🎉

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,16 @@
 language: php
 
+os: linux
+
+dist: xenial
+
 php: [7.2, 7.3, 7.4]
 
 cache:
   directories:
     - $HOME/.composer/cache
 
-matrix:
+jobs:
   include:
     - php: 7.2
       env: SKIP_INTEROP='~@interop&&'
@@ -38,9 +42,9 @@ before_deploy:
 
 deploy:
   provider: releases
-  api_key: $GITHUB_API_KEY
+  token: $GITHUB_API_KEY
   file: behat.phar
-  skip_cleanup: true
+  cleanup: false
   on:
     tags: true
     php: 5.6
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,10 +12,6 @@ matrix:
       env: SKIP_INTEROP='~@interop&&'
     - php: 7.2
       env: DEPENDENCIES='low'
-    - php: 7.2
-      env: SYMFONY_VERSION='^4.4'
-    - php: 7.3
-      env: SYMFONY_VERSION='^4.4'
     - php: 7.4
       env: SYMFONY_VERSION='^4.4'
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ cache:
 jobs:
   include:
     - php: 7.2
-      env: SKIP_INTEROP='~@interop&&'
+      env: SKIP_INTEROP=true
     - php: 7.2
       env: DEPENDENCIES='low'
     - php: 7.4
@@ -31,7 +31,7 @@ before_script:
 
 script:
   - ./vendor/bin/phpunit
-  - behat -fprogress --strict --tags "${SKIP_INTEROP}~@random"
+  - behat -fprogress --strict $([ -z "$SKIP_INTEROP" ] || echo "--tags ~@interop")
 
 before_deploy:
   - curl -LSs https://box-project.github.io/box2/installer.php | php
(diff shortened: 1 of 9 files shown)


.travis.yml
```diff
@@ -45,12 +27,11 @@ before_script:
   - if [ "$SKIP_INTEROP" != "" ]; then composer remove --dev --no-update container-interop/container-interop; fi;
   - if [ "$DEPENDENCIES" != "low" ]; then composer update; fi;
   - if [ "$DEPENDENCIES" = "low" ]; then composer update --prefer-lowest; fi;
-  - if [ ! "$PHP_VERSION" ]; then export PHP_VERSION=`php -r 'echo PHP_MAJOR_VERSION;'`; fi;
   - export PATH=./bin:$PATH
 
 script:
   - ./vendor/bin/phpunit
-  - behat -fprogress --strict --tags "${SKIP_INTEROP}~@php-version,@php${PHP_VERSION}"
+  - behat -fprogress --strict --tags "${SKIP_INTEROP}~@random"
```
pamil, line 34: We only need to pass `--tags "~@interop"` for build skipping the interop, any ideas how to make it cleaner?

features/junit_format.feature
```diff
@@ -526,66 +526,6 @@
       """
     And the file "junit/default.xml" should be a valid document according to "junit.xsd"
 
-  @php-version @php53 @php54+
```
pamil, line 529: It has been only run with PHP versions that are unsupported now.

.travis.yml
```diff
@@ -1,41 +1,23 @@
...
-    - php: 7.3
-      env: SYMFONY_VERSION='^3.4'
-    - php: 7.3
-      env: SYMFONY_VERSION='^4.3'
+      env: DEPENDENCIES='low'
+    - php: 7.2
+      env: SYMFONY_VERSION='^4.4'
     - php: 7.3
-      env: SYMFONY_VERSION='^5.0'
+      env: SYMFONY_VERSION='^4.4'
```
stof, line 18: Do we actually need 3 jobs running against the Symfony 4.4 LTS ?

pamil: I guess not really, don't have a better solution for now though - we could iterate over it later.

stof: I mean, just remove some of these jobs, keeping only one testing against the 4.4 LTS. We already have earlier jobs testing each PHP version.

pamil: Left PHP 7.4 & Symfony 4.4 only.

pamil:
Also fixes some deprecations in existing Travis configuration:
CleanShot 2020-06-05 at 12 00 39

.travis.yml
ciaranmcnulty: Remove the `&&` from the var and then something like `$([ -z "$SKIP_INTEROP" ] || echo "$SKIP_INTEROP&&")` (note outer parentheses)

composer.json
```diff
@@ -14,22 +14,22 @@
...
+        "symfony/translation": "^4.4 || ^5.0",
+        "symfony/yaml": "^4.4 || ^5.0",
         "psr/container": "^1.0"
     },
 
     "require-dev": {
-        "symfony/process": "~2.5 || ^3.0 || ^4.0 || ^5.0",
-        "phpunit/phpunit": "^4.8.36 || ^6.5.14 || ^7.5.20",
+        "symfony/process": "^4.4 || ^5.0",
+        "phpunit/phpunit": "^7.5.20",
```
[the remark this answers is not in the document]
stof: Well, a next step would be to migrate to PHPUnit 8: https://phpunit.de/supported-versions.html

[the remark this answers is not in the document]
ciaranmcnulty: oh it's require-dev, ignore me we can bump :)

PR #1309, opened 2020-06-18
Print example row step text with exception

davidgrayston:
### Changed
- Prints failing step text above the exception message when an example row fails

I'm not sure if this is the correct approach to fix – happy to refine after some feedback.

#### Notes:
- The _progress_ format already outputs the failing step text along with exception message.
- This isn't needed for the step printer as the exception is printed directly below the failing step text.


ciaranmcnulty:
I like the idea but I think we should give some more context

e.g. something like:

```
    Failed on step: "Then I should have 8 apples"
    Failed asserting that 7 matches expected 8.
```

davidgrayston:
If we add additional text, I think we'd also want to translate it. Perhaps to simplify this it could just be:
```
    Step: "Then I should have 8 apples"
    Failed asserting that 7 matches expected 8.
```
(Then translate _"Step"_)

davidgrayston:
I've attempted to add a translated title/label in 968b18e

I think it might now be a breaking change - I can revert back to no label/title if we think it's unnecessary?

ciaranmcnulty:
I like it, thanks. Don't think it's particularly a BC break

PR #1310, opened 2020-06-19
Use new PHP > 5.3 features

rpkamp:
A first batch of refactorings to use newer PHP construction where possible.

Notably absent are:

- Scalar type hints
- Use short array notation
- `::class` constants instead of strings with FQCNs

I think these deserve PRs of their own, given their impact.

### DIFF
--- a/features/error_reporting.feature
+++ b/features/error_reporting.feature
@@ -148,7 +148,6 @@ Feature: Error Reporting
           When an exception is thrown # features/exception_in_scenario.feature:7
             Exception: Exception is thrown in features/bootstrap/FeatureContext.php:56
             Stack trace:
-            #0 [internal function]: FeatureContext->anExceptionIsThrown()
-            #1 src/Behat/Testwork/Call/Handler/RuntimeCallHandler.php(109): call_user_func_array(Array, Array)
-            #2 src/Behat/Testwork/Call/Handler/RuntimeCallHandler.php(64): Behat\Testwork\Call\Handler\RuntimeCallHandler->executeCall(Object(Behat\Behat\Definition\Call\DefinitionCall))
+            #0 src/Behat/Testwork/Call/Handler/RuntimeCallHandler.php(109): FeatureContext->anExceptionIsThrown()
+            #1 src/Behat/Testwork/Call/Handler/RuntimeCallHandler.php(64): Behat\Testwork\Call\Handler\RuntimeCallHandler->executeCall()
     """
--- a/features/error_reporting.feature
+++ b/features/error_reporting.feature
@@ -132,7 +132,6 @@ Feature: Error Reporting
           When an exception is thrown # features/exception_in_scenario.feature:7
             Exception: Exception is thrown in features/bootstrap/FeatureContext.php:56
             Stack trace:
-            #0 [internal function]: FeatureContext->anExceptionIsThrown()
 
     1 scenario (1 failed)
     1 step (1 failed)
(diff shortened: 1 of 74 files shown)


rpkamp:
Build failure seems unrelated; locally `master` fails on the same feature ...

ciaranmcnulty:
@rpkamp I re-ran travis on master and it passed, so not sure what's going on

ciaranmcnulty:
Bisecting locally, the failure is in commit d96dd39ad20a8d0dcfdb

ciaranmcnulty:
Maybe you just need to apply a change like this one to the scenario after it https://github.com/Behat/Behat/pull/1310/commits/d96dd39ad20a8d0dcfdb4c5162f163bfdf80f160#diff-fc346e939877ba944b146dfe72567079L135

features/error_reporting.feature
```diff
@@ -132,7 +132,6 @@ Feature: Error Reporting
           When an exception is thrown # features/exception_in_scenario.feature:7
             Exception: Exception is thrown in features/bootstrap/FeatureContext.php:56
             Stack trace:
-            #0 [internal function]: FeatureContext->anExceptionIsThrown()
```
ciaranmcnulty, line 135: To fix the tests you need to apply a change like this to the scenario below

src/Behat/Behat/Transformation/Context/Annotation/TransformationAnnotationReader.php
```diff
@@ -64,18 +64,13 @@ public function readCallee($contextClass, ReflectionMethod $method, $docLine, $d
...
-
-        if (PHP_VERSION_ID >= 70000) {
-            $transformations[] = 'Behat\Behat\Transformation\Transformation\TokenNameAndReturnTypeTransformation';
-            $transformations[] = 'Behat\Behat\Transformation\Transformation\ReturnTypeTransformation';
-        }
-
-        $transformations[] = 'Behat\Behat\Transformation\Transformation\TokenNameTransformation';
-
-        return $transformations;
+        return array(
```
ciaranmcnulty, line 67: you may as well use `[]` for the array

rpkamp: I've thought about that, but the risk of doing that is that you'll end up with a mix of both styles throughout the entire codebase. So I thought it would be better to at least keep consistent now with how it was and than later replace all of it in one swoop. Probably using something like Rector.

features/error_reporting.feature
rpkamp: That seems to have fixed it 🎉 

Weirdest thing though, locally it doesn't work 😕 

Locally I get
```
#1 src/Behat/Testwork/Call/Handler/RuntimeCallHandler.php(64): Behat\Testwork\Call\Handler\RuntimeCallHandler->executeCall()
```

instead of 
```
#1 src/Behat/Testwork/Call/Handler/RuntimeCallHandler.php(64): Behat\Testwork\Call\Handler\RuntimeCallHandler->executeCall(Object(Behat\Behat\Definition\Call\DefinitionCall))
```

(so `Object(Behat\Behat\Definition\Call\DefinitionCall)` is missing from the output)

but Travis and AppVoyer both have it in the output.

No idea why that is. I've tried with and without XDebug locally, doesn't matter.

I'm running PHP 7.4.3 in Ubuntu 20.04 in WSL2.

guilliamxavier: @rpkamp: you probably have the php.ini setting `zend.exception_ignore_args` (introduced in PHP 7.4) to `On`, try turning it `Off` (source: https://stackoverflow.com/questions/60000391/php-missing-function-arguments-in-exception-stack-trace)

rpkamp: I'll give that a whirl later, thanks! If that is indeed the case we should force that option to `Off` in the CI as well (if that isn't the case already).

rpkamp: @guilliamxavier Yup, that was it 🙂 
I've created a PR to ensure this setting is off in CI: https://github.com/Behat/Behat/pull/1312

pamil:
Thanks, Remon! :1st_place_medal:

PR #1312, opened 2020-06-23
Ensure zend.exception_ignore_args is disabled in CI

rpkamp:
When enabled the stack trace output for exceptions changes, which causes
the tests to fail as the changed output does not match the expected
output.


pamil:
Thank you, Remon! :tada:

PR #1313, opened 2020-06-25
PHP8 support

ciaranmcnulty:
This is deprecated in PHP 8 and generates deprecation warnings (https://github.com/php/php-src/pull/5209). 

Because we require PHP >=7.1 we can rely on getType returning a ReflectionNamedType

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,8 @@ jobs:
       env: DEPENDENCIES='low'
     - php: 7.4
       env: SYMFONY_VERSION='^4.4'
+    - php: nightly
+      env: PHP8=true
 
 before_install:
   - phpenv config-rm xdebug.ini || echo "XDebug is not enabled"
@@ -25,14 +27,15 @@ before_install:
 before_script:
   - if [ "$SYMFONY_VERSION" != "" ]; then composer require --no-update "symfony/symfony:${SYMFONY_VERSION}"; fi;
   - if [ "$SKIP_INTEROP" != "" ]; then composer remove --dev --no-update container-interop/container-interop; fi;
-  - if [ "$DEPENDENCIES" != "low" ]; then composer update; fi;
+  - if [ "$DEPENDENCIES" != "low" ]; then composer update $([ -z "$PHP8" ] || echo "--ignore-platform-reqs") ; fi;
   - if [ "$DEPENDENCIES" = "low" ]; then composer update --prefer-lowest; fi;
   - echo 'zend.exception_ignore_args=Off' >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
   - export PATH=./bin:$PATH
 
 script:
   - ./vendor/bin/phpunit
-  - behat -fprogress --strict $([ -z "$SKIP_INTEROP" ] || echo "--tags ~@interop")
+  - behat -fprogress --strict $([ -z "$SKIP_INTEROP" ] || echo "--tags ~@interop&&~@php8")
+  - if [ "$PHP8" = true ]; then behat -fprogress --strict --tags=@php8; fi;
 
 before_deploy:
   - curl -LSs https://box-project.github.io/box2/installer.php | php
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,11 +15,13 @@ jobs:
     - php: 7.2
       env: SKIP_INTEROP=true
     - php: 7.2
-      env: DEPENDENCIES='low'
+      env: COMPOSER_FLAGS='--prefer-lowest'
     - php: 7.4
       env: SYMFONY_VERSION='^4.4'
     - php: nightly
-      env: PHP8=true
+      env:
+        - PHP8=true
+        - COMPOSER_FLAGS='--ignore-platform-reqs'
 
 before_install:
   - phpenv config-rm xdebug.ini || echo "XDebug is not enabled"
@@ -27,8 +29,7 @@ before_install:
 before_script:
   - if [ "$SYMFONY_VERSION" != "" ]; then composer require --no-update "symfony/symfony:${SYMFONY_VERSION}"; fi;
   - if [ "$SKIP_INTEROP" != "" ]; then composer remove --dev --no-update container-interop/container-interop; fi;
-  - if [ "$DEPENDENCIES" != "low" ]; then composer update $([ -z "$PHP8" ] || echo "--ignore-platform-reqs") ; fi;
-  - if [ "$DEPENDENCIES" = "low" ]; then composer update --prefer-lowest; fi;
+  - composer update $COMPOSER_FLAGS
   - echo 'zend.exception_ignore_args=Off' >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
   - export PATH=./bin:$PATH
 
--- a/features/autowire.feature
+++ b/features/autowire.feature
@@ -247,7 +247,7 @@ Feature: Helper services autowire
     When I run "behat --no-colors -f progress features/autowire.feature"
     Then it should pass
 
-    @php8
+  @php8
   Scenario: Union constructor arguments
     Given a file named "features/autowire.feature" with:
       """
--- a/features/autowire.feature
+++ b/features/autowire.feature
@@ -12,6 +12,7 @@ Feature: Helper services autowire
     - It only wires arguments that weren't otherwise set
     - Services must be last arguments in step definitions
     - Services must be last arguments in transformations
+    - Autowiring is not yet triggered for union types
 
   Background:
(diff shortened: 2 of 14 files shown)


src/Behat/Testwork/Argument/MixedArgumentOrganiser.php
```diff
@@ -227,7 +227,7 @@ private function filterApplicableTypehintedParameters(array $parameters)
             }
 
             try {
-                $reflectionClass = $parameter->getClass();
+                $reflectionClass = $parameter->getType() ? $parameter->getType()->getName() : null;
```
rpkamp, line 230: As far as I can see none of these methods throws an exception, does it still need a `try / catch` here?

Also, it looks like this can now also return non-class types, like `array`, `int`, etc. Does that matter?

ciaranmcnulty:
For PHP8 this needs to cope with union types so we need to check we are getting a `ReflectionNamedType` back from `getType`  - in PHP8 we may get a `ReflectionUnionType` back which instead has `getTypes`

It'll be sufficient for now to throw some sort of exception or return null (as the current Behat won't install in PHP8)

We need to also decide what the union types behaviour should be for PHP8...

stof:
AFAICT, the typehint is used to find a predicate acceptable by the expected type. So I would say that for union types, we should loop over the various types to find a candidate.

src/Behat/Testwork/Argument/MixedArgumentOrganiser.php
ciaranmcnulty: I'm not sure what throws an exception here TBH, it's poorly documented

ciaranmcnulty: Good point about scalars, what a pain

stof: `$parameter->getClass()` is still used on line 286. The needed refactoring might be bigger, to avoid returning `ReflectionParameter` objects here...

ciaranmcnulty: Yeah I've done a proper search and there are a bunch more places it's used in other files - this was the only one breaking the tests when I ran initially

stof: also, be careful about the special `self` and `parent` types. `$parameter->getClass()` had them resolved to the actual class, but `ReflectionNamedType` does not resolve them.

ciaranmcnulty:
I have one more set of getClass usages to find and only a few other scenarios are failing and I can see how to fix them, so will try and push through until the test suite passes

.travis.yml
```diff
@@ -18,21 +18,24 @@ jobs:
...
+  - if [ "$DEPENDENCIES" != "low" ]; then composer update $([ -z "$PHP8" ] || echo "--ignore-platform-reqs") ; fi;
   - if [ "$DEPENDENCIES" = "low" ]; then composer update --prefer-lowest; fi;
   - echo 'zend.exception_ignore_args=Off' >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
   - export PATH=./bin:$PATH
 
 script:
   - ./vendor/bin/phpunit
-  - behat -fprogress --strict $([ -z "$SKIP_INTEROP" ] || echo "--tags ~@interop")
+  - behat -fprogress --strict $([ -z "$SKIP_INTEROP" ] || echo "--tags ~@interop&&~@php8")
+  - if [ "$PHP8" = true ]; then behat -fprogress --strict --tags=@php8; fi;
```
stof, line 38: shouldn't we rather restore the old way of automatically setting the right `@php` filters ?

src/Behat/Behat/Transformation/Transformation/ReturnTypeTransformation.php
```diff
@@ -152,9 +154,12 @@ private function getParameterClassNameByIndex(DefinitionCall $definitionCall, $a
...
                 $this->hasIndex($argumentIndex)
             ),
-            $this->isClass()
+            $this->getClassReflection()
         );
-        return count($parameters) ? current($parameters)->getClass()->getName() : null;
+
+        if (count($parameters)) {
+            return ($this->getClassReflection())(current($parameters))->getName();
+        }
```
stof, line 162: I would keep an explicit `return null;` to be more ready for return types

src/Behat/Testwork/Argument/MixedArgumentOrganiser.php
```diff
@@ -217,29 +216,59 @@ private function prepareTypehintedArguments(array $parameters, array $typehinted
...
+            return $classes;
+        }
+
+        $type = $parameter->getType();
+
+        if ($type instanceof \ReflectionNamedType) {
+            $types = [$type];
+        }
+        elseif ($parameter->getType() instanceof \ReflectionUnionType) {
+            $types = $type->getTypes();
```
stof, line 248: are we sure that this always returns a `\ReflectionNamedType[]` ?

tests/Behat/Tests/Testwork/Argument/MixedArgumentOrganiserTest.php
```diff
@@ -0,0 +1,118 @@
+<?php
+declare(strict_types=1);
```
stof, line 2: AFAIK, we don't use strict types right now

tests/Behat/Tests/Testwork/Argument/MixedArgumentOrganiserTest.php
```diff
@@ -0,0 +1,118 @@
...
+
+        $organised = $this->organiser->organiseArguments($r, $args);
+
+        $this->assertSame([1, 'date' => $date], $organised);
+    }
+
+    /** @test */
+    function it_matches_union_types()
+    {
+        if (\PHP_VERSION_ID < 80000) {
```
stof, line 99: could use `@requires php 8`

.travis.yml
```diff
@@ -18,21 +18,24 @@ jobs:
...
+      env: PHP8=true
 
 before_install:
   - phpenv config-rm xdebug.ini || echo "XDebug is not enabled"
 
 before_script:
   - if [ "$SYMFONY_VERSION" != "" ]; then composer require --no-update "symfony/symfony:${SYMFONY_VERSION}"; fi;
   - if [ "$SKIP_INTEROP" != "" ]; then composer remove --dev --no-update container-interop/container-interop; fi;
-  - if [ "$DEPENDENCIES" != "low" ]; then composer update; fi;
+  - if [ "$DEPENDENCIES" != "low" ]; then composer update $([ -z "$PHP8" ] || echo "--ignore-platform-reqs") ; fi;
```
stof, line 30: I'd rather add support for a `$COMPOSER_FLAGS` variable (that could also solve the case of `--prefer-lowest`

src/Behat/Testwork/Argument/MixedArgumentOrganiser.php
ciaranmcnulty: ReflectionUnionType is only documented in https://wiki.php.net/rfc/union_types_v2 so far where it's documented as returning `ReflectionType[]` It seems to always be ReflectionNamedType currently as we don't have unions of unions.

It won't hurt to add a check here as it could break in future with generics etc.

.travis.yml
ciaranmcnulty: @stof is it ok to leave this for now and bring in that extension when I've had time to update it?

ciaranmcnulty:
The scrutinizer issues are to do with it not knowing about Union Types yet

ciaranmcnulty:
@stof will you get a chance to review?

src/Behat/Behat/HelperContainer/ArgumentAutowirer.php
```diff
@@ -81,10 +83,32 @@ private function isArgumentWireable(array $arguments, $index, ReflectionParamete
...
+        }
+
+        return (bool) $this->getClassFromParameter($parameter);
+    }
+
+    private function getClassFromParameter(ReflectionParameter $parameter) : ?string
+    {
         try {
-            return (bool) $parameter->getClass();
+            $typeString = $parameter->getType()->getName();
```
Naktibalda, line 96: `$parameter->getType()` can return `null`.

vincentchalamon (approved):
:+1:

features/autowire.feature
```diff
@@ -245,3 +246,28 @@ Feature: Helper services autowire
       """
     When I run "behat --no-colors -f progress features/autowire.feature"
     Then it should pass
+
+    @php8
```
dunglas, line 250:
```suggestion
  @php8
```

dunglas (approved):

Issue #1305, opened 2020-06-17
Remove old compatibility layers for PHP <7.2 and Symfony <4.4


pamil:
#1299 has been merged, so we can remove the code that has been introduced to allow compatibility with PHP <7.2 and Symfony <4.4.

pamil:
Fixed by #1349.

PR #1315, opened 2020-07-04
Remove BC Layer for Symfony Dispatcher versions 4.3 and lower

rpkamp:


rpkamp:
This PR partially resolves https://github.com/Behat/Behat/issues/1305

rpkamp:
Scrutenizer errors not introduced by this PR, they were already there.

src/Behat/Testwork/EventDispatcher/TestworkEventDispatcherSymfonyLegacy.php
```diff
@@ -13,6 +13,9 @@
  * if the old symfony interface is detected.
  *
  * @deprecated Do not reference this class directly, use TestworkEventDispatcher
+ *
+ * @todo Remove this class in the next major version. It's not used anymore, but just left here
+ *       for BC purposes.
```
Taluu, line 18: add some `trigger_error` instead of such comments

rpkamp: Good call, I've removed the comment and added some `trigger_error` with `E_USER_DEPRECATED` 👍

ciaranmcnulty:
I think longer term we should be looking at decoupling from the symfony interfaces here, but this looks fine

rpkamp:
Anything needed from me here to push this forward?

ciaranmcnulty:
Thanks!

PR #1317, opened 2020-07-09
Use PHPUnit8 for unit testing

phil-davis:
- `.phpunit.result.cache` is produced locally when running modern versions of `phpunit`. Add it to `.gitignore` so nobody accidentally commits it.
- use `phpunit` 8 in `composer.json` (PHP 7.2+ supports PHPUnit 8, so we do not need PHPUnit 7 at all)
- refactor deprecated `expectedException` and `expectedExceptionMessage` tags in unit tests
- use new `assertStringContainsString` method in tests
- remove deprecated `assertInternalType` and use `assertIsArray` `assertIsInt` etc instead.

https://phpunit.de/supported-versions.html
PHPUnit8 is for PHP 7.2+
PHPUnit9 is for PHP 7.3+

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 behat.yml
 vendor
 composer.lock
+.phpunit.result.cache
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
 
     "require-dev": {
         "symfony/process": "^4.4 || ^5.0",
-        "phpunit/phpunit": "^7.5.20",
+        "phpunit/phpunit": "^8.5",
         "herrera-io/box": "~1.6.1",
         "container-interop/container-interop": "^1.2"
     },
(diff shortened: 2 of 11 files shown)


phil-davis:
@stof or anyone ping for review

phil-davis:
@ciaranmcnulty @stof I added some more refactoring of the tests for things that are deprecated in phpunit8 and can be refactored here in preparation for phpunit9.

stof:
@phil-davis see my comment in #1319, which actually applies here.

phil-davis:
> @phil-davis see my comment in #1319, which actually applies here.

Done - see latest commit. That is nicer, if the exception unexpectedly happens earlier the test will fail, which is good.

phil-davis:
@stof ready for review again.

PR #1318, opened 2020-07-10
Scenario Hook should allowed negated filters (fixes #1235)

### LINKED ISSUE #1235: Negation of tags for before and after scenario not working?
Documentation doesn't say anything how to negate a tag on a `@beforeScenario`. Previously it was enough to use `~` for negation. I.e.:

```
  /**
   * @beforeScenario ~@javascript
   */
```

But from 3.x this doesn't work anymore. Is there any other way than checking if the scenario has the tag inside the hook?

andrewnicols:
Scenario hooks support filtering by tag. The tag can be filtered on
either the Feature tags, or the Scenario tags.

When checking these tags, the tags must be combined before checking to
allow support of negated tags.

The `TagFilter::isScenarioMatch()` function in Gherkin takes both the
FeatureNode, and ScenarioNode as arguments, and then combines the tags
on both of these.

This fix addresses an issue whereby the Feature does not contain a tag,
but the Scenario does, and the hook is filtered on the absence of that
Scenario tag.

For example, Given the following Feature:
```
Feature:

  @doNotMatch
  Scenario:
    I should not be called with the @~doNotMatch filter
```

And the following Hooks:
```
/**
 * @BeforeScenario @~doNotMatch
 */
public negatedBeforeScenarioHook($event) {
  throw new \Exception('I should not have been called');
}
```

The Feature contains no hooks, but the Scenario contains the
`@doNotMatch` hook.

When processed before applying this patch the call to
`TagFilter::isFeatureMatch($feature)` will not not match (i.e. it _will_ match),
the `isMatchTagFilter` function returns early with a true value, and
therefore the hook incorrectly applies.

Following this change the, `TagFilter::isScenarioMatch($scenario)`
function is called and checks the values of all tags on _both_ the
Feature, _and_ the Scenario. The tag does _not_ match, the
`isMatchTagFilter()` function return false,a dn the tag _does not_ apply
which is the correct behaviour.

### DIFF
--- a/src/Behat/Behat/Hook/Call/RuntimeScenarioHook.php
+++ b/src/Behat/Behat/Hook/Call/RuntimeScenarioHook.php
@@ -76,10 +76,6 @@ abstract class RuntimeScenarioHook extends RuntimeFilterableHook
     {
         $filter = new TagFilter($filterString);
 
-        if ($filter->isFeatureMatch($feature)) {
-            return true;
-        }
-
         return $filter->isScenarioMatch($feature, $scenario);
     }
 
--- a/features/hooks.feature
+++ b/features/hooks.feature
@@ -84,6 +84,20 @@ Feature: hooks
           }
 
           /**
+           * @BeforeScenario @filtered
+           */
+          public function beforeScenarioFiltered($event) {
+              $this->scenarioFilter = 'filtered';
+          }
+
+          /**
+           * @BeforeScenario @~filtered
+           */
+          public function beforeScenarioNotFiltered($event) {
+              $this->scenarioFilter = '~filtered';
+          }
+
+          /**
            * @BeforeStep I must have 100
            */
           public function beforeStep100($event) {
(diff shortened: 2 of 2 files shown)


src/Behat/Behat/Hook/Call/RuntimeScenarioHook.php
```diff
@@ -76,10 +76,6 @@ protected function isMatchTagFilter(FeatureNode $feature, ScenarioInterface $sce
     {
         $filter = new TagFilter($filterString);
 
-        if ($filter->isFeatureMatch($feature)) {
```
ciaranmcnulty, line 79: It's a worry that this is just deletable 😅 

Makes sense if it's subsequently checked on the next line

ciaranmcnulty (approved):

andrewnicols:
Any chance that this can land?

ciaranmcnulty:
Thanks!

PR #1319, opened 2020-07-16
Phpunit9

phil-davis:
Make an attempt to use phpunit9.

It might be a little difficult to have code that works cleanly with both phpunit 8 and 9 - let's see.

https://phpunit.de/supported-versions.html
PHPUnit8 is for PHP 7.2+
PHPUnit9 is for PHP 7.3+

### DIFF
--- a/tests/Behat/Tests/Definition/Pattern/PatternTransformerTest.php
+++ b/tests/Behat/Tests/Definition/Pattern/PatternTransformerTest.php
@@ -66,12 +66,10 @@ class PatternTransformerTest extends TestCase
         $this->assertEquals('/hello world/', $regex3);
     }
 
-    /**
-     * @expectedException \Behat\Behat\Definition\Exception\UnknownPatternException
-     * @expectedExceptionMessage Can not find policy for a pattern `hello world`.
-     */
     public function testTransformPatternToRegexNoMatch()
     {
+        $this->expectException('\Behat\Behat\Definition\Exception\UnknownPatternException');
+        $this->expectExceptionMessage("Can not find policy for a pattern `hello world`.");
         // first pattern
         $policy1Prophecy = $this->prophesize('Behat\Behat\Definition\Pattern\Policy\PatternPolicy');
         $policy1Prophecy->supportsPattern('hello world')->willReturn(false);
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 behat.yml
 vendor
 composer.lock
+.phpunit.result.cache
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
 
     "require-dev": {
         "symfony/process": "^4.4 || ^5.0",
-        "phpunit/phpunit": "^8.5",
+        "phpunit/phpunit": "^8.5|^9.0",
         "herrera-io/box": "~1.6.1",
         "container-interop/container-interop": "^1.2"
     },
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
 
     "require-dev": {
         "symfony/process": "^4.4 || ^5.0",
-        "phpunit/phpunit": "^7.5.20",
+        "phpunit/phpunit": "^8.5",
         "herrera-io/box": "~1.6.1",
         "container-interop/container-interop": "^1.2"
     },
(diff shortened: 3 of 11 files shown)


phil-davis:
This is a potential follow-on from PR #1317 

Note that phpunit 9 emits some warnings about deprecated stuff:
```
$ ./vendor/bin/phpunit

PHPUnit 9.2.6 by Sebastian Bergmann and contributors.

WWWWW                                                               5 / 5 (100%)

Time: 00:00.013, Memory: 6.00 MB

There were 5 warnings:

1) Behat\Tests\Definition\Pattern\PatternTransformerTest::testTransformPatternToRegexCache

PHPUnit\Framework\TestCase::prophesize() is deprecated and will be removed in PHPUnit 10. Please use the trait provided by phpspec/prophecy-phpunit.

2) Behat\Tests\Definition\Pattern\PatternTransformerTest::testTransformPatternToRegexCacheAndRegisterNewPolicy

PHPUnit\Framework\TestCase::prophesize() is deprecated and will be removed in PHPUnit 10. Please use the trait provided by phpspec/prophecy-phpunit.

3) Behat\Tests\Definition\Pattern\PatternTransformerTest::testTransformPatternToRegexNoMatch

PHPUnit\Framework\TestCase::prophesize() is deprecated and will be removed in PHPUnit 10. Please use the trait provided by phpspec/prophecy-phpunit.

4) Behat\Tests\Testwork\Subject\GroupedSubjectIteratorTest::testIterationWithEmptyAtBeginning

PHPUnit\Framework\TestCase::prophesize() is deprecated and will be removed in PHPUnit 10. Please use the trait provided by phpspec/prophecy-phpunit.

5) Behat\Tests\Testwork\Subject\GroupedSubjectIteratorTest::testIterationWithEmptyInMiddle

PHPUnit\Framework\TestCase::prophesize() is deprecated and will be removed in PHPUnit 10. Please use the trait provided by phpspec/prophecy-phpunit.

WARNINGS!

Tests: 5, Assertions: 16, Warnings: 5.

The command "./vendor/bin/phpunit" exited with 0.
```
But those do not fail the CI.

It might be a bit difficult to "Please use the trait provided by phpspec/prophecy-phpunit" while still supporting phpunit8 for PHP 7.2

Taluu:
IMO we shouldn't bother supporting phpunit 9. As it's a "dev" dependency, this dep will be installed only if you install / clone `behat/gherkin` from the root. If it is installed through a dep... the dev dep won't ever be installed.

So it is not useful IMO to support multiple major versions of any dev deps. Just support the minimal one, unless you are supporting multiple php versions AND the dev deps are strict on their php requirements (like ocramius' proxy manager in which patch / major versions only support ONE php minor version at a time). 

This isn't the case here, so only support phpunit 8 should be fine.

phil-davis:
Yes, there is no urgent need to use phpunit9 in CI. I was doihg this mainly to find out what phpunit9 will complain about, and then update the phpunit8 test code to refactor away as many of the future-deprecated stuff now as reasonable.

tests/Behat/Tests/Definition/Pattern/PatternTransformerTest.php
```diff
@@ -66,12 +66,10 @@ public function testTransformPatternToRegexCacheAndRegisterNewPolicy()
...
     }
 
-    /**
-     * @expectedException \Behat\Behat\Definition\Exception\UnknownPatternException
-     * @expectedExceptionMessage Can not find policy for a pattern `hello world`.
-     */
     public function testTransformPatternToRegexNoMatch()
     {
+        $this->expectException('\Behat\Behat\Definition\Exception\UnknownPatternException');
+        $this->expectExceptionMessage("Can not find policy for a pattern `hello world`.");
```
stof, line 72: I'd put these just before the call in which we expect the exception instead, so that any code running before that cannot satisfy the expectation.

phil-davis:
See #1317 that moves us to phpunit 8. The possible refactorings of test code from here are in that PR, which is ready t review/merge. That should be far enough to go for now. Actually implementing phpunit 9 is not urgent.

dvdoug:
FYI You will need PHPUnit 9.3+ to support PHP8

PR #1322, opened 2020-09-23
Add feature title as classname to junit testcase output

steefmin:
You can setup gitlab to parse junit output of testruns. When doing this for behat, the "Suite" name is empty. Normally this is filled by parsing the classname tag of the testcase. This PR adds that to the behat junit formatter.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -916,9 +916,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   * Initial release
 
 [Unreleased]: https://github.com/Behat/Behat/compare/v3.7.0...master
-### Changed
-  * Added feature title as classname of testcase in junit output
-
 [3.7.0]: https://github.com/Behat/Behat/compare/v3.6.1...v3.7.0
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -916,6 +916,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   * Initial release
 
 [Unreleased]: https://github.com/Behat/Behat/compare/v3.7.0...master
+### Changed
```
ciaranmcnulty, line 919: This shouldn't go here, you can remove it and we'll add as part of the release

ciaranmcnulty (changes requested):
Looks ok to me - feels weird making it the 'class name' and using the title but I guess it's ok

CHANGELOG.md
steefmin: Okido, removed. I followed https://github.com/Behat/Behat/blob/master/CONTRIBUTING.md :thinking:

ciaranmcnulty: Yeah it's meant to go at the top

PR #1323, opened 2020-10-15
Solves ambiguous usage of named arguments in PHP 8

dgafka:
Screenshot from 2020-10-15 15-52-05

### DIFF
--- a/src/Behat/Testwork/Call/Handler/RuntimeCallHandler.php
+++ b/src/Behat/Testwork/Call/Handler/RuntimeCallHandler.php
@@ -107,7 +107,7 @@ final class RuntimeCallHandler implements CallHandler
         try {
             $arguments = array_values($arguments);
             $this->validator->validateArguments($reflection, $arguments);
-            $return = $callable(...array_values($arguments));
+            $return = $callable(...$arguments);
         } catch (Exception $caught) {
             $exception = $caught;
         }
--- a/src/Behat/Testwork/Call/Handler/RuntimeCallHandler.php
+++ b/src/Behat/Testwork/Call/Handler/RuntimeCallHandler.php
@@ -105,6 +105,7 @@ final class RuntimeCallHandler implements CallHandler
         $return = $exception = null;
 
         try {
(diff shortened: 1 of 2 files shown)


ciaranmcnulty:
As this isn't picked up by the tests, can you add a test that would trigger the behaviour previously?

src/Behat/Testwork/Call/Handler/RuntimeCallHandler.php
```diff
@@ -105,6 +105,7 @@ private function executeCall(Call $call)
         $return = $exception = null;
 
         try {
+            $arguments = array_values($arguments);
             $this->validator->validateArguments($reflection, $arguments);
             $return = $callable(...array_values($arguments));
```
stof, line 110: This `array_values` can be removed now that you moved it earlier (but is it actually an issue in `validateArguments` or are you using a Behat version which does not have this `array_values` yet ?)

dgafka:
> As this isn't picked up by the tests, can you add a test that would trigger the behaviour previously?

If you set up testing for PHP 8, I think a lot of existing tests will fail. 
With current setup for <= 7.4 there is no way to show this issue, as it's related to PHP 8 functionality, named parameters.

src/Behat/Testwork/Call/Handler/RuntimeCallHandler.php
dgafka: Good catch, I have removed duplicated call :)

I am not sure, where this issue comes from I could not track it all the way. 
However I am pretty sure, you will see the issue, when test suite will be set up for PHP 8.

ciaranmcnulty:
The reason I ask about a test is that #1313 I believe fixes all the outstanding failures under PHP 8 so we are probably missing coverage of this case

ciaranmcnulty:
Any idea why the build is failing?

ciaranmcnulty:
It looks like the test was just sensitive to the line numbers in a stack trace (that changed because you added a line)

ciaranmcnulty:
Thanks @dgafka !

PR #1327, opened 2020-11-02
Migrate builds from Travis CI to GitHub actions

jakzal:
image


ciaranmcnulty:
You may as well remove the travis config as part of the PR if it's comprehensively replaced

jakzal:
@ciaranmcnulty I did...

ciaranmcnulty:
@jakzal was confused why it looked like it was running but it’s a required check for the master branch

.github/workflows/build.yml
```diff
@@ -0,0 +1,95 @@
...
+                run: ./bin/behat -fprogress --strict $([ -z "${{ matrix.skip-interop }}" ] || echo "--tags ~@interop&&~@php8")
+
+            -   name: Run tests (Behat for PHP 8.0)
+                if: matrix.php >= 8.0
+                run: ./bin/behat -fprogress --strict --tags=@php8
+
+            -   name: Build the PHAR
+                if: matrix.php != '8.0'
+                run: |
+                    curl -LSs https://box-project.github.io/box2/installer.php | php &&
```
smoench, line 64: This downloads a very old version of box. As far as I understand that script it downloads one of these promoted files https://box-project.github.io/box2/manifest.json.

jakzal: Yes, my approach is to migrate everything as is and do any updates as a separate step.

Behat still uses box1. I think @ciaranmcnulty would appreciate if someone upgraded it and added phar scoping ;)

stof: I don't think we can scope phars: behat extensions have to deal with Symfony components, not only with the Behat code itself.

pamil (approved):

.github/workflows/build.yml
jakzal: This only means we can't scope PHARs fully (particularly in places where the contract of a Symfony component is exposed to the end-users).

stof: if we don't scope Behat packages and we don't scope Symfony components, there isn't much left to scope: https://github.com/Behat/Behat/blob/304661d4bc00934314c835c79fe36c0ce38da268/composer.json#L16-L28

ciaranmcnulty: Hm, I'd seen phar scoping as a good reason for dropping older symfony versions

PR #1333, opened 2021-01-05
Upgrade deprecated dependencies

szepeviktor:


ciaranmcnulty:
container-interop is included here as a dev dependency so we can test against it

szepeviktor:
kép

I've just read it in PSR 11: https://www.php-fig.org/psr/psr-11/#2-package

composer.json
```diff
@@ -30,8 +30,7 @@
     "require-dev": {
         "symfony/process": "^4.4 || ^5.0",
         "phpunit/phpunit": "^8.5 || ^9.0",
-        "herrera-io/box": "~1.6.1",
-        "container-interop/container-interop": "^1.2"
+        "psr/container-implementation": "^1.0"
```
stof, line 33: that does not bring any benefit, as behat does not ask to choose an arbitrary implementation (and that's even more useless to require the virtual package as a dev dependency as the behat composer.json has to fulfill it anyway)

stof:
Behat does not ask for "any implementation of `psr/container`, which is what `psr/container-implementation` is about.

szepeviktor:
I'm lost.
Could you help me why to list `container-interop/container-interop` in `require-dev`?

PR #1334, opened 2021-02-14
Add support for specification of a preferred profile

andrewnicols:
In many situations it is desirable to have a specific profile that is
used when none is specified, however because of the way in which all
profiles inherit from the `default` profile, this is not currently
possible.

For example, if you have different MinkExtension configurations for a
range of different browsers, then each browser may only be specified in
its own profile. This works well, but often you will want to specify a
default browser configuration.

To achieve this you may think that you can set the Extension
configuration for your preferred browser into the `default` profile, but
this means that all other profiles will include this configuration, and
may not work as expected.

This change introduces the concept of a preferred profile, which is the
profile which will be used unless the `--profile` CLI argument is
provided to override it.

This means that if you have profiles for chrome, headlesschrome,
firefox, headlessfirefox, edge, and headlessedge, you can choose to
specify a preferred profile of `headlesschrome`, which will be used
unless you explicitly set the `--profile` argument when running behat.

### DIFF
--- a/features/preferred_profile.feature
+++ b/features/preferred_profile.feature
@@ -1,0 +1,144 @@
+Feature: Preferred Profiles
+  In order to test my preferred profile
+  As a tester
+  I need to be able to specify a preferred profile
+
+  Background:
+    Given a file named "features/bootstrap/FeatureContext.php" with:
+      """
+      <?php
+
+      use Behat\Behat\Context\Context;
+
+      class FeatureContext implements Context
+      {
+          private $result;
+          private $numbers;
+
+          /**
+           * @Given /I have basic calculator/
+           */
+          public function iHaveBasicCalculator() {
+              $this->result  = 0;
+              $this->numbers = array();
+          }
+
+          /**
+           * @Given /I have entered (\d+)/
+           */
+          public function iHaveEntered($number) {
+              $this->numbers[] = intval($number);
+          }
+
+          /**
+           * @When /I add/
+           */
+          public function iAdd() {
+              $this->result  = array_sum($this->numbers);
+              $this->numbers = array();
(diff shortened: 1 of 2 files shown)


andrewnicols:
The failures above seem to be reproducible against master, before my patch is applied.

andrewnicols:
Most of the tests are now passing with the exception of Scrutinizer. Not sure how best to solve that one.

Any chance of a review on this change?

andrewnicols:
Is there anything that I can do to help get this over the line?

It's a real cause of frustration for developers on our project.

andrewnicols:
Hi @stof, @everzet, @ciaranmcnulty ,

I'm just wondering if there's any chance that we might see this merged at some point, or if there's anything that I can do to help things along.

All tests are passing, and there's a real and valid use-case for it.

Thanks

andrewnicols:
Pinging @stof, @ciaranmcnulty, @everzet again.

This issue has been sat waiting for attention for 5 months now. What can we do to help?

ciaranmcnulty:
Can you take the changelog entry out please? (I need to change the guidance on that)

I'm not sure about the use case - this is to avoid having to type `--profile` each time?

carlos-granados (changes requested):
@andrewnicols,  as suggested by @ciaranmcnulty, could you please remove the changes to CHANGELOG.md? This is handled separately. Otherwise it looks good to me

carlos-granados (approved):
@andrewnicols I rebased your change and removed the change to the CHANGELOG file. It all looks OK now for me

acoulton (approved):
LGTM, thanks for the contribution @andrewnicols

PR #1335, opened 2021-02-14
Adding "file"-attribute to "testcase"-tag in JUnit format

ppaulis:
This PR adds a "file" attribute to the "testcase" tags in the JUnit output format. This is e.g. necessary when trying to parallelize testruns on CircleCI that can split tests automatically by timing data.

Example of output:
```
<?xml version="1.0" encoding="UTF-8"?>
<testsuites name="default">
    <testsuite name="World consistency" tests="1" skipped="0" failures="1" errors="0" time="-IGNORE-VALUE-">
        <testcase name="Failed" classname="World consistency" file="features/World.feature" status="failed" time="-IGNORE-VALUE-">
            <failure message="Given I have entered 10: failure (Exception)" type="teardown"></failure>
        </testcase>
    </testsuite>
</testsuites>
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [3.9.0] - 2021-02-14
+
+### Added
+ * "file"-attribute in JUnit "testcase"-tag containing the relative filepath to the feature file
+
 ## [3.8.1] - 2020-11-07
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [3.9.0] - 2021-02-14
 
 ### Added
- * "file"-attribute in JUnit "testcase"-tag containing the relative filepath to the feature file
+ * [1335](https://github.com/Behat/Behat/pull/1335): "file"-attribute in JUnit "testcase"-tag containing the relative filepath to the feature file
 
 ## [3.8.1] - 2020-11-07
 
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
     "require": {
         "php": "^7.2 || ^8.0",
         "ext-mbstring": "*",
-        "behat/gherkin": "^4.6.0",
+        "behat/gherkin": "^4.8.0",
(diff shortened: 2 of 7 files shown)


ciaranmcnulty:
I fixed the test failure in #1336, perhaps you could review that? Then once it’s merged you can rebase this PR

ppaulis:
> I fixed the test failure in #1336, perhaps you could review that? Then once it’s merged you can rebase this PR

done :+1:

ppaulis:
@ciaranmcnulty I need to take a look at appveyor too. The tests that are passing on Github Actions, are not yet passing on appveyor. I have no experience with that tool yet.

ciaranmcnulty:
It may be easier to get windows builds on GA

ppaulis:
@ciaranmcnulty Found it:

`file="features/World.feature"` on GitHub
vs.
`file="features\World.feature"` on AppVeyor

I'm using the DIRECTORY_SEPARATOR constant (like it is used in other parts of the code). But it seems that AppVeyor is running on operating systems using Backslashes as separators...

Should I use simple slashes ("/") as separators instead?

ppaulis:
@ciaranmcnulty In any case, I'll close this PR and open a fresh one after having rebased your merge. I already tinkered around too much in this one.

ciaranmcnulty:
Yes Appveyor is there to test Windows

ppaulis:
@ciaranmcnulty To make this work, the only way I currently see is to duplicate the concerned Scenarios, adapt them to their respective operating system and to exclude them via tags in the builds where they don't belong.

Something like this:
```
behat --tags="~@windows-only" (on GA)
behat --tags="~@unix-only" (on AV)
```

Would that be acceptable?

ciaranmcnulty:
I'm not a Windows user, but it may be fine to make it `/` in the report - a lot of apps seem to cope ok with forward slashes.

We could also 'fix' it in FeatureContext:: fileXmlShouldBeLike - there's already some manipulation there to remove timestamps

ppaulis:
@ciaranmcnulty Yes, I was currently examining this option too. Using some sort of Placeholder for the Directory Separator should do the trick here. I'll get back to you with a fresh PR in the next couple of hours, having a fix for this.

Thanks :+1:

ciaranmcnulty:
I guess what it comes down to is not which OS you're running on, but which directory separator is appropriate in the XML format; i.e. which separators JUnit tools will work best with. My assumption is that `/` everywhere makes the most sense

ppaulis:
closed in favor of #1337

PR #1336, opened 2021-02-15
Fix syntax help test and bump gherkin dependency

ciaranmcnulty:
The change in the sort order of the translations broke this test

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
     "require": {
         "php": "^7.2 || ^8.0",
         "ext-mbstring": "*",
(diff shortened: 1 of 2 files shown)


ppaulis (approved):
seems fine to me :+1:

stof:
@ciaranmcnulty this test is quite fragile, as adding a new translation (which is not considered a BC break in Gherkin, and should not) will still break this test

ciaranmcnulty:
Yeah, I had to bump the gherkin version here too. Maybe it should be looking for patterns etc? I'm not sure the best way

PR #1340, opened 2021-04-09
Added Chiniese language

54853315:
Added Chiniese language


i18n.php
```diff
@@ -264,4 +264,21 @@
         'undefined_count'         => '{1,21,31} %count% не определен|]1,Inf] %count% не определено',
         'skipped_count'           => '{1,21,31} %count% пропущен|]1,Inf] %count% пропущено',
     ),
+    'cn'    => array(
```
ciaranmcnulty, line 267: Isn’t it `zh`?

54853315: `zh` will be nice ,I will fix it .

ciaranmcnulty: I don’t know enough to say if it’s correct :)

You could consider contributing gherkin translations to cucumber/cucumber too

ciaranmcnulty:
It would be good to get another Chinese speaker (reader?) to check this and confirm the translations are ok

ciaranmcnulty:
Thank you!

PR #1341, opened 2021-04-09
Allow profiles to override extensions

### LINKED ISSUE #1339: Unrecognized option "extensions" under "testwork".
Hi,

What I'm trying to have is a suite that does not need extensions and executes fast.
My default suite uses Symfony extension and because it's a big app, creating the container takes a lot of time.
I have a suite "domain" of "acceptance tests" that does not rely on symfony in any way and that is really really fast so I'd like to run this suite before the other on CI. 
The fact that my "domain" suite load the extensions makes it slow.

So I tried to create a "domain" profile instead of a suite with this configuration:

```
default:
  suites:
    default: ....

  extensions:
    Behat\Symfony2Extension:
      kernel:
        bootstrap: config/bootstrap.php
        env: behat
        debug: true
        path: src/Kernel.php
        class: Kernel

domain:
  suites:
    default: ....

  extensions: ~
```

But I have this issue:

```
Unrecognized option "extensions" under "testwork". Available options are "argument", "autoload", "calls", "cli", "contexts", "definitions", "environments", "events", "exceptions", "filesystem", "formatters", "gherkin", "gherkin_translations", "helper_container", "hooks", "ordering", "snippets", "specifications", "suites", "testers", "transformations", "translation".
```

A solution I found is to rename the `default` profile to something else and remove the `extensions` key from my `domain` profile but this breaks the `default` profile.

This issue is that the `default` profile is BOTH "the default profile if no profile is specified" AND "the profile that others inherit from" :/

Is there a solution for my need that is not a second configuration file ?

If this is considered a bug that needs to fixed I can make a PR if someone guides me a little on how to implement a fix.

Thanks for your time.

Zayon:
Fixes #1339

### DIFF
--- a/features/extension_inheritance.feature
+++ b/features/extension_inheritance.feature
@@ -1,0 +1,184 @@
+Feature: Profile extension overrides
+    In order to organize my profiles
+    As a tester
+    I need to be able to override extensions in non-default profiles
+
+  Background:
+    Given a file named "custom_extension.php" with:
+      """
+      <?php
+
+      use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
+      use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+      class CustomInitializer implements Behat\Behat\Context\Initializer\ContextInitializer
+      {
+          public function supportsContext(Behat\Behat\Context\Context $context)
+          {
+              return true;
+          }
+
+          public function initializeContext(Behat\Behat\Context\Context $context)
+          {
+              $context->addExtension('custom extension 1');
+          }
+      }
+
+      class CustomExtension implements Behat\Testwork\ServiceContainer\Extension {
+          public function getConfigKey()
+          {
+              return 'custom_extension';
+          }
+
+          public function configure(ArrayNodeDefinition $builder) {}
+
+          public function initialize(Behat\Testwork\ServiceContainer\ExtensionManager $extensionManager) {}
+
+          public function load(ContainerBuilder $container, array $config)
+          {
+              $definition = $container->register('custom_initializer', 'CustomInitializer');
+              $definition->addTag('context.initializer', array('priority' => 100));
+          }
+
+          public function process(ContainerBuilder $container) {}
+      }
+
+      return new CustomExtension;
+      """
+    And a file named "custom_extension2.php" with:
+      """
+      <?php
+
+      use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
+      use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+      class CustomInitializer2 implements Behat\Behat\Context\Initializer\ContextInitializer
+      {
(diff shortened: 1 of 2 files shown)


Zayon:
Following #1339 

I added a feature file that demonstrate the issue I have.

I fixed the `Unrecognized option` error by moving the `unset($configs[$i]['extensions']);` line but the extensions in the default profile are still loaded in the custom profile.

My investigations tends toward the `ConfigurationLoader` here https://github.com/Behat/Behat/blob/689f1e43efd6efa5f1dbde56a1cb967f1a512cad/src/Behat/Testwork/ServiceContainer/Configuration/ConfigurationLoader.php#L212 where changes could be done.

One thing I have to test is loading a second extension specifically in the custom profile, that would be interesting to know if it works or if extensions are activated only when declared on the default profile.

 Any thoughts ?

ciaranmcnulty:
This looks useful, it's certainly something I've struggled with 👍

ciaranmcnulty:
Can you please rebase this? I think maybe the feature files could be simplified as well

Zayon:
I rebased :)

Zayon:
From the tests I have written I would say that it's possible to :

1. Disable all extensions from the default profile with 

```yaml
      default:
        extensions:
          custom_extension.php: ~
      custom_profile:
        extensions: ~
```

Here, `custom_profile` doesn't have any extension loaded

2. Load an extension on a custom profile but not on the default one

```yaml
      default:
        extensions:
          custom_extension.php: ~
      custom_profile:
        extensions:
          custom_extension2.php: ~
```

Here, `custom_profile` has **both** extensions loaded and the default profile **only** has `custom_extension` loaded and **NOT** `custom_extension2` 

It would be intersting to add a tests for overriding a specific configuration of an extension, while keeping the rest of the configuration.

Zayon:
To be clearer : 

> how do I inherit extension configuration from the default profile?

It's done by default

carlos-granados (approved):
Looks good to me

acoulton (changes requested):
I've rebased this onto current master as the older commits were not showing any build/test history.

Unfortunately it is failing CI - one failure is due to #1483, but the other looks like a new failure from this feature. Not sure if that was always broken, or if something in the behat config loading has changed since this was first proposed.

I'd be happy to merge if you can get it green.

Zayon:
I've repushed (and squashed into a single commit) there is now a single feature file with all the scenarios that needed to be covered.

The change is a lot bigger than previously, specifically: being able to disable all extensions required that I changed the way extensions are activated.

Maybe this feature (disabling all extensions) is not worth the change. Also there may be issues with imports that I did not test.

The github workflow seems to be blocked due to the "changes requested" status.

acoulton (approved):
This looks OK to me and worth having the bugfix, it is a bigger change but reading through the code of the changed method it looks safe enough.

I'll ask @carlos-granados to re-review since it's a bit different from what he saw before.

carlos-granados (approved):
The new changes good look to me and they are well tested

PR #1342, opened 2021-04-27
Introduce reading PHP8 Attributes for Given, When and Then steps

rpkamp:


src/Behat/Behat/Context/Reader/AttributedContextReader.php
```diff
@@ -0,0 +1,70 @@
...
+use Behat\Behat\Context\Environment\ContextEnvironment;
+use ReflectionClass;
+use ReflectionMethod;
+
+/**
+ * Reads context callees by Attributes using registered Attribute readers.
+ *
+ * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ */
+class AttributedContextReader implements ContextReader
```
rpkamp, line 23: Naming things ...

I don't really like `AttributedContextReader`, but I don't have any better ideas either.
Pro is that it's discoverable wrt `AnnotationContextReader` so it kind of makes sense.

src/Behat/Behat/Context/ServiceContainer/ContextExtension.php
```diff
@@ -252,6 +254,20 @@ private function loadDefaultClassGenerators(ContainerBuilder $container)
      * @param ContainerBuilder $container
      */
     private function loadDefaultContextReaders(ContainerBuilder $container)
+    {
+        $this->loadAnnotatedContextReader($container);
```
rpkamp, line 258: This method got really big and pretty unreadable so I split it up.

src/Behat/Behat/Definition/Context/Attribute/DefinitionAttributeReader.php
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        $attributes = $method->getAttributes(Definition::class, \ReflectionAttribute::IS_INSTANCEOF);
+
+        $callees = [];
+        foreach ($attributes as $attribute) {
+            $class = static::$classes[$attribute->getName()];
+            $callable = array($contextClass, $method->getName());
+
+            $callees[] = new $class($attribute->newInstance()->pattern, $callable, null);
```
rpkamp, line 52: Description is `NULL` now by default, because we can't do docblock descriptions with Attributes.
We could opt to still check any annotations to get the description, but I don't really like that as Attributes are supposed to replace docblocks, not to live side-by-side.

Would be interested to hear what others think.

rpkamp: From what I can gather they only seem to be used when running `[vendor/]bin/behat --definitions l`

src/Behat/Behat/Context/Reader/AttributedContextReader.php
stof: `AttributeContextReader` seems better to me

src/Behat/Behat/Definition/Context/Attribute/DefinitionAttributeReader.php
stof: well, the description is still perfectly stored in the docblock, as it is about showing the documentation of the method

src/Behat/Behat/Definition/Context/Attribute/DefinitionAttributeReader.php
```diff
@@ -0,0 +1,57 @@
...
+    {
+        if (PHP_MAJOR_VERSION < 8) {
+            return [];
+        }
+
+        $attributes = $method->getAttributes(Definition::class, \ReflectionAttribute::IS_INSTANCEOF);
+
+        $callees = [];
+        foreach ($attributes as $attribute) {
+            $class = static::$classes[$attribute->getName()];
```
stof, line 49: accessing private static properties must always use `self::`, not `static::`. Otherwise, child classes are broken, as they try to access `ChildClass::$classes`

src/Behat/Behat/Definition/Attribute/When.php
```diff
@@ -0,0 +1,28 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Behat\Behat\Definition\Attribute;
+
+/**
+ * Represents an Attribute for When steps
+ */
+#[\Attribute(\Attribute::TARGET_METHOD)]
```
stof, line 16: it must be allowed to be repeated as the same method can be registered multiple times with different patterns

src/Behat/Behat/Definition/Attribute/Definition.php
```diff
@@ -0,0 +1,19 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Behat\Behat\Definition\Attribute;
+
+/**
+ * Marker interface for all Attributes regarding
+ * Call definitions
+ */
+interface Definition
```
stof, line 17: This interface should be marked `@internal` as the implementation does not actually support having more implementations

src/Behat/Behat/Definition/Context/Attribute/DefinitionAttributeReader.php
rpkamp: True. I'll see if I can grab the docblock from the `ReflectionMethod` here and fetch the description from it.

src/Behat/Behat/Definition/Context/Attribute/DefinitionAttributeReader.php
rpkamp: Nice catch! I've been working with final and abstract classes so long I forgot that was even a thing :slightly_smiling_face:

ciaranmcnulty:
As we currently support two pattern formats (and may in future want to support Cucumber Expressions) is there an opportunity here to use the attributes to be specific about what pattern is being used?

i.e. rather than loading the string and then guessing if it's a regex or turnip, we could be having `Behat\...\Regex\When` and so on?

On another point, although the attribute's namespace  makes sense for the code organisation I'd like to see something shorter like `Behat\Step\When`

src/Behat/Behat/Context/Attribute/AttributeReader.php
```diff
@@ -0,0 +1,33 @@
...
+{
+    /**
+     * Reads all callees associated with a provided method.
+     *
+     * @param string           $contextClass
+     * @param ReflectionMethod $method
+     *
+     * @return array
+     */
+    public function readCallees($contextClass, ReflectionMethod $method);
```
Taluu, line 32: shouldn't it be safe to add a type here ? And how about a generator instead ?

Taluu: (Because IIRC, the project was bumped to 7.2+)

Taluu (approved):

src/Behat/Behat/Context/Attribute/AttributeReader.php
rpkamp: Yes that would work 👍 

I don't see a point in using a `Generator` though, given that is only called once and then stored in a caching layer for future calls anyway.

Taluu: Oh if it's stored it's indeed useless to use a generator. So :+1:

pamil:
> i.e. rather than loading the string and then guessing if it's a regex or turnip, we could be having Behat\...\Regex\When and so on?

In Sylius projects, we usually have both of them used in a single file - this way we would end up aliasing the attributes and having `#[RegexWhen]` and `#[TurnipWhen]`.

rpkamp:
I'm a big fan of one change at a time - doing multiple things at the same time will always muddy the waters and more often than not strand in mundane discussions about minor details that weren't initially in scope anyway. 

Also, TBH I don't see a reason why Behat should bother userland with Turnip vs Regex if Behat can perfectly well detect the difference. If the difference should be made (I'm not versed enough into Behat code to make that call) than it can also be done _between_ parsing the annotations and Attributes and converting it into (an) internal object(s).

1. Read annotations / attribute
2. Decide whether the format is Regex or Turnip
3. Instantiate correct Given/When/Then based on what was detected in step 2 (ie `TurnipGiven`, `RegexGiven`, etc)

rpkamp:
> On another point, although the attribute's namespace makes sense for the code organisation I'd like to see something shorter like Behat\Step\When

That certainly makes it a lot nicer to `use` the attributes in FeatureContexts. I've moved them 👍

src/Behat/Behat/Definition/Context/Attribute/DefinitionAttributeReader.php
rpkamp: Done. I've introduced a `DocBlockHelper` to avoid having the exact same logic in two spots.

src/Behat/Behat/Definition/Context/Attribute/DefinitionAttributeReader.php
rpkamp: Changed to `self` and made all classes `final` as well.

src/Behat/Behat/Definition/Attribute/When.php
rpkamp: They are now repeatable (which is also verified through `features/attributes.feature`) 👍

src/Behat/Behat/Definition/Attribute/Definition.php
rpkamp: It's been marked as internal.

src/Behat/Behat/Context/Attribute/AttributeReader.php
rpkamp: I've added the `string` typehint

stof:
@rpkamp detecting turnip vs regexp can be done with guessing thanks to regexps having a delimiter (and writing a turnip pattern that can be confused with delimiters means you write weird sentences, which is not the expected using of Gherkin. however, if we add another type of pattern in the future, guessing might not work.
And the author of the attribute knows whether they wrote their pattern as regexp or as turnip syntax.

ciaranmcnulty:
Yes that was my thinking - Cucumber Expressions don’t have a delimiter

rpkamp:
So as far as I understand it, within the scope of this PR, the issue of supporting multiple notations for steps is that we might up choosing a namespace for the attributes now that we may end up having to change later it is decided that different notations should get different attributes (i.e. rather then just `Behat\Step\Given` it would be `Behat\Step\Regex\Given`, `Behat\Step\Tunip\Given`, `Behat\Step\Expression\Given`, etc). 

That to me seems a fair point. We shouldn't introduce something we already know will have to change later, leading to a BC break. 

So the question then becomes, is that a route that's going to be taken? And also, how does this fit with annotation for PHP7.x?

As an alternative to namespaces for different notations, the Attributes could also get a second property to indicate which notation is being used. I.e.

```
#[Given('I have :count apple(s)', 'tunip')]
```

That value could default to "auto", which is what Behat is doing now: auto-detect what is being used. 

Since the attributes are `final` adding a property with a default value isn't a BC, so that can be implemented safely later, after this PR. 

Having the value of "auto" also aligns nicely with the way Behat currently abstracts away the notation of multiple notations, and doesn't all of a sudden force users of Behat to care about this. 

The drawback is that there is no way to support this in the same way with annotations. And given the adaption of PHP8 so far I'm thinking PHP7 will have to be supported for quite a while.

rpkamp:
Okay, so how do we proceed from here? Should this be merged? Should the namespace of the Attributes be changed before it can be merged? Should I add Attributes for the Hooks too (BeforeScenario et al) before this can be merged?

stof:
> Should this be merged? Should the namespace of the Attributes be changed before it can be merged?

I think we need to reach a decision on the handling of types of steps here

> Should I add Attributes for the Hooks too (BeforeScenario et al) before this can be merged?

this can be done in a separate PR.

rpkamp:
Can someone make a decision here please? Personally I would opt for the second argument that defaults to `auto`, but n=1 and I'm not too well versed in the ideas / architecture behind Behat.

ciaranmcnulty:
I think auto detection is fine for now; we can tackle handling extra dialects if and when we add cucumber-expressions - I was thinking too far ahead

rpkamp:
Cool. Want me to add that in now, so we don't forget? It won't do/change anything, but just so in the future we know what was decided by looking at the code rather than having go dredging GitHub for this MR.

ciaranmcnulty:
I mean parameter-less detection like you have

rpkamp:
Alright, in this case I believe this MR is done and be reviewed to be merged 😄

rpkamp:
Is there anything needed from me to push this forward?

rpkamp:
@stof @ciaranmcnulty @pamil I'd like to get to work on Attributes for the hooks, but I don't want to as long as this PR is open. Can this PR please be reviewed and/or merged?

src/Behat/Behat/Context/Annotation/DocBlockHelper.php
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Helper class for DocBlock parsing
+ */
+class DocBlockHelper
+{
+    /**
+     * Extracts a description from the provided docblock,
+     * with support for multiline descriptions.
+     */
+    public static function extractDescription(string $docBlock): string
```
ciaranmcnulty, line 22: I'd prefer to see this non-static and the helper injected into the class via constructor - this is to allow it to be replaced by an Extension if desired

ciaranmcnulty (approved):

src/Behat/Behat/Context/Annotation/DocBlockHelper.php
rpkamp: Apologies for the late reply. My development machine was borked and I was to lazy to re-configure it. I did do that after all, so now I could process this. 

I wasn't sure where to define a service for the `DocBlockHelper` so I defined it twice, once for the `ContextExtension` (service ID `context.docblock_helper`) and once for the `DefinitionExtension` (service ID `definition.doc_block_helper`). That way extensions can override one without necessarily also overriding the other. Gives a bit more freedom.

rpkamp: With that, I think this MR is ready to be merged now 🙂 

It would be nice to write a rector at some point that can replace annotations with attributes. 
But that wouldn't belong in this MR. I even doubt it belongs in the Behat repository at all.

rpkamp:
IMO this is finished and can be merged. Anything I can do to push this forward?

rpkamp:
@pamil @ciaranmcnulty can this be merged please?

pamil (approved):

PR #1346, opened 2021-06-20
Allow Symfony 6

Kocal:
Symfony 6 will be released in 5 months, but I think it would be a nice thing to prepare Behat for Symfony 6.

I had to update a test, the stacktrace displayed when an exception is thrown does not contains parameters anymore. Changed `RuntimeCallHandler->executeCall(Object(Behat\Behat\Definition\Call\DefinitionCall))` to `RuntimeCallHandler->executeCall(Object(` to fix that:

WDYT?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,6 +26,9 @@ jobs:
                     -   php: 7.4
                         symfony-version: ^4.4
                     -   php: 8.0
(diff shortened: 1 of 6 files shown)


pamil (approved):

pamil:
Thanks a lot, Hugo! 🎉

PR #1356, opened 2021-09-14
Remove support for PHP 7.2 and 7.3

pamil:
PHP 7.2 is not maintained at all anymore and PHP 7.3's security support ends in 2 months.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,16 +15,15 @@ jobs:
             fail-fast: false
             matrix:
                 include:
-                    -   php: 7.2
-                        publish-phar: true
-                    -   php: 7.2
-                        skip-interop: true
-                    -   php: 7.2
-                        composer-flags: --prefer-lowest
-                    -   php: 7.3
                     -   php: 7.4
+                        publish-phar: true
+                    -   php: 7.4
+                        skip-interop: true
+                    -   php: 7.4
+                        composer-flags: --prefer-lowest
                     -   php: 7.4
                         symfony-version: ^4.4
+                    -   php: 7.4
                     -   php: 8.0
                     -   php: 8.1
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -27,7 +27,7 @@ init:
 install:
     - ps: |
         if (!(Test-Path c:\tools\php)) {
-            appveyor-retry cinst --params '""/InstallDir:C:\tools\php""' --ignore-checksums -y php --version 7.2
+            appveyor-retry cinst --params '""/InstallDir:C:\tools\php""' --ignore-checksums -y php --version 7.4
             Get-ChildItem -Path c:\tools\php
             cd c:\tools\php
 
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     ],
 
     "require": {
-        "php": "^7.2 || ^8.0",
+        "php": "^7.4 || ^8.0",
         "ext-mbstring": "*",
         "behat/gherkin": "^4.8.0",
         "behat/transliterator": "^1.2",


stloyd (approved):

stof:
does this actually allow to reduce the maintenance work on Behat (by simplifying the code) or would it only push projects to ask for a longer support of Behat 3.8 due to needing to run their CI on those PHP versions ?

pamil:
I'd say it makes maintenance and contributing easier since we wouldn't have to think whether a feature was introduced in PHP 7.2 or 7.3. At the same time, we'd be able to introduce typed properties in new or refactored code.

mannion007 (approved):

mvorisek:
Please keep at least PHP 7.3 until some major unavailable language features are needed

acoulton:
> does this actually allow to reduce the maintenance work on Behat (by simplifying the code) or would it only push projects to ask for a longer support of Behat 3.8 due to needing to run their CI on those PHP versions ?

FWIW my opinion is that:

* 7.3 stops receiving security updates less than a month from now, so folk should be moving off it anyway
* There might be people with legacy code that won't / can't drop 7.2 / 7.3 yet - but if so, their CI is presumably running happily with the current Behat 3.8 release, or have already found ways to work around any open issues. There's therefore no particular reason they'd need a future 3.8 release.
* Likewise those users probably aren't making major updates to other dependencies, so they're not likely to need a new 3.8 to resolve a new issue with some future release of a third-party package.
* If they hit some as-yet undiscovered behat bug, there are plenty options for them to work round / fork / patch / otherwise fix it to run on their legacy platform.

So I'm not really sure why anyone on <= 7.3 would ever have a pressing need for a new 3.8 release.

The people that need new releases of behat are the ones that *are* upgrading their own projects. And the work for maintainers to provide those is significantly lower if it only supports current PHP versions - primarily because of the impact on behat's own dependencies, and the potential for conflicts with end-user apps.

IMO, if you make a business decision to keep an app on an unsupported PHP version, the cost / hassle of working around that has to fall on you as the end-user. It's not reasonable to expect the open-source community to facilitate that long-term.

Essentially I guess the question is whether you prefer:

* Core behat/behat keeps moving & supported, potentially a minority of users do a community fork (or an official divergent 3.8 branch) to allow them to stay on 7.2 / 7.3, which eventually fades away.

* Core behat/behat gets stalled on the complexity of supporting very different PHP versions & dependencies, lots of users end up doing personal forks / hacks to get it to work on PHP 8 / 8.1 / 9 / whatever, which fragments the community & pushes everyone off into silos of repeated/duplicated effort long-term.

Just my $0.02.

B4rb4ross4:
> Please keep at least PHP 7.3 until some major unavailable language features are needed

Currently you could not use behat in a Symfony 6 Project, because Symfony has a dependency on psr/container ^2 with has >=7.4

> symfony/dependency-injection  v6.0.2  requires  psr/container (^1.1|^2.0)  
> symfony/service-contracts     v3.0.0  requires  psr/container (^2.0)

If you want to use behat you can still use an old version, but keeping unsupported PHP versions simply locks out everyone who wants current software.

I think doing a new Behat major version and dropping the php 7 support entirely should be the way to go.

stof:
@B4rb4ross4 `symfony/service-contracts v3.0.0` indeed requires `psr/container (^2.0)`. but Symfony 6 requires `symfony/service-contracts (^1.1.6|^2.0|^3.0)`. So it is still possible to install things.

and Behat itself could add support for newer versions of psr/container without dropping support for older versions at the same time. That's 2 separate things

stronk7:
Yeah, I think that's the key (from @stof)

> and Behat itself could add support for newer versions of psr/container without dropping support for older versions at the same time. That's 2 separate things.

Whenever PHP requirements are going to change, you can create the stable branch `3.10`, keeping requirements unmodified and then go for all the new and breaking stuff in master, which first release will be `4.0`.

And, at the same time, in the `3.10` branch, without changing requirements, fixes can be accepted towards allowing them to work with php81, php82, php83 (while possible and / or there are contributions to it) while keeping the old versions working.

So, that supports `^3`, `^4`... "semantic" versioning, with people adhering to the majors that their projects really need, being immune to "sudden" changes to requirements.

In fact, I recently commented @ "sister" https://github.com/FriendsOfBehat/MinkExtension/pull/11#issuecomment-1005970629 because of the very same problem (sudden raise of requirements).

I really think that things can be better organised, in fact I'm sure that if we go towards the branched alternative, people will be keen to try to keep the "old" ones working, because in practice, everybody out there uses "old" versions (because of distros, because of libraries...) a lot.

Anyway, that's the way I see it, how we could easily make things better for everybody.

Ciao :-)

Taluu:
There's no need to bump the major version for a dependency upgrade (bunmping requirements) if the api doesn't change (this is the case here), as stated in the semver document : https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api

So if you can't use the newest php version, you will be stuck on an inferior minor version, as composer won't install it. End of story.

If you want the newest *features* that are brought by with the requirements increase... then upgrade your stuff. You can't have both, as it's a hassle to maintain something for a version that is not even supported anymore by the upstream of php for example, and forces to use hacks to support newest and actual versions.

stronk7:
I agree there is not "need". Just saying it "may work", with a major or a minor branching, I don't mind how the "next", incompatible with old php versions, is named (3.11 or 4.0), just interested into keeping 3.10.x working as long as possible and accepting stuff there that may be interesting.

All this because using a "radical/quick" pace adopting new php versions (and discarding old ones, specially this) is not how the world, in practice, works.

Of course, I'm looking to this from the "client" POV, heh.

Also note that all this discussion is relatively non-sense beacuse, what I'm interested on, is about testing my products with old PHP versions (web servers using them). That's completely different from the PHP version to run behat (just it's a general commodity for both to be the same).

Thanks for all the hard work!

Ciao :-)

ciaranmcnulty:
We've created a `4.x` branch, @pamil can you please PR there with what you think a sensible support level would be?

PR #1370, opened 2021-10-26
Fix SYMFONY_REQUIRE for github action

dmaicher:
I made a mistake in https://github.com/Behat/Behat/pull/1368

Now the proper Symfony version is actually installed.


pamil:
Thanks @dmaicher! 🥇

PR #1398, opened 2022-07-06
Fix CI fails because symfony/flex is blocked

Chekote:
## Problem:
When running the workflow, it fails at the "Install symfony/flex" step with the following error:

Error: symfony/flex contains a Composer plugin which is blocked by your allow-plugins config. You may add it to the list if you consider it safe.

## Cause:
A newer version of Composer has introduced a strict check whereby all plugins have to be explicitly allowed for security reasons.

## Fix:
Explicitly allow the global plugin before installing it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,7 +61,9 @@ jobs:
                 run: composer config minimum-stability ${{ matrix.stability }}
 
             -   name: Install symfony/flex
-                run: composer global require symfony/flex
+                run: |
+                    composer config --global --no-plugins allow-plugins.symfony/flex true &&
+                    composer global require symfony/flex
 
             -   name: Uninstall Psalm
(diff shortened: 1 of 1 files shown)


stof (approved):

stof:
This is used to restrict Symfony versions with `SYMFONY_REQUIRE`

ciaranmcnulty:
ok but why do we need that outside of what composer does?

stof:
Composer does not have a built-in way to restrict `symfony/*` packages to the LTS version (as 6.1 is a higher version number than 5.4). This requires using a plugin to filter the pool (which is one of the features of symfony/flex).

ciaranmcnulty:
@stof I'll bother you on slack about this, I'm totally in the dark as to why this is necessary / isn't something can be done with composer

ciaranmcnulty:
Thanks @Chekote

PR #1401, opened 2022-07-07
Don't always install symfony/flex / add separate symfony LTS jobs

ciaranmcnulty:
I am concerned Flex may be masking some issues for non-symfony projects; this removes the step that always installs Flex, and instead only installs it for specific symfony LTS builds

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,33 +18,49 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
+                php: [7.2, 7.3, 7.4, 8.0, 8.1]
+                os: [ubuntu-latest]
+                composer-mode: [update]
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -18,33 +18,49 @@ jobs:
...
+                        composer-mode: update
+                        symfony-version: '4.4'
+                    -   php: 8.1
+                        os: ubuntu-latest
+                        composer-mode: update
+                        symfony-version: '5.4'
+                    -   php: 8.1
+                        os: ubuntu-latest
+                        composer-mode: update
+                        symfony-version: '6.0'
```
stof, line 63: I would drop this one, keeping only jobs for LTS versions

PR #1404, opened 2022-07-10
Draft: Use cucumber/gherkin parser in Behat

ciaranmcnulty:
This incorporates the compatibility layer from https://github.com/Behat/Gherkin/pull/253 into Behat, as a configuration file flag

```yml
default:
  gherkin:
    parser: cucumber
```

The expectation is that at some point in the future we could flip the default behaviour to be 'use the cucumber one if cucumber/gherkin is installed'

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,6 +22,7 @@ jobs:
                 os: [ubuntu-latest]
                 composer-mode: [update]
                 symfony-version: ['']
+                cucumber-parser: [false]
                 include:
                     # Get the existing 7.2 to publish the phar
                     -   php: 7.2
@@ -29,38 +30,52 @@ jobs:
                         composer-mode: update
                         publish-phar: true
                         symfony-version: ''
+                        cucumber-parser: false
 
                     # 7.2 build with lowest dependencies
                     -   php: 7.2
                         os: ubuntu-latest
                         composer-mode: lowest
                         symfony-version: ''
+                        cucumber-parser: false
+
+                    # Cucumber-parser build
+                    -   php: 8.1
+                        os: ubuntu-latest
+                        composer-mode: update
+                        symfony-version: ''
+                        cucumber-parser: true
 
                     # MacOS on latest PHP only
                     -   php: 8.1
                         os: macos-latest
                         composer-mode: update
                         symfony-version: ''
+                        cucumber-parser: false
 
                     # Windows on latest PHP only
                     -   php: 8.1
                         os: windows-latest
                         composer-mode: update
                         symfony-version: ''
+                        cucumber-parser: false
 
                     # Symfony jobs:
                     -   php: 8.1
                         os: ubuntu-latest
                         composer-mode: update
                         symfony-version: '4.4'
+                        cucumber-parser: false
                     -   php: 8.1
                         os: ubuntu-latest
                         composer-mode: update
                         symfony-version: '5.4'
+                        cucumber-parser: false
                     -   php: 8.1
                         os: ubuntu-latest
                         composer-mode: update
                         symfony-version: '6.0'
+                        cucumber-parser: false
 
         steps:
             -   uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -16,7 +16,7 @@
     "require": {
         "php": "^7.2 || ^8.0",
         "ext-mbstring": "*",
-        "behat/gherkin": "^4.9.0",
+        "behat/gherkin": "dev-cucumber-gherkin",
```
ciaranmcnulty, line 19: DO NOT MERGE with this

.github/workflows/build.yml
```diff
@@ -72,12 +87,24 @@ jobs:
...
                     coverage: none
 
-            -   name: Install symfony/flex
+            -   name: Require symfony/flex
                 if: matrix.symfony-version != ''
                 run: |
                     composer config --global --no-plugins allow-plugins.symfony/flex true &&
                     composer global require symfony/flex
 
+            -   name: Temporary - add ciaran fork
```
ciaranmcnulty, line 96: DO NOT MERGE with this

src/Behat/Behat/Gherkin/ServiceContainer/GherkinExtension.php
```diff
@@ -97,6 +99,11 @@ public function configure(ArrayNodeDefinition $builder)
                     ->useAttributeAsKey('name')
                     ->prototype('scalar')->end()
                 ->end()
+                ->enumNode('parser')
+                    ->info('Sets which gherkin parser to use if not specified at CLI')
+                    ->values(['auto', 'cucumber', 'legacy'])
```
pamil, line 104: Legacy -> Behat?

src/Behat/Behat/Gherkin/ServiceContainer/GherkinExtension.php
```diff
@@ -208,26 +215,40 @@ private function loadParser(ContainerBuilder $container)
...
+        }
+
+        if ($parserName == 'cucumber') {
+            if (!class_exists(CucumberGherkinLoader::class) || !CucumberGherkinLoader::isAvailable()) {
+                throw new ParserNotAvailable('Can not specify the cucumber parser without cucumber/gherkin');
+            }
+
+            $definition = new Definition('Behat\Gherkin\Loader\CucumberGherkinLoader');
+            $definition->addMethodCall('setBasePath', array('%paths.base%'));
+            $definition->addTag(self::LOADER_TAG, array('priority' => 100));
```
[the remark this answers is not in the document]
ciaranmcnulty: Probably. I think if we were doing 'auto' mode more cleanly we could register both but have older parser be higher priority?

src/Behat/Behat/Gherkin/ServiceContainer/GherkinExtension.php
ciaranmcnulty: Yeah I don't like either name :)

Legacy is too harsh but 'behat' is confusing as the new one is used by behat too. Any suggestions?

src/Behat/Behat/Gherkin/ServiceContainer/GherkinExtension.php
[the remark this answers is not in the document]
pamil: Should it be even possible to have two Gherkin parsers? How would that work together?

src/Behat/Behat/Gherkin/ServiceContainer/GherkinExtension.php
pamil: Yeah, `behat` is not my favourite too, but makes it consistent with `cucumber`? What about if we make it a package name? `behat/gherkin` or `cucumber/gherkin` - could be more discoverable this way.

src/Behat/Behat/Gherkin/ServiceContainer/GherkinExtension.php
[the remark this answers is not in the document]
ciaranmcnulty: The highest priority that `supports()` the resource being loaded wins

IRL the resource is always a file and both parsers will say they can load it

src/Behat/Behat/Gherkin/ServiceContainer/GherkinExtension.php
ciaranmcnulty: How about `classic`?

ciaranmcnulty:
I want to come back to this and implement a CLI flag too; however it's not quite clear how to do that in the container so it might need to be reworked as a runtime switch

behat.yml.dist
```diff
@@ -1,4 +1,4 @@
 default:
   gherkin:
     filters:
-      tags: ~@php8
+      tags: ~@php8&&~@cucumber-parser
```
stof, line 4: We should probably handle that with 2 suites running the same scenarios with each parser instead of having only a few scenarios running with the new parser. The tag would then be used only to tag scenarios that _require_ the cucumber parser

ciaranmcnulty: Hm how about a second profile too?

`default` -  runs `~@php8&&~@cucumber-parser`
`cucumber` - includes `@cucumber-parser` and also enables the config setting?

stof: yeah, a profile rather than a suite might do it.

PR #1414, opened 2022-10-15
Add .editorconfig file

chapeupreto:
Add .editorconfig file

See: https://editorconfig.org/

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,15 @@
+root = true
+
+[*]
+charset = utf-8
+indent_size = 4
+indent_style = space
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false
+
+[*.{yml,yaml}]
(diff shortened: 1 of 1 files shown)


ciaranmcnulty:
I don't see a strong reason for this, sorry

stof:
@ciaranmcnulty I actually find this quite convenient as it avoids having to reconfigure manually the formatting setting of the IDE for new contributors (is this project using 2 spaces or 4 spaces in Yaml files ? Editorconfig handles that automatically)

chapeupreto:
It is worth mentioning that big projects are already using such a file, e.g., `Composer`:

https://github.com/composer/composer/blob/main/.editorconfig

Why not use it here as well? I mean, what harm can it do anyways?

ciaranmcnulty:
ok, I'm convinced. I can't re-open 'cause the branch is deleted

chapeupreto:
@ciaranmcnulty my bad! I submitted another PR here: https://github.com/Behat/Behat/pull/1418

PR #1417, opened 2022-11-29
PHP 8.2 support

ciaranmcnulty:


.github/workflows/build.yml
```diff
@@ -37,27 +37,27 @@ jobs:
...
                     # Windows on latest PHP only
-                    -   php: 8.1
+                    -   php: 8.2
                         os: windows-latest
                         composer-mode: update
                         symfony-version: ''
 
                     # Symfony jobs:
-                    -   php: 8.1
+                    -   php: 8.2
```
stof, line 52: As Symfony 4.4 reaches EOM before the 8.2 release, I suggest keeping that job on PHP 8.1

ciaranmcnulty: Aha, we should think about dropping it too then?

PR #1420, opened 2023-01-11
feature: `--allow-no-tests` option to pass even if no specifications found

### LINKED ISSUE #1389: pass test cases having scenarios with skipped tags
I have migrated from 2.5 to 3.10 and I am executing test cases via following command:

`behat  --format=progress --tags '~@wip' features/feature-file.feature`

Feature file has tag @wip and has one scenario.

So when I execute test case, it skips the scenario which is right but I get error

`In ExerciseController.php line 111: No specifications found at path(s) features/feature-file.feature. This might be because of incorrect paths configuration in your suites.`

How can I pass all those features file which are actually skipped? These skipped test cases are appearing as failed on pipeline.

Kingdutch:
In our CI we split out test jobs based on folder. Additionally we run each `.feature` file individually so we can create output test groups. However, due to how we've arranged our tags and our test matrices it could be that Behat is called with a path to a file that has all tests excluded using a tag.

The change in this PR provides an option to mute this error and prevent a CI from failing when no tests were executed deliberately.

Fixes #1389

### DIFF
--- a/src/Behat/Testwork/Tester/Cli/ExerciseController.php
+++ b/src/Behat/Testwork/Tester/Cli/ExerciseController.php
@@ -102,6 +102,12 @@ final class ExerciseController implements Controller
                 null,
                 InputOption::VALUE_NONE,
                 'Invokes formatters without executing the tests and hooks.'
+            )
+            ->addOption(
+                '--allow-no-tests',
+                null,
+                InputOption::VALUE_NONE,
+                'Will not fail if no specifications are found.'
             );
     }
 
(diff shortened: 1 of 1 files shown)


acoulton (changes requested):
This feels risky to me - personally I'd want my tooling to explicitly skip combinations of files / tags that were expected not to run any tests. Otherwise there's a chance at some point behat stops finding tests (due to changes in the project or e.g. a behat upgrade) but goes green without running anything.

However, happy to merge it as an opt-in if it's useful to others.

But I find `--allow-empty` a bit unclear. Could it be `--allow-no-tests` instead?

src/Behat/Testwork/Tester/Cli/ExerciseController.php
```diff
@@ -94,6 +94,9 @@ function ($locator) {
             ->addArgument('paths', InputArgument::OPTIONAL,
                 'Optional path(s) to execute. Could be:' . PHP_EOL . $locatorsExamples
             )
+            ->addOption('--allow-empty', null, InputOption::VALUE_NONE,
+              'Will not fail if no specification are found.'
```
carlos-granados, line 98: Should be "if no specifications are found"

carlos-granados (changes requested):

Kingdutch:
@acoulton @carlos-granados Thanks for your feedback! Took me a bit of time to get to it. I've renamed the flag to `--allow-no-tests`, that's a clearer name. I've also fixed the description to make `specifications` plural.

carlos-granados (approved):
Thanks @Kingdutch

acoulton (approved):
Great, thanks for coming back to this @Kingdutch

Issue #1421, opened 2023-02-08
PHPUnit 10 Assertions failure results in "Uncaught Error: Class "PHPUnit\Framework\TestFailure" not found"


mnocon:
Hi!

When a PHPUnit assertion fails in a Behat Context class the tests fail with:
```
~/Desktop/repos/behat-phpunit10 on main !2 ?2 ❯ vendor/bin/behat                                                                                                                                                       14.17.0 at 10:36:00
Feature: PHPUnit 10 exception

  Scenario: Trigger PHPUnitExceptionStringer # features/Reproducer.feature:3
    Given ExceptionStringer is triggered     # FeatureContext::triggerExceptionStringer()
PHP Fatal error:  Uncaught Error: Class "PHPUnit\Framework\TestFailure" not found in /Users/mareknocon/Desktop/repos/behat-phpunit10/behat/src/Behat/Testwork/Exception/Stringer/PHPUnitExceptionStringer.php:49
Stack trace:
#0 /Users/mareknocon/Desktop/repos/behat-phpunit10/behat/src/Behat/Testwork/Exception/ExceptionPresenter.php(93): Behat\Testwork\Exception\Stringer\PHPUnitExceptionStringer->stringException(Object(PHPUnit\Framework\ExpectationFailedException), 1)
#1 /Users/mareknocon/Desktop/repos/behat-phpunit10/behat/src/Behat/Behat/Output/Node/Printer/Pretty/PrettyStepPrinter.php(174): Behat\Testwork\Exception\ExceptionPresenter->presentException(Object(PHPUnit\Framework\ExpectationFailedException))
#2 /Users/mareknocon/Desktop/repos/behat-phpunit10/behat/src/Behat/Behat/Output/Node/Printer/Pretty/PrettyStepPrinter.php(95): Behat\Behat\Output\Node\Printer\Pretty\PrettyStepPrinter->printException(Object(Behat\Testwork\Output\Printer\StreamOutputPrinter), Object(Behat\Behat\Tester\Result\ExecutedStepResult))
[...]

PR #1422, opened 2023-02-08
Added support for ThrowableToStringMapper from PHPUnit 10

mnocon:
My attempt at solving https://github.com/Behat/Behat/issues/1421

This is my first contribution to Behat, so if I've missed anything please let me know 🙇

PR #1426, opened 2023-04-08
Exception stringification followup

uuf6429:
This is the follow up for https://github.com/Behat/Behat/issues/1421#issuecomment-1500089547

On a side note, I'm wondering if `supportsException` should also check for the stringification classes.
Although its name relates to the exception, the usage actually relates to if it can stringify an exception or not.

### DIFF
--- a/src/Behat/Testwork/Exception/Stringer/PHPUnitExceptionStringer.php
+++ b/src/Behat/Testwork/Exception/Stringer/PHPUnitExceptionStringer.php
@@ -35,17 +35,24 @@ final class PHPUnitExceptionStringer implements ExceptionStringer
      */
     public function stringException(Exception $exception, $verbosity)
     {
-        if (class_exists('PHPUnit\\Util\\ThrowableToStringMapper')) {
-            return trim(\PHPUnit\Util\ThrowableToStringMapper::map($exception));
-        }
-
-        if (!class_exists('PHPUnit\\Framework\\TestFailure')) {
-            return trim(\PHPUnit_Framework_TestFailure::exceptionToString($exception));
-        }
-
         // PHPUnit assertion exceptions do not include expected / observed info in their
         // messages, but expect the test listeners to format that info like the following
         // (see e.g. PHPUnit_TextUI_ResultPrinter::printDefectTrace)
-        return trim(\PHPUnit\Framework\TestFailure::exceptionToString($exception));
+
+        switch (true) {
+            case class_exists(\PHPUnit\Util\ThrowableToStringMapper::class):
+                return trim(\PHPUnit\Util\ThrowableToStringMapper::map($exception));
+
+            case class_exists(\PHPUnit\Framework\TestFailure::class):
+                return trim(\PHPUnit\Framework\TestFailure::exceptionToString($exception));
+
+            case class_exists(\PHPUnit_Framework_TestFailure::class):
+                return trim(\PHPUnit_Framework_TestFailure::exceptionToString($exception));
+
+            default:
+                throw new \RuntimeException(
+                    'Cannot stringify PHPUnit exception, a valid stringifier class could not be found'
+                );
+        }
     }
 }


ciaranmcnulty:
Thanks @uuf6429 I like the idea to add an exception, but not a fan of the change to a switch/case

I'll try and look at the test suite and add something to the matrix that would have caught the issue before #1422 was merged, so we can check that this change is preserving the same behaviour

carlos-granados (approved):
LGTM

src/Behat/Testwork/Exception/Stringer/PHPUnitExceptionStringer.php
```diff
@@ -35,17 +35,24 @@ public function supportsException(Exception $exception)
...
+            case class_exists(\PHPUnit\Framework\TestFailure::class):
+                return trim(\PHPUnit\Framework\TestFailure::exceptionToString($exception));
+
+            case class_exists(\PHPUnit_Framework_TestFailure::class):
+                return trim(\PHPUnit_Framework_TestFailure::exceptionToString($exception));
+
+            default:
+                throw new \RuntimeException(
+                    'Cannot stringify PHPUnit exception, a valid stringifier class could not be found'
+                );
```
acoulton, line 55: Throwing here means if a scenario fails, users won't have any idea why / where the exception was actually thrown - which could be breaking in some cases.

I'd prefer we still just always return a string with as much info as we can safely provide e.g. `Encountered a {exception::class} with the message {..} but could not find a valid stringifier to render the details`.

This avoids any breakage and means that users will be able to see what went wrong with their actual tests. They can then choose whether to just fix the failing test, or figure out how to get more detail stringified out of the exception, depending what they need.

src/Behat/Testwork/Exception/Stringer/PHPUnitExceptionStringer.php
```diff
@@ -35,17 +35,24 @@ public function supportsException(Exception $exception)
...
         // messages, but expect the test listeners to format that info like the following
         // (see e.g. PHPUnit_TextUI_ResultPrinter::printDefectTrace)
-        return trim(\PHPUnit\Framework\TestFailure::exceptionToString($exception));
+
+        switch (true) {
+            case class_exists(\PHPUnit\Util\ThrowableToStringMapper::class):
+                return trim(\PHPUnit\Util\ThrowableToStringMapper::map($exception));
+
+            case class_exists(\PHPUnit\Framework\TestFailure::class):
+                return trim(\PHPUnit\Framework\TestFailure::exceptionToString($exception));
```
acoulton, line 47: These classes are both [annotated that they're not covered by the phpunit Backward Compatibility promise](https://github.com/sebastianbergmann/phpunit/blob/f72a244153efe365f97c053f7b328ec811b93fe9/src/Util/ThrowableToStringMapper.php#L22C4-L22C89).

I don't think there's an easy way to replicate the phpunit behaviour while guaranteeing BC. But (as with my other comment) I think we should do our best to return some sort of string that reflects the user's own test failure even if e.g. they upgrade to a new phpunit with a different method signature. So we should wrap this whole thing in a `try {...} catch (Throwable $e) { return 'a useful string incorporating the original exception class & message'}`.

acoulton (changes requested):
Thanks for this, but I'm not sure it's the right solution yet.

src/Behat/Testwork/Exception/Stringer/PHPUnitExceptionStringer.php
```diff
@@ -35,17 +35,24 @@ public function supportsException(Exception $exception)
...
+
+            case class_exists(\PHPUnit_Framework_TestFailure::class):
+                return trim(\PHPUnit_Framework_TestFailure::exceptionToString($exception));
```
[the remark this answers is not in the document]
uuf6429: I'm also not a fan of this, but back in the day there seemed to be a lot of pressure on maintaining BC compatibility, even if it meant supporting extremely old unsupported PHP versions. I can tell you this class wasn't meant for PHPUnit 6, but even older. 😅 

I'm fine with removing this, if others agree as well.

src/Behat/Testwork/Exception/Stringer/PHPUnitExceptionStringer.php
uuf6429: If we want to be absolutely safe (by catching all exceptions), I would still consider rethrowing an exception with the original exception as "previous" (as per my previous comment). That should provide the user with all the relevant info and feels like a rather standard way to do things.

src/Behat/Testwork/Exception/Stringer/PHPUnitExceptionStringer.php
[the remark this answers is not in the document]
acoulton: I think there's a difference between runtime BC - things shouldn't start throwing exceptions or failing scenarios - and cosmetic BC - if a scenario fails, we should always render exactly the same failure message.

Basically, would the change break someone's workflow or force them to upgrade a php version / other tool?

I don't think that applies here.

src/Behat/Testwork/Exception/Stringer/PHPUnitExceptionStringer.php
acoulton: If this was a library that people were doing their own try/catch I'd agree with using $previous (and it certainly wouldn't hurt to include it).

But in this case it depends on whether behat renders the class/trace of the previous exception - I think it doesn't by default?

Basically, if someone's CI gets a transient failure, is the output in their CI logs going to look like there's a problem with their test/app, or is it going to look like behat is broken and send them down a rabbit hole thinking they need to upgrade/downgrade dependencies immediately?

As a developer in that situation I would much prefer to see the basic info of the assertion that failed, even if I don't get the full diff, than a generic message about stringification.

uuf6429: > But in this case it depends on whether behat renders the class/trace of the previous exception - I think it doesn't by default?

I don't know yet, I would need to check that first. If this isn't handled well, I'd argue it would be a problem in Behat that should be solved.

> Basically, if someone's CI gets a transient failure, is the output in their CI logs going to look like there's a problem with their test/app, or is it going to look like behat is broken and send them down a rabbit hole thinking they need to upgrade/downgrade dependencies immediately?

I get what you mean, but lacking knowledge how to stringify the PHPUnit exception means there's always going to be a limit to what we can stringify.

That limit is essentially going to be our opinion of what we can do. If we decide to present `(string) $exception` to the user, there could always be a complaint that we should have an `$exception instanceof Diffable ? $exception->diff() : (string)$exception` (naive example) - so I believe the expected case is to provide all the details possible, in a standard way (rethrowing an exception). That of course only applies if the first point you made is working fine.

> As a developer in that situation I would much prefer to see the basic info of the assertion that failed, even if I don't get the full diff, than a generic message about stringification.

I would argue that the first priority (for the end user) would be that Behat is operating in a consistent state. If they get these two exceptions, they shouldn't just fix the test and ignore the rest.

----

By the way, if we had to be more pragmatic about this, I'd argue that my change is already and improvement over the current situation. Right now you'd get a somewhat ambiguous fatal error because it assumes `TestFailure` must exist... so maybe the rest of the discussion we're having might make more sense to discussed separately to unblock this PR. Just a thought.

src/Behat/Testwork/Exception/Stringer/PHPUnitExceptionStringer.php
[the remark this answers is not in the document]
uuf6429: I'm sorry, I'm not really following your point here. 🤔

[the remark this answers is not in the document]
acoulton: Sorry for being unclear.

My point was just that I don't believe changing the content of a failure message (which is all that would happen if we removed support for the PHPUnit_Framework_XXX class names) is a BC break. 

And therefore I think they should go, to keep our own code simpler & more maintainable.

src/Behat/Testwork/Exception/Stringer/PHPUnitExceptionStringer.php
acoulton: I've been thinking about all this, but I disagree.

> If behat doesn't render `previous` we should solve that elsewhere. (paraphrasing)

Usually, I wouldn't want behat to render the full `$previous` stack - unless I'd specified a higher-than-default verbosity. Lots of the assertions / potential failures in our own suites have a `$previous` stack and it would produce a lot of noise that we don't usually need.

> I get what you mean, but lacking knowledge how to stringify the PHPUnit exception means there's always going to be a limit to what we can stringify.

Correct. But behat doesn't use or offer phpunit assertions itself (which is why we don't have a hard dependency on a specific phpunit version). We are stringifying here as a convenience to the user, but the tradeoff is we might not always know what to do with them if they're using a phpunit version we don't expect.

If the user really wants to use an unexpected / unsupported phpunit version, then they can always wrap assertions in their own helper function to format the failures (which they might have to do if they were using some other third-party assertion library that didn't put the detail in the exception message itself).

So if e.g. there was ` a complaint that we should have an $exception instanceof Diffable ? $exception->diff() : (string)$exception ` I'd say no - we are not phpunit, we support some common setups, if those don't work for you then it's safer & better all round if you handle it yourself.

> I would argue that the first priority (for the end user) would be that Behat is operating in a consistent state. If they get these two exceptions, they shouldn't just fix the test and ignore the rest.

It depends on the situation. Suppose you are making a small change, upgrading a single package, or bumping PHP version in a project.

You might get a set of build failures related to that which actually only need a very easy fix to resolve them. That's clearly the business priority.

If you can't easily even find out what those are, or it looks like you can't get the build green without changing your phpunit version - which could mean major changes to your unit tests - that's a problem. I've seen variations of this situation many times over the years and it is incredibly frustrating.

Sure, you should also ideally fix the issue that is preventing behat giving you detailed stringification, but it's easy to imagine lots of times where you'd want to do that after and separately from fixing your own actual product.

The more I think about it, the more convinced I am that our inability to stringify the detail should be a warning at most, and the priority should be rendering the type, message and trace of the actual test failure to allow the end-user to get on with their day.

>  I'd argue that this PR is already an improvement over the current situation (when no major downsides). Right now you'd get a somewhat ambiguous fatal error because it assumes TestFailure must exist...

I'm afraid I disagree. The status quo is you get a somewhat unhelpful message about behat/phpunit internals that masks your own test failure. With this PR, you still get a somewhat unhelpful message about behat/phpunit internals that masks your own test failure - it's just a different one. Neither of them tell you what you as a user can do about it.

src/Behat/Testwork/Exception/Stringer/PHPUnitExceptionStringer.php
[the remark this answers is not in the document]
uuf6429: In that case I would disagree - since we are changing existing behaviour, it's still a BC break for me. The end user might be depending on the exact string, for one.

But anyway, as said earlier I'm fine with removing this as long as there's broad agreement.

src/Behat/Testwork/Exception/Stringer/PHPUnitExceptionStringer.php
uuf6429: > Usually, I wouldn't want behat to render the full $previous stack...

What I meant was, if the previous exception is not rendered in the relevant verbose mode (because we forgot about it, for example), then it should be fixed. I also wouldn't show all these details by default, which is why I've said since the start that the advantage of an exception is that the user gets to see as much (or as little) details as they want.

If we simply return a string, we are deciding for the user what they will see. If we return the stacktrace, the user may complain that it's too much. If we don't, the user may complain that it's too little and not really understandable why the output is not properly formatted. Even then, we might miss very useful information for the user.. I wouldn't be surprised that the exception message would be "Failed asserting that false is true" and the stack trace pointing to somewhere deep inside PHPUnit.

> If the user really wants to use an unexpected / unsupported phpunit version, then they can always wrap assertions in their own helper function to format the failures

That's all fine... but only if we are also hinting that it's not supported, otherwise the user won't know that and we'll be getting bug reports from them about unexpected strings (if/when they realise it).

> the priority should be rendering the type, message and trace of the actual test failure

The problem with that is as I said earlier, it may be too much (as you said, you might not want the full trace) or too little (as mentioned in the code, the message might misss useful info that would have been printed out by PHPUnit).

> I'm afraid I disagree. The status quo is you get a somewhat unhelpful message about behat/phpunit internals that masks your own test failure. With this PR, you still get a somewhat unhelpful message about behat/phpunit internals that masks your own test failure - it's just a different one. Neither of them tell you what you as a user can do about it.

Well we cannot fix everything in one pull request... the point of this PR was to improve some code - not fix a bug nor add some feature. So it's feasible to release it as long as the change is useful (even to internal devs) and not breaking anything.
And frankly:
1. having a default case means being clear about the intention and not an "oops we forgot this case might happen" (having a "class not found" php error tends to smell of sloppy coding or some major breakage)
2. passing $previous is already potentially helpful to the end user (as said, only when it becomes somehow visible, e.g. by increasing verbosity)


In any case, feel free to suggest some code.

carlos-granados:
@uuf6429 @acoulton what's needed to get this over the line?

uuf6429:
Closing this in favour of #1521, which handles the situation very well.

acoulton:
@uuf6429 thanks for highlighting the issue and for the time preparing this PR and the discussion around it, was very helpful getting to the solution.

PR #1427, opened 2023-04-08
Improved unit tests CS; fixed autoload for tests

uuf6429:


composer.json
```diff
@@ -50,7 +50,7 @@
 
     "autoload-dev": {
         "psr-4": {
-            "Behat\\Tests\\": "tests/"
+            "Behat\\Tests\\": "tests/Behat/Tests"
```
uuf6429: The `autoload-dev` config for tests was wrong/broken. This can be contrasted with the `autoload` config:
https://github.com/Behat/Behat/blob/29344fdf8dfaf818ad4a111d3d3121536d1e1ef1/composer.json#L44

uuf6429:
@stof @pamil can someone please review this and get it merged?

acoulton (approved):
I've resolved the merge conflicts, this looks OK to me.

tests/Behat/Tests/Testwork/Argument/MixedArgumentOrganiserTest.php
```diff
@@ -1,24 +1,26 @@
 <?php
 
+/** @noinspection PhpUnhandledExceptionInspection */
```
carlos-granados, line 3: I don't think we should add IDE specific tags to the code so I would rather remove this

carlos-granados (changes requested):

tests/Behat/Tests/Testwork/Argument/MixedArgumentOrganiserTest.php
acoulton: Good spot, I missed that.

uuf6429: Since that doesn't affect end users, I'd say it depends mostly on the maintainers/contributors since it would only serve to help development.
In any case, it seems the feeling is unanimous, so out it goes. :)

acoulton: I think it's a "slippery slope" argument. If people are using different tooling, you end up with a codebase full of different IDE-specific comments and eventually they start to create noise/diff conflicts/etc. Or pulls that just upgrade or change comments to suit a new version of an IDE that pollute the history for end-users.

So I agree it's better to just have a policy to stick to standard code & annotations in public projects.

carlos-granados (approved):
All good now. Thanks @uuf6429!

PR #1428, opened 2023-04-08
Allow duplicate steps on same method

### LINKED ISSUE #1425: Same step pattern, different step types are not allowed; additionally step type is ignored in feature files
Steps defining all step type like so:
```php
    #[Given('/^I take a page screenshot(?: as (.+))?$/')]
    #[Then('/^I take a page screenshot(?: as (.+))?$/')]
    #[When('/^I take a page screenshot(?: as (.+))?$/')]
    public function takeScreenshot(?string $name = null): void
    {
```
are causing errors like so:
```gherkin
  Scenario: clicking on something  # features/something.feature:3
    Given that I do something
      Step "/^I take a page screenshot(?: as (.+))?$/" is already defined in <...>\Debug\Context::takeScreenshot()
      
      <...>\Debug\Context::takeScreenshot()
      <...>\Debug\Context::takeScreenshot()
    And I am on page 1
      Step "/^I take a page screenshot(?: as (.+))?$/" is already defined in <...>\Debug\Context::takeScreenshot()
```

The same thing is happening for PHPDoc-based step annotations.

If I remove the `#[Given..` and `#[Then..`, the errors are gone, and behat doesn't complain that I have a step defined as `When` but is used as `Given` in the feature file, for example.

Why does this happen? What's the point of using the step type (e.g. `@Given`) if it is ultimately ignored? (why not just have a `@Step` annotation/attribute in that case?)

In my opinion, steps should be consider unique by *step pattern* AND *step type*. Additionally, misusing steps (defined as `Given`, but used as `Then`) should trigger a runtime warning.

uuf6429:
Fixes #1425

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -268,8 +268,8 @@ EOL;
      */
     public function itShouldPassWith($success, PyStringNode $text)
     {
-        $this->itShouldFail($success);
         $this->theOutputShouldContain($text);
+        $this->itShouldFail($success);
     }
 
     /**
@@ -281,8 +281,8 @@ EOL;
      */
     public function itShouldPassWithNoOutput($success)
     {
-        $this->itShouldFail($success);
         Assert::assertEmpty($this->getOutput());
+        $this->itShouldFail($success);
     }
 
     /**
--- a/src/Behat/Behat/Context/Reader/AnnotatedContextReader.php
+++ b/src/Behat/Behat/Context/Reader/AnnotatedContextReader.php
@@ -13,6 +13,8 @@ namespace Behat\Behat\Context\Reader;
 use Behat\Behat\Context\Annotation\AnnotationReader;
 use Behat\Behat\Context\Annotation\DocBlockHelper;
 use Behat\Behat\Context\Environment\ContextEnvironment;
+use Behat\Behat\Definition\Call\RuntimeDefinition;
+use Behat\Behat\Definition\Definition;
 use Behat\Testwork\Call\Callee;
 use ReflectionClass;
 use ReflectionException;
@@ -129,6 +131,7 @@ final class AnnotatedContextReader implements ContextReader
         $description = $this->docBlockHelper->extractDescription($docBlock);
         $docBlock = $this->mergeMultilines($docBlock);
 
+        $patterns = [];
         foreach (explode("\n", $docBlock) as $docLine) {
             $docLine = preg_replace(self::DOCLINE_TRIMMER_REGEX, '', $docLine);
 
@@ -140,9 +143,18 @@ final class AnnotatedContextReader implements ContextReader
                 continue;
             }
 
-            if ($callee = $this->readDocLineCallee($class, $method, $docLine, $description)) {
-                $callees[] = $callee;
+            if (!($callee = $this->readDocLineCallee($class, $method, $docLine, $description))) {
+                continue;
             }
+
+            if ($callee instanceof Definition) {
+                if (isset($patterns[$callee->getPattern()])) {
+                    continue;
+                }
+                $patterns[$callee->getPattern()] = true;
+            }
+
+            $callees[] = $callee;
         }
 
         return $callees;
--- a/features/attributes.feature
+++ b/features/attributes.feature
@@ -66,6 +66,52 @@ Feature: attributes
       """
 
   @php8
+  Scenario: PHP 8 Multi-Step-Type Attributes
+    Given a file named "features/bootstrap/FeatureContext.php" with:
+      """
+      <?php
+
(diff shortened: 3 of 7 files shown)


features/bootstrap/FeatureContext.php
```diff
@@ -268,8 +268,8 @@ public function iRunBehatInDebugMode()
      */
     public function itShouldPassWith($success, PyStringNode $text)
     {
-        $this->itShouldFail($success);
         $this->theOutputShouldContain($text);
+        $this->itShouldFail($success);
```
uuf6429, line 272: I changed the order of this and the check below since comparing the output first is more informative and tends to be more helpful when debugging tests.

src/Behat/Behat/Context/Reader/AnnotatedContextReader.php
```diff
@@ -140,9 +143,18 @@ private function readDocBlockCallees($class, ReflectionMethod $method, $docBlock
...
-                $callees[] = $callee;
+            if (!($callee = $this->readDocLineCallee($class, $method, $docLine, $description))) {
+                continue;
             }
+
+            if ($callee instanceof Definition) {
+                if (isset($patterns[$callee->getPattern()])) {
+                    continue;
+                }
+                $patterns[$callee->getPattern()] = true;
```
uuf6429, line 154: I'm using `[$pattern => true]` & `isset()` since it should be faster than `in_array`, at the cost of readability.

acoulton:
@uuf6429 I'm not convinced this is useful - personally while Given / Then / When etc are interchangeable I think it is better to be clear that the bit we actually do match has to be unique. Otherwise it just creates more scope for confusion.

I think it's only worth / sensible to do this if there was a plan to either make the `Given / Then / When` part of matching in a future BC-Breaking release - but my preference would probably just be to have a single `Step` annotation in Contexts instead (which could be introduced now and deprecate the Given/Then/When).

If you'd like to discuss that further it's probably worth opening a separate issue?

features/bootstrap/FeatureContext.php
acoulton: @uuf6429 this part looks reasonable, if you want to submit as a separate PR?

For future contributions, it would help if you could group unrelated changes into separate commits, with the explanation in the commit message rather than a github comment. That way we don't lose that information if/when the PR is merged. Thanks :)

uuf6429: > For future contributions, it would help if you could group unrelated changes into separate commits

It's unfortunate that these sort of things pop up during development of unrelated tasks. Normally when the change is too big or controversial, I would open a separate issue. When it's something really minor (and directly covered by tests), I'd just go for it since it doesn't seem to be worth the effort to have a different PR.

But anyway, it's fine, let's take that approach. I'll extract the uncontroversial changes into separate PRs.

uuf6429:
> @uuf6429 I'm not convinced this is useful - personally while Given / Then / When etc are interchangeable I think it is better to be clear that the bit we actually do match has to be unique. Otherwise it just creates more scope for confusion.
> 
> I think it's only worth / sensible to do this if there was a plan to either make the `Given / Then / When` part of matching in a future BC-Breaking release - but my preference would probably just be to have a single `Step` annotation in Contexts instead (which could be introduced now and deprecate the Given/Then/When).
> 
> If you'd like to discuss that further it's probably worth opening a separate issue?

This was discussed in the original issue linked to this PR (except the single Step part). I'm rather opinionated in the sense that there should be a functional difference between those 3, but it's something we can live without at the moment.

features/bootstrap/FeatureContext.php
acoulton: Just to be clear, I'm fine with small things being in the same PR so long as separate concerns are in separate commits (with their own commit messages). That way even if we disagree on what's small/groupable then worst case is we split the commits to separate branches during review, which is easy, rather than having to ask people to go back over things.

I'd only suggested putting this into a separate PR to get it merged even though I've declined the other changes. If it was a separate commit already I'd probably just have cherry picked it myself.

stof:
Behat has **never** had different behavior, and Cucumber has never had them either (btw, you could write your gherkin file using `And` for every step if you want, and the Cucumber parser even supports using `*` to start steps, even though I don't think behat/gherkin supports that).

features/bootstrap/FeatureContext.php
uuf6429: > Just to be clear, I'm fine with small things being in the same PR so long as separate concerns are in separate commits (with their own commit messages).

Ah, I didn't get that. All good then.

I've extracted the code, by the way. Thanks.

uuf6429:
@stof I know, that's why I said I'm rather opinionated on this.

I didn't know about `*` though. Interesting. :)

Issue #1306, opened 2020-06-17
Introduce a coding standard tool


pamil:
It'd be nice to have some automated coding standard tool to help us maintain a consistent codebase and get rid of ancient `array()` syntax. What coding standard should Behat follow? I'm open to suggestions.

gmponos:
IMHO you should use both codesniffer and PHP-CS-Fixer for different reasons.

**Codesniffer**
- More configurable than php-cs-fixer
- It's not only about "style" but more into introducing a coding standard as you say on the title of your issue. For instance you can ban functions using codesniffer.. like `die` but it will not autofix them..
- The maintainer of the package chooses not to auto-fix rules that he considers risky (at least that the feeling I get).. for instance if there is a rule that you are required to add declare strict types he will not introduce a fix regarding that. But there is slevomat/coding standard that does a lot of them

**PHP-CS-fixer**

- is all about fixing.. rules are related with fixing stuff. I don't think it reports things that can not be fixed.
- not so handly config (or at least that's what I know) if I want to disable a specific rule per file.. or enable a specific rule per file.

nice coding standards:
- https://github.com/doctrine/coding-standard
- https://github.com/slevomat/coding-standard

and also this for php-cs-fixer
https://mlocati.github.io/php-cs-fixer-configurator/#version:2.16

Taluu:
Meh, not sure CS-Fixer is needed. You can always run `CBF` that is installed with Code-Sniffer.

gmponos:
Well.. in a way I agree with @Taluu  .. many things will be covered by codesniffer if you install slevomat

Yozhef:
On my projects I use this analyzer [PHP Code Sniffer](https://packagist.org/packages/squizlabs/php_codesniffer) - it helps to cover many cases + there is an automatic fix of errors.
Ready to help with integration and take on library style improvements.

pamil:
In Sylius, we use ECS with https://github.com/SyliusLabs/CodingStandard, which has only the autofixable coding standard rules included. Most of the issues that should be handled manually are discovered by PHPStan and Psalm.

ciaranmcnulty:
I have a worry around keeping version history understandable - how can we apply a tool sensibly around areas that are changing?

rpkamp:
I would say start with PSR-12 and work up from there if the need arises (maybe Doctrine CS)?

bitwombat:
[...]

PR #1433, opened 2023-05-18
Uniformize files coding style

helyakin:
Hello everyone

I'm very found of this project and try to convert everyone I know to use it when it's necessary
I've long wanted to do a contribution, and this is a complicated one (to me at list).
But beforehand I've noticed numerous inconsistencies in the coding styles files are using or use of syntax that are kinda old fashioned.

Such as array() which is useful for backward compatibility yes, but do we want to keep using it when using typed parameters in some files ?

Anyway this PR is here to discuss about this topic, and once this one will be settled I'll gladly make another one to uniformise all the typed parameters and attribute in all files.

Here are a few rule I've set using cs-fixer:
no_superfluous_phpdoc_tags: It's kind of pointless to comment the type of a typed parameter
array_syntax: use [] instead of array()
blank_line_before_statement: Some blank lines were missing between attributes
header_comment: Some file are missing the header comment
single_quote: Some file are using double quote when single is enough
visibility_required: Some function visibility were not declared
php_unit_method_casing: Enforce CamelCase case even for tests

Anyway I'll be glad to here from you before spending more time on this !

Cheers

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# Enforce php-cs-fixer new rules
+59dbff8de6a45fa6cbd7fb6c37075c151562b875
--- a/.php-cs-fixer.php
+++ b/.php-cs-fixer.php
@@ -1,0 +1,81 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Behat.
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use PhpCsFixer\Config;
+use PhpCsFixer\Finder;
+
+$header = <<<'OEF'
+This file is part of the Behat.
+(c) Konstantin Kudryashov <ever.zet@gmail.com>
+
+For the full copyright and license information, please view the LICENSE
+file that was distributed with this source code.
+OEF;
+
+$rules = [
+    '@PSR2' => true,
+    '@PSR12' => true,
+    '@PhpCsFixer' => true,
+    'header_comment' => ['header' => $header],
+    'array_syntax' => ['syntax' => 'short'],
+    'class_definition' => ['multi_line_extends_each_single_line' => true],
+	'concat_space' => false,
+    'heredoc_to_nowdoc' => true,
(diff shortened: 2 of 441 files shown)


ciaranmcnulty:
Most of these changes LGTM on their own, my only historical hesitation has been the affect it has on `git blame` - is there a sensible way around that?

If we do make a change like this I'd also want to see enforcement of the rules in future by committing relevant config files and enforcing them in the GitHub workflows

ciaranmcnulty:
Because we're using Psalm already, if some of these rules (mostly type things) can be enforced by that instead / as well it'd be good to add them to the psalm config

ciaranmcnulty:
> my only historical hesitation has been the affect it has on git blame - is there a sensible way around that?
Aha https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view seems to address it

ciaranmcnulty:
If you can get the build passing and add a `.git-blame-ignore-revs` file to root as described above, we can merge this sort of change.

At some point we should/could set up a CS CI job

helyakin:
Hello @ciaranmcnulty 

First of all thank you so much for your attention regarding this PR.

I have pushed:
- [x] The .php-cs-fixer.php with some rules, we can talk furthermore about it, and the CI could be set up using the following command  `php vendor/bin/php-cs-fixer fix --using-cache=no --verbose --diff --dry-run`
- [x] Added a .git-blame-ignore-revs as requested but haven't successfully used it. I don't know what i'm doing wrong
- [x] Fixed the failing test

helyakin:
I also think a tools like phpstan and/or php code style might do some good.
What do you think ?

acoulton:
@helyakin thank you for your work on this. I agree that we should clean up the code style & add tooling to keep that in place in future.

However I'm going to close this because:
* It's been left stale for so long that it obviously now has a lot of merge conflicts with master.
* There are also quite a few other old PRs outstanding, some of which might not make the next release, and I don't want to make it harder to get them up to date.
* Once we have the next point release out, I think we need to review our PHP version support policy, and that might affect what code rules / refactorings we want to apply.

There's an open issue in #1306 on this topic - this PR will be useful and we'd really welcome your help when we're ready to look at that once we've caught up with the backlog a bit.

Issue #1431, opened 2023-04-25
Formatter return value null when Symfony error


that-guy-iain:
Since upgrade to `3.13.0` in a Symfony application when there is an error with the Doctrine ORM it results in a HTML symfony error output with `<!-- Behat\Behat\Output\Printer\Formatter\ConsoleFormatter::format(): Return value must be of type string, null returned (500 Internal Server Error) -->`

Easy way to produce is checkout commit of [`1f935cda935c1aa473fa55b53b6dc6801f1b9c85`](https://github.com/BillaBear/billabear/commit/1f935cda935c1aa473fa55b53b6dc6801f1b9c85) of https://github.com/BillaBear/billabear. The error output from behat can be seen at https://github.com/BillaBear/billabear/actions/runs/4797470877/jobs/8534498033

acoulton:
fixed by #1439

that-guy-iain:
Thanks!

PR #1439, opened 2023-08-22
bug: Fix error when attempting to format long output messages

jonpugh:
I am getting a strange failure when testing some old Drupal code:

> TypeError: Behat\Behat\Output\Printer\Formatter\ConsoleFormatter::format(): Return value must be of type string, null returned in Behat\Behat\Output\Printer\Formatter\ConsoleFormatter->format() (line 33 of /usr/share/devshop/src/DevShop/Control/vendor/behat/behat/src/Behat/Behat/Output/Printer/Formatter/ConsoleFormatter.php).

Looking into it, I discovered format() calls `preg_replace_callback`, which returns null|string|string[].

This change returns an empty string if preg_replace_callback returns null.

Should we check for an array as well?

### DIFF
--- a/src/Behat/Behat/Output/Printer/Formatter/ConsoleFormatter.php
+++ b/src/Behat/Behat/Output/Printer/Formatter/ConsoleFormatter.php
@@ -30,7 +30,7 @@ final class ConsoleFormatter extends BaseOutputFormatter
      */
     public function format($message): string
     {
-        return preg_replace_callback(self::CUSTOM_PATTERN, array($this, 'replaceStyle'), $message);
+        return preg_replace_callback(self::CUSTOM_PATTERN, array($this, 'replaceStyle'), $message) ?? '';
     }
 
     /**
--- a/src/Behat/Behat/Output/Printer/Formatter/ConsoleFormatter.php
+++ b/src/Behat/Behat/Output/Printer/Formatter/ConsoleFormatter.php
@@ -30,7 +30,8 @@ final class ConsoleFormatter extends BaseOutputFormatter
      */
     public function format($message): string
     {
-        return preg_replace_callback(self::CUSTOM_PATTERN, array($this, 'replaceStyle'), $message) ?? '';
+        return preg_replace_callback(self::CUSTOM_PATTERN, array($this, 'replaceStyle'), $message) ??
+            'Error formatting output: ' . preg_last_error_msg();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


stof:
Can you share a reproducer for a case making the regex fail ?

stof:
> Should we check for an array as well?

No. `preg_replace_callback` returns an array when its `subject` argument is an array, which is not the case here.

mdio:
I ran into this when the response body (which was checked) was generated by https://github.com/filp/whoops due to an exception being thrown. Unfortunately I cannot share the content because it's full of internal information.
The HTML was 1 MB large with a 6 level stack trace and roughly a thousand environment variables.
The error for `preg_replace_callback` was `PREG_BACKTRACK_LIMIT_ERROR` with `pcre.backtrack_limit` being `1000000` (the default).

I believe issue #1431 is about the same problem.

Maybe this helps to craft a test case.

Nek-:
Just hit the same issue! And this fix is working fine!

src/Behat/Behat/Output/Printer/Formatter/ConsoleFormatter.php
```diff
@@ -30,7 +30,7 @@ final class ConsoleFormatter extends BaseOutputFormatter
      */
     public function format($message): string
     {
-        return preg_replace_callback(self::CUSTOM_PATTERN, array($this, 'replaceStyle'), $message);
+        return preg_replace_callback(self::CUSTOM_PATTERN, array($this, 'replaceStyle'), $message) ?? '';
```
Nek-, line 33:
```suggestion
        return preg_replace_callback(self::CUSTOM_PATTERN, array($this, 'replaceStyle'), $message) ?? 'Warning: behat formatter cannot handle the output the content is too big to be printed';
```

Maybe this would be a nice addition?

acoulton: I don't think we can guarantee that it's a size issue, but I agree the message would ideally include the reason it failed rather than just an empty string.

Something including `preg_last_error_msg()` would probably do the trick, although we'd need to add symfony/polyfill-php80 until we drop support for the 7.x series

acoulton:
It'd be good to get this updated & merged. Ideally with a test to prove it's working as expected.

carlos-granados:
@acoulton I updated this to the latest master, added an error message if the preg function fails and added a unit test (we can now safely use `preg_last_error_msg` as we are now using PHP8)

acoulton (approved):
Great, thanks

PR #904, opened 2016-07-24
Provide clearer exception message when long token names used

everzet:
TurnipPatternPolicy defines the transformation of turnip notation into a
regular expression using named captures.

PHP forbids usage of capture names longer than 32 symbols. That was
generating cryptic error message. This commit provides clearer
exception.

Closes #872

Issue #872, opened 2016-04-05
Warning: preg_match(): Compilation failed: subpattern name is too long


ladhadha:
There is my .feature file:

```
Feature: Education Calculator
    As a customer
    In order to count how much money to save in a period of time to get my kids the best education

    Scenario: Count how much to save
        Given My child age now is 7 years old
        And My child age when she/he will using the money is 19 years old
        And He/she will use the money for 12 years
        And I need totally 600000000 for his/him education
        And My saving for his/him education is 25000000
        And My monthly saving for his/him education is 0
        And the annual interest rate is 6%
        And the inflation rate per year is 4%
        When I calculate
        Then I should get 960619331 as my totally need (after adjusted with inflation)
        And I should get 40025805 as the Future Value of my current saving
        And I should get 0 as the Future Value of my educational monthly savings
        And I should get 40025805 as my total saving when the education start
        And I should get 920593526 more money to save
        And I should save more 5014346 monthly
        And I should save more 60172155 yearly
```

But when i run `behat` it gave me PHP warning:

```
PHP Warning:  preg_match(): Compilation failed: subpattern name is too long (maximum 32 characters) at offset 56 in /home/baboon/.composer/vendor/behat/behat/src/Behat/Behat/Definition/Search/RepositorySearchEngine.php on line 120
```

Is any one got similar problem? Or whats wrong with my feature file?

stof:
The issue seems to be with one of your step definition, not with your feature file.

Please run Behat in verbose mode (with ``-v``), to get a stack trace for the error, and then paste it here

ladhadha:
There is the console output when i run `behat -v`

```
Feature: Education Calculator
    As a customer
    In order to count how much money to save in a period of time to get my kids the best education

  Scenario: Count how much to save                                                 # htdocs/content/themes/Astralife/tests/features/educalc.feature:5
PHP Warning:  preg_match(): Compilation failed: subpattern name is too long (maximum 32 characters) at offset 56 in /home/baboon/.composer/vendor/behat/behat/src/Behat/Behat/Definition/Search/RepositorySearchEngine.php on line 121

Warning: preg_match(): Compilation failed: subpattern name is too long (maximum 32 characters) at offset 56 in /home/baboon/.composer/vendor/behat/behat/src/Behat/Behat/Definition/Search/RepositorySearchEngine.php on line 121
    Given My child age now is 7 years old                                          # DomainContext::myChildAgeNowIsYearsOld()
      TODO: write pending definition
[...]

PR #1457, opened 2024-01-18
Fix a bug around maximum token length in turnip notation implementation.

ivastly:
PHP allows up to 32 characters (inclusive) for the pattern name. 
The error message from Behat is correct: `Token name should not exceed 32 characters, but {$tokenMatch[1]} was used.`.
However, the code itself is not correct, because it allows up to 31 character, not 32 characters.

This PR provides a fix.

Originally this check was introduced [here](https://github.com/Behat/Behat/pull/904), triggered by [this issue](https://github.com/Behat/Behat/issues/872).

### DIFF
--- a/features/definitions_patterns.feature
+++ b/features/definitions_patterns.feature
@@ -220,7 +220,7 @@ Feature: Step Definition Pattern
       class FeatureContext implements Context
       {
           /**
-           * @Given I provide parameter :too1234567891123456789012345678901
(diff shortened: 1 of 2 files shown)


acoulton (approved):
Thanks for the PR - LGTM :)

carlos-granados:
Thanks @ivastly

carlos-granados (approved):

Issue #1024, opened 2017-04-18
Scenario outline with <parameters> from Examples section in title.


unekhriukova:
Hello, guys.
 I've found a possibility of using the <parameters> from Examples section in the Scenario Outline title in the Cucumber: https://github.com/cucumber/cucumber-js/issues/267
I've tried to use this in behat, but it does not work for me. 
In PhpStorm when I use this parameter in title, it is highlighted and 'go to source' command moves the cursor to the same name column in Examples section. 
When tests are run, I see the parameter is not substituted to its value, just stayed as it is, 
e.g.  

Scenario Outline: Login with &lt;email&gt;.

Seems like this feature is not implemented yet in behat, may be it will be a good idea to do it.

spolischook:
Is it just for title customizing or it also will be a step implementation?

unekhriukova:
As a step implementation this feature is already exist. So just for title customizing.

bwowk:
Well, I can't see how the `Scenario Outline: Login with <email>.` would have it's token replaced. From which Example line would it get its replacement?

For the pretty formatter the ExampleNode titles are not even printed, only the OutlineNode title.

On the other hand, I believe it makes sense in case any other formatter uses the ExampleNode title.

Anyway, this would be done on the Gherkin package (https://github.com/Behat/Gherkin/blob/master/src/Behat/Gherkin/Node/OutlineNode.php#L207), not on Behat core.
You should close this issue and open another one on https://github.com/Behat/Gherkin

bwowk:
`For the pretty formatter the ExampleNode titles are not even printed, only the OutlineNode title.`
Correction: ExampleNode titles are printed, and they're the corresponding line on the examples table

everzet:
@bwowk once again - thanks for nudging in :)

PR #1459, opened 2024-03-03
Display outline values instead of index in junit output

### LINKED ISSUE #1448: [Bug?] Reruns of Scenario outlines lead to potentially confusing scenario names
To explain what is going on, I am going to use a silly example, just to keep it short. Suppose random.html is a pointless page which displays 1 or 2 at random, and we test it like this.

```cucumber
  Scenario Outline: Unreliable test
    When I load "random.html"
    Then I should see "<expected>"
    Examples:
      | expected |
      | 1        |
      | 2        |
```

Suppose that we run this, and get the results:
`Unreliable test #1` -> PASS
`Unreliable test #2` -> FAIL

Suppose we are using the --rerun option to the test-runner. When we do the re-run, it correctly only runs the second example. However, in the result, it names it #1. That is

`Unreliable test #1` -> PASS (this did correctly check for and find 2 in the page).

This makes it difficult to link the result of the re-run to the original failure, which is leading to occasional false-positives from our CI system, which is a minor annoyance.

I am not sure that this is actually a bug, but I thought I woudl raise it.

I don't know much about Behat internals. Issue #154 might be related? If this is not considered to be a Behat probelm we can probably change our CI scripts which handle the test results to deal with our issue. Thank you for considering this.

uuf6429:
Replaces "example name #N" with "example name (XX)" in JUnit output.

Closes #1448

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -394,18 +394,19 @@ EOL;
      */
     public function itShouldFail($success)
     {
+        $message = '';
         if ('fail' === $success) {
             if (0 === $this->getExitCode()) {
-                echo 'Actual output:' . PHP_EOL . PHP_EOL . $this->getOutput();
+                $message = 'Actual output:' . PHP_EOL . PHP_EOL . $this->getOutput();
             }
 
-            Assert::assertNotEquals(0, $this->getExitCode());
+            Assert::assertNotEquals(0, $this->getExitCode(), $message);
         } else {
             if (0 !== $this->getExitCode()) {
-                echo 'Actual output:' . PHP_EOL . PHP_EOL . $this->getOutput();
+                $message = 'Actual output:' . PHP_EOL . PHP_EOL . $this->getOutput();
             }
 
-            Assert::assertEquals(0, $this->getExitCode());
+            Assert::assertEquals(0, $this->getExitCode(), $message);
         }
     }
 
--- a/features/junit_format.feature
+++ b/features/junit_format.feature
@@ -114,15 +114,15 @@
           <testcase name="Failed" classname="World consistency" status="failed" time="-IGNORE-VALUE-" file="features-DIRECTORY-SEPARATOR-World.feature">
             <failure message="Then I must have 13: Failed asserting that 14 matches expected '13'."/>
           </testcase>
-          <testcase name="Passed &amp; Failed #1" classname="World consistency" status="failed" time="-IGNORE-VALUE-" file="features-DIRECTORY-SEPARATOR-World.feature">
+          <testcase name="Passed &amp; Failed (value: 5, result: 16)" classname="World consistency" status="failed" time="-IGNORE-VALUE-" file="features-DIRECTORY-SEPARATOR-World.feature">
             <failure message="Then I must have 16: Failed asserting that 15 matches expected '16'."/>
           </testcase>
-          <testcase name="Passed &amp; Failed #2" classname="World consistency" status="passed" time="-IGNORE-VALUE-" file="features-DIRECTORY-SEPARATOR-World.feature"/>
-          <testcase name="Passed &amp; Failed #3" classname="World consistency" status="failed" time="-IGNORE-VALUE-" file="features-DIRECTORY-SEPARATOR-World.feature">
+          <testcase name="Passed &amp; Failed (value: 10, result: 20)" classname="World consistency" status="passed" time="-IGNORE-VALUE-" file="features\World.feature"></testcase>
+          <testcase name="Passed &amp; Failed (value: 23, result: 32)" classname="World consistency" status="failed" time="-IGNORE-VALUE-" file="features\World.feature">
             <failure message="Then I must have 32: Failed asserting that 33 matches expected '32'."/>
           </testcase>
-          <testcase name="Another Outline #1" classname="World consistency" status="passed" time="-IGNORE-VALUE-" file="features-DIRECTORY-SEPARATOR-World.feature"/>
-          <testcase name="Another Outline #2" classname="World consistency" status="passed" time="-IGNORE-VALUE-" file="features-DIRECTORY-SEPARATOR-World.feature"/>
+          <testcase name="Another Outline (value: 5, result: 15)" classname="World consistency" status="passed" time="-IGNORE-VALUE-" file="features\World.feature"></testcase>
+          <testcase name="Another Outline (value: 10, result: 20)" classname="World consistency" status="passed" time="-IGNORE-VALUE-" file="features\World.feature"></testcase>
         </testsuite>
       </testsuites>
       """
--- a/src/Behat/Behat/Output/Node/Printer/JUnit/JUnitScenarioPrinter.php
+++ b/src/Behat/Behat/Output/Node/Printer/JUnit/JUnitScenarioPrinter.php
@@ -10,12 +10,10 @@
 
 namespace Behat\Behat\Output\Node\Printer\JUnit;
 
-use Behat\Behat\Output\Node\EventListener\JUnit\JUnitOutlineStoreListener;
 use Behat\Behat\Output\Node\EventListener\JUnit\JUnitDurationListener;
 use Behat\Behat\Output\Node\Printer\Helper\ResultToStringConverter;
 use Behat\Gherkin\Node\ExampleNode;
 use Behat\Gherkin\Node\FeatureNode;
-use Behat\Gherkin\Node\OutlineNode;
 use Behat\Gherkin\Node\ScenarioLikeInterface;
 use Behat\Testwork\Output\Formatter;
 use Behat\Testwork\Output\Printer\JUnitOutputPrinter;
@@ -34,44 +32,26 @@ final class JUnitScenarioPrinter
     private $resultConverter;
 
     /**
-     * @var JUnitOutlineStoreListener
-     */
-    private $outlineStoreListener;
-
-    /**
-     * @var OutlineNode
-     */
-    private $lastOutline;
-
-    /**
-     * @var int
-     */
-    private $outlineStepCount;
-
-    /**
      * @var JUnitDurationListener|null
      */
     private $durationListener;
 
-    public function __construct(ResultToStringConverter $resultConverter, JUnitOutlineStoreListener $outlineListener, JUnitDurationListener $durationListener = null)
+    public function __construct(ResultToStringConverter $resultConverter, JUnitDurationListener $durationListener = null)
     {
         $this->resultConverter = $resultConverter;
-        $this->outlineStoreListener = $outlineListener;
         $this->durationListener = $durationListener;
     }
 
     /**
      * {@inheritDoc}
      */
-    public function printOpenTag(Formatter $formatter, FeatureNode $feature, ScenarioLikeInterface $scenario, TestResult $result, string $file = null)
+    public function printOpenTag(Formatter $formatter, FeatureNode $feature, ScenarioLikeInterface $scenario, TestResult $result, string $file = null): void
     {
-        $name = implode(' ', array_map(function ($l) {
-            return trim($l);
-        }, explode("\n", $scenario->getTitle())));
-
-        if ($scenario instanceof ExampleNode) {
-            $name = $this->buildExampleName($scenario);
-        }
+        $name = $this->convertMultipleLinesToOne(
+            $scenario instanceof ExampleNode
+                ? $this->buildExampleName($scenario)
+                : $scenario->getTitle()
+        );
 
         /** @var JUnitOutputPrinter $outputPrinter */
         $outputPrinter = $formatter->getOutputPrinter();
@@ -85,29 +65,42 @@ final class JUnitScenarioPrinter
 
         if ($file) {
             $cwd = realpath(getcwd());
-            $testCaseAttributes['file'] =
-                substr($file, 0, strlen($cwd)) === $cwd ?
-                    ltrim(substr($file, strlen($cwd)), DIRECTORY_SEPARATOR) : $file;
+            $testCaseAttributes['file'] = substr($file, 0, strlen($cwd)) === $cwd
+                ? ltrim(substr($file, strlen($cwd)), DIRECTORY_SEPARATOR)
+                : $file;
         }
 
         $outputPrinter->addTestcase($testCaseAttributes);
     }
 
-    /**
-     * @param ExampleNode $scenario
-     * @return string
-     */
-    private function buildExampleName(ExampleNode $scenario)
+    private function buildExampleName(ExampleNode $scenario): string
     {
-        $currentOutline = $this->outlineStoreListener->getCurrentOutline($scenario);
-        if ($currentOutline === $this->lastOutline) {
-            $this->outlineStepCount++;
-        } else {
-            $this->lastOutline = $currentOutline;
-            $this->outlineStepCount = 1;
-        }
+        return sprintf(
+            '%s (%s)',
+            $scenario->getOutlineTitle(),
+            implode(
+                ', ',
+                array_map(
+                    static function ($key, $val) {
+                        return "$key: $val";
+                    },
+                    array_keys($scenario->getTokens()),
+                    array_values($scenario->getTokens())
+                )
+            )
+        );
+    }
 
-        $name = $currentOutline->getTitle() . ' #' . $this->outlineStepCount;
-        return $name;
+    private function convertMultipleLinesToOne(string $string): string
+    {
+        return implode(
+            ' ',
+            array_map(
+                static function ($line) {
+                    return trim($line);
+                },
+                explode("\n", $string)
+            )
+        );
     }
 }
--- a/src/Behat/Behat/Output/ServiceContainer/Formatter/JUnitFormatterFactory.php
+++ b/src/Behat/Behat/Output/ServiceContainer/Formatter/JUnitFormatterFactory.php
@@ -79,7 +79,6 @@ final class JUnitFormatterFactory implements FormatterFactory
 
         $definition = new Definition('Behat\Behat\Output\Node\Printer\JUnit\JUnitScenarioPrinter', array(
             new Reference(self::RESULT_TO_STRING_CONVERTER_ID),
-            new Reference('output.node.listener.junit.outline'),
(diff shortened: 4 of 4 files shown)


features/bootstrap/FeatureContext.php
```diff
@@ -394,18 +394,19 @@ private function getExpectedOutput(PyStringNode $expectedText)
...
-            Assert::assertNotEquals(0, $this->getExitCode());
+            Assert::assertNotEquals(0, $this->getExitCode(), $message);
         } else {
             if (0 !== $this->getExitCode()) {
-                echo 'Actual output:' . PHP_EOL . PHP_EOL . $this->getOutput();
+                $message = 'Actual output:' . PHP_EOL . PHP_EOL . $this->getOutput();
             }
 
-            Assert::assertEquals(0, $this->getExitCode());
+            Assert::assertEquals(0, $this->getExitCode(), $message);
```
uuf6429, line 409: Not related to this PR, but when tests failed, the output was hidden since echo/output is not the right way to influence PHPUnit test results (the `$message` assertion parameter should be used instead).

features/junit_format.feature
```diff
@@ -114,15 +114,15 @@
...
+          <testcase name="Passed &amp; Failed (value: 10, result: 20)" classname="World consistency" status="passed" time="-IGNORE-VALUE-" file="features\World.feature"></testcase>
+          <testcase name="Passed &amp; Failed (value: 23, result: 32)" classname="World consistency" status="failed" time="-IGNORE-VALUE-" file="features\World.feature">
             <failure message="Then I must have 32: Failed asserting that 33 matches expected '32'."/>
           </testcase>
-          <testcase name="Another Outline #1" classname="World consistency" status="passed" time="-IGNORE-VALUE-" file="features-DIRECTORY-SEPARATOR-World.feature"/>
-          <testcase name="Another Outline #2" classname="World consistency" status="passed" time="-IGNORE-VALUE-" file="features-DIRECTORY-SEPARATOR-World.feature"/>
+          <testcase name="Another Outline (value: 5, result: 15)" classname="World consistency" status="passed" time="-IGNORE-VALUE-" file="features\World.feature"></testcase>
+          <testcase name="Another Outline (value: 10, result: 20)" classname="World consistency" status="passed" time="-IGNORE-VALUE-" file="features\World.feature"></testcase>
```
uuf6429, line 125: While the tests run successfully with this change on my (Windows) machine, `DIRECTORY-SEPARATOR` and the end tag seem suspicious. 🤔

src/Behat/Behat/Output/Node/Printer/JUnit/JUnitScenarioPrinter.php
```diff
@@ -85,29 +65,42 @@ public function printOpenTag(Formatter $formatter, FeatureNode $feature, Scenari
...
 
         $outputPrinter->addTestcase($testCaseAttributes);
     }
 
-    /**
-     * @param ExampleNode $scenario
-     * @return string
-     */
-    private function buildExampleName(ExampleNode $scenario)
+    private function buildExampleName(ExampleNode $scenario): string
```
uuf6429, line 76: I'm wondering if it would make sense to have this function inside `ExampleNode` - it might be useful elsewhere.

I have implemented something similar for a custom report at work, for example.

acoulton (changes requested):
Thanks @uuf6429, I've got a few comments plus there are some merge conflicts with master to resolve.

features/bootstrap/FeatureContext.php
uuf6429: Done: https://github.com/Behat/Behat/pull/1487

src/Behat/Behat/Output/Node/Printer/JUnit/JUnitScenarioPrinter.php
```diff
@@ -33,45 +31,27 @@ final class JUnitScenarioPrinter
...
-     */
-    private $outlineStepCount;
-
     /**
      * @var JUnitDurationListener|null
      */
     private $durationListener;
 
-    public function __construct(ResultToStringConverter $resultConverter, JUnitOutlineStoreListener $outlineListener, JUnitDurationListener $durationListener = null)
+    public function __construct(ResultToStringConverter $resultConverter, JUnitDurationListener $durationListener = null)
```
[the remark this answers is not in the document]
uuf6429: I share some concern about BC. This could also go in hand with [your other comment](https://github.com/Behat/Behat/pull/1459#discussion_r1811575754).
Could it make sense that we release this entire feature in a major version instead?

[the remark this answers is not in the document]
acoulton: I think we'll need to do a major at some point, but I'm not convinced we should do it now/soon. There's no point releasing a major if the version constraints across the ecosystem will prevent people from upgrading. With the project having been quiet for a while, I think we need to prioritise getting through the backlog on the 3.x series and getting people used to the idea there will be regular releases before we think about what needs a 4.x bump.

That's obviously a different discussion to have elsewhere.

I think there's 4 possibilities:

* Pin this for a major, but it could be waiting a long time
* Release it as-is (apart from fixing the BC break on the constructor) and assume/hope the impact on any reports will be minimal
* Rework it to take a formatter option (as an optional final constructor arg) so it is entirely BC unless people can opt in to the new naming (which could become default in the next major)
* Rework it to a simpler solution for the specific bug, so that we keep the `Some scenario #1` naming but we just ensure the numbering is tied to the feature file sequence, not the test execution. I don't know how feasible that would be though.

My preference is probably to release it with an opt-in formatter option if that can be done without too much work. Or if not then to release it as-is.

What do you / @carlos-granados think?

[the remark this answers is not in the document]
carlos-granados: I think that releasing it as is (removing the BC break in the constructor) should be good enough, I think the impact should be minimal

carlos-granados:
@uuf6429 @acoulton what's needed to get this over the line?

src/Behat/Behat/Output/Node/Printer/JUnit/JUnitScenarioPrinter.php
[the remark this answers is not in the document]
uuf6429: Hang on, I just realised this class has been defined final, so no one can extend it and change the constructor signature.

The only problem is if someone does initialize the class with the wrong(old) set of arguments, so here's an idea that preserves BC and shows a deprecation message:
```php
    /**
     * @param null|JUnitDurationListener|JUnitOutlineStoreListener $durationListener
     */
    public function __construct(ResultToStringConverter $resultConverter, $durationListener = null)
    {
        if ($durationListener instanceof JUnitOutlineStoreListener ) {
            @trigger_error('2nd argument is deprecated blabla', E_USER_DEPRECATED);
            $durationListener = func_get_args()[2] ?? null;
        }

        $this->resultConverter = $resultConverter;
        $this->durationListener = $durationListener;
    }
```
wdyt?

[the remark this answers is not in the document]
acoulton: Yes, it's specifically the concern about external construction (even if it wasn't a final class, the `__construct` method does not enforce signature comparability between children and parents).

Out of the box, behat's error reporting is configured to fail on deprecations so triggering a deprecation is also a BC break IMO.

I would just keep accepting the parameter but leave it unused with a marker to review for the next major. We will probably want to think about how to do deprecations ahead of that anyway (maybe using one of the userland deprecation packages rather than native php errors) and I don't think we should hold this up for that.

features/junit_format.feature
```diff
@@ -114,15 +114,15 @@
...
+          <testcase name="Passed &amp; Failed (value: 23, result: 32)" classname="World consistency" status="failed" time="-IGNORE-VALUE-" file="features-DIRECTORY-SEPARATOR-World.feature">
             <failure message="Then I must have 32: Failed asserting that 33 matches expected '32'."/>
           </testcase>
-          <testcase name="Another Outline #1" classname="World consistency" status="passed" time="-IGNORE-VALUE-" file="features-DIRECTORY-SEPARATOR-World.feature"/>
-          <testcase name="Another Outline #2" classname="World consistency" status="passed" time="-IGNORE-VALUE-" file="features-DIRECTORY-SEPARATOR-World.feature"/>
+          <testcase name="Another Outline (value: 5, result: 15)" classname="World consistency" status="passed" time="-IGNORE-VALUE-" file="features-DIRECTORY-SEPARATOR-World.feature"/>
+          <testcase name="Another Outline (value: 10, result: 20)" classname="World consistency" status="passed" time="-IGNORE-VALUE-" file="features-DIRECTORY-SEPARATOR-World.feature"/>
```
uuf6429, line 125: I've come to realise something else. At work we create custom reports that reuse some of this functionality (title, example values etc).

We realised that the nicest solution (in our opinion) is to replace placeholders in the title - instead of add `#<number>` or in this case ` (<example values>)`.

Consider this feature:
```gherkin
feature:
    Scenario Outline: <count> items should <passOrFail>
        When I have <count> items
        Then I should <passOrFail>
    Examples:
        | count  | passOrFail |
        | 0      | pass       |
        | 1      | pass       |
        | 2      | fail       |
```
- Currently, the result will be:
  ```text
      <count> items should <passOrFail> #1
      <count> items should <passOrFail> #2
      <count> items should <passOrFail> #3
  ```
  The number is not particularly useful.
- With this PR, it would look like:
  ```text
      <count> items should <passOrFail> (count: 0, passOrFail: pass)
      <count> items should <passOrFail> (count: 1, passOrFail: pass)
      <count> items should <passOrFail> (count: 2, passOrFail: fail)
  ```
  This is more useful, but not very readable and easily/unavoidably verbose
- With my latest proposal:
  ```text
      0 items should pass
      1 items should pass
      2 items should fail
  ```
  Couldn't get any better. :)

I think the 3rd would be the best overall approach, but I've a feeling this is not supported in gherkin though (although I guess it's not contradicting it either). What do you think @stof?

src/Behat/Behat/Output/Node/Printer/JUnit/JUnitScenarioPrinter.php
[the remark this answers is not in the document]
uuf6429: @acoulton fine by me - I was only wondering what's the best way to do this deprecation or marking, as you named it.

I don't know if we have anything like it, one way could be to make a new issue saying that parameter is now unused and can be removed and tagging it with a milestone.

Just let me know exactly what's the best way forward, I'm fully fine with reverting this part.

features/junit_format.feature
acoulton: I like the idea of the third approach (it's similar to how e.g. `it.each(` works in jest / vitest). 

Off the top of my head I don't know how feasible it is with gherkin syntax - although presumably could find a placeholder string that didn't contain anything gherkin considers to be special characters.

The other question would be what you do if the title *doesn't* contain placeholders - would you try to detect that and append the values as in this PR, or would you stick with the current `#0` etc suffix, or something else?

src/Behat/Behat/Output/Node/Printer/JUnit/JUnitScenarioPrinter.php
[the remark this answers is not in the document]
acoulton: An issue is a good idea but I think will be too noisy if we have one for each thing. I've added a master issue at #1519 to track small refactorings / removals and added this.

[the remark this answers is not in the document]
acoulton: So then could just have the constructor as something like:
```
public function __construct(ResultToStringConverter $resultConverter, JUnitOutlineStoreListener $outlineListener, JUnitDurationListener $durationListener = null)
{
    // The unused $outlineListener parameter has been kept temporarily for BC and will be removed in the next major.
    // See https://github.com/Behat/Behat/issues/1519

File: features/junit_format.feature (hunk 5 above)
Comment (uuf6429) [reply features/junit_format.feature:125]: > The other question would be what you do if the title doesn't contain placeholders - would you try to detect that and append the values as in this PR, or would you stick with the current #0 etc suffix, or something else?

I was hoping I wouldn't have to answer that. 😆

When I think about it, it might make sense to add the numbering when the title already exists - it would be backward compatible and somewhat unsurprising.

```php
    private $outlineTitles = [];

    private function buildExampleName(ExampleNode $scenario): string
    {
        $currentOutline = $this->outlineStoreListener->getCurrentOutline($scenario);

        $title = $this->replacePlaceholders($scenario->getOutlineTitle(), $scenario->getTokens());

        $number = $this->outlineTitles[$title] ?? 0;

        $this->outlineTitles[$title] = $number + 1;

        if ($number > 0) {
            $title .= " #$number";
        }

        if ($currentOutline === $this->lastOutline) {
            $this->outlineTitles = [];
        }

        return $title;
    }
```

acoulton: That could work quite well, so it becomes a definitely-BC update path:

* If you do nothing, you'll get the same test names as before (including that the numbers are inconsistent between initial and re-run?)
* But to get consistent - and clearer - naming, you just have to reword your scenario outlines to include placeholders

That's a really nice idea if you can find a syntax for it that works in gherkin. Using `<title>` would seem logical as it follows the pattern of how the parameters are referenced in the steps within the outline. But might be risky (now or in future) because of that.

I don't *think* `{}` has any special meaning in gherkin, so that might be a possibility...

uuf6429: > so it becomes a definitely-BC update path

Exactly.

> But might be risky (now or in future) because of that.

If I didn't misunderstand, it is supported/implemented in cucumber already: https://github.com/cucumber/common/issues/2149 and https://github.com/cucumber/cucumber-js/issues/267. The request here seems to lend credence to that too: https://youtrack.jetbrains.com/issue/IDEA-320963/Cucumber-example-variable-usage-in-scenario-outline-title

On the whole, I think I'd go for the `<...>` syntax since it seems to also be what people there were expecting.

I see mixed signals in Behat though: https://github.com/Behat/Behat/issues/1024

acoulton: Oh perfect, if cucumber is already using the tags like that for this purpose then that's exactly what we need. There's also an example of it in [cucumber/gherkin](https://raw.githubusercontent.com/cucumber/gherkin/refs/heads/main/testdata/good/example_tokens_everywhere.feature).

If I've understood right, most of the mixed signals in #1024 look to me like they're specific to the pretty output format (where I can see there's no logical place to do the substitution in the current output, because it only prints the Scenario Outline block once and just renders the results into and below the Examples: table).

The suggestion this should be implemented in Gherkin is probably right, but:
* I'm not sure why [this comment](https://github.com/Behat/Behat/issues/1024#issuecomment-307594867) is linking to `OutlineNode::getKeyword` as I'm pretty sure that will (correctly) always just return `Scenario Outline`.
* It can't change the return of `OutlineNode::getTitle()` because that needs to be the title that represents all the examples  e.g. without replacing any placeholders, so that e.g. the pretty formatter can use it as above.

I actually think the logical place to do this is inside ExampleNode. That is the class that's responsible for [replacing placeholders in the step definitions](https://github.com/Behat/Gherkin/blob/master/src/Behat/Gherkin/Node/ExampleNode.php#L186), so replacing placeholders in the title there as well feels sensible.

ExampleNode already has a [`getOutlineTitle`](https://github.com/Behat/Gherkin/blob/master/src/Behat/Gherkin/Node/ExampleNode.php#L176) which currently just returns the parent title as-is. I can't find any usages of that method in behat - it could of course be used by other consumers. I don't know if we'd consider changing the string content a BC break - the safest option of course would be to add a new `getOutlineTitleWithPlaceholders` (with a better name :laughing:) and perhaps deprecate the existing `getOutlineTitle`.

The other advantage of doing it in ExampleNode is that when it's created by OutlineNode we presumably know which index it is in the gherkin file - I think filtering what will actually run comes later than parsing. So we could pass that into ExampleNode as well and use that to append a consistent number to the end of the title which would also then solve the JUnit bug even if people don't put placeholders into their titles.

WDYT?

uuf6429: I also feel the best place to put it would be in `ExampleNode`. I'm a bit hesitant with the semantics though.
- `getTitle` - this feels like it should return the definitive title, e.g. to be shown in reports and so on. _In the ideal world,_ I think this should be the one replacing placeholders and adding numbers for deduplication.
- `getOutlineTitle` - Internally (e.g. even in gherkin), it seems that there shouldn't be a concept of a scenario outline - instead they should be flattened into regular scenarios. With that in mind, I think the "outline title" should be the original (parent?) title - since that parent is the "scenario outline".
- `getOutlineTitleWithPlaceholders` (& co.) - I would go for such a method if we want to be absolutely BC safe. That said, I do feel that the majority of end users would rather have the BC break than the current approach, but that's my hypothesis.

PS: with regards to `getTitle`, if it did behave like that, #1470 might have been avoided. I'd appreciate @cavo789's thoughts on this as well.

acoulton: The challenge is `getTitle` is the title of the "example" and is currently the content of the table row for the example e.g. `1 | 5 | 4` and existing formatters rely on it

Regarding collapsing scenario and scenario outline - that could also be valid : gherkin now treats them as synonyms. But again existing formatters display them differently.

Maybe we should move this to an issue in gherkin to keep the discussion visible & clear?

src/Behat/Behat/Output/Node/Printer/JUnit/JUnitScenarioPrinter.php
uuf6429: Relates to the title discussion, so this can be closed.

src/Behat/Behat/Output/Node/Printer/JUnit/JUnitScenarioPrinter.php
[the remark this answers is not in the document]
uuf6429: Done: https://github.com/Behat/Behat/pull/1459/commits/576d5eef5a98a9bf0cf87d11ab3c5088e9f92170

features/junit_format.feature
uuf6429: I've updated the PR to fix conflicts and apply the last review suggestions, so that is can be "frozen" while a discussion can take place in behat/gherkin.

PR #1460, opened 2024-03-03
feature: Improve JUnit test duration precision

### LINKED ISSUE #1443: [bug] jUnit formatter returns time as int but should use float
Currently all my tests last `time="0"`

uuf6429:
Closes #1443.

CC @spolischook

### DIFF
--- a/src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitDurationListener.php
+++ b/src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitDurationListener.php
@@ -78,6 +78,7 @@ final class JUnitDurationListener implements EventListener
         if ($timer instanceof Timer) {
             $timer->stop();
             $this->resultStore[$key] = $timer->getTime();
+            unset($this->scenarioTimerStore[$key]);
         }
     }
 
@@ -92,6 +93,7 @@ final class JUnitDurationListener implements EventListener
         if ($timer instanceof Timer) {
             $timer->stop();
             $this->featureResultStore[$key] = $timer->getTime();
+            unset($this->featureTimerStore[$key]);
         }
     }
 
--- a/src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitDurationListener.php
+++ b/src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitDurationListener.php
@@ -15,13 +15,17 @@ use Behat\Testwork\Output\Node\EventListener\EventListener;
 
 final class JUnitDurationListener implements EventListener
 {
+    /** @var array<string, Timer> */
     private $scenarioTimerStore = array();
+    /** @var array<string, Timer> */
     private $featureTimerStore = array();
+    /** @var array<string, float> */
     private $resultStore = array();
+    /** @var array<string, float> */
     private $featureResultStore = array();
 
     /** @inheritdoc */
-    public function listenEvent(Formatter $formatter, Event $event, $eventName)
+    public function listenEvent(Formatter $formatter, Event $event, $eventName): void
     {
         $this->captureBeforeScenarioEvent($event);
         $this->captureBeforeFeatureTested($event);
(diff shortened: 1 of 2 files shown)


src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitDurationListener.php
```diff
@@ -83,17 +91,16 @@ private function captureAfterFeatureEvent(Event $event)
         $timer = $this->featureTimerStore[$key];
         if ($timer instanceof Timer) {
             $timer->stop();
-            $this->featureResultStore[$key] = round($timer->getTime());
+            $this->featureResultStore[$key] = $timer->getTime();
```
uuf6429, line 94: I'm wondering, why do we keep a reference to the timer (after finishing) AND store the timer result in a separate property/array?

- If we want to free up memory, we should `unset($this->featureTimerStore[$key]);` after `$this->featureResultStore[$key] = $timer->getTime();`
- if we don't care about memory, we should remove `featureResultStore` entirely and just call `$timer->getTime()` when needed

CC @higrow

acoulton: @uuf6429 I agree, I can't see a reason to keep both. I doubt the timers take a lot of memory, but we might as well unset them (your first suggestion).

acoulton (changes requested):

src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitDurationListener.php
uuf6429: I think the 2nd approach would have been more idiomatic.

Otherwise, the first approach would be greatly simplified with direct `microtime(true)` calls.

But just my .2 cents.

(I implemented the first suggestion, so if everyone's fine with that, we can proceed forward)

acoulton (changes requested):
My only thought is there's no test coverage. If I were writing this from scratch I'd inject a fake clock to allow asserting actual values in unit tests at least. But that's probably a separate enhancement to consider some time rather than part of this fix.

However can we at least update the regex at https://github.com/Behat/Behat/blob/58681da9fda01f857a8b4bc90ee0c5848df2a842/features/bootstrap/FeatureContext.php#L325 to be stricter about expected time format?

I think the [scenarios that test junit](https://github.com/Behat/Behat/blob/58681da9fda01f857a8b4bc90ee0c5848df2a842/features/junit_format.feature) should all be quick to run, we should allow for occasional CI slowness but even so something like  `/^[012]\.\d{3}$/` would allow up to 2.999 seconds and give us a lot more confidence than we have now?

acoulton:
I think just updating the test?

uuf6429:
> I think just updating the test regex?

Sorry, I only noticed the original comment about regex now. 

I'm wondering why you want to have such a specific regex? I would go for any valid numeric values, no?

E.g. `/^\d+(?:\.\d+)?$/`

PS: apparently PHP prints non-fraction floating numbers and integers, which is why I made that optional in the regex.

acoulton:
It would just give confidence that the number is in the right units (seconds, not eg milliseconds) and formatted correctly e.g. ignoring any localisation etc.

You're right that print renders floats as ints sometimes but you're using number_format to 3dp so there are (should) always guaranteed to be 3 decimals https://3v4l.org/242AQ

uuf6429:
> It would just give confidence that the number is in the right units (seconds, not eg milliseconds) and formatted correctly e.g. ignoring any localisation etc.

I think it's a good idea to check that, but I would rather have a specialized test (e.g. with fake timer, as you suggested), otherwise that regex risks confusing bugs (abnormally slow test fails) and developers (wtf-factor when seeing the regex).

> You're right that print renders floats as ints sometimes but you're using number_format to 3dp so there are (should) always guaranteed to be 3 decimals https://3v4l.org/242AQ

I swear I didn't see the `number_format` part in the code. 😄

acoulton:
Ok, compromise solution `^\d\.\d{3}$`? Even on an extremely slow system if that test takes 9.999 seconds there are bigger problems and it's specific enough to cover the format and order of magnitude

carlos-granados (approved):

Issue #1456, opened 2024-01-12
phar is broken for the 3.14.0 release


jakzal:
## Steps to reproduce

```bash
wget https://github.com/Behat/Behat/releases/download/v3.14.0/behat.phar
php behat.phar --version
```

## Expected result 

Outputs the version

```
behat 3.14.0
```

## Actual result

Fatal error:

```
Fatal error: Uncaught PharException: manifest cannot be larger than 100 MB in phar "/project/behat.phar" in /project/behat.phar:9
Stack trace:
#0 /project/behat.phar(9): Phar::mapPhar('default.phar')
#1 {main}
  thrown in /project/behat.phar on line 9
```

![Reproduction steps](https://github.com/Behat/Behat/assets/190447/b0ab1f0c-86e1-4fef-af89-01271a3ee068)

## Environment

PHP 8.3 from the [official Docker image](https://hub.docker.com/_/php).

jakzal:
Looking at the recent changes, the only relevant change that could break the phar during the upload is #1430.

theofidry:
I'll check this ASAP, but I would highly recommend Behat to do a smoke test of the PHAR at least, e.g. calling `behat.phar --version`

theofidry:
Looking at [this](https://github.com/stof/Behat/blob/7470472ccbb346928a248cf86fa1222cf28f3460/.github/workflows/build.yml#L104C61-L104C65) it's not really worth for me to try to investigate this any further.

I'm happy to help if there is issues with upgrading, but I'm not willing to support a version I had no hands in and that is [7 years old](https://github.com/box-project/box2/releases/2.7.5).

acoulton:
The workflow is now fixed (by upgrading to latest box) by @heiglandreas in #1462 - there will be a valid phar for the next release, coming soon.

PR #1462, opened 2024-03-11
Update PHAR creation to more up-to-date version

heiglandreas:
This PR updates the PHAR creation process to use a rather current version of box.

It moves 
a) from the old box2 to the current box 
b) from downloading and preparing the build-tools ourselves to using the build-tools provided by shivamatur
c) towards using more internal logic from box to determine which files are to be included into the PHAR

The config file has been cleaned up and the github-action files have been updated accordingly.

Note: This just replaces the old process with an updated one. The PHAR file is still unscoped which might cause problems when the application executes code that calls classes that are available in the PHAR as well as in the tested codes dependencies as the ones from the PHAR will be used. But that is a separate issue from the update

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -114,6 +114,11 @@ jobs:
                     composer install --no-dev -o &&
                     box build
 
+            -   name: Test the PHAR
+                if: matrix.publish-phar == true
+                run: |
+                    behat.phar --version
+
             -   uses: actions/upload-artifact@v3
                 name: Publish the PHAR
                 if: matrix.publish-phar == true
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -75,6 +75,7 @@ jobs:
                     php-version: "${{ matrix.php }}"
                     ini-values: "phar.readonly=0,zend.exception_ignore_args=Off"
                     coverage: none
+                    tools: box
 
             -   name: Install symfony/flex
                 if: matrix.symfony-version != ''
@@ -110,11 +111,8 @@ jobs:
             -   name: Build the PHAR
                 if: matrix.publish-phar == true
                 run: |
-                    curl -LSs https://box-project.github.io/box2/installer.php | php &&
-                    export PATH=.:$PATH &&
-                    rm -Rf ./vendor &&
                     composer install --no-dev -o &&
-                    box.phar build
+                    box build
 
             -   uses: actions/upload-artifact@v3
                 name: Publish the PHAR
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,8 +23,8 @@ jobs:
                 composer-mode: [update]
                 symfony-version: ['']
                 include:
-                    # Get the existing 7.2 to publish the phar
-                    -   php: 7.2
+                    # Get the latest stable PHP to publish the phar
+                    -   php: 8.3
                         os: ubuntu-latest
                         composer-mode: update
                         publish-phar: true
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -121,7 +121,6 @@ jobs:
               run: ./vendor/bin/psalm --output-format=github
 
     build-phar:
-        needs: tests
         runs-on: ubuntu-latest
         name: Build PHAR file
         steps:
@@ -184,7 +183,7 @@ jobs:
     publish-phar:
         runs-on: ubuntu-latest
         name: Publish the PHAR for release
-        needs: test-phar
+        needs: [test-phar, tests]
         if: github.event_name == 'release'
         steps:
             -   uses: actions/download-artifact@v4
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -180,8 +180,6 @@ jobs:
         - name: Test the PHAR
           run: |
             php ./behat.phar --version
-            php ./behat.phar
-
 
     publish-phar:
         runs-on: ubuntu-latest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,7 @@ jobs:
         defaults:
             run:
                 shell: bash
-        name: Build and test
+        name: Test
         strategy:
             fail-fast: false
             matrix:
@@ -75,7 +75,6 @@ jobs:
                     php-version: "${{ matrix.php }}"
                     ini-values: "phar.readonly=0,zend.exception_ignore_args=Off"
                     coverage: none
-                    tools: box
 
             -   name: Install symfony/flex
                 if: matrix.symfony-version != ''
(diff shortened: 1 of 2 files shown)


box.json
```diff
@@ -1,19 +1,10 @@
...
         "i18n.php"
     ],
-    "finder": [
-        {
-            "name": ["*.php", "*.xsd", "LICENSE"],
-            "exclude": ["Tests", "tests", "sebastian", "phpunit", "phpspec", "process", "filesystem"],
-            "in": "vendor"
-        }
+    "compactors": [
+        "KevinGH\\Box\\Compactor\\Php"
```
ciaranmcnulty, line 7: Is there a TL;DR somewhere of what difference this makes to the other compactor? Or are they both just minifying classes and we should just use it?

ciaranmcnulty (commented):
There's one failing build, looks like a dependencies issue?

The actual changes look fine to me if that's possible to resolve.

How can we test it works, aside from manually? It'd be good to test the PHAR as well, similar to [how phpspec does it](https://github.com/phpspec/phpspec/blob/main/behat.yml.dist#L13-L19) but that could be a future PR if this one's been manually verified.

stof:
Scoping the phar would be really hard because the API exposed to Behat extensions is not isolated from the symfony/dependency-injection API.

box.json
heiglandreas: They are both just minifying the PHP-Code (stripping code-comments basically). And both are part of the respective package. As this now uses `box` and no longer `box2` the classes are named differently but in essence still do the same.

.github/workflows/build.yml
```diff
@@ -11,6 +11,7 @@ on:
 jobs:
     tests:
         runs-on: ${{ matrix.os }}
+        continue-on-error: ${{ matrix.experimental }}
```
stof, line 14: this is not the right approach for making experimental builds: this would continue the job even if `setup-php` fails for instance.

stof:
@heiglandreas this is used as an easy way of ensuring that the dependencies inside the phar are resolved as compatible with PHP 7.2.

But this could be achieved in a different way by using `config.platform` to force a PHP version in that job (but it would require the building of the phar to be a different job than the one running tests)

.github/workflows/build.yml
heiglandreas: According to the [GithubAction docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) this setting allows a job to "pass" even though one step failed. The job will still be marked as failed but the whole workflow will not fgail due to that one failed step.

If you have a different way of achieving this I am all ears!

heiglandreas:
> But this could be achieved in a different way by using config.platform to force a PHP version in that job (but it would require the building of the phar to be a different job than the one running tests)

Which might make sense anyhow.

.github/workflows/build.yml
stof: oh, it looks like they have separated the `continue-on-error` at the job level and at the step level now.

.github/workflows/build.yml
```diff
@@ -23,8 +23,8 @@ jobs:
                 composer-mode: [update]
                 symfony-version: ['']
                 include:
-                    # Get the existing 7.2 to publish the phar
-                    -   php: 7.2
+                    # Get the latest stable PHP to publish the phar
+                    -   php: 8.3
                         os: ubuntu-latest
                         composer-mode: update
                         publish-phar: true
```
acoulton, line 30: If I've understood correctly, I think this block might all now be redundant as the publish-phar no longer does anything in this job?

.github/workflows/build.yml
```diff
@@ -143,10 +127,42 @@ jobs:
...
+              uses: shivammathur/setup-php@v2
+              with:
+                  php-version: "8.3"
+                  ini-values: "phar.readonly=0,zend.exception_ignore_args=Off"
+                  coverage: none
+                  tools: box
+
+            - name: Build the PHAR
+              run: |
+                  composer config platform.php ^7.2.34
```
acoulton, line 147: This is failing in github actions as an invalid version - I think it needs to be just `7.2.34`  as you're faking the running PHP version rather than setting a requirement?

.github/workflows/build.yml
```diff
@@ -143,10 +127,42 @@ jobs:
...
+
+            - name: Build the PHAR
+              run: |
+                  composer config platform.php ^7.2.34 
+                  composer update --no-dev -o
+                  box build
+
+            - name: Test the PHAR
+              run: |
+                  behat.phar --version
```
acoulton, line 153: Two things:

* As it stands, this is still running in PHP 8.3 - should we run a second `setup-php` step to downgrade back to 7.2 to make sure the phar actually runs on the lowest supported version? 
* I *think* it would be straightforward here to actually run our behat suite (or perhaps tag a few features as `@phar-smoketest` or something) rather than just printing the version, which would give a lot more confidence the phar was working? It might be worth removing the `./vendor` directory first to be sure it's not taking anything from composer?

.github/workflows/build.yml
```diff
@@ -143,10 +127,42 @@ jobs:
...
+
+            - uses: actions/upload-artifact@v3
```
acoulton, line 155: Need to bump to upload-artifact@v4

acoulton (changes requested):
@heiglandreas thanks very much for all your work on this so far. I've merged up to the latest master & resolved a couple of github actions conflicts. I've made a couple of suggestions that I think would get this over the line.

acoulton:
This fixes #1456 and as far as I can see is probably the only blocker for getting the next point release out.

.github/workflows/build.yml
heiglandreas: Yes and Yes 😁

.github/workflows/build.yml
```diff
@@ -143,10 +126,68 @@ jobs:
...
+                  tools: box
+
+            - name: Build the PHAR
+              run: |
+                  composer config platform.php 7.2.34
+                  composer update --no-dev -o
+                  box compile
+
+            - name: cache artifact
+              uses: actions/upload-artifact@master
```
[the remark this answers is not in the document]
heiglandreas: Thanks! I forgot to replace **that** one with the actual version

.github/workflows/build.yml
```diff
@@ -143,10 +120,71 @@ jobs:
             - name: Run Psalm
               run: ./vendor/bin/psalm --output-format=github
 
+    build-phar:
+        needs: tests
```
acoulton, line 124:
```suggestion
```

acoulton (approved):
@heiglandreas thanks for this - it's looking good to me apart from a minor suggestion that we should just build the phar straight away, to reduce the overall time waiting for CI (which I can apply if everyone's happy).

Thanks for checking if we can run the actual behat, I agree since that's not working "out of the box" we should look at that separately. What you've got here already gives us a lot more confidence than we had before.

Really appreciate your work on this.

.github/workflows/build.yml
heiglandreas: I have no issue at all with that change.

heiglandreas: I though made the publish-stage dependend on the tests as well as on the phar-tests so that no untested phar will be published.

carlos-granados (approved):
Thanks @heiglandreas

theofidry:
👏

PR #1463, opened 2024-04-15
Adds void return type to generated snippets

carlos-granados:
Add `void` return type to the snippets that are generated when a step definition is missing. Even though this return type does not offer any runtime advantage and is of limited use for static analysis, it provides completeness if, for example, we are calculating the type coverage of our application

### DIFF
--- a/features/append_snippets.feature
+++ b/features/append_snippets.feature
@@ -185,7 +185,7 @@ Feature: Append snippets option
           /**
            * @Then /^do something undefined with \$$/
            */
-          public function doSomethingUndefinedWith2()
+          public function doSomethingUndefinedWith2(): void
           {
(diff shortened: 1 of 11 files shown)


pierresh:
It is a good idea, PHPStan always reminds me that new methods have no return type specified.

acoulton (approved):
LGTM

acoulton:
@carlos-granados I see this still needs a second review - are you able to review your own PR?

carlos-granados:
@acoulton no, I am unable to approve a PR that I created myself. @everzet how do we solve this problem? It is going to be an issue if @acoulton and myself start working on pending issues...

stof (approved):

PR #1464, opened 2024-05-05
Support behat.dist.y[a]ml format

uuf6429:
Many common tools (phpunit, phpstan, php-cs-fixer) use config files like `<toolname>.<ext>`. To ease collaboration, they and Behat support checking for and using `<toolname>.<ext>.dist`.

However, those tools also support the `<toolname>.dist.<ext>`, which is arguably superior when it comes to extension conventions (e.g. IDEs need to be taught that `.<ext>.dist` is really of type `.<ext>`). This PR adds support for that format to Behat too.

----

On a sidenote: none of those alternatives are mentioned in any Behat documentation - it seems to be limited to [`behat.yml`](https://behat.org/en/latest/user_guide/configuration.html#behat-yml).

### DIFF
--- a/src/Behat/Behat/ApplicationFactory.php
+++ b/src/Behat/Behat/ApplicationFactory.php
@@ -117,10 +117,14 @@ final class ApplicationFactory extends BaseFactory
             $cwd . 'behat.yml',
             $cwd . 'behat.yaml.dist',
             $cwd . 'behat.yml.dist',
+            $cwd . 'behat.dist.yaml',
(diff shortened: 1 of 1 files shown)


acoulton (approved):
@uuf6429 thanks - LGTM

carlos-granados (approved):
Thanks @uuf6429

stof:
don't hesitate to send a PR on https://github.com/Behat/docs to mention the alternate names

PR #1466, opened 2024-05-06
Feature/option rerun only

Treast:
This PR adds a new CLI option: `--rerun-only`

The purpose of the command is the same as `--rerun`, but currently, if there is no `.rerun` file for the current test, all tests are executed.

Instead, this command exits with a code 0 if this is the case.

### DIFF
--- a/features/rerun-only.feature
+++ b/features/rerun-only.feature
@@ -1,0 +1,180 @@
+Feature: Rerun
+  In order to test only failed scenarios and exit if none is found
+  As a feature developer
+  I need to have an ability to rerun failed previously scenarios
+
+  Background:
+    Given a file named "features/bootstrap/FeatureContext.php" with:
+      """
+      <?php
+
+      use Behat\Behat\Context\Context;
+
+      class FeatureContext implements Context
+      {
+          private $apples = 0;
+          private $parameters;
+
+          public function __construct(array $parameters = array()) {
+              $this->parameters = $parameters;
+          }
+
+          /**
+           * @Given /^I have (\d+) apples?$/
+           */
+          public function iHaveApples($count) {
+              $this->apples = intval($count);
+          }
+
+          /**
+           * @When /^I ate (\d+) apples?$/
+           */
+          public function iAteApples($count) {
+              $this->apples -= intval($count);
+          }
+
+          /**
+           * @When /^I found (\d+) apples?$/
+           */
+          public function iFoundApples($count) {
+              $this->apples += intval($count);
+          }
+
+          /**
+           * @Then /^I should have (\d+) apples$/
+           */
+          public function iShouldHaveApples($count) {
+              PHPUnit\Framework\Assert::assertEquals(intval($count), $this->apples);
+          }
+
+          /**
+           * @Then /^context parameter "([^"]*)" should be equal to "([^"]*)"$/
+           */
+          public function contextParameterShouldBeEqualTo($key, $val) {
+              PHPUnit\Framework\Assert::assertEquals($val, $this->parameters[$key]);
+          }
+
+          /**
+           * @Given /^context parameter "([^"]*)" should be array with (\d+) elements$/
+           */
+          public function contextParameterShouldBeArrayWithElements($key, $count) {
+              PHPUnit\Framework\Assert::assertIsArray($this->parameters[$key]);
+              PHPUnit\Framework\Assert::assertEquals(2, count($this->parameters[$key]));
+          }
+      }
+      """
+    And a file named "features/apples.feature" with:
+      """
+      Feature: Apples story
+        In order to eat apple
+        As a little kid
+        I need to have an apple in my pocket
+
+        Background:
+          Given I have 3 apples
+
+        Scenario: I'm little hungry
+          When I ate 1 apple
+          Then I should have 3 apples
+
+        Scenario: Found more apples
+          When I found 5 apples
+          Then I should have 8 apples
+
+        Scenario: Found more apples
+          When I found 2 apples
+          Then I should have 5 apples
+
+        Scenario Outline: Other situations
+          When I ate <ate> apples
+          And I found <found> apples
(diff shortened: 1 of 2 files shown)


carlos-granados:
Wouldn't it be better to fix the current behaviour so that if you use `--rerun` and there is no `.rerun` file then no tests are executed??

features/rerun-only.feature
```diff
@@ -0,0 +1,180 @@
...
+          Examples:
+            | ate | found | result |
+            | 3   | 1     | 1      |
+            | 0   | 4     | 7      |
+            | 2   | 2     | 3      |
+      """
+    When I run "behat --no-colors -f progress features/apples.feature"
+    And I run "behat --no-colors -f progress features/apples.feature --rerun-only"
+    Then it should pass with:
+    """
```
acoulton, line 179: If we added this feature, I think it should render an explicit message to output - otherwise it could be difficult to distinguish from an unexpected exit.

acoulton:
@carlos-granados I guess the question is whether people are relying on `--rerun` behaving as it does at the moment - changing that could be a BC break? I would suggest it's safer to add this and (perhaps) deprecate the existing `--rerun` option.

carlos-granados:
@acoulton ok, I said that if no `.rerun` file was present then no tests should be run because the current help for this option is "Re-run scenarios that failed during last execution". So if when there is no `.rerun` file it is running scenarios that did not fail, it did not seem to be doing what was expected. 
But then I saw that in Behat 2 the help from this option said "Save list of failed scenarios into new file or use existing file to run only scenarios from it.". And then it added "If the file contains no scenarios, then all scenarios will executed". So I guess that the current functionality is as expected, it is just that the documentation does not show it clearly.

So I would say that the best course of action would be:
a) Modify the help message printed by the `rerun` option to something like "Re-run the list of scenarios that failed during last execution or save a list of those scenarios if it is not present"
b) Add this new `rerun-only` option as suggested
c) Modify the code so that if this option is set and no rerun file is present, it presents an informative message as you suggested

What do you think?

acoulton:
@carlos-granados I agree, except these days (probably since the behat 3 rewrite) the rerun file is written every time, regardless of the `--rerun` option. So I think the help text for the old `--rerun` should be slightly different.

src/Behat/Behat/Tester/Cli/RerunController.php
```diff
@@ -74,6 +74,9 @@ public function configure(Command $command)
         $command->addOption('--rerun', null, InputOption::VALUE_NONE,
             'Re-run scenarios that failed during last execution.'
```
acoulton, line 75: Please could you change the existing `--rerun` help text to "Re-run scenarios that failed during last execution, or run everything if there were no failures"

src/Behat/Behat/Tester/Cli/RerunController.php
```diff
@@ -74,6 +74,9 @@ public function configure(Command $command)
...
         );
+        $command->addOption('--rerun-only', null, InputOption::VALUE_NONE,
+            'Re-run scenarios that failed during last execution or exit if no file is found.'
```
acoulton, line 78: I think this is clearer than "no file is found" (which is an implementation detail):
```suggestion
            'Re-run scenarios that failed during last execution, or exit if there were no failures.'
```

src/Behat/Behat/Tester/Cli/RerunController.php
```diff
@@ -92,12 +95,12 @@ public function execute(InputInterface $input, OutputInterface $output)
...
         $this->key = $this->generateKey($input);
 
-        if (!$input->getOption('rerun')) {
+        if (!$input->getOption('rerun') && !$input->getOption('rerun-only')) {
             return;
         }
 
         if (!$this->getFileName() || !file_exists($this->getFileName())) {
-            return;
+            return $input->getOption('rerun-only') ? 0 : null;
```
acoulton, line 103: Please add an output message here to explain the early termination (and update the feature accordingly).

acoulton (changes requested):
I'd be happy to merge this (and sounds like Carlos would too) if you could just make these tweaks? Thanks :)

src/Behat/Behat/Tester/Cli/RerunController.php
```diff
@@ -71,9 +71,18 @@ public function __construct(EventDispatcherInterface $eventDispatcher, $cachePat
      */
     public function configure(Command $command)
     {
-        $command->addOption('--rerun', null, InputOption::VALUE_NONE,
+        $command->addOption(
+            '--rerun',
+            null,
+            InputOption::VALUE_NONE,
             'Re-run scenarios that failed during last execution.'
```
acoulton, line 78: Please could you update this text per the previous comment?

src/Behat/Behat/Tester/Cli/RerunController.php
```diff
@@ -156,12 +170,12 @@ private function generateKey(InputInterface $input)
...
-            implode(' ', $input->getOption('tags')) .
-            $input->getOption('role') .
-            $input->getArgument('paths') .
-            $this->basepath
+                $input->getOption('suite') .
+                implode(' ', $input->getOption('name')) .
+                implode(' ', $input->getOption('tags')) .
+                $input->getOption('role') .
+                $input->getArgument('paths') .
+                $this->basepath
```
acoulton, line 178: I think this is just formatting change? Please revert if so

acoulton (changes requested):

src/Behat/Behat/Tester/Cli/RerunController.php
Treast: Yeah, sorry. My formatting extension is kind of messed up.

acoulton (approved):
Thanks very much for updating this, it looks good to me now.

carlos-granados:
Looks good to me as well, thanks @Treast !

PR #1467, opened 2024-05-14
perf: add an array cache for callees

jeremyFreeAgent:
The more context we have the more `EnvironmentManager::readEnvironmentCallees()` method take time.

The aim is to avoid doing the work several time by run.


jeremyFreeAgent:
I close it because it may be specific to other lib.

PR #1471, opened 2024-06-23
FEATURE: Output custom stdout in `--format progress` mode when very verbose

mhsdesign:
Currently the progress formatter completely omits the possible stdout from the executed step. This makes debugging harder than it needs.

I know the default 'pretty' output dumps the stdout by default, but i dont like to use it as its tooo verbose for 10000+ steps and even imo for 5.

### DIFF
--- a/src/Behat/Behat/Output/Node/Printer/Progress/ProgressStepPrinter.php
+++ b/src/Behat/Behat/Output/Node/Printer/Progress/ProgressStepPrinter.php
@@ -52,6 +52,17 @@ final class ProgressStepPrinter implements StepPrinter
         $printer = $formatter->getOutputPrinter();
         $style = $this->resultConverter->convertResultToString($result);
 
+        // FIXME when refactored to symphony output use OutputInterface->getVerbosity() here instead of the magic global
+        $isVeryVerbose = $_SERVER['SHELL_VERBOSITY'] ?? 0 >= 2;
+        if (
+            $isVeryVerbose
+            && $result instanceof ExecutedStepResult
+            && ($callResult = $result->getCallResult())
+            && $callResult->hasStdOut()
+        ) {
+            $printer->writeln(['', $callResult->getStdOut()]);
+        }
+
         switch ($result->getResultCode()) {
             case TestResult::PASSED:
                 $printer->write("{+$style}.{-$style}");


src/Behat/Behat/Output/Node/Printer/Progress/ProgressStepPrinter.php
```diff
@@ -52,6 +52,17 @@ public function printStep(Formatter $formatter, Scenario $scenario, StepNode $st
         $printer = $formatter->getOutputPrinter();
         $style = $this->resultConverter->convertResultToString($result);
 
+        // FIXME when refactored to symphony output use OutputInterface->getVerbosity() here instead of the magic global
+        $isVeryVerbose = $_SERVER['SHELL_VERBOSITY'] ?? 0 >= 2;
```
mhsdesign, line 56: im very aware that this looks hacky, but i thought it would increase the odds of this being merged as bugfix?

Anyway i was not able to get the symphony output instance here as it seems behat is using an completely custom outputPrinter (probably due to history).

So i assume that once symphony is used here we can cleanly access the verbosity.

On the other hand, i think we could even drop the verbosity thing here and always dump the stdout as this is helpful for debugging, and in CI there should ideally be no output in general.

One more thing: The behat output has a verbosity option via `$printer->getOutputVerbosity()` which can only be set via: `--format-settings '{"output_verbosity":2}'` (it will not respect the global app option `-vv`)
But `getOutputVerbosity` is declared as `@deprecated since 3.1, to be removed in 4.0` so i dindt want to use that.

src/Behat/Behat/Output/Node/Printer/Progress/ProgressStepPrinter.php
```diff
@@ -52,6 +52,17 @@ public function printStep(Formatter $formatter, Scenario $scenario, StepNode $st
...
+        if (
+            $isVeryVerbose
+            && $result instanceof ExecutedStepResult
```
carlos-granados, line 59: Forgot to include the `use` statement for this class

carlos-granados (changes requested):
@mhsdesign thanks for your submission but I see several problems with it:
- I don't think that using the verbosity setting to activate this feature is a good idea as I believe that verbosity is designed to provide more information about Behat operation, not to enable output from the test itself
- I can't see how this can be useful. If you see some output from the tests you won't be able to say which test was the one that produced the output. The output will be a mix of dots and test output, rendering both the progress and the output itself unreadable
- If you need to see the output of some tests I think there are better options, for example use the pretty printer and the rerun feature or the pretty printer and some tags

@acoulton my vote is to close this PR, let me know what you think

acoulton:
@carlos-granados actually, I think the concept could be very useful.

We have a project that does need to output some debug info in CI (mostly around interactions with other services). We just use `fwrite(STDOUT, "info")` to bypass output buffering (based on the old `printDebug()` that used to exist in behat 2.

As you say, the output is pretty awkward merged in with the dots, but when something fails transiently it's better than nothing. The pretty printer is far too verbose for CI jobs in our usecase.

However I agree the implementation here isn't quite right:

* I agree it shouldn't tie to verbosity - an explicit formatter option would be better IMO.
* It should render a concise header before any output - maybe something like "\nOutput from: features/foo/bar.feature:42:\n" - that would make the feature worth having over the existing `fwrite` based approach
* In a perfect world a three-way option - no output (default), output only if scenario fails, output always - would be very helpful.

What do you think?

carlos-granados:
@acoulton I see what you mean and I agree that if we can separate the output from the formatter from the test output by using a header that clearly marks it, then it could work.

And yes, a three-way option would be great. Something like `--show-output=true`, `--show-output==false` and `--show-output=on-fail`

What about making this an option also available for the pretty formatter as well? This formatter always prints the output and I have found this sometimes inconvenient when there is a test that produces a lot of output. If this option was not present then the formatters would default to their current behaviour (true for pretty, false for progress)

In any case, whatever ends up finally implemented this PR needs some tests that confirm that the functionality is working as expected

carlos-granados:
@mhsdesign thanks for your work so far and for proposing this change. Do you think you might work to incorporate the suggested changes?

src/Behat/Behat/Output/Node/Printer/Progress/ProgressStepPrinter.php
```diff
@@ -52,6 +52,17 @@ public function printStep(Formatter $formatter, Scenario $scenario, StepNode $st
         $printer = $formatter->getOutputPrinter();
         $style = $this->resultConverter->convertResultToString($result);
 
+        // FIXME when refactored to symphony output use OutputInterface->getVerbosity() here instead of the magic global
```
ciaranmcnulty, line 55: Wrong (probably autocorrected) spelling of Symfony here

ciaranmcnulty (commented):
I'm not super-keen on stdout as a debug channel for Step Definitions, but equally just swallowing the output can be unexpected.

See #1491 for some ideas

Outputting it in the middle of the dots is not as neat as a nice summary at the end.

acoulton:
> Outputting it in the middle of the dots is not as neat as a nice summary at the end.

Although not as neat, there are cases where debug output during the run is extremely useful. Two off the top of my head:

* We had a project a while ago that was occasionally fataling out of the whole behat process in CI. Rendering debug output at key points during the run was the only way to figure out what was going on.
* We have a project that has a fairly long suite with integrations with an external service. Seeing that this is misbehaving in the streaming logs early in the build is a big timesaver compared to waiting till the end.

So although I like the idea of a more structured debug logging solution, I think we should let end-users have some control over when/where it appears - otherwise they'll just end up ignoring our feature and using `fwrite(STDERR,...)`.

acoulton:
sorry, I closed this by accident.

carlos-granados:
@mhsdesign any progress on this? Thanks!

acoulton:
@mhsdesign thanks for proposing this and the time discussing it, @carlos-granados has implemented an alternative version of this with an explicit config option as #1576 which will be in the next minor.

PR #1473, opened 2024-08-18
Remove deprecations in PHP8.4

heiglandreas:
PHP8.4 has deprecated implicit nullable parameters. Some of those are used within Behat and now render a lot of deprecation messages.

This PR replaces the implicit nullables with an explicit nullable wherever they are encountered. This should not have any influence on the code-execution but merely updates the function signature to make more explicit what is expected.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,7 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php: [7.2, 7.3, 7.4, 8.0, 8.1, 8.2, 8.3]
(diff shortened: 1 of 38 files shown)


jrfnl (approved):
This looks good @heiglandreas!

Maintainers: would be great if this could get merged & released ASAP as Behat is currently a blocker for other project their tests on PHP 8.4 and waiting to fail on deprecation notices. :pray: :heart:

andrewnicols:
I'll secodn that request for a release. This one is blocking us from properly testing and tracking down other issues.

Thanks @stof

jrfnl:
@andrewnicols FYI: I posted an action list for the release here: https://github.com/Behat/Behat/pull/1475#issuecomment-2387181346

PR #1474, opened 2024-10-01
Add dependabot configuration file

jrfnl:
Node 16 is deprecated on GitHub Actions for action runners. Most, if not all, action runners have released new majors to run on Node 20 to fix this, but the workflows in this repo still use outdated versions.

This commit adds an initial Dependabot configuration to:
* Submit pull requests for security updates and version updates for GH Action runner dependencies.

This PR doesn't include a config for the Composer dependencies, but this could be added later, if so desired.

The configuration has been set up to:
* Run weekly (for now).
* Submit a maximum of 5 pull requests at a time.
    If additional pull requests are needed, these will subsequently be submitted the next time Dependabot runs after one or more of the open pull requests have been merged.

Refs:
* https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
* https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy

Once this PR is merged, at least one Dependabot PR will be opened (for `actions/checkout`).

Merging that will fix the following notice which currently shows in Actions run summaries:
```
The following actions use a deprecated Node.js version and will be forced to run on node20: actions/checkout@v3. For more info: https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+# Dependabot configuration.
+#
+# Please see the documentation for all configuration options:
+# https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  # Maintain dependencies for GitHub Actions.
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 5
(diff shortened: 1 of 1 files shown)


stof:
Can it be configured to do grouped updates instead of submitting many PRs ?

jrfnl:
> Can it be configured to do grouped updates instead of submitting many PRs ?

Good question. I haven't used that feature before. Also not sure it's really worth it as the workflow uses only four action runners and is set to only need updates on majors.

Either way, I've added a config for this now, but this is untested.

PR #1475, opened 2024-10-01
GH Actions/tests: fix workflow issues

jrfnl:
### GH Actions/tests: use "develop" ini file

By default setupPHP uses the `production` ini file, which silences some errors.
Changing this to the `development` ini file should surface all errors when running the tests.

### GH Actions/test: use ignore-platform-reqs

... on "PHP next", not on "PHP latest".

This should fix the `composer install` failure on PHP 8.4.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -92,7 +92,7 @@ jobs:
                 if: matrix.composer-mode == 'update'
                 env:
                     SYMFONY_REQUIRE: ${{ matrix.symfony-version }}.*
-                run: composer update ${{ matrix.php == '8.3' && '--ignore-platform-req=php+' || '' }}
+                run: composer update ${{ matrix.php == '8.4' && '--ignore-platform-req=php+' || '' }}
 
             -   name: Install lowest dependencies
                 if: matrix.composer-mode == 'lowest'
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -73,6 +73,7 @@ jobs:
                 uses: shivammathur/setup-php@v2
                 with:
                     php-version: "${{ matrix.php }}"
+                    ini-file: "development"
                     ini-values: "phar.readonly=0,zend.exception_ignore_args=Off"
                     coverage: none
 


.github/workflows/build.yml
```diff
@@ -73,7 +73,7 @@ jobs:
                 uses: shivammathur/setup-php@v2
                 with:
                     php-version: "${{ matrix.php }}"
-                    ini-values: "phar.readonly=0,zend.exception_ignore_args=Off"
+                    ini-values: "zend.assertions=1,error_reporting=-1,display_errors=On,phar.readonly=0,zend.exception_ignore_args=Off"
```
stof, line 76: better way for that is to use `ini-file: development`, to have all the configuration tailored for development environments rather than production ones.

We will still need those specific `ini-values` that we have for our CI setup, but it would avoid having to maintain manually the list of development settings.

jrfnl: That's anothér way to do it, yes. Ive updated the commit now for this preferences.

stof (approved):

jrfnl:
Note: to get a passing build on PHP 8.4 in combination with the changes in this PR, the following prerequisites should be full-filled:
- [ ] New tag for `behat/gherkin` which includes the fixes made in behat/gherkin#258 and behat/gherkin#262
- [x] Merge PR #1473
- [ ] Merge PR #1476
- [ ] Merge PR #1477
- [ ] Merge PR #1478

I've tested this combination locally and managed to get a passing build.

Would also be great, if once the above mentioned actions have been taken and this PR has been merged too, a new release of Behat would be tagged to unblock users who want to start tested on PHP 8.4.

stof:
I would first focus on getting the needed PRs merged in the dev version. People wanting to test things on PHP 8.4 which is not released yet can use the dev version rather than a stable release (and this would help making sure that all necessary work is done before making a release claiming to support PHP 8.4)

jrfnl:
> I would first focus on getting the needed PRs merged in the dev version. 

@stof As I said, I've done my due diligence by testing the combination of the above tasks.

> > I've tested this combination locally and managed to get a passing build.

Other than that, there's not much I can do to get the needed PRs merged. That's up to the maintainers for whom I've made this as easy as possible by submitting small, single-decision, atomic PRs, which should allow for a quick and straight-forward review.

Also note that I've also scanned the project(s) with bleeding edge [PHPCompatibility](https://github.com/PHPCompatibility/PHPCompatibility), so I'm fairly confident most things will now have been found between the tests and the scans.


> People wanting to test things on PHP 8.4 which is not released yet can use the dev version rather than a stable release (and this would help making sure that all necessary work is done before making a release claiming to support PHP 8.4)

Once everything is merged, I'll happily run a test run with the dev version on a few projects, but that shouldn't block a release to get the available fixes out.

The release doesn't need to claim full PHP 8.4 compatibility, just that the "known issues" have been fixed.

If new issues would be found after, a new tag can be released with additional fixes.

Keep in mind, most projects will not accept running their tests on a `dev` version as their composer config doesn't even allow for dev versions (`minimum-stability: stable`).

The fact that PHP 8.4 is not released is irrelevant, as Behat is a test tool and open source projects will want to be ready for PHP 8.4 in time and Behat is currently blocking them from doing _their_ due diligence for that.

Lastly, I'm not a maintainer of your project. I shouldn't have to be the one pushing this and going in to make these fixes and fix CI.
I'm just a "drive-by" contributor trying to help. Don't put tasks which belong with the maintainers on me.

andrewnicols:
Just to echo @jrfnl's comments, a release is really necessary here. We are not expecting _all_ things to be fixed, but this will make it at least possible to start testing against PHP 8.4 without dropping our minimum stability.

I have also taken a look at both of the pending PRs.

PR #1476, opened 2024-10-01
PHP 8.4 | CallExtension: remove use of `E_STRICT` [1]

jrfnl:
The `E_STRICT` constant is deprecated as of PHP 8.4 and will be removed in PHP 9.0.

The error level hasn't been in use since PHP 8.0 anyway and was only barely still used in PHP 7.x, so removing the exclusion from the default value in the `CallExtension` script shouldn't really make any difference in practice.

Ref:
* https://wiki.php.net/rfc/deprecations_php_8_4#remove_e_strict_error_level_and_deprecate_e_strict_constant


andrewnicols (approved):

everzet:
I accidentally fixed this myself when getting back up-to-speed with the repo. Thank you for the contribution though!

jrfnl:
@everzet Well, at least it's fixed now 🤷🏻‍♀️

PR #1480, opened 2024-10-18
Add PRIORITY constant in RowBasedTableTransformation.php

ablazejuk:


stof:
What is the reason to add such a public constant ?

ablazejuk:
> What is the reason to add such a public constant ?

Changed it to private

acoulton:
@ablazejuk thanks for the PR. Please can you explain the purpose of this change? As this is a final class, and now you've made the constant private, it doesn't seem to add any benefit?

acoulton:
@ablazejuk I'm going to close this to help us see what's still outstanding, as I don't understand the argument for the change. If I've missed something and you think it's important please feel free to reopen it.

PR #1483, opened 2024-10-22
Fix expected assertion failure for max token length test

acoulton:
#1457 updated the policy for token lengths, and updated the test to trigger the assertion, but it missed changing the expected error message so master is now failing builds.

Looks like we missed this before the merge, because the PR was so stale that the CI either hadn't run on the commit, or ran but has since been archived from github.

### DIFF
--- a/features/definitions_patterns.feature
+++ b/features/definitions_patterns.feature
@@ -234,7 +234,7 @@ Feature: Step Definition Pattern
     When I run "behat -f progress --no-colors"
     Then it should fail with:
       """
-      Token name should not exceed 32 characters, but `too1234567891123456789012345678901` was used.
+      Token name should not exceed 32 characters, but `too12345678911234567890123456789012` was used.
       """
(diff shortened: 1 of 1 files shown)


carlos-granados:
Thanks @acoulton not sure how it was merged with this failing test, must be something like what you described

acoulton:
@carlos-granados it looks like a lot of the PRs don't have a CI status, either because it ran and has now been archived, or because the workflow needs admin approval for new contributors and they weren't granted at the time. And it hadn't occurred to me till this morning that of course some of the open PRs are quite a bit behind master, so even if they went green originally they might not now.

We'll just need to make sure any older PRs are rebased / merged up to date & pass CI before merging.

I enabled the "Always suggest updating pull request branches" feature in the repo settings this morning so (as long as the contributor has granted us permission to edit the PR) we can easily do that from the github web interface now.

PR #1487, opened 2024-10-23
Avoid arbitrary output in tests

uuf6429:
When the relevant tests failed, the output was hidden since echo/output is not the right way to influence PHPUnit test results (the `$message` assertion parameter should be used instead).


acoulton (approved):
I see what you mean about other ways this could be done, but I think for what it is it's good enough. Thanks!

carlos-granados (approved):
Looks good, thanks @uuf6429 !

PR #1468, opened 2024-05-15
[Chore]: Fixed redirect to documentation URL in readme file.

Ar-Monta:
Dear Behat Team,

I am writing to express my heartfelt gratitude for the incredible work you have done in developing and maintaining the Behat framework. Your dedication and innovation in creating this Behavior-Driven Development (BDD) tool have made a significant impact on the PHP community and on countless projects around the world.

Behat has been an invaluable asset in helping us align our development efforts with business expectations. The clarity and precision it brings to our testing processes have not only improved our code quality but also enhanced collaboration between developers and stakeholders. Your commitment to making BDD accessible and effective has truly set a high standard in the industry.

Thank you for your hard work, continuous improvements, and the support you provide to the community. Your contributions are deeply appreciated, and we look forward to seeing how Behat continues to evolve and inspire better software development practices.

With sincere thanks,

Arshia Montakhabi

Backend dev.

G2Tech

Change log:
- Fixed redirect to the documentation of Behat URL in the readme file.

PR #1492, opened 2024-10-27
chore: Update links in README to https, drop dead channels

acoulton:
The freenode IRC no longer exists (as far as I know), and the google group exists but appears to be entirely spam emails.

Neither of these are likely to be good destinations for community / support in 2024 so I have removed them.

All other links are now https (some were http).

Replaces #1468


carlos-granados (approved):
LGTM

PR #1499, opened 2024-10-29
chore: Prepare for 3.15.0 release

acoulton:
Updates the changelog ahead of releasing 3.15.0.

@carlos-granados if you're happy with this I can go ahead and cut the release on github.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,21 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [3.14.0] - 2024-01-10
+
+### Added
+
+* 🎉 Symfony 7 is now supported 🎉 by @dmaicher in [#1442](https://github.com/Behat/Behat/pull/1442)
+* PHP 8.3 is now supported (no code changes were required) by @jrfnl in [#1440](https://github.com/Behat/Behat/pull/1440)
+
+### Fixed
+
+* Renamed method parameters to match signatures from interfaces by @ciaranmcnulty in [#1434](https://github.com/Behat/Behat/pull/1434)
+
+### Internal
+
+* CI improvements by @stof in [#1430](https://github.com/Behat/Behat/pull/1430)
+
 ## [3.13.0] - 2023-04-18
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,37 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [3.15.0] - 2024-10-29
+
+### Added
+
+* PHP 8.4 support by @heiglandreas in [#1473](https://github.com/Behat/Behat/pull/1473), @jrfnl in [#1478](https://github.com/Behat/Behat/pull/1478),
+  @jrfnl in [#1477](https://github.com/Behat/Behat/pull/1477)
+* Support config files named `behat.dist.y[a]ml` by @uuf6429 in [#1464](https://github.com/Behat/Behat/pull/1464)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,52 @@ All notable changes to this project will be documented in this file.
...
+  by @andrewnicols in [#1334](https://github.com/Behat/Behat/pull/1334)
+* Add void return type when generating new snippets by @carlos-granados in [#1463](https://github.com/Behat/Behat/pull/1463)
+
+### Fixed
+
+* Fix enforcing that 32 character (or longer) turnip pattern names are not allowed
+  by @ivastly in [#1457](https://github.com/Behat/Behat/pull/1457) and @acoulton in [#1483](https://github.com/Behat/Behat/pull/1483)
+* Fix generating the PHAR for releases by upgrading the build tooling by @heiglandreas in [#1462](https://github.com/Behat/Behat/pull/1462)
+
+### Internal
```
stof, line 27: Does this even belong to a changelog if it is internal ?

CHANGELOG.md
```diff
@@ -4,6 +4,52 @@ All notable changes to this project will be documented in this file.
...
+* 🎉 Symfony 7 is now supported 🎉 by @dmaicher in [#1442](https://github.com/Behat/Behat/pull/1442)
+* PHP 8.3 is now supported (no code changes were required) by @jrfnl in [#1440](https://github.com/Behat/Behat/pull/1440)
+
+### Fixed
+
+* Renamed method parameters to match signatures from interfaces by @ciaranmcnulty in [#1434](https://github.com/Behat/Behat/pull/1434)
+
+### Internal
+
+* CI and Psalm improvements by @stof in [#1430](https://github.com/Behat/Behat/pull/1430)
```
stof, line 51: There is no Psalm changes in this PR.

acoulton: Good catch. I'd originally grouped in the psalm changes from 1434 then decided since it included that parameters (and therefore technically could affect people calling with named params) it should be listed as a bugfix but forgot to update this line.

CHANGELOG.md
acoulton: It's a valid question - personally, I would probably lean closer to "no" on some of these, but thought that might be controversial with others (conscious the convention these days is increasingly to list absolutely everything). 

That said, there's always the chance someone is using the package in an unexpected way (including maybe applying a manual patch to a file as part of composer install, or something) so there might be an argument to note these since at least one changes runtime code.

It also feels a safer convention to say everything is listed, but some things might be "Internal". If we make the wrong call on a PR that just refactors things or updates code style, at least it's there somewhere rather than missing outright. Plus it shows it's been considered, rather than just overlooked.

Finally there's the increasing convention to credit contributors in the changelog - I don't want to downplay people's work just because the immediate benefit is to us rather than end-users.

I did, however, decide not to include dependabot PRs at all.

stof: Listing everything is often caused by the fact that it is generated automatically, without a human review step.

acoulton: I'm very aware of that - that's my motivation to find a release process that does have a review step.

Personally, I think I've stuck a balance between including everything and excess verbosity that I'd be happy with as an end user.

carlos-granados (commented):
@acoulton overall it looks great, nice job. My only question would be about the item about the update to the contributing and readme docs. Those changes do not really look "internal" to me but not really sure where to put them. Maybe a "Documentation" section?

acoulton:
@carlos-granados if they were more substantial I'd put them in a Documentation section. But for this:

* The readme changes are just changing some links to avoid an http -> https redirect, and dropping a couple that were 404 / dead. So really the same kind of change as renaming a private variable or reformatting code.
* The contributing guide is only for the relationship between maintainers and contributors - they're for "developers " not "users" - and the changes here really just clarify the existing process.

So I think Internal is right - they're part of the history, but not something an end user needs to pay attention to when upgrading their own projects to the new behat version?

carlos-granados (approved):

acoulton:
@carlos-granados thanks - released! :fireworks: https://github.com/Behat/Behat/releases/tag/v3.15.0

Issue #1470, opened 2024-05-31
getScenario() has to return a Scenario but returns an ExampleNode so the getTitle() function didn't works as expected


cavo789:
Hi all

I've multiple scenarios and some are `Scenario Outline`.

I'm using a hook for `@BeforeScenario` like 

```php
public function beforeScenario(BeforeScenarioScope $scope): void 
{
}
```

Below my debugging session in vscode:

![image](https://github.com/Behat/Behat/assets/355999/d6400a32-ba1c-43db-89fc-8cb92e904010)

We can see `$scope` is well a `BeforeScenarioScope` and in my console, `$scope->getScenario()` is returning an `ExampleNode` and that's **wrong** since the docblock for `getScenario()` tells it should be a Scenario

File src/Behat/Behat/Hook/Scope/BeforeScenarioScope.php
```php
    /**
     * Returns scenario.
     *
     * @return Scenario
     */
    public function getScenario()
    {
        return $this->scenario;
    }
```

(the docblock is not correct)

By dumping the `$scope` object, we can see there is a private attribute `outlineTitle` and I've retrieved a `getOutlineTitle()` so I'm unblocked...

But it is normal?  I'll need to check if I'm facing a Scenario or an ExampleNode and depending on what, use `getTitle()` in the first case and `getOutlineTitle()` in the second case.

Tanks

uuf6429:
See also #1459 ... it might be related..

carlos-granados:
@cavo789 the code is actually correct. The scenario scope declares that it returns a `Scenario` but this is actually an alias for a `ScenarioInterface`. `ExampleNode` implements the `ScenarioInterface` interface, so it is fine that this function returns an `ExampleNode`. You should be able to use the methods in the interface to retrieve any information that you need.

Closing this issue, thanks for reporting it.

stof:
I would suggest removing the alias in the code, to avoid such confusion

uuf6429:
@stof I believe that removing the alias won't solve the confusion/problem, it just replaces the wording:
- `wrong since the docblock for getScenario() tells it should be a Scenario` -> `... tells it should be a ScenarioInterface`
- `I'll need to check if I'm facing a Scenario or an ExampleNode` -> `... ScenarioInterface or an ExampleNode`

I would propose the following isntead (although I'm not against removing the alias either):
- `ExampleNode` is too vague. `ScenarioExampleNode` (or similar) might have avoided this issue altogether.
[...]

PR #1500, opened 2024-10-29
Removes the unnecessary alias of the ScenarioInterface as it just causes confusion

carlos-granados:
In the ScenarioScope classes the `ScenarioInterface` interface had been aliased as `Scenario`. This alias is unnecessary and just causes confusion on people expecting a specific class and not a class that implements this interface, see issue https://github.com/Behat/Behat/issues/1470 for an example.

This PR just removes this alias in all these classes


acoulton (approved):
LGTM thanks

PR #1501, opened 2024-10-30
Add `ConfigurationExtension`

### LINKED ISSUE #1488: Allow stop-on-failure to be a parameter of the profile.
Hi,

I love the `stop-on-failure` cli parameter. But this is only a cli parameter.
By searching, I think that it does decorate the cli and hook into behat.

It would be great if we could configure this feature in the profile: to be honest, I nearly launch behat on dev environment ONLY with stop-on-failure, but I do not want that in CI.

Thank you

jdeniau:
## Current status

We have a lot of cli parameters, but no way to get them configured directly in the behat.yml file (and to configure different options for different profile).

## Proposed evolution

This PR propose to add a `configuration` key in the `behat.yml` that can handle a subset of the cli parameters (for now only `stop_on_failure`).

This way we can configure our behat.yml file like this:

```
default:
  configuration:
    stop_on_failure: true

ci:
  configuration:
    stop_on_failure: false
```

We could add more configuration option in the future (`--strict` or `--snippet-xxx` for example).

Fixes #1488

### DIFF
--- a/features/stop_on_failure_configuration.feature
+++ b/features/stop_on_failure_configuration.feature
@@ -1,0 +1,113 @@
+Feature: Stop on failure via configuration
+  In order to stop further execution of steps when first step fails
+  As a feature developer
+  I need to have a configuration.stop_on_failure sets to true
+
+  Background:
+    Given a file named "features/bootstrap/FeatureContext.php" with:
+      """
+      <?php
+
+      use Behat\Behat\Context\Context;
+
+      class FeatureContext implements Context
+      {
+          /**
+           * @Given /^I have (?:a|another) step that passes?$/
+           * @Then /^I should have a scenario that passed$/
+           */
+          public function passing() {
+          }
+
+          /**
+           * @Given /^I have (?:a|another) step that fails?$/
+           * @Then /^I should have a scenario that failed$/
+           */
+          public function failing() {
+              throw new Exception("step failed as supposed");
+          }
+
+      }
+      """
+    And a file named "features/failing.feature" with:
+      """
+      Feature: Failing Feature
+        In order to test the stop-on-failure feature
+        As a behat developer
+        I need to have a feature that fails
+
+        Background:
+          Given I have a step that passes
+
+        Scenario: 1st Passing
+          When I have a step that passes
+          Then I should have a scenario that passed
+
+        Scenario: 1st Failing
+          When I have a step that passes
+           And I have another step that fails
+          Then I should have a scenario that failed
+
+        Scenario: 2st Failing
+          When I have a step that fails
+          Then I should have a scenario that failed
+
+        Scenario: 2nd Passing
+          When I have a step that passes
+           And I have another step that passes
+          Then I should have a scenario that passed
+      """
+
+  Scenario: with stop_on_failure set to false
+   Given a file named "behat.yml" with:
+      """
+      default:
+        configuration:
+          stop_on_failure: false
+      """
+    When I run "behat --no-colors --format-settings='{\"paths\": false}' features/failing.feature"
+    Then it should fail with:
+      """
+      --- Failed scenarios:
+
+          features/failing.feature:13
+          features/failing.feature:18
+
+      4 scenarios (2 passed, 2 failed)
+      14 steps (10 passed, 2 failed, 2 skipped)
+      """
+
+  Scenario: with stop_on_failure default value (false)
+   Given a file named "behat.yml" with:
+      """
+      default: ~
+      """
+    When I run "behat --no-colors --format-settings='{\"paths\": false}' features/failing.feature"
+    Then it should fail with:
+      """
+      --- Failed scenarios:
+
+          features/failing.feature:13
+          features/failing.feature:18
+
+      4 scenarios (2 passed, 2 failed)
+      14 steps (10 passed, 2 failed, 2 skipped)
+      """
+
+  Scenario: with stop_on_failure set to true
+   Given a file named "behat.yml" with:
+      """
+      default:
+        configuration:
+          stop_on_failure: true
+      """
+    When I run "behat --no-colors --format-settings='{\"paths\": false}' features/failing.feature"
+    Then it should fail with:
+      """
+      --- Failed scenarios:
+
+          features/failing.feature:13
+
+      2 scenarios (1 passed, 1 failed)
+      7 steps (5 passed, 1 failed, 1 skipped)
+      """
--- a/src/Behat/Behat/EventDispatcher/Cli/StopOnFailureController.php
+++ b/src/Behat/Behat/EventDispatcher/Cli/StopOnFailureController.php
@@ -10,7 +10,7 @@
 
 namespace Behat\Behat\EventDispatcher\Cli;
 
-use Behat\Behat\Configuration\Handler\StopOnFailureHandler;
+use Behat\Behat\Config\Handler\StopOnFailureHandler;
 use Behat\Behat\EventDispatcher\Event\AfterScenarioTested;
 use Behat\Behat\EventDispatcher\Event\ExampleTested;
 use Behat\Behat\EventDispatcher\Event\ScenarioTested;
@@ -27,6 +27,7 @@ use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 
 /**
  * Stops tests on first scenario failure.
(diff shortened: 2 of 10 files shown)


carlos-granados:
@jdeniau how does this interact with the parameter sent on the command line? Which one takes precedence if both the command line parameter and the config file value are set to different values?

jdeniau:
@carlos-granados good question : I didn't think of that.
The way it's coded, it's a OR (if one is set to true, then it will stop on failure. As a matter of fact, the listeners will be registered twice).

We might want to change that to let the cli have the last word, but in that case we may want to implement `--no-stop-on-failure`

carlos-granados:
> @carlos-granados good question : I didn't think of that. The way it's coded, it's a OR (if one is set to true, then it will stop on failure. As a matter of fact, the listeners will be registered twice).
> 
> We might want to change that to let the cli have the last word, but in that case we may want to implement `--no-stop-on-failure`

Yes, I would like to make sure that there is a clear priority so you may need to work on your code. Don't forget to add new tests that check this

jdeniau:
@carlos-granados no problem to improve that.
Are you OK on the global idea of the `configuration` key in behat.yml to be sure that I do not develop something that is not related to behat's way of think?

carlos-granados:
> @carlos-granados no problem to improve that. Are you OK on the global idea of the `configuration` key in behat.yml to be sure that I do not develop something that is not related to behat's way of think?

I believe it is a good idea but would also like to hear the opinion of @acoulton so please wait until we hear from him

carlos-granados:
Ah, and I would probably use `config:` instead of `configuration:`, that seems to be the term used elsewhere in Behat

src/Behat/Behat/EventDispatcher/Cli/StopOnFailureController.php
```diff
@@ -36,24 +36,18 @@
...
+    private $stopOnFailureHandler;
 
     /**
      * Initializes controller.
      *
-     * @param EventDispatcherInterface $eventDispatcher
+     * @param StopOnFailureHandler $stopOnFailureHandler
      */
-    public function __construct(EventDispatcherInterface $eventDispatcher)
+    public function __construct(StopOnFailureHandler $stopOnFailureHandler)
```
acoulton, line 48: This looks like a BC break, can it be avoided? e.g. by still passing in the dispatcher & creating the StopOnFailureHandler in the constructor?

src/Behat/Behat/EventDispatcher/Cli/StopOnFailureController.php
```diff
@@ -82,28 +76,14 @@ public function execute(InputInterface $input, OutputInterface $output)
...
-        $this->eventDispatcher->addListener(ExampleTested::AFTER, array($this, 'exitOnFailure'), -100);
+        $this->stopOnFailureHandler->registerListeners($resultInterpretation);
     }
 
-    /**
-     * Exits if scenario is a failure and if stopper is enabled.
-     *
-     * @param AfterScenarioTested $event
-     */
-    public function exitOnFailure(AfterScenarioTested $event)
```
acoulton, line 98: This looks like a BC break? Can we keep the method and proxy to the StopOnFailureHandler?

acoulton (commented):
Only had time for a very quick skim of the code but I really like the concept. Would be extremely useful (in a future iteration, obviously) to set a default for the snippets options as I pretty much always want to disable them at runtime.

I think `config:` / `configuration:` might be a bit too generic - maybe `command_options`: or `cli_options:` to make the relationship clear? And call it `DefaultCommandOptionsExtension`? Just a thought...

With the boolean flags for priority I think:

* The default is the value from behat.yml (so false is valid, but redundant, as it is the same as not giving the flag)
* Specifying `--stop-on-failure` on the CLI overrides `false` in the behat.yml
* If behat.yml specifies `true` then at present there is no direct way to override that

Which I think works out the same as the `OR` you have at the moment?

If people do need to force a flag back to false, they can do that by using a different profile from the behat.yml. Especially now that we can set a preferred profile, that should be easy.

I'd prefer to avoid lots of CLI flags, so wouldn't add a `--no-stop-on-failure` unless it's requested. And then would probably instead change `--stop-on-failure` to take an `=0` or `=1` (default for BC).

features/stop_on_failure_configuration.feature
```diff
@@ -0,0 +1,113 @@
...
+           And I have another step that passes
+          Then I should have a scenario that passed
+      """
+
+  Scenario: with stop_on_failure set to false
+   Given a file named "behat.yml" with:
+      """
+      default:
+        configuration:
+          stop_on_failure: false
```
acoulton, line 66: Oh also, although we generally use `_` for map keys, perhaps we should use `-` (e.g. `stop-on-failure`) for these so that they exactly match the CLI arguments?

carlos-granados:
> I think `config:` / `configuration:` might be a bit too generic - maybe `command_options`: or `cli_options:` to make the relationship clear? And call it `DefaultCommandOptionsExtension`? Just a thought...

In my case I find `config:` to be clearer: what you are defining is the configuration. The fact that this can also be defined on the command line should not direct the naming in my opinion

features/stop_on_failure_configuration.feature
jdeniau: I questioned myself on that, and preferred to be consistent with the rest of the yaml configuration. But it joins the other discussion about the naming, because it's the same philosophy.

carlos-granados: I think I would find it weird to see `-` used here when we use `_` everywhere else

carlos-granados: The relationship between these options and the command line ones should be obvious

stof:
symfony/console 5.3 introduced support for negatable options for flag options, adding automatically the ability to use `--no-foo` to set the `foo` flag to `false` even if its default is `true` (or `null`). I think this feature should be used, to preserve the ability for the CLI flags to disable some features again.
This will require dropping support for Symfony 4.4. I think this is fine to do for a next minor version of Behat (maybe releasing first the current merged changes), given that Symfony 4.4 is EOL since 11 months.

acoulton:
Happy with all your arguments. I didn't know about auto-negatable-flags, in that case perfect. Agree with dropping symfony 4.4 in a minor - and we have released up to date so that can happen soon. 

Carry on @jdeniau 😀

src/Behat/Behat/EventDispatcher/Cli/StopOnFailureController.php
acoulton: Anything *can* be used outside of behat - user code is running quite tightly coupled with the behat code, and for example extensions have full access to reconfigure / use the DI container so we can't guarantee what people are doing.

I prefer to take a safe / strict view of semver unless things are explicitly marked as not part of the public API (for example, [phpunit is very clear about this](https://github.com/sebastianbergmann/phpunit/blob/921ca7c95d440437da5ce077cf691ebf90c116dc/src/Framework/Exception/Exception.php#L41))

So I would like to see the class constructor keep BC, as you say ideally with a comment to change it in future.

This `exitOnFailure` method is tricky though - it looks like it was only ever public because callbacks had to be in older PHP versions, and it was never intended to be called directly. So looking at it again I realise proxying to the new listener doesn't really make sense.

I think therefore it's extremely unlikely anyone will be calling that, and we could safely remove it.

stof:
well, negatable flags still require an opt-in when defining the option, but it is quite easy: https://symfony.com/blog/new-in-symfony-5-3-negatable-command-options

src/Behat/Behat/EventDispatcher/Cli/StopOnFailureController.php
jdeniau: OK I will do that 👍

jdeniau: I finally kept the event dispatcher in the constructor and add a setter as `@required` that is called via the DI.

jdeniau:
@acoulton @carlos-granados PR is updated

carlos-granados (changes requested):
Just some minor comments

carlos-granados (approved):
Everything looks good to me now

acoulton (approved):
@jdeniau I owe you an apology - I reviewed this on my phone between other things yesterday (behat has been busier than I expected when I signed up so I've been juggling a bit 🤣).

I completely missed that of course the new `StopOnFailureHandler` needs to be a single global listener that has to be injected into the `StopOnFailureController` somehow so a technical BC break is inevitable.

Thanks for switching it to setter injection.

I'm satisfied this class is sufficiently internal to behat that the risk of causing problems is extremely low. And it's easy for end users to inject the handler if they are creating a `StopOnFailureController` in an unexpected way.

So I'm happy for this to go into the next minor as-is.

Thanks for your contribution & apologies again for the rushed review.

PR #1502, opened 2024-10-30
chore: Fixed indentation in GHA.

AlexSkrypnyk:
This PR fixes indentation in GHA to be inline with the `.editorconfig` rules where we set the indentation to `2` spaces.

Also, moves the job name to the top to clearly see what it is.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,200 +1,199 @@
 name: Build
 
 on:
-    push:
-        branches: [master]
-    pull_request: ~
-    release:
-        types: [created]
-    workflow_dispatch: ~
+  push:
+    branches: [master]
+  pull_request: ~
+  release:
+    types: [created]
+  workflow_dispatch: ~
 
 jobs:
-    tests:
-        runs-on: ${{ matrix.os }}
-        defaults:
-            run:
-                shell: bash
-        name: Test
-        strategy:
-            fail-fast: false
-            matrix:
-                php: [7.2, 7.3, 7.4, 8.0, 8.1, 8.2, 8.3, 8.4]
-                os: [ubuntu-latest]
-                composer-mode: [update]
-                symfony-version: ['']
-                include:
-                    # 7.2 build with lowest dependencies
-                    -   php: 7.2
-                        os: ubuntu-latest
-                        composer-mode: lowest
-                        symfony-version: ''
+  tests:
+    name: Tests
+    runs-on: ${{ matrix.os }}
+    defaults:
+      run:
+        shell: bash
+    strategy:
+      fail-fast: false
+      matrix:
+        php: [7.2, 7.3, 7.4, 8.0, 8.1, 8.2, 8.3, 8.4]
+        os: [ubuntu-latest]
+        composer-mode: [update]
+        symfony-version: ['']
+        include:
+          # 7.2 build with lowest dependencies
+          - php: 7.2
+            os: ubuntu-latest
+            composer-mode: lowest
+            symfony-version: ''
 
-                    # MacOS on latest PHP only
-                    -   php: 8.2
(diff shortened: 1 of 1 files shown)


AlexSkrypnyk:
What is the reason to have the approval for workflow runs? Doesn't this puts a barrier for contributions?

stof:
@AlexSkrypnyk requesting approval for new contributors is the default behavior of GitHub, as a way to prevent abuse where malicious users were sending PRs to many repositories modifying the workflows to make them do cryptocurrency mining for them.

AlexSkrypnyk:
Thanks @stof 
So it looks like the approval is required only **once** for a first time contributors to the project.

stof:
The approval is required until you have your first PR merged in the repo (which might require approval more than once, if the contribution is not accepted immediately but requires extra changes triggering another run)

acoulton (approved):

stof (approved):

acoulton:
Thanks @AlexSkrypnyk

trampgeek:
I suspect this change is the cause of my behat scripts suddenly breaking. I am using behat with Moodle, using moodle-ci on my developmnent system and github actions when pushing changes. Both are now breaking, with behat code that has worked for years.

I have the following behat statement, which inserts Python code into a form field:
<pre>
    And I set the field "id_template" to:
  """
  import subprocess, json, sys
  {{ STUDENT_ANSWER | e('py') }}
  tests = json.loads('''{{ TESTCASES | json_encode | e('py') }}''')
  test_results = [['Test', 'Expected', 'Got', 'iscorrect']]
  num_right = 0
  for test in tests:
    code = test['testcode']
    expected = int(test['expected'])
    got = eval(code)
    test_results.append([code, expected, got, expected == got])
    if expected == got:
      num_right += 1
  print(json.dumps({
    'testresults': test_results,
    'fraction': num_right / len(tests)
  }))
  """
</pre>

What that now produces is:

<pre>import subprocess, json, sys
{{ STUDENT_ANSWER | e('py') }}
tests = json.loads('''{{ TESTCASES | json_encode | e('py') }}''')
test_results = [['Test', 'Expected', 'Got', 'iscorrect']]
num_right = 0
for test in tests:
  code = test['testcode']
    expected = int(test['expected'])
      got = eval(code)
        test_results.append([code, expected, got, expected == got])
          if expected == got:
              num_right += 1
              print(json.dumps({
                'testresults': test_results,
                  'fraction': num_right / len(tests)
                      }))
</pre>

It appears that behat is autoindenting each line to match the previous line, then *adding* my own indentation to that. This is contrary to the specified behaviour:

"Indentation of the opening """ is not important, although common practice is two spaces in from the enclosing step. The indentation inside the triple quotes, however, is significant. Each line of the string passed to the step definition’s callback will be de-indented according to the opening """. Indentation beyond the column of the opening """ will therefore be preserved."

AlexSkrypnyk:
@trampgeek
this PR only updates the indentation of GHA of this repository.

trampgeek:
Ah, thanks. I should have looked more closely. Sorry.

carlos-granados:
@trampgeek can you please create a new issue with this information and we will check it to see if this is something that we should fix on our side? Thanks!

trampgeek:
OK, thanks for the interest. I'll add the issue, though the Moodle + behat environment is too complex for me to give you a sequence of "steps to reproduce".

trampgeek:
I posted the issue, but then closed it again. It turned out that the problem was that we had recently added JavaScript to autoindent the field concerned. The "I set the field to ..." statement apparently simulates this by a series of keypresses rather than a direct setting of the text. Hence the multiple indents. 

Again, thank for taking an interest and of course thanks for Behat - I don't think Moodle could exist without it.

Issue #1503, opened 2024-10-30
PHP / symfony version support policy


acoulton:
* [x] #1504 
* [x] Behat/docs#173
* [x] Update version constraints / CI in other behat packages

We need to agree a policy for supporting versions of PHP & Symfony (and any other runtime dependencies)

In summary, I propose:

* We only support PHP versions until they reach the end of security support from php.net - _so currently only 8.1, 8.2, 8.3 (and 8.4)._
* Our composer.json should explicitly **not** support unknown future PHP versions - `8.1.* || 8.2.* || 8.3.* || 8.4.*`, not `^8.1`.
* We only support symfony versions that are currently maintained or receiving security fixes - _so currently only 5.4, 6.4, and 7.1._
* For other dependencies we only support versions that work with our PHP / symfony versions. _At the moment this only applies to psr/container where we support `^1||^2` - we'd drop v1 as soon as none of our supported symfony versions support it (which is a long way off)._
* Dropping support for a PHP / dependency version is a **minor/feature** release _(obviously unless we are making BC breaks to our own code)_
* Bugfixes will only be released for the latest minor version of behat. The only exception is security-related bugs (unlikely, but...) which we'll fix and release from the oldest possible version so that people using behat with unsupported PHP / symfony versions can still get them.

If agreed, the 3.16.0 release of behat will:

* Drop support for PHP 7.2 - 8.0 inclusive
* Prevent people from using behat with PHP >= 8.5
* Drop support for symfony 4.*
* Drop support for symfony 5.0 - 5.3, 6.0 - 6.3, and 7.0 (all inclusive)

I will submit PRs to update composer.json and CI etc, and also a PR with a version support policy to provide users with clear expectations on what to expect going forward.

### PHP versions

I am aware that PHP versions < 8.1 are still available and receiving critical security fixes through paid support packages / LTS distros. However, I don't think there is any obligation for us to provide non-security support (free) for these platforms. If these users want general bugfixes and new features they will need to upgrade, or backport patches themselves. This is the position they're already in with PHP itself, so I think it's reasonable.

Being explicit about versions in our composer.json will prevent people from running behat on php 8.5 until we have tested it and announced support. That puts a little pressure on us (or maybe not - it might prompt a PR from the first person to try it). But it protects users from the frustration/risk of thinking they can upgrade PHP and then finding unexpected problems.

[...]

PR #1504, opened 2024-10-30
change: Drop support for PHP < 8.1, Symfony < 5.4 and Symfony 6.0 - 6.3

AlexSkrypnyk:
Updated minimum version requirements to match the new version support policy in #1503.

## Changes

1. Drop support for PHP versions before 8.1
2. Drop support for Symfony 4 (EOL), and the non-LTS versions of Symfony 5 and 6 (also EOL)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,13 +18,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: [7.2, 7.3, 7.4, 8.0, 8.1, 8.2, 8.3, 8.4]
+        php: [8.1, 8.2, 8.3, 8.4]
         os: [ubuntu-latest]
         composer-mode: [update]
         symfony-version: ['']
         include:
-          # 7.2 build with lowest dependencies
-          - php: 7.2
+          # Build on the lowest supported PHP with the lowest supported dependencies
+          - php: 8.1
             os: ubuntu-latest
             composer-mode: lowest
             symfony-version: ''
@@ -41,23 +41,16 @@ jobs:
             composer-mode: update
             symfony-version: ''
 
-          # Symfony jobs:
-          - php: 8.1
-            os: ubuntu-latest
-            composer-mode: update
-            symfony-version: '4.4'
+          # Symfony LTS jobs:
           - php: 8.2
             os: ubuntu-latest
             composer-mode: update
             symfony-version: '5.4'
+
           - php: 8.2
             os: ubuntu-latest
             composer-mode: update
-            symfony-version: '6.0'
-          - php: 8.2
-            os: ubuntu-latest
-            composer-mode: update
-            symfony-version: '7.0'
+            symfony-version: '6.4'
 
     steps:
       - uses: actions/checkout@v4
@@ -110,11 +103,19 @@ jobs:
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: 7.2
+          php-version: 8.1
           ini-values: "zend.exception_ignore_args=Off"
           coverage: none
 
+      # until psalm is updated to v5 which is compatible with Symfony 7
+      - name: Force symfony version for psalm compatibility
+        run: |
+          composer config --global --no-plugins allow-plugins.symfony/flex true &&
+          composer global require symfony/flex
+
       - name: Install dependencies
+        env:
+          SYMFONY_REQUIRE: 5.4.*
         run: composer update
 
       - name: Run Psalm
@@ -129,16 +130,21 @@ jobs:
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: "8.3"
+          php-version: "8.2"
           ini-values: "phar.readonly=0,zend.exception_ignore_args=Off"
           coverage: none
           tools: box
 
-      - name: Build the PHAR
+      # We have to force the composer platform because box only runs on ^8.2, but we need to make sure the dependencies
+      # inside the phar support 8.1.0 (our lowest supported PHP version).
+      # When we drop 8.1, we may be able to drop this (if we can still get box to run on 8.2).
+      - name: Force dependencies for lowest supported PHP version
         run: |
-          composer config platform.php 7.2.34
+          composer config platform.php 8.1.0
           composer update --no-dev -o
-          box compile
+
+      - name: Build the PHAR
+        run: box compile
 
       - name: cache artifact
         uses: actions/upload-artifact@v4
@@ -153,7 +159,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: ['7.2', '7.3', '7.4', '8.0', '8.1', '8.2', '8.3', '8.4']
+        php: ['8.1', '8.2', '8.3', '8.4']
     steps:
       - uses: actions/checkout@v4
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -156,6 +156,7 @@ jobs:
         php: ['7.2', '7.3', '7.4', '8.0', '8.1', '8.2', '8.3', '8.4']
     steps:
       - uses: actions/checkout@v4
+
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
@@ -166,18 +167,15 @@ jobs:
       - name: Install dependencies
         run: composer install
 
-
       - uses: actions/download-artifact@v4
         with:
           name: behat.phar
 
       - name: Check content
-        run: |
-          ls -R .
+        run: ls -R .
 
       - name: Test the PHAR
-        run: |
-          php ./behat.phar --version
+        run: php ./behat.phar --version
 
   publish-phar:
     name: Publish the PHAR for release
@@ -189,6 +187,7 @@ jobs:
         with:
           name: behat.phar
           path: .
+
       - name: Upload behat.phar
         uses: basefas/upload-release-asset-action@v1
         env:
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
 
     "require-dev": {
         "symfony/process": "^5.4 || ^6.4 || ^7.0",
-        "phpunit/phpunit": "^8.5 || ^9.0",
+        "phpunit/phpunit": "^9.6",
         "herrera-io/box": "~1.6.1",
         "vimeo/psalm": "^4.8"
     },
--- a/composer.json
+++ b/composer.json
@@ -14,21 +14,21 @@
     ],
 
     "require": {
-        "php": "^7.2 || ^8.0",
+        "php": "8.1.* || 8.2.* || 8.3.* || 8.4.* ",
         "ext-mbstring": "*",
         "behat/gherkin": "^4.10.0",
-        "behat/transliterator": "^1.2",
-        "symfony/console": "^4.4 || ^5.0 || ^6.0 || ^7.0",
-        "symfony/config": "^4.4 || ^5.0 || ^6.0 || ^7.0",
-        "symfony/dependency-injection": "^4.4 || ^5.0 || ^6.0 || ^7.0",
-        "symfony/event-dispatcher": "^4.4 || ^5.0 || ^6.0 || ^7.0",
-        "symfony/translation": "^4.4 || ^5.0 || ^6.0 || ^7.0",
-        "symfony/yaml": "^4.4 || ^5.0 || ^6.0 || ^7.0",
+        "behat/transliterator": "^1.5",
+        "symfony/console": "^5.4 || ^6.4 || ^7.0",
+        "symfony/config": "^5.4 || ^6.4 || ^7.0",
+        "symfony/dependency-injection": "^5.4 || ^6.4 || ^7.0",
+        "symfony/event-dispatcher": "^5.4 || ^6.4 || ^7.0",
+        "symfony/translation": "^5.4 || ^6.4 || ^7.0",
+        "symfony/yaml": "^5.4 || ^6.4 || ^7.0",
         "psr/container": "^1.0 || ^2.0"
     },
 
     "require-dev": {
-        "symfony/process": "^4.4 || ^5.0 || ^6.0 || ^7.0",
+        "symfony/process": "^5.4 || ^6.4 || ^7.0",
         "phpunit/phpunit": "^8.5 || ^9.0",
         "herrera-io/box": "~1.6.1",
         "vimeo/psalm": "^4.8"


ttomdewit:
Thanks for the contribution! I'll give it a proper look tomorrow and will try to verify the EOL policy of the org, which I'm currently unfamiliar with.

AlexSkrypnyk:
This is failing on https://github.com/Behat/Transliterator not being able to run on PHP8+ without warnings.
I can submit a PR into that repo to update the code, but before I do that, could someone please confirm that such update would be accepted and that the maintainers would have the time to make a release in that repo. Thanks

.github/workflows/build.yml
```diff
@@ -42,18 +42,16 @@ jobs:
             symfony-version: ''
 
           # Symfony jobs:
-          - php: 8.1
-            os: ubuntu-latest
-            composer-mode: update
-            symfony-version: '4.4'
```
AlexSkrypnyk, line 48: Symfony 4 dropped, so no need for this

.github/workflows/build.yml
```diff
@@ -110,15 +107,15 @@ jobs:
...
+          php-version: 8.3
           ini-values: "zend.exception_ignore_args=Off"
           coverage: none
 
       - name: Install dependencies
         run: composer update
 
       - name: Run Psalm
-        run: ./vendor/bin/psalm --output-format=github
+        run: ./vendor/bin/psalm --output-format=github || true
```
AlexSkrypnyk, line 118: this is set to not fail CI as psalm has revealed issues in code. I can fix those in this PR or a follow up PR.

.github/workflows/build.yml
```diff
@@ -136,7 +133,7 @@ jobs:
 
       - name: Build the PHAR
         run: |
-          composer config platform.php 7.2.34
+          composer config platform.php 8.2.25
```
AlexSkrypnyk, line 136: mostly used version of the lowest non-EOL version to date

composer.json
```diff
@@ -14,24 +14,24 @@
     ],
 
     "require": {
-        "php": "^7.2 || ^8.0",
+        "php": "^8.0",
```
AlexSkrypnyk, line 17: i think this should be 8.2, but I'm leaving as 8.0 for now

.github/workflows/build.yml
```diff
@@ -76,9 +74,8 @@ jobs:
           composer config --global --no-plugins allow-plugins.symfony/flex true &&
           composer global require symfony/flex
 
-      # until psalm is updated to v5 which is compatible with Symfony 7
-      - name: Remove vimeo/psalm
-        if: matrix.symfony-version == '7.0'
+      - name: Remove vimeo/psalm until it supports PHP 8.4
```
AlexSkrypnyk, line 77: `vimeo/psalm` version 5 is  compatible with Symfony 7, but not compatible with PHP 8.4 :) so have to keep the removal still

.github/workflows/build.yml
```diff
@@ -18,13 +18,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: [7.2, 7.3, 7.4, 8.0, 8.1, 8.2, 8.3, 8.4]
+        php: [8.0, 8.1, 8.2, 8.3, 8.4]
```
acoulton, line 21: Should also drop 8.0
```suggestion
        php: [8.1, 8.2, 8.3, 8.4]
```

.github/workflows/build.yml
```diff
@@ -18,13 +18,13 @@ jobs:
...
         os: [ubuntu-latest]
         composer-mode: [update]
         symfony-version: ['']
         include:
-          # 7.2 build with lowest dependencies
-          - php: 7.2
+          # 8.0 build with the lowest dependencies
+          - php: 8.0
```
acoulton, line 27:
```suggestion
          # 8.1 build with the lowest dependencies
          - php: 8.1
```

.github/workflows/build.yml
acoulton: Please can we keep psalm at v4 in this branch and revert the psalm-related changes? The Psalm upgrade can then happen separately and alongside fixing any issues that v5 is now finding.

.github/workflows/build.yml
acoulton: Also here please revert and we'll upgrade psalm in a separate PR

.github/workflows/build.yml
acoulton: This was only there as a workaround to allow box (which I think requires ^8.2) to run but build with deps that were valid for 7.2.

It would be better to remove this and change the setup-php step to run on 8.2.  As we are now keeping pace with PHP version support, we should always be able to find a version of box that will run on our lowest supported version, so we won't need this workaround in future.

.github/workflows/build.yml
```diff
@@ -153,31 +150,33 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: ['7.2', '7.3', '7.4', '8.0', '8.1', '8.2', '8.3', '8.4']
+        php: ['8.0', '8.1', '8.2', '8.3', '8.4']
```
acoulton, line 153: Drop 8.0

```suggestion
        php: ['8.1', '8.2', '8.3', '8.4']
```

.github/workflows/build.yml
```diff
@@ -153,31 +150,33 @@ jobs:
...
         uses: shivammathur/setup-php@v2
         with:
           php-version: "${{ matrix.php }}"
           ini-values: "phar.readonly=0,zend.exception_ignore_args=Off"
           coverage: none
 
+      - name: Remove vimeo/psalm until it supports PHP 8.4
+        if: matrix.php == '8.4'
+        run: composer remove vimeo/psalm --no-update --dev
+
```
acoulton, line 167: This shouldn't be needed - we're not running psalm in this job?

.github/workflows/build.yml
```diff
@@ -153,31 +150,33 @@ jobs:
...
 
       - name: Check content
-        run: |
-          ls -R .
+        run: ls -R .
 
       - name: Test the PHAR
-        run: |
-          php ./behat.phar --version
+        run: php ./behat.phar --version
```
acoulton, line 179: Please revert this reformatting to keep this commit diff focused on the actual changes. Likewise the added & removed empty lines elsewhere in the file.

composer.json
acoulton: In line with our new policy to be explicit about the exact supported PHP range.
```suggestion
        "php": "8.1.* || 8.2.* || 8.3.* || 8.4.* ",
```

composer.json
```diff
@@ -14,24 +14,24 @@
...
-        "symfony/event-dispatcher": "^4.4 || ^5.0 || ^6.0 || ^7.0",
-        "symfony/translation": "^4.4 || ^5.0 || ^6.0 || ^7.0",
-        "symfony/yaml": "^4.4 || ^5.0 || ^6.0 || ^7.0",
+        "behat/transliterator": "^1.5",
+        "symfony/console": "^5.0 || ^6.0 || ^7.0",
+        "symfony/config": "^5.0 || ^6.0 || ^7.0",
+        "symfony/dependency-injection": "^5.0 || ^6.0 || ^7.0",
+        "symfony/event-dispatcher": "^5.0 || ^6.0 || ^7.0",
+        "symfony/translation": "^5.0 || ^6.0 || ^7.0",
+        "symfony/yaml": "^5.0 || ^6.0 || ^7.0",
```
acoulton, line 26: symfony versions should all be `^5.4 || ^6.4 || ^7.0` - same in the dev dependencies

acoulton (changes requested):
@AlexSkrypnyk thanks - you've saved me a job for tomorrow :) As you've seen in #1503 we've just agreed a new policy on version support so we can make your actions matrix and composer requirements stricter still.

Also, there's not a real agreement on what `chore:` means but it would generally be only for things that don't affect end-users.

We need to make this more obvious - please reword the commit as something like

`change: Drop support for PHP < 8.1, Symfony < 5.4 and Symfony 6.0 - 6.3`

I'm aware `change` isn't in any of the common conventional commits specs but I don't think any of the normal categories cover this kind of change (which is technically a BC break, even though we're relying on composer to manage that for us).

AlexSkrypnyk:
@acoulton
I will action all of the comments shortly. Please leave this with me.

.github/workflows/build.yml
AlexSkrypnyk: I do not see these things as issues as they are improving the overall readability.

.github/workflows/build.yml
AlexSkrypnyk: I would not update psalm if it was possible to avoid such an update.

Psalm 4 requires Symfony <5. Not having it updated in this PR would mean removing it from each CI matrix as it will simply not be able to be installed.

AlexSkrypnyk:
@acoulton
I've actioned the items that you requested that were relevant to my initial intent - improve CI.
I can see that phpunit is now failing in one of the jobs and you have question about downgrading of psalm.

I personally think that this PR should contain all required changes to do the updates in one go instead of splitting them into multiple PRs and keep trying to fix the dependencies issues.

I think it will be best for you to commit any other changes directly into this PR (the branch is open for maintainer's commits).

.github/workflows/build.yml
AlexSkrypnyk: see above

.github/workflows/build.yml
AlexSkrypnyk: it is needed to remove from dev deps or `composer install` will fail

composer.json
```diff
@@ -14,24 +14,24 @@
     ],
 
     "require": {
-        "php": "^7.2 || ^8.0",
+        "php": "8.1.* || 8.2.* || 8.3.* || 8.4.* ",
```
AlexSkrypnyk, line 17: I've added this as per @acoulton's request, but this is not an industry standard.
Normally, this line would be contain the lowest version and above: `php: >= 8.1`
There is nothing special about this project that requires to put an upper boundary to be `8.4`

.github/workflows/build.yml
acoulton: But we must have been installing psalm 4 with the symfony 5 / 6 jobs until now?

https://packagist.org/packages/vimeo/psalm#4.30.0 looks like it only requires `symfony/console:  ^3.4.17 || ^4.1.6 || ^5.0 || ^6.0` unless there's a constraint hiding in one of the other dependencies?

composer.json
acoulton: Nobody knows what will be in PHP 8.5, so we have no idea if behat will work with it. We do know that behat required changes to work on 8.4 (and 8.3 ISTR). If we're going to advertise supported PHP versions, we should be accurate.

acoulton:
@AlexSkrypnyk thanks for the updates. I'll check this out and finish it off tomorrow as you suggest.

composer.json
AlexSkrypnyk: Other projects are in the same boat and they do not declare such dependencies. Adding an upper bound will mean that a code change will be required in the future to allow supporting PHP 8.5. Minor PHP versions should be compatible between each other and only add deprecations. Setting `^8.1` lock the major version where we truly do not know what to expect. But it should be totally okay for minor php versions.

.github/workflows/build.yml
AlexSkrypnyk: let me try to revert this in another PR against this PR and I will report back if it works.

AlexSkrypnyk: psalm was reverted. it now produces a build failure because of the outdated source code. I am not fixing that code in this PR.

AlexSkrypnyk:
@acoulton 
Psalm change was reverted and now produces a build failure due to outdated source code. Leaving this to you to resolve. thanks

composer.json
acoulton: > Other projects are in the same boat and they do not declare such dependencies.

IMO they should. As a user, it is much easier if composer can just tell you when/whether it's worth considering bumping a project to the next PHP than to start the process and then find problems that need to be fixed upstream before you can continue.

> Adding an upper bound will mean that a code change will be required in the future to allow supporting PHP 8.5. 

Not a code change. 

Best case, once a year we will need to add a version to the build matrix and a version to the composer.json - a very easy task, and an easy contribution for someone. If it goes green, fine. If it fails, then we know what needs fixing before it's worth people running it in their own projects.

> Minor PHP versions should be compatible between each other and only add deprecations. Setting ^8.1 lock the major version where we truly do not know what to expect.

Every PHP minor has [backwards-incompatible changes](https://www.php.net/manual/en/migration83.incompatible.php) (as well as deprecations). They're often edge cases, but there's no way to know in advance which might affect us.

Also behat - unlike some other projects - explicitly sets and runs an error_reporting level including E_DEPRECATED, so the deprecations will also break a behat build. People might have opinions on that, but changing it would be a BC break so for the 3.x series at least that will always be the case.

> But it should be totally okay for minor php versions.

The historical evidence is that assumption is false : as you can see in the CHANGELOG, PHP 8.2 was the only minor version behat was able to support without code changes.

.github/workflows/build.yml
```diff
@@ -42,18 +42,16 @@ jobs:
...
-            symfony-version: '4.4'
           - php: 8.2
             os: ubuntu-latest
             composer-mode: update
             symfony-version: '5.4'
+
           - php: 8.2
             os: ubuntu-latest
             composer-mode: update
             symfony-version: '6.0'
```
stof, line 53: This should be a job using `6.4`, not `6.0`. `6.0` is not a LTS version (and not receiving fixes anymore)

stof: btw, we should not have a job for `7.0` either, as it is not a LTS either.

.github/workflows/build.yml
```diff
@@ -129,16 +127,13 @@ jobs:
...
         with:
-          php-version: "8.3"
+          php-version: "8.2"
           ini-values: "phar.readonly=0,zend.exception_ignore_args=Off"
           coverage: none
           tools: box
 
       - name: Build the PHAR
-        run: |
-          composer config platform.php 7.2.34
```
stof, line 139: Resolving dependencies on PHP 8.2 without a platform config ensuring compat with PHP 8.1 is a no-go, as it would not guarantee that the phar is compatible with PHP 8.1 (it would likely install Symfony 7 which requires PHP 8.2)

composer.json
```diff
@@ -14,21 +14,21 @@
...
+        "symfony/event-dispatcher": "^5.4 || ^6.4 || ^7.0",
+        "symfony/translation": "^5.4 || ^6.4 || ^7.0",
+        "symfony/yaml": "^5.4 || ^6.4 || ^7.0",
         "psr/container": "^1.0 || ^2.0"
     },
 
     "require-dev": {
-        "symfony/process": "^4.4 || ^5.0 || ^6.0 || ^7.0",
+        "symfony/process": "^5.4 || ^6.4 || ^7.0",
         "phpunit/phpunit": "^8.5 || ^9.0",
```
stof, line 32: We can simplify this. We don't need to support phpunit 8 in our testsuite anymore if we don't support older PHP versions anymore.

acoulton: @stof I have that locally already

.github/workflows/build.yml
acoulton: That's my bad, did the initial review on my phone and forgot box needs 8.2 and we're supporting 8.1 still. Will fix.

AlexSkrypnyk:
@acoulton 
I see you force-pushed to my branch and did not leave any credit for me. Can you please update your commits to list me as an author on at least one of them. Thanks

acoulton:
@AlexSkrypnyk good spot, apologies that was accidental. I rebased to squash up the changes and update onto latest master and I thought git would leave you as author & me as committer as it usually does. It may be because I edited a couple of times during the rebase. I should have checked - I was rushing a bit. I'll fix that now.

acoulton:
@AlexSkrypnyk please can you confirm the credit looks OK to you now? The overall PR will also be listed as yours in the changelog / release notes.

AlexSkrypnyk:
@acoulton 
Thank you so much for fixing the credit. Sorry for the extra noise.

I'm super happy that we are moving Behat into the future!

acoulton:
@AlexSkrypnyk no apology needed, I'm glad you flagged my mistake.

carlos-granados (approved):
Looks great, thanks @AlexSkrypnyk and @acoulton

Issue #1507, opened 2024-11-01
Latest phar is still broken


carlos-granados:
When you try to use it you get a message saying `Uncaught PharException: manifest cannot be larger than 100 MB`, same thing happens with the phar from version 3.14. Version 3.13 works fine. So it seems we did not manage to fix the issues with the phar file

carlos-granados:
The phar file built for the latest master and which can be found here https://github.com/Behat/Behat/actions/runs/11627410394/artifacts/2132427642 
works fine

carlos-granados:
Actually the phar file built when the 3.15.0 tag was created also works fine:
https://github.com/Behat/Behat/actions/runs/11589077326/artifacts/2121684168
It is like the phar publish action is the one working incorrectly 🤔

carlos-granados:
For now until we have been able to fix this issue I have manually updated the release with the working phar file for v3.15.0

PR #1509, opened 2024-11-01
Fix phar upload for the release

carlos-granados:
The phar file uploaded for the 3.15.0 release did not work correctly, it threw an error when you tried to use it. Using my fork to perform some checks I confirmed that the problem happens in the job that uploads the created phar to the release which somehow seems to bork it. To fix this, I have changed our actions code to use a different "upload to release" action. Testing with my fork, it seems to work fine.

(The new action seems to be much more popular, it has more than 600 GH stars while the one we were using previously only has one)

Fixes https://github.com/Behat/Behat/issues/1507

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -195,10 +195,9 @@ jobs:
           path: .
 
       - name: Upload behat.phar
-        uses: basefas/upload-release-asset-action@v1
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        uses: svenstaro/upload-release-action@v2
         with:
-          release_id: ${{ github.event.release.id }}
-          asset_path: behat.phar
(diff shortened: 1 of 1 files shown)


acoulton (approved):
Thanks for sorting this @carlos-granados 

I see the new action [sets an explicit ](https://github.com/svenstaro/upload-release-action/blob/84f3aed82331a2c0c665cf8e9707228f766aba9d/src/main.ts#L169-L170) `content-type` and `content-length` header [but the old one doesn't](https://github.com/basefas/upload-release-asset-action/blob/844c1f19487cc073073fceede9bdd3f2fd08d4de/src/main.ts#L31-L37) so presumably it/github was guessing at a content type. I wonder if the file was getting mangled there for some reason. Alternatively, it could be to do with how the different actions read and pass the content of the phar - they're passing slightly different types into the octokit upload method.

Anyway, like you say the action being more popular is also a benefit. Let's see how we get on using the new publish action on the next release.

PR #1512, opened 2024-11-03
handle --strict option with stop_on_failure config

jdeniau:
With the new config parameter, the "strict" mode was not handled correctly : 

The default mode is a `SoftInterpretation` and is "strictified" with the `--stict` cli parameter.

The difference between the two interpretation is how we handled the undefined steps: SoftInterpretation does not fail if a step is missing, but do fail otherwise.

https://github.com/Behat/Behat/pull/1501/files wrongly set the default interpreter to the `Strict` one, and even if the right interpreter was set with the cli parameters, it was not with the `config:` one.

This PR does fix that, and it will be followed by another one implementing `strict` in the config.

### DIFF
--- a/features/stop_on_failure_config.feature
+++ b/features/stop_on_failure_config.feature
@@ -57,6 +57,24 @@ Feature: Stop on failure via config
            And I have another step that passes
           Then I should have a scenario that passed
       """
+    And a file named "features/missing-step.feature" with:
+      """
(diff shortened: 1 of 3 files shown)


acoulton (approved):
Oops... Good catch, thanks for adding the additional test coverage.

carlos-granados (approved):
Thanks @jdeniau

PR #1513, opened 2024-11-03
Handle "strict" in behat configuration file

jdeniau:
Follows https://github.com/Behat/Behat/pull/1501 and #1512 

Implement the fact that we can set the `strict` settings in the behat.yml in `config:` section.

### DIFF
--- a/src/Behat/Behat/Config/Handler/StrictHandler.php
+++ b/src/Behat/Behat/Config/Handler/StrictHandler.php
@@ -18,14 +18,9 @@ use Behat\Testwork\Tester\Result\ResultInterpreter;
  */
 final class StrictHandler
 {
-    /**
-     * @var ResultInterpreter
-     */
-    private $resultInterpreter;
-
-    public function __construct(ResultInterpreter $resultInterpreter)
-    {
-        $this->resultInterpreter = $resultInterpreter;
+    public function __construct(
+        private readonly ResultInterpreter $resultInterpreter
+    ) {
     }
 
     public function registerStrictInterpretation()
--- a/src/Behat/Behat/Config/Handler/StrictHandler.php
+++ b/src/Behat/Behat/Config/Handler/StrictHandler.php
@@ -1,0 +1,35 @@
+<?php
+
+/*
+ * This file is part of the Behat Testwork.
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Behat\Behat\Config\Handler;
+
+use Behat\Testwork\Tester\Result\Interpretation\StrictInterpretation;
+use Behat\Testwork\Tester\Result\ResultInterpreter;
+
+/**
+ * Enables strict mode via config.
+ */
+final class StrictHandler
+{
+    /**
+     * @var ResultInterpreter
+     */
+    private $resultInterpreter;
+
+    public function __construct(ResultInterpreter $resultInterpreter)
+    {
+        $this->resultInterpreter = $resultInterpreter;
+    }
+
+    public function registerStrictInterpretation()
+    {
+        $this->resultInterpreter->registerResultInterpretation(new StrictInterpretation());
+    }
+}
--- a/src/Behat/Testwork/Tester/Cli/StrictController.php
+++ b/src/Behat/Testwork/Tester/Cli/StrictController.php
@@ -10,6 +10,7 @@
 
 namespace Behat\Testwork\Tester\Cli;
 
+use Behat\Behat\Config\Handler\StrictHandler;
 use Behat\Testwork\Cli\Controller;
 use Behat\Testwork\Tester\Result\Interpretation\StrictInterpretation;
 use Behat\Testwork\Tester\Result\ResultInterpreter;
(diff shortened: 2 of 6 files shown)


src/Behat/Behat/Config/Handler/StrictHandler.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    /**
+     * @var ResultInterpreter
+     */
+    private $resultInterpreter;
+
+    public function __construct(ResultInterpreter $resultInterpreter)
+    {
+        $this->resultInterpreter = $resultInterpreter;
+    }
```
acoulton, line 29: Now that we only support PHP ^8.1, we should probably use typed properties / constructor promotion / readonly properties in new code? (we can't yet use readonly classes as they are 8.2 only).

```suggestion

    public function __construct(
        private readonly ResultInterpreter $resultInterpreter,
    ) {}
```

src/Behat/Testwork/Tester/Cli/StrictController.php
```diff
@@ -34,6 +36,11 @@ final class StrictController implements Controller
      */
     private $strict;
 
+    /**
+     * @var StrictHandler
+     */
+    private $strictHandler;
+
```
acoulton, line 43: As above, I'd suggest using strict types for new code (but leave the existing props with `@var` declarations for now - modernising old code is a separate concern IMO).

```suggestion
    private StrictHandler $strictHandler;
```

acoulton (commented):
LGTM, just a couple of suggestions on using new PHP features now that we can, but maybe wait to see if @carlos-granados and @ttomdewit agree with me before applying them as this would be the start of a new convention.

acoulton:
It would be great if you could also submit a PR to add the stop on failure & strict config options to https://github.com/Behat/docs/blob/v3.0/user_guide/configuration.rst? Thanks

jdeniau:
> It would be great if you could also submit a PR to add the stop on failure & strict config options to [Behat/docs@v3.0/user_guide/configuration.rst](https://github.com/Behat/docs/blob/v3.0/user_guide/configuration.rst?rgh-link-date=2024-11-04T08%3A14%3A50Z)? Thanks

@acoulton yes, it's on my mind, I just wanted to add `strict` mode first to avoid making two different PR on the documentation 👍

src/Behat/Behat/Config/Handler/StrictHandler.php
jdeniau: I was wondering what was the policy on that kind of feature.

If you did drop older version of PHP, that's great, and I can of course migrate typing. You might also consider migrate all codebase to typed property using rectorphp and this kind of set: https://getrector.com/find-rule?rectorSet=core-type-declarations&activeRectorSetGroup=core (but if you consider that it might be breaking, it may be only on the next major).

Another "new" php feature I was wondering is the `array('foo')` vs `['foo']`. Behat does use `array` everywhere, but I think that the standard is more `['foo`]` now (but I kept `array` to be consistent too).

jdeniau:
@acoulton I did change to more modern PHP code. I converted the the StopOnFailure handler from #1501 as it has not been released in a version for now.

src/Behat/Behat/Config/Handler/StrictHandler.php
```diff
@@ -0,0 +1,30 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Behat\Behat\Config\Handler;
+
+use Behat\Testwork\Tester\Result\Interpretation\StrictInterpretation;
+use Behat\Testwork\Tester\Result\ResultInterpreter;
+
+/**
+ * Enables stop on failure via config.
```
carlos-granados, line 17: Should be something like "Enables strict via config"

carlos-granados (changes requested):
Just a small change needed, otherwise looking good

carlos-granados (approved):
Looking good now, thanks @jdeniau

jdeniau:
> It would be great if you could also submit a PR to add the stop on failure & strict config options to [Behat/docs@`v3.0`/user_guide/configuration.rst](https://github.com/Behat/docs/blob/v3.0/user_guide/configuration.rst?rgh-link-date=2024-11-04T08%3A14%3A50Z)? Thanks

https://github.com/Behat/docs/pull/155

carlos-granados:
@jdeniau I took a new look at your PR and thought of something: instead of handling the strict flag and config setting in the stop on failure controller and handler, why not just pass the result interpreter to the handler and use it to check the interpretation? This would provide a better separation of concerns as all the handling of the strict flag and config setting would be just in the strict controller and handler. What do you think?

jdeniau:
@carlos-granados Well I am not really confident with `ResultInterpreter` because it does register multiple interpretations, and because it does ["Interprets result as a UNIX return code (0 for success, 1 for failure)."](https://github.com/jdeniau/Behat/blob/be2b1432969dce440ba67f7a7aa451f01a05292c/src/Behat/Testwork/Tester/Result/ResultInterpreter.php#L38) : it seems less readable to check `=== 0` instead of `=== true`.

I will try an implementation in another PR 🤞

jdeniau:
@carlos-granados You were right, this seems to be a way better implementation: https://github.com/Behat/Behat/pull/1516

ttomdewit:
Thanks for the contribution!

Please do not wait on my review and/or approval, I haven't been able to dig deep enough to fully understand what is going on. I trust @acoulton @carlos-granados when they approve this.

acoulton:
Closed in favour of #1516

Issue #1491, opened 2024-10-26
Feature: Explicit logger for sub-step debugging


ciaranmcnulty:
A subject that's been discussed a few times is whether we can add 'debug' information during steps

That is, within a step like 'When I create an Invoice' it could be useful to see what actual actions have been executed.

For instance browser automation might want to report something like:
```
 - Went to /invoices
 - Clicked 'new invoice' button
 - etc
```

Equally a Step Definition that just calls methods could still want some debug
```
 - Called createInvoice with parameters <some var_dump output>
 - Got ID ABC1234 for the invoice
```

A simple way to do this would be to provide some sort of Logger object (PSR?) that is available in the container, which could be injected by default into Contexts

User-defined steps could explicitly call `$this->logger->log("Got ID $id")`, while helpers like those in the MinkExtension could hook into the same Logger object at a deeper level.

I'd expect this to be hidden in the UI behind either some new flag, or a verbosity setting.

ciaranmcnulty:
If we actually implemented something like this we should/could then deprecate+remove stdout output from steps

acoulton:
As noted in #1471 this could be very useful, and the PSR logger is both familiar and could allow users to filter what gets displayed depending on the logevel which is a big improvement on STDOUT. But users should also be in charge of whether the output comes in realtime or is buffered to the end of the build.

acoulton:
While this could still be considered for the future, in the meantime @carlos-granados has now added a config option to the Pretty & Progress printers in #1576 that allows the user to control whether / when STDOUT from steps will be displayed.

PR #1515, opened 2024-11-04
test: Improve output when behat's own tests pass or fail unexpectedly

jdeniau:
### Current

When the expected "pass" or "fail" result is wrong, it's quite difficult to understand what is going on.

Imagine a step that should fail but is actually passing, the output is the following:

```gherkin
    Then it should pass with:                                                                                        # FeatureContext::itShouldPassWith()
      """
      Feature: Failing Feature
        In order to test the stop-on-failure feature
        As a behat developer
        I need to have a feature that fails
      
        Background:
          Given I have a step that passes
      
        Scenario: 1st Passing
          When I have a step that passes
          Then I should have a scenario that passed
      
        Scenario: 2nd Passing
          When I have a step that passes
          And I have another step that passes
          Then I should have a scenario that passed
      
        Scenario: 1st Failing
          When I have a step that passes
          And I have another step that fails
            step failed as supposed (Exception)
          Then I should have a scenario that failed
      
      --- Failed scenarios:
      
          features/failing.feature:18
      
      3 scenarios (2 passed, 1 failed)
      11 steps (9 passed, 1 failed, 1 skipped)
      """
      Actual output:
      
      Feature: Failing Feature
        In order to test the stop-on-failure feature
        As a behat developer
        I need to have a feature that fails
      
        Background:
          Given I have a step that passes
      
        Scenario: 1st Passing
          When I have a step that passes
          Then I should have a scenario that passed
      
        Scenario: 2nd Passing
          When I have a step that passes
          And I have another step that passes
          Then I should have a scenario that passed
      
        Scenario: 1st Failing
          When I have a step that passes
          And I have another step that fails
            step failed as supposed (Exception)
          Then I should have a scenario that failed
      
      --- Failed scenarios:
      
          features/failing.feature:18
      
      3 scenarios (2 passed, 1 failed)
      11 steps (9 passed, 1 failed, 1 skipped)
      Failed asserting that 1 matches expected 0.

--- Scénarios échoués:

    features/stop_on_failure.feature:114
```

The `Actual Output` is exactly the same as the expected one, but the fact is that we do test two things:
- the actual output
- the fact that the step is passing or failing !

When I did implement my latest PR, I really got stucked with that output, doing manual diff, trying to clear some cache etc.

## Proposed improvement

I propose to first test the "fail" / "pass" status, before testing the content.

This way the ouput will be the following:

```gherkin
    Then it should pass with:                                                                                        # FeatureContext::itShouldPassWith()
      """
      Feature: Failing Feature
        In order to test the stop-on-failure feature
        As a behat developer
        I need to have a feature that fails
      
        Background:
          Given I have a step that passes
      
        Scenario: 1st Passing
          When I have a step that passes
          Then I should have a scenario that passed
      
        Scenario: 2nd Passing
          When I have a step that passes
          And I have another step that passes
          Then I should have a scenario that passed
      
        Scenario: 1st Failing
          When I have a step that passes
          And I have another step that fails
            step failed as supposed (Exception)
          Then I should have a scenario that failed
      
      --- Failed scenarios:
      
          features/failing.feature:18
      
      3 scenarios (2 passed, 1 failed)
      11 steps (9 passed, 1 failed, 1 skipped)
      """
      Step should PASS but it FAIL.
      Failed asserting that false is true.

--- Scénarios échoués:

    features/stop_on_failure.feature:114
```

In this case, we now understand that the step is passing when we expect it to fail. If the content is not the same, it will be displayed only if the `$success` value is in a correct state.

Here is a diff of the two output to understand easily what did change in the following outputs:

image

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -263,13 +263,12 @@ EOL;
      *
      * @Then /^it should (fail|pass) with:$/
      *
-     * @param string       $success "fail" or "pass"
-     * @param PyStringNode $text    PyString text instance
+     * @param 'pass'|'fail' $success
      */
-    public function itShouldPassWith($success, PyStringNode $text)
+    public function itShouldPassOrFailWith($success, PyStringNode $text)
     {
         $this->theOutputShouldContain($text);
-        $this->itShouldFail($success);
+        $this->itShouldPassOrFail($success);
     }
 
     /**
@@ -277,12 +276,12 @@ EOL;
      *
      * @Then /^it should (fail|pass) with no output$/
      *
-     * @param string $success "fail" or "pass"
+     * @param 'pass'|'fail' $success
      */
-    public function itShouldPassWithNoOutput($success)
+    public function itShouldPassOrFailWithNoOutput($success)
     {
         Assert::assertEmpty($this->getOutput());
-        $this->itShouldFail($success);
+        $this->itShouldPassOrFail($success);
     }
 
     /**
@@ -390,30 +389,26 @@ EOL;
      *
      * @Then /^it should (fail|pass)$/
      *
-     * @param string $success "fail" or "pass"
+     * @param 'pass'|'fail' $success  
      */
-    public function itShouldFail($success)
+    public function itShouldPassOrFail($success)
     {
-        $message = '';
-
-        if ('fail' === $success) {
-            Assert::assertNotEquals(0, $this->getExitCode(), 'Step should FAIL but it PASS.');
-
-            if (0 === $this->getExitCode()) {
-                $message = 'Actual output:' . PHP_EOL . PHP_EOL . $this->getOutput();
-            }
-
-            Assert::assertNotEquals(0, $this->getExitCode(), $message);
-        } else {
-            Assert::assertEquals(0, $this->getExitCode(), 'Step should PASS but it FAIL.');
-
-
-            if (0 !== $this->getExitCode()) {
-                $message = 'Actual output:' . PHP_EOL . PHP_EOL . $this->getOutput();
-            }
-
-            Assert::assertEquals(0, $this->getExitCode(), $message);
-        }
+        $is_correct = match($success) {
+            'fail' => 0 !== $this->getExitCode(),
+            'pass' => 0 === $this->getExitCode(),
+       };
+     
+       if ($is_correct) {
+          return;
+       }
+    
+       throw new UnexpectedValueException(
+            'Expected previous command to ' . strtoupper($success) . ' but got exit code ' . $this->getExitCode() . PHP_EOL
+            . PHP_EOL
+            . '##### Actual output #####' . PHP_EOL
+            . '> ' . str_replace(PHP_EOL, PHP_EOL  .  '> ', $this->getOutput()) . PHP_EOL
+            . '##### End of output #####'
+        );
     }
 
     /**
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -397,12 +397,17 @@ EOL;
         $message = '';
 
         if ('fail' === $success) {
+            Assert::assertNotEquals(0, $this->getExitCode(), 'Step should FAIL but it PASS.');
+
             if (0 === $this->getExitCode()) {
                 $message = 'Actual output:' . PHP_EOL . PHP_EOL . $this->getOutput();
             }
 
             Assert::assertNotEquals(0, $this->getExitCode(), $message);
         } else {
+            Assert::assertEquals(0, $this->getExitCode(), 'Step should PASS but it FAIL.');
+
+
             if (0 !== $this->getExitCode()) {
                 $message = 'Actual output:' . PHP_EOL . PHP_EOL . $this->getOutput();
             }


features/bootstrap/FeatureContext.php
```diff
@@ -397,12 +397,17 @@ public function itShouldFail($success)
...
             }
 
             Assert::assertNotEquals(0, $this->getExitCode(), $message);
         } else {
+            Assert::assertEquals(0, $this->getExitCode(), 'Step should PASS but it FAIL.');
+
+
             if (0 !== $this->getExitCode()) {
                 $message = 'Actual output:' . PHP_EOL . PHP_EOL . $this->getOutput();
             }
```
acoulton, line 413: So this was changed quite recently in #1487 by @uuf6429 as with the previous behaviour it was hard to see *why* the scenario had failed without including the test output. I think this will reintroduce that problem.

I think a better solution might be to just throw our own exception here so that we can render the message ourselves in a more logical sequence.

Perhaps something like (haven't tested this):

```php
public function itShouldFail($success)
{
    $is_correct = match($success) {
        'fail' => 0 !== $this->getExitCode(),
        'pass' => 0 === $this->getExitCode(),
   }
 
   if ($is_correct) {
    return;
   }

   throw new UnexpectedValueException(
      'Expected previous command to '.$success.' but got exit code '.$this->getExitCode().PHP_EOL
      .PHP_EOL
      .'Actual command output: '.PHP_EOL
      .$this->getOutput()
    );
}
```

acoulton (changes requested):

features/bootstrap/FeatureContext.php
acoulton: Would need to look at the features in detail as explicitly asserting the output could be neater but my hunch is there are times when only a partial match on the output can be used, so having something that captures the full output could still be useful if the scenario fails unexpectedly (less so if it passes unexpectedly, probably).

Behat has no concept of Failed Vs Errored steps, and no specific domain for exceptions - it can't, because it relies entirely on the user supplying an assertion library (if they want one).
 https://docs.behat.org/en/latest/user_guide/context/definitions.html#failed-steps

carlos-granados: We could look into modifying what we do with the output but for now this solution would provide a more informative output, so I would probably just go ahead with it, either as it is or using the change proposed by @acoulton . Just one more thing, I would probably change the name of the function to `itShouldPassOrFail()` so that it is clearer that it's testing both possibilities

uuf6429: That is, the proposed change could be simplified to just the following:
```php

    /**
     * Checks whether previously ran command failed|passed.
     *
     * @Then /^it should (fail|pass)$/
     *
     * @param 'fail'|'pass' $passOrFail
     */
    public function itShouldPassOrFail($passOrFail)
    {
        if ('fail' === $success) {
            Assert::assertNotEquals(0, $this->getExitCode(), 'Command should fail, but it succeeded.');
        } else {
            Assert::assertEquals(0, $this->getExitCode(), 'Command should succeed, but it failed.');
        }
    }
```
(I also fixed a few things, you're welcome. 😄)

stof: I would actually implement those as 2 separate steps instead of this `passOfFail` argument that switches the whole body of the method anyway

acoulton: Personally I would prefer to keep the output for debugging regardless of whether previous steps have done an assertion on partial output matches. There is nothing worse than getting a `it should have worked but it didn't` failure with nothing else to go on, IMO. Especially as in my experience that often happens either flickering or during dependency/platform updates.

If we had #1491 already this would be a good usecase for it. As we don't, I think we should just make sure the failure message is clear & first and the debug output follows it.

> I would actually implement those as 2 separate steps instead of this passOfFail argument that switches the whole body of the method anyway

Me too, but bear in mind this step is also called from other steps that follow the same pattern so those would have to be refactored too.

acoulton: I agree which is why I requested changes to this PR and suggested we just throw an exception ourselves so the output is clear.

features/bootstrap/FeatureContext.php
```diff
@@ -394,21 +394,23 @@ private function getExpectedOutput(PyStringNode $expectedText)
      */
     public function itShouldFail($success)
```
carlos-granados, line 395: I would change the name of the function to `itShouldPassOrFail` to make it clear that it manages both cases

features/bootstrap/FeatureContext.php
jdeniau: @uuf6429 @acoulton  I changed the output to be 

```
    When I run "behat --no-colors --format-settings='{\"paths\": false}' features/failing.feature" # FeatureContext::iRunBehat()
    Then it should pass with:                                                                      # FeatureContext::itShouldPassWith()
      """
      --- Failed scenarios:
      
          features/failing.feature:13
          features/failing.feature:18
      
      4 scenarios (2 passed, 2 failed)
      14 steps (10 passed, 2 failed, 2 skipped)
      """
      Expected previous command to PASS but got exit code 1
      
      ##### Actual output #####
      > Feature: Failing Feature
      >   In order to test the stop-on-failure feature
      >   As a behat developer
      >   I need to have a feature that fails
      > 
      >   Background:
      >     Given I have a step that passes
      > 
      >   Scenario: 1st Passing
      >     When I have a step that passes
      >     Then I should have a scenario that passed
      > 
      >   Scenario: 1st Failing
      >     When I have a step that passes
      >     And I have another step that fails
      >       step failed as supposed (Exception)
      >     Then I should have a scenario that failed
      > 
      >   Scenario: 2nd Failing
      >     When I have a step that fails
      >       step failed as supposed (Exception)
      >     Then I should have a scenario that failed
      > 
      >   Scenario: 2nd Passing
      >     When I have a step that passes
      >     And I have another step that passes
      >     Then I should have a scenario that passed
      > 
      > --- Failed scenarios:
      > 
      >     features/failing.feature:13
      >     features/failing.feature:18
      > 
      > 4 scenarios (2 passed, 2 failed)
      > 14 steps (10 passed, 2 failed, 2 skipped)
      ##### End of output ##### (UnexpectedValueException)
```

It's the best scenario I think, but I saw that there is two different steps `itShouldFail` and `itShouldFailWith`. If the output is not right, then the output does show only the diff in the output.

carlos-granados (approved):
Small suggestion, looking good otherwise

features/bootstrap/FeatureContext.php
```diff
@@ -394,21 +394,23 @@ private function getExpectedOutput(PyStringNode $expectedText)
...
+        return;
+       }
+    
+       throw new UnexpectedValueException(
+          'Expected previous command to '.strtoupper($success).' but got exit code '.$this->getExitCode().PHP_EOL
+          .PHP_EOL
+          .'##### Actual output #####'.PHP_EOL
+          .'> '.str_replace(PHP_EOL, PHP_EOL . '> ', $this->getOutput()).PHP_EOL
+          .'##### End of output #####'
+        );
```
uuf6429, line 413: Looks good, but please fix code style - indents are 4 spaces and there should be a space around concat (period) operator:
```php
        $is_correct = match ($success) {
            'fail' => 0 !== $this->getExitCode(),
            'pass' => 0 === $this->getExitCode(),
        };

        if ($is_correct) {
            return;
        }

        throw new UnexpectedValueException(
            'Expected previous command to ' . strtoupper($success) . ' but got exit code ' . $this->getExitCode() . PHP_EOL
            . PHP_EOL
            . '##### Actual output #####' . PHP_EOL
            . '> ' . str_replace(PHP_EOL, PHP_EOL . '> ', $this->getOutput()) . PHP_EOL
            . '##### End of output #####'
        );
```

features/bootstrap/FeatureContext.php
uuf6429: I would fix that whole part, like so:
```php
    /**
     * Checks whether previously ran command passed or failed.
     *
     * @Then /^it should (fail|pass)$/
     *
     * @param 'pass'|'fail' $success
     */
    public function itShouldPassOrFail(string $success): void
```

features/bootstrap/FeatureContext.php
jdeniau: Arg, so accustomed to use prettier-php that fixes automatically all this boring code style stuff 😄

carlos-granados: We are hoping to have a coding standard tool available here soon 😉

features/bootstrap/FeatureContext.php
```diff
@@ -263,26 +263,26 @@ public function iRunBehatInDebugMode()
      *
      * @Then /^it should (fail|pass) with:$/
      *
-     * @param string       $success "fail" or "pass"
-     * @param PyStringNode $text    PyString text instance
+     * @param 'pass'|'fail' $success
+     * @param PyStringNode  $text    PyString text instance
```
carlos-granados, line 267: I don't think this extra space is needed...

features/bootstrap/FeatureContext.php
```diff
@@ -390,25 +390,26 @@ private function getExpectedOutput(PyStringNode $expectedText)
...
-
-            Assert::assertEquals(0, $this->getExitCode(), $message);
-        }
+        $is_correct = match($success) {
+            'fail' => 0 !== $this->getExitCode(),
+            'pass' => 0 === $this->getExitCode(),
+       };
+     
+       if ($is_correct) {
+        return;
```
carlos-granados, line 403: Still need to fix the indentation here

carlos-granados (changes requested):

features/bootstrap/FeatureContext.php
jdeniau: I followed the rest of the file (vertical alignment of the types, the variable names and the description)

uuf6429: Actually I would remover that entire line, it does not provide any added benefit.

carlos-granados: @uuf6429 you are right, the information is all redundant

carlos-granados (approved):
Looking good now

acoulton (approved):
Great, thanks for this.

PR #1516, opened 2024-11-04
feature: make new `stop_on_failure` config consistent with `strict`

jdeniau:
Following @carlos-granados advice, I try to implement the same PR as #1513 but using the global `ResultInterpreter` service instead of custom definitions everywhere.


acoulton:
@jdeniau I agree this looks cleaner, thanks for taking the time to try the other approach.

However, while reviewing how this ties in with the existing code, I found this in Behat/Testwork/Tester/ServiceContainer/TesterExtension https://github.com/Behat/Behat/blob/c46aae716b165a24e2d7d9a6b86b0e95223a3f11/src/Behat/Testwork/Tester/ServiceContainer/TesterExtension.php#L86-L89

It seems there is **already** an undocumented option to set `strict` in the config file, under the `testers` key - I've just checked and that is working as expected...

jdeniau:
> It seems there is **already** an undocumented option to set `strict` in the config file, under the `testers` key - I've just checked and that is working as expected...

@acoulton Yes I did saw that, but to be honest, I do not know what the TestWork workspace does and the `testers` key is undocumented, so I just ignored this (as it was ignored with the cli arguments).

There is also a `skip` configuration that seems to do exacly the same as the `--dry-run` cli parameter : https://github.com/jdeniau/Behat/blob/2de69c64d84117b801fa66a13bc421fed063eb47/src/Behat/Testwork/Tester/Cli/ExerciseController.php#L145

Do you think that we should move the `config:` (and the `stop_on_failure`) to the TesterExtension where it seems to implement the exact same things that I wanted to implement (cli parameter in configuration file)?

jdeniau:
No problem for the time I spent. I prefer taking some more time and beeing consistant and have a mere simple codebase.
I will open a third PR replacing those two to rework both options according to this PR logic.

src/Behat/Behat/EventDispatcher/Cli/StopOnFailureController.php
```diff
@@ -32,6 +32,8 @@
 
 /**
  * Stops tests on first scenario failure.
+ * 
+ * TODO this should be moved in the Behat\Testwork\Tester\Cli namespace
```
jdeniau, line 36: I added this comment as the --stop-on-failure is related to the tester extension, and not the EventDispatcher.

To avoid BC, we might move the class in the Tester namespace and this one could extend the new class and be marked as deprecated (but it might be done in another PR)

src/Behat/Testwork/Tester/ServiceContainer/TesterExtension.php
```diff
@@ -159,14 +181,21 @@ protected function loadStrictController(ContainerBuilder $container, $strict = f
...
         $container->setDefinition(self::RESULT_INTERPRETER_ID, $definition);
 
         $definition = new Definition('Behat\Testwork\Tester\Result\Interpretation\SoftInterpretation');
         $definition->addTag(self::RESULT_INTERPRETATION_TAG);
         $container->setDefinition(self::RESULT_INTERPRETATION_TAG . '.soft', $definition);
+
+        
+        if ($strict) {
+            $definition = new Definition('Behat\Testwork\Tester\Result\Interpretation\StrictInterpretation');
+            $definition->addTag(TesterExtension::RESULT_INTERPRETATION_TAG);
```
jdeniau, line 196: service with RESULT_INTERPRETATION_TAG are automatically registered in the result interpreter, so we do not need the StrictHandler anymore

jdeniau:
@carlos-granados @acoulton I finally made another commit on that PR that migrate stop-on-failure into the TesterExtension.

The good news is that the tests do pass by just renaming the key in behat.yml file 🤞

src/Behat/Testwork/Tester/Cli/StrictController.php
```diff
@@ -29,21 +29,16 @@ final class StrictController implements Controller
...
      * Initializes controller.
      *
      * @param ResultInterpreter $resultInterpreter
      * @param bool           $strict
      */
-    public function __construct(ResultInterpreter $resultInterpreter, $strict = false)
+    public function __construct(ResultInterpreter $resultInterpreter)
     {
         $this->resultInterpreter = $resultInterpreter;
-        $this->strict = $strict;
```
jdeniau, line 46: I removed the "strict" injection, and moved the implementation in the TesterExtension (with a service that is tagged with `RESULT_INTERPRETATION_TAG`), which seems to be a better implementation. This way we avoid having core logic into the command domain.

The `ExerciseController` (that handle the `dry-run`) is not now migrated though. It can be done later if you think that it's a good vision. The other possibility is to let the Controller code "as-is", but in that case I would suggest to move the StopOnFailureController into the testwork namespace and remove the StopOnFailureHandler then (and keep the implementation in the controller).

acoulton: I'm not sure about this.

If it was new code, I prefer it the way you have done it. But it introduces another BC break - for example if someone is decorating/wrapping the StrictController and expecting it to respect the `$strict` parameter.

IMO keeping the controller as-is is not a huge downside for the benefit of minimising the risk of breaking someone's existing usage.

I agree for purity then it would make sense to rework the StopOnFailureController / StopOnFailureHandler to follow the same pattern - although I could also live with those being inconsistent if we made a note to reapply these changes for the next major.

acoulton (commented):
@jdeniau thanks for continuing this work. I've looked at the diff here and also at the overall comparison from https://github.com/Behat/Behat/compare/v3.15.0...abe71a1 to get a clear picture of what is actually changed since the last release.

src/Behat/Testwork/Tester/Cli/StrictController.php
jdeniau: OK, I rolledback the changes in the StrictController.
In the end, I just added a test for this feature 😄

acoulton (approved):
@jdeniau thanks - this is looking good to me now.

I have again reviewed in parallel with the overall diff since v3.15.0 https://github.com/Behat/Behat/compare/v3.15.0...e8d1dd9 to confirm that the things that look like BC breaks in this PR are limited to the classes added since the last release.

src/Behat/Testwork/Tester/ServiceContainer/TesterExtension.php
```diff
@@ -137,6 +143,23 @@ protected function loadExerciseController(ContainerBuilder $container, $skip = f
         $definition->addTag(CliExtension::CONTROLLER_TAG, array('priority' => 0));
         $container->setDefinition(CliExtension::CONTROLLER_TAG . '.exercise', $definition);
     }
+    
+    
+    /**
+     * Loads stop on failure controller.
```
carlos-granados, line 149: Should be "Loads stop on failure handler"

src/Behat/Testwork/Tester/ServiceContainer/TesterExtension.php
```diff
@@ -137,6 +143,23 @@ protected function loadExerciseController(ContainerBuilder $container, $skip = f
...
+     */
+    private function loadStopOnFailureHandler(ContainerBuilder $container, ?bool $stopOnFailure)
+    {
+        $definition = new Definition('Behat\Testwork\Tester\Handler\StopOnFailureHandler', array(
+            new Reference(EventDispatcherExtension::DISPATCHER_ID)
+        ));
+        $definition->addMethodCall('setResultInterpreter', array(new Reference(TesterExtension::RESULT_INTERPRETER_ID)));
```
carlos-granados, line 156: You can probably just set this in the constructor instead of having to do a separate call

carlos-granados (changes requested):
Almost ready, just a couple of comments

src/Behat/Testwork/Tester/ServiceContainer/TesterExtension.php
jdeniau: Yes, probably due to all the changes 😉 
This has been changed

carlos-granados (approved):
All good now, thanks @jdeniau ! 🎉

acoulton (approved):
Thanks @jdeniau I have also just updated the PR title to better reflect what it now does (we got there in the end 🤣)

acoulton:
@carlos-granados now that this is solved we can go for another release I think. I'll draft a changleog this morning.

carlos-granados:
@jdeniau also, could you please update https://github.com/Behat/docs/pull/155? Thanks!!

jdeniau:
@carlos-granados this has been done already 😉

Issue #1514, opened 2024-11-04
Breaking(?) change from Behat/Gherkin should be outlined in changelog?


nickvergessen:
Behat/Behat 3.15.0 bumped Behat/Gherkin
The "simple" change of https://github.com/Behat/Gherkin/commit/5a0836db6fd4f0d6fe1db6dfff5118ab615681b4 however breaks all TableNodes that use JSON data with escaped backslashes.

In our case we had:
```gherkin
    And user "admin" creates global flow with 200
      | name      | Admin flow                       |
      | class     | OCA\FilesAccessControl\Operation |
      | entity    | OCA\WorkflowEngine\Entity\File   |
      | events    | []                               |
      | operation | deny                             |
      | checks-0  | {"class":"OCA\\WorkflowEngine\\Check\\FileName","operator":"is","value":"foobar.txt"} |
```

working for years, and the above change makes touch all tables and escape the slashes another time
```gherkin
    And user "admin" creates global flow with 200
      | name      | Admin flow                       |
      | class     | OCA\FilesAccessControl\Operation |
      | entity    | OCA\WorkflowEngine\Entity\File   |
      | events    | []                               |
      | operation | deny                             |
      | checks-0  | {"class":"OCA\\\\WorkflowEngine\\\\Check\\\\FileName","operator":"is","value":"foobar.txt"} |
```

- We can of course do that, but I think it should at least be added to the changelog for clarity?
There already was another question about it in https://github.com/Behat/Gherkin/issues/268
But the library issues/changelog might not be what most users check.

carlos-granados:
Thanks @nickvergessen your suggestion to mention this in the changelog seems reasonable. @acoulton how should we handle this?

nickvergessen:
*My suggestion would be to add a "Breaking" section to https://github.com/Behat/Behat/blob/master/CHANGELOG.md#3150---2024-10-29 and manually edit it also into the existing tag/release description (without a further release for now)*

acoulton:
Hmm - I'm not sure. 

We could flag it in the behat changelog (I hadn't actually noticed that the minimum requirement was bumped, as it was done direct on master before we took over).

However, ultimately the break is in behat/gherkin@4.10.0 and will affect people even if they stick with behat/behat@3.14.0.

[...]

PR #1517, opened 2024-11-05
chore: Update CHANGELOG to mention behaviour change in Gherkin

acoulton:
Further to #1514 and Behat/Gherkin#269, add a note to Behat's own CHANGELOG to mention the behaviour change in gherkin 4.10.0.

I don't think we should set a precedent of incorporating the changelog for other packages into the change for this repo. Not least because the release timings are often going to be different (people would have got the problematic Gherkin update on their next composer update even if Behat hadn't bumped the minimum version).

However in this case given that behat packages have been infrequently released for a while, the releases were close in time so people might have got them both at once and assumed the problem was from Behat rather than Gherkin, and the Behat/Gherkin changelog was not clear originally, I agree we should make an exception and point people in the right direction.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [3.15.0] - 2024-10-29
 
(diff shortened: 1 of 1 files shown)


nickvergessen:
👍

carlos-granados (approved):

Issue #1065, opened 2017-08-08
Using strict flag with rerun doesn't prevent the undefined steps from passing


relte:
Given I have an undefined step and run:
```
bin/behat --tags="my_feature" --colors --format=progress -n --strict
```
- if all is passing besides the undefined step, the execution doesn't pass (exit code 1), then I run the command with `--rerun`:
```
bin/behat --tags="my_feature" --colors --format=progress -n --strict --rerun
```
which produces the same output, so in this case `--strict` does work
- if some test fails, the execution of course doesn't pass (exit code 1), but running `--rerun` makes it pass (it ignores the undefined step)

Here is what I use if it makes any difference:
```
"require-dev": {
        "behat/behat": "^3.2",
        "behat/mink": "^1.7",
        "behat/mink-browserkit-driver": "^1.3",
        "behat/mink-extension": "^2.2",
        "behat/mink-selenium2-driver": "^1.3",
        "friends-of-behat/context-service-extension": "^0.3",
        "friends-of-behat/cross-container-extension": "^0.3.0",
        "friends-of-behat/performance-extension": "^1.0",
        "friends-of-behat/service-container-extension": "^0.3",
        "friends-of-behat/suite-settings-extension": "^1.0",
        "friends-of-behat/symfony-extension": "^0.2.1",
        "friends-of-behat/variadic-extension": "^0.1",
        //...
},
```

jon-acker:
It seems to be that this is expected behaviour.

If I understand correctly this would require a  feature with at least two scenarios: the first contains a failure and the second has an undefined step.

Using the --rerun switch, *only* the scenario that failed is run, and if it gets fixed and now passes, the result is success (exit code 0), because that single scenario passes.

ciaranmcnulty:
This is still counter-intuitive behaviour

PR #1518, opened 2024-11-05
change: Rerun all tests that should be considered failed (including undefined, when strict)

carlos-granados:
We were saving in the rerun file only the tests that strictly failed. We should be saving any test that the result interpreter considers failed so that they can be rerun. This will include unimplemented tests if the strict option is used.

Fixes https://github.com/Behat/Behat/issues/1065

### DIFF
--- a/src/Behat/Behat/Tester/Cli/RerunController.php
+++ b/src/Behat/Behat/Tester/Cli/RerunController.php
@@ -15,6 +15,7 @@ use Behat\Behat\EventDispatcher\Event\ExampleTested;
 use Behat\Behat\EventDispatcher\Event\ScenarioTested;
 use Behat\Testwork\Cli\Controller;
 use Behat\Testwork\EventDispatcher\Event\ExerciseCompleted;
+use Behat\Testwork\Tester\Result\ResultInterpreter;
 use Behat\Testwork\Tester\Result\TestResult;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
@@ -50,6 +51,8 @@ final class RerunController implements Controller
      */
     private $basepath;
 
+    private ResultInterpreter $resultInterpreter;
+
     /**
      * Initializes controller.
      *
@@ -62,6 +65,11 @@ final class RerunController implements Controller
         $this->eventDispatcher = $eventDispatcher;
         $this->cachePath = null !== $cachePath ? rtrim($cachePath, DIRECTORY_SEPARATOR) : null;
         $this->basepath = $basepath;
+    }
+
+    public function setResultInterpreter(ResultInterpreter $resultInterpreter)
+    {
+        $this->resultInterpreter = $resultInterpreter;
     }
 
     /**
@@ -122,7 +130,7 @@ final class RerunController implements Controller
             return;
         }
 
-        if ($event->getTestResult()->getResultCode() !== TestResult::FAILED) {
+        if ($this->resultInterpreter->interpretResult($event->getTestResult()) === 0) {
             return;
         }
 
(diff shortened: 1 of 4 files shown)


src/Behat/Behat/Tester/Cli/RerunController.php
```diff
@@ -122,7 +130,7 @@ public function collectFailedScenario(AfterScenarioTested $event)
             return;
         }
 
-        if ($event->getTestResult()->getResultCode() !== TestResult::FAILED) {
+        if ($this->resultInterpreter->interpretResult($event->getTestResult()) === 0) {
```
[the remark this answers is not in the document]
carlos-granados: This is just the value used by the ResultInterpreter (https://github.com/Behat/Behat/blob/57723fee4a9a498ce855edd5e0b94ec456c958e1/src/Behat/Testwork/Tester/Result/ResultInterpreter.php#L44-L53)
If we want to use a CONST, I would probably prefer to add a couple of public CONST in the ResultInterpreter, for example:
```
    public const PASS = 0;
    public const FAIL = 1;
```
Then modify the `interpretResult()` function to use these consts and modify my code to use it as well. What do you think?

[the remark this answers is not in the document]
stof: AFAIK, the ResultInterpreter returns an int that is the CLI exit code.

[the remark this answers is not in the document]
carlos-granados: > AFAIK, the ResultInterpreter returns an int that is the CLI exit code.

Yes, that's correct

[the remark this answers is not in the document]
carlos-granados: @ttomdewit consts added 😁

ttomdewit (approved):

src/Behat/Behat/Tester/Cli/RerunController.php
```diff
@@ -64,6 +67,11 @@ public function __construct(EventDispatcherInterface $eventDispatcher, $cachePat
         $this->basepath = $basepath;
     }
 
+    public function setResultInterpreter(ResultInterpreter $resultInterpreter)
```
stloyd, line 70: As this class is final, why not changing the `__construct()`?

carlos-granados: Even though it is final, there is the possibility that some extension is using it and creating an instance. We prefer to be careful and not introduce a BC break if possible. In a future major version we can probably move this to the constructor

acoulton: @carlos-granados I've been thinking about this pattern some more and actually using the setter injection we still have a technical BC break if people are using the class without providing the interpreter. It will just come at runtime when `resultInterpreter` is undefined, instead of when they create the class.

There is a way to make this one fully BC by doing `if (isset($this->result interpreter))` and falling back to the old logic if not.

But I wonder if it's actually worth the indirection, or if we should just work on the basis that these classes are internal to Behat and not part of the public API (even though that has never been explicitly said), and go for the neater/more visible solution of passing this in the constructor and highlighting it in the changelog...

What do you think?

carlos-granados: @acoulton mmm, yeah you may be right. I did a search on Github for RerunController and Behat and the only extension that I could find that did something with it was the https://github.com/Daniel-Marynicz/BehatParallelExtension extension. But this extension just decorates the controller, so should continue working fine if we modify the controller constructor.
Let me update the PR with this change

carlos-granados: @acoulton updated

acoulton: @carlos-granados thanks - I've had a look at that extension and agree this shouldn't affect it. And that also proves it's possible to decorate our controllers with the DI container config, rather than creating them manually, so hopefully any other extensions will be following that pattern.

acoulton (approved):

PR #1520, opened 2024-11-07
change: deprecate ApplicationFactory::VERSION, auto-detect behat version

acoulton:
As discussed in https://github.com/Behat/Behat/discussions/1484#discussioncomment-11141724

The ApplicationFactory::VERSION constant has not been bumped for the last couple of releases, and updating it manually every time is an avoidable maintenance burden.

Deprecate the constant - it will not be updated in the future, and will be removed in next major.

Instead, use composer's runtime API to detect the version at runtime (only really used for inclusion in tool output). This follows the pattern used by e.g. [doctrine](
https://github.com/doctrine/orm/blob/44fa8bbde87238ad9b574e87325c7ff89c75837e/src/Tools/Console/ConsoleRunner.php#L45-L46)

Composer's version is generated to a file at the time of `composer install` and therefore is baked into the `.phar` so that also works as expected.

Note that in some cases composer versions are non-numeric (e.g. a dev branch may report a version of `dev-c629b4a773cf2dc11e0930bd8930adf22eae5031`).

### DIFF
--- a/src/Behat/Behat/ApplicationFactory.php
+++ b/src/Behat/Behat/ApplicationFactory.php
@@ -39,6 +39,7 @@ use Behat\Testwork\ServiceContainer\ServiceProcessor;
 use Behat\Testwork\Specification\ServiceContainer\SpecificationExtension;
 use Behat\Testwork\Suite\ServiceContainer\SuiteExtension;
 use Behat\Testwork\Translator\ServiceContainer\TranslatorExtension;
+use Composer\InstalledVersions;
 
 /**
  * Defines the way behat is created.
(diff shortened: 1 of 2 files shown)


acoulton:
I have tested this with a fake release on my fork and confirmed that the correct version number is detected by composer and baked into the .phar - see the github actions logs at https://github.com/acoulton/Behat/actions/runs/11713947654/job/32627706336

image

src/Behat/Behat/ApplicationFactory.php
```diff
@@ -47,6 +48,10 @@
  */
 final class ApplicationFactory extends BaseFactory
 {
+    /**
+     * @deprecated this constant will not be updated for releases after 3.13.0 and will be removed in the next major.
+     * You can use composer's runtime API if you need to check the installed behat version.
```
carlos-granados, line 53: I would probably change this second line to `Call the getVersion() function in this class to get the installed Behat version using composer's runtime API`

carlos-granados (commented):
Small suggestion, feel free to ignore

src/Behat/Behat/ApplicationFactory.php
carlos-granados: Ah, yes of course, missed the point that getVersion() has protected visibility. Your new suggestion sounds great

acoulton: @carlos-granados done

carlos-granados (approved):
Good stuff, thanks @acoulton

Issue #1257, opened 2019-11-25
Single-quoted strings generate incorrect method names when using regex snippets


ShaneMcC:
If I have a feature that looks like:

```gherkin
Feature: Test

Scenario: Testing Replacements
  When I run 'foo'
  When I run 'bar'
  When I run "baz"
  When I run "qux"
```

behat output looks like:
```bash
$ behat --snippets-type=regex
Feature: Test

  Scenario: Testing Replacements # features/test.feature:3
    When I run 'foo'
    When I run 'bar'
    When I run "baz"
    When I run "qux"

1 scenario (1 undefined)
4 steps (4 undefined)
0m0.01s (7.85Mb)

 >> default suite has undefined steps. Please choose the context to generate snippets:

  [0] None
  [1] FeatureContext
 >

--- FeatureContext has missing steps. Define them with these snippets:

    /**
     * @When /^I run '([^']*)'$/
     */
    public function iRunFoo($arg1)
    {
        throw new PendingException();
    }

    /**
     * @When /^I run '([^']*)'$/
     */
    public function iRunBar($arg1)
    {
        throw new PendingException();
    }

    /**
     * @When /^I run "([^"]*)"$/
     */
    public function iRun($arg1)
    {
        throw new PendingException();
    }

$
```

This should only generate a single `iRun` method rather than generating `iRunFoo` and `iRunBar`

Turnip snippets do the correct thing:

```bash
$ behat --snippets-type=turnip
Feature: Test

  Scenario: Testing Replacements # features/test.feature:3
    When I run 'foo'
    When I run 'bar'
    When I run "baz"
    When I run "qux"

1 scenario (1 undefined)
4 steps (4 undefined)
0m0.01s (7.85Mb)

 >> default suite has undefined steps. Please choose the context to generate snippets:

  [0] None
  [1] FeatureContext
 >

--- FeatureContext has missing steps. Define them with these snippets:

    /**
     * @When I run :arg1
     */
    public function iRun($arg1)
    {
        throw new PendingException();
    }

$
```

carlos-granados:
Note: if you use `--snippets-type=turnip` then it works fine, you just get:
```
--- FeatureContext has missing steps. Define them with these snippets:

    /**
     * @When I run :arg1
     */
    public function iRun($arg1): void
    {
        throw new PendingException();
    }
```

PR #1524, opened 2024-11-08
fix: regex snippet generation for steps defined with single quotes

carlos-granados:
The regex expressions used when generating the regex step definition expected that the step text had been "escaped" (`\` character added before all regex characters). But when trying to extract the canonical text for the function name these regex expressions were applied to the unescaped text. This meant that it failed to properly ignore the values within single quotes and they ended up being part of the function name. So for example a step like `When I run 'foo'` created a function named `iRunFoo`. This PR fixes this by making sure that the regex expression is applied to the escaped text and thus correctly removes the parameter text.

One thing that this PR does not do is to change our code so that steps like `When I run 'foo'` and `When I run "foo"` generate the same function when using regex as it happens when using turnip. This would have required that our regex expressions were more complex and I think we should avoid that. If you have those two steps it will generate an `iRun` function and an `iRun2` function. I don't expect this to be a problem as users will probably stick to one style or the other

Modifies the tests to take into account this new behaviour. 

Note: this will not constitute a BC break. Users who have generated tests with the old functionality will continue to have functions incorrectly named but they will continue to work

Fixes https://github.com/Behat/Behat/issues/1257

### DIFF
--- a/features/legacy_snippets.feature
+++ b/features/legacy_snippets.feature
@@ -12,7 +12,7 @@ Feature: Legacy Snippets
           Given I have magically created 10$
 
         Scenario: Single quotes
-          When I have chose 'coffee with turkey' in coffee machine
+          When I have chosen 'coffee with turkey' in coffee machine
           Then I should have 'turkey with coffee sauce'
           And I should get a 'super/string':
             '''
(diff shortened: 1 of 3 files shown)


acoulton (approved):
Thanks @carlos-granados, looks good to me. I agree with you that we should avoid making the regex any more complex (they are already quite complex enough for me on a Friday night :laughing: ) and I think getting `iRun` and `iRun2` is an acceptable compromise for that edge case and still a lot better than status quo,

PR #1410, opened 2022-08-23
Deprecated error PHP 8.1 compatibility.

albeorte96:
Issue #1411 
The following error appears working with a Drupal when launching behat tests:

`Deprecated function: DOMElement::setAttribute(): Passing null to parameter #2 ($value) of type string is deprecated in Behat\Testwork\Output\Printer\JUnitOutputPrinter->addAttributesToNode() (line 118 of /var/www/html/vendor/behat/behat/src/Behat/Testwork/Output/Printer/JUnitOutputPrinter.php)`

The problem is solved by substituting an empty string for those cases where the value is NULL, which is now deprecated.

PR #1525, opened 2024-11-08
Bugfix: Deprecated passing null to `explode()`

mvhirsch:
I just ran into this issue, running my behat suite lately.

> PHP Deprecated:  explode(): Passing null to parameter #2 ($string) of type string is deprecated in /home/meh/projects/protas/orf/vendor/behat/behat/src/Behat/Behat/Output/Node/Printer/JUnit/JUnitScenarioPrinter.php on line 70

This was already mentioned in https://github.com/Behat/Behat/pull/1410#issuecomment-1273207509 but somehow never hit `master` :-D


stof (approved):

acoulton:
Thanks for the fix @mvhirsch - looks like it was only added as a comment on that PR but nobody turned that into a commit till now.

Issue #1495, opened 2024-10-28
Output the name of the corresponding step, scenario, feature or suite when a failure on a hook is recorded


carlos-granados:
When using the --progress formatter there may be was a situation where a test failed in an `after` or `before` hook. The information printed to the output may not be enough to diagnose the issue because we are just told which hook function failed but not the related step, scenario, feature or suite.

When printing out the failed hooks, we should print:
- For `beforeStep` or `afterStep` hooks, the name of the step where the failure happened
- For `beforeScenario` or `afterScenario` hooks, the name of the scenario where the failure happened
- For `beforeFeature` or `afterFeature` hooks, the name of the feature where the failure happened
- For `beforeSuite` or `afterSuite` hooks, the name of the suite where the failure happened

Related to https://github.com/Behat/Behat/pull/1392 and https://github.com/Behat/Behat/issues/842

acoulton:
My only thought is for the step-level hooks it would still help to have more context than just the step, since the same step will probably be used in multiple places and the hook failure could well relate to specific scenarios / features.

It might be enough to give just `path/to/feature.feature:23` in all cases, similar to how the file & line get rendered for failures at the end of the progress printer?

carlos-granados:
Ah, yes that might be the best option

PR #1526, opened 2024-11-09
feature: display hook failures location in progress printer

carlos-granados:
When an error happens in a hook, in the pretty printer it is very easy to see where the error originated because the hook output is interleaved with the pretty output. However, in the progress printer, hook errors are only printed at the end of the run and they only reference the hook function where the error happened. If the function depends on the specific item (suite, feature, scenario or step) for which the hook is called, it is not possible to see this "location", which makes debugging the issue much more difficult.

In this PR we print the "location" of the call to the hook as part of the hook results printed at the end of the progress run. This location is:
- For suite hooks, the name of the suite
- For feature hooks, the path of the feature file
- For scenario and step hooks, the path of the feature file and the line number in that file

One example output is:
```
--- Failed hooks:

    BeforeScenario "features/append_snippets.feature:120" # FeatureContext::prepareTestFolders()
      Error in scenario (Exception)
```

Fixes https://github.com/Behat/Behat/issues/1495

### DIFF
--- a/features/hooks_failures.feature
+++ b/features/hooks_failures.feature
@@ -1,0 +1,213 @@
+Feature: Display hook failures location in progress printer
+  In order to be able to locate the code that generated a failure
+  As a feature developer using the progress printer
+  When a hook throws an error I want to see the related item where the code failed
+
+  Background:
+    Given a file named "features/simple.feature" with:
+      """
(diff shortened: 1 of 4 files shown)


acoulton (approved):
Great, thanks :)

PR #1529, opened 2024-11-09
chore: fix some whitespace in newly added code

carlos-granados:
Removes some trailing spaces in some lines


acoulton (approved):
Good spot, thanks.

PR #1530, opened 2024-11-11
feature: print failed hooks summary at the end of the pretty format run

carlos-granados:
As a continuation of https://github.com/Behat/Behat/issues/1495, currently when you run your tests using the pretty formatter, the output of failed hooks is interleaved with the output of the printer, which is fine. But these failures are not included in the summary that is printed at the end of the run. So, if nothing else fails, you may not notice that something has failed if you only look at the final summary. And if you realize that something has failed, you will need to go through all the output to find out exactly what. This PR modifies the pretty printer statistics so that they also record the failed hooks and they print a summary of the failed hooks (with just a single line for each failed hook, no output or exception details, you can find those in the previous output of each test)

The output looks like this:
```
--- Failed hooks:

    AfterScenario "features/append_snippets.feature:120" # FeatureContext::afterScenario()
    AfterScenario "features/arguments.feature:145" # FeatureContext::afterScenario()
    ...
```

### DIFF
--- a/features/hooks.feature
+++ b/features/hooks.feature
@@ -383,6 +383,11 @@ Feature: hooks
       │
       └─ @AfterFeature # FeatureContext::doSomethingAfterFeature()
 
+      --- Failed hooks:
+
+          BeforeScenario @exception "features/background.feature:4" # FeatureContext::beforeScenarioException()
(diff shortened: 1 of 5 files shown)


acoulton (approved):
Brilliant, that's a very helpful improvement. Thanks :)

PR #1532, opened 2024-11-12
DX: better output for wrong status code or unexpected output

### LINKED ISSUE #1522: Improve test failure output when behat output does not match the expected output
#1515 improved the output for the basic `It should pass` and `It should fail` steps - but also highlighted that these are generally "fallback" assertions because in many cases we are (or should be) using `It should pass with:` or `It should fail with:` to assert that the behat output includes expected content.

These steps have to perform partial matches on the output to avoid excess verbosity in our feature files and coupling to irrelevant details.

However, working on #1521 I realised that the output of these steps is also quite hard to work with if the expectation fails. They're implemented using PHPUnit's `Assert::assertStringContainsString` which isn't ideal for this usecase because:

* It can only render the complete actual and expected strings, which in our case are often big blocks of text. If the difference is a single word (or e.g. a missing `.`) it's virtually impossible to see without copying the two blocks and pasting into a diffing tool.
* It renders explicit `\n` for newlines in the actual content (maybe other special characters) but prints the expected content as-is. As well as making the difference harder to spot, it also makes it difficult to paste the actual into a diff tool or indeed back into the scenario if you're happy with the actual output and just want to update the expectation.

We could improve DX quite a bit by making these assertions produce a more helpful output.

One option would be to experiment with `sebastian/diff` and see if we can get it to produce a diff which ignores any extra content before or after the expected string. Alternatively we could maybe use `str_contains` for the actual assertion, and then just show a diff of the whole output if it fails - which would probably still be easier to read than the status quo.

jdeniau:
Fixes #1522

### Then it should pass/fail with: + expected status code + wrong output

image

### Then it should pass/fail with: + wrong status code + wrong output

image

### Then it should pass/fail with: + wrong status code + expected output

image

### Then the output should contain with unexpected value

image

### Then it should pass/fail

image

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -14,6 +14,8 @@ use Behat\Gherkin\Node\PyStringNode;
 use PHPUnit\Framework\Assert;
 use Symfony\Component\Process\PhpExecutableFinder;
 use Symfony\Component\Process\Process;
+use SebastianBergmann\Diff\Differ;
+use SebastianBergmann\Diff\Output\DiffOnlyOutputBuilder;
 
 /**
  * Behat test suite context.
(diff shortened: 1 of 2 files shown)


features/bootstrap/FeatureContext.php
```diff
@@ -478,6 +503,24 @@ private function moveToNewPath($path)
 
         $this->workingDir = $newWorkingDir;
     }
+    
+    /**
+     * @param 'fail'|'pass' $success
+     */
+    public function exitCodeIsCorrect(string $success): bool
```
carlos-granados, line 510: I don't think this function needs to be public

features/bootstrap/FeatureContext.php
```diff
@@ -343,7 +367,15 @@ public function fileXmlShouldBeLike($path, PyStringNode $text)
      */
     public function theOutputShouldContain(PyStringNode $text)
     {
-        Assert::assertStringContainsString($this->getExpectedOutput($text), $this->getOutput());
+        
+        if (str_contains($this->getOutput(), $this->getExpectedOutput($text))) {
+            return;
+        }
+
+
```
carlos-granados, line 375: And this one as well

carlos-granados (changes requested):
Mostly looking good, just a few tiny comments

acoulton (approved):
@jdeniau thanks - this looks great and will be very helpful, I don't have anything to add to @carlos-granados changes

jdeniau:
@carlos-granados this is updated

carlos-granados (approved):
Thanks @jdeniau !

Issue #1408, opened 2022-08-19
Unhelpful exception message when --format=junit is used without specifying --out


dkarlovi:
```
 ̇$ vendor/bin/behat --format=junit

Deprecated: is_file(): Passing null to parameter #1 ($filename) of type string is deprecated in /app/vendor/behat/behat/src/Behat/Testwork/Output/Printer/Factory/FilesystemOutputFactory.php on line 47

Deprecated: is_dir(): Passing null to parameter #1 ($filename) of type string is deprecated in /app/vendor/behat/behat/src/Behat/Testwork/Output/Printer/Factory/FilesystemOutputFactory.php on line 52

Deprecated: mkdir(): Passing null to parameter #1 ($directory) of type string is deprecated in /app/vendor/behat/behat/src/Behat/Testwork/Output/Printer/Factory/FilesystemOutputFactory.php on line 53

Warning: mkdir(): Invalid path in /app/vendor/behat/behat/src/Behat/Testwork/Output/Printer/Factory/FilesystemOutputFactory.php on line 53

Warning: fopen(/file.xml): Failed to open stream: Permission denied in /app/vendor/behat/behat/src/Behat/Testwork/Output/Printer/Factory/FilesystemOutputFactory.php on line 63

                                                                
  The StreamOutput class needs a stream as its first argument.  
  ```
  
  It requires you to also pass `--out`, but that's not clear from the error message at all.

Options from discussion in Symfony Slack:
1. default to CWD
2. clear exception that `--out` is now required

dkarlovi:
@carlos-granados this is IMO a bug, not a feature suggestion.

acoulton:
@carlos-granados @dkarlovi I have reproduced this and agree it's a bug as far as the exception message is very unclear (and in my configuration, a lot more verbose than the example in the issue description - I got multiple pages of stack trace in the console for each of the Deprecated / Warning lines before the final `The StreamOutput class needs a stream as its first argument`.

I have renamed the issue, as I think the fix is just to make the failure message clear e.g. `You must specify --out as the path to write your junit file` (or similar).

Giving the formatter a default would be a feature request - and I'm not convinced defaulting to CWD is a good idea, I don't think we should be writing to arbitrary filesystem locations without the user explicitly configuring that.

PR #1533, opened 2024-11-13
fix: show more meaningful message if no `output_path` is specified for the junit formatter

carlos-granados:
Previously it would throw a cryptic `The StreamOutput class needs a stream as its first argument` exception.

Fixes https://github.com/Behat/Behat/issues/1408

### DIFF
--- a/src/Behat/Testwork/Output/Printer/JUnitOutputPrinter.php
+++ b/src/Behat/Testwork/Output/Printer/JUnitOutputPrinter.php
@@ -11,6 +11,7 @@
 namespace Behat\Testwork\Output\Printer;
 
 use Behat\Testwork\Output\Exception\MissingExtensionException;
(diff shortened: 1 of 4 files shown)


src/Behat/Testwork/Output/Printer/JUnitOutputPrinter.php
```diff
@@ -151,11 +152,17 @@ public function setFileName($fileName, $extension = 'xml')
     public function flush()
     {
         if($this->domDocument instanceof \DOMDocument){
-            $this->getWritingStream()->write(
-                $this->domDocument->saveXML(null, LIBXML_NOEMPTYTAG),
-                false,
-                OutputInterface::OUTPUT_RAW
-            );
+            try {
```
carlos-granados, line 155: And in the junit formatter we catch this exception and throw a more specific error

acoulton (approved):
Looks great, thank you :)

PR #1537, opened 2024-11-18
Add PHP config

loic425:
Referenced issue : #1539

To start this new feature, I've just created the minimalist way to add a PHP config.

**Planned for future PRs:**
To know more about the next things to improve, look at the [referenced issue](#1539).


src/Behat/Behat/ApplicationFactory.php
```diff
@@ -120,6 +120,8 @@ protected function getConfigPath()
         $cwd = rtrim(getcwd(), DIRECTORY_SEPARATOR) . DIRECTORY_SEPARATOR;
         $configDir = $cwd . 'config' . DIRECTORY_SEPARATOR;
         $paths = array(
+            $cwd . 'behat.php',
+            $cwd . 'behat.php.dist',
```
Kocal, line 124: Maybe add those lines line 130-131? :) 
```php
            $configDir . 'behat.php',
            $configDir . 'behat.php.dist',
```

loic425: done, you're totally right.

loic425: thx @Kocal

behat.yml.dist
acoulton, line 1: Removing this file is a BC break

features/config.feature
```diff
@@ -92,6 +92,28 @@ Feature: Config
...
+  Scenario: Prioritize *.php config file instead of yaml
+    Given a file named "behat.php"
+    Given a file named "behat.yaml"
+    Given a some feature context
+    And a some feature scenarios
+    When I run behat in debug mode
+    Then the output should contain:
+      """
+      behat.php
+      """
```
acoulton, line 104: Prioritising behat.php over behat.yaml is a BC break. It's possible end-users already have a behat.php for some other purpose.

tests/Behat/Tests/Config/ConfigTest.php
```diff
@@ -0,0 +1,33 @@
...
+    public function testConfigCanBeConvertedIntoAnArray(): void
+    {
+        $config = new Config();
+
+        $this->assertIsArray($config->toArray());
+    }
+
+    public function testItReturnsSettings(): void
+    {
+        $settings = [
```
acoulton, line 19: Not really a comment on this line, but I think we'd need at least some .feature coverage that the config is actually parsed and used as expected (not just asserting which file is picked).

This is where I imagine maintenance burden will start to creep in if the `Config` object adds more features, as we would need to be sure the `.php` and `.yaml` files produced the same effect.

acoulton (changes requested):
@loic425 thanks for the proposal. I would suggest it's usually worth opening an issue to discuss a new feature before spending too much time working on implementation.

I'm in two minds about this. 

I can see from the ♥️ that there are obviously other users who already think this would be useful. And I can imagine for myself that things like being able to reference `MyContext::class` or `SomeExtension::class` from a php config file might help with things like IDE file usage detection.

At the same time I'm nervous about having two different ways to define config (both for code maintenance/QA and for docs/user DX). Especially because there are some structures that look/parse differently in PHP and yaml.

I'm tempted to say PHP would only really make sense if the `Config` class were strictly typed to give autocompletion etc, rather than just a bag of `$settings` - but obviously that would be very difficult to implement with behat's extensible config schema.

Could you explain a bit more why you think this feature would be worth having?

I have only looked very quickly at the PR itself and there are a couple of definite issues I've marked, but I would wait to hear other maintainers' opinions on the overall concept before spending more time addressing them.

tests/Behat/Tests/Config/ConfigTest.php
loic425: Yes, I've added a first scenario using a PHP file configuration.

loic425:
@acoulton 
I've created the issue to discuss about the whole implementation.
#1539

behat.php.dist
```diff
@@ -0,0 +1,13 @@
+<?php
+
+$config = new Behat\Config\Config([
```
carlos-granados, line 3: Wouldn't it be better to use 
```
use Behat\Config\Config;

$config = Config([
...
```
?

features/config.feature
```diff
@@ -92,6 +92,43 @@ Feature: Config
       The requested config file does not exist
       """
 
+  Scenario: PHP configuration file
+    Given a file named "behat.php" with:
+      """
+      <?php
+
+      return new Behat\Config\Config();
```
carlos-granados, line 100: Again I'd rather use a `use` statement here

features/config.feature
```diff
@@ -92,6 +92,43 @@ Feature: Config
...
+      {
+      }
+      """
+    And a file named "features/config.feature" with:
+      """
+      Feature:
+        Scenario:
+          When this scenario executes
+      """
+    When I run "behat -f progress --no-colors --append-snippets --config=behat.php"
```
carlos-granados, line 119: Do we need to specify a config file here? Shouldn't Behat be able to pick it up automatically?

src/Behat/Behat/ApplicationFactory.php
```diff
@@ -126,12 +126,16 @@ protected function getConfigPath()
             $cwd . 'behat.yml.dist',
             $cwd . 'behat.dist.yaml',
             $cwd . 'behat.dist.yml',
+            $cwd . 'behat.php',
+            $cwd . 'behat.php.dist',
```
carlos-granados, line 130: Should also add the `behat.dist.php` possibilty

features/config.feature
```diff
@@ -92,6 +92,43 @@ Feature: Config
...
+      U
+
+      1 scenario (1 undefined)
+      1 step (1 undefined)
+
+      --- Use --snippets-for CLI option to generate snippets for following default suite steps:
+
+          When this scenario executes
+      """
+
```
carlos-granados, line 131: I would probably add a test to confirm that the `--config` command line option lets you pick a config file different from the default behat.php

carlos-granados (changes requested):
Overall looking good, some small comments

behat.yml.dist
loic425: reverted

features/config.feature
loic425: I've moved the priorities.

src/Behat/Behat/ApplicationFactory.php
acoulton: I'd actually propose `behat.dist.php` **instead of** `behat.php.dist`. We mostly support `behat.yml.dist` for compatibility reasons, but if the goal is to get IDE highlighting and tool support, forcing people to use a recognisable `.php` final extension is reasonable IMO

features/config.feature
loic425: Exactly, I was based on previous scenarios.

src/Behat/Behat/ApplicationFactory.php
carlos-granados: With Yaml we support all three possibilities, with the .dist.yml listed last, I was suggesting to do the same here, support all three options, with .dist.php listed last, to keep it similar to Yaml

features/config.feature
loic425: done

src/Behat/Config/Config.php
```diff
@@ -0,0 +1,18 @@
...
+{
+    public function __construct(
+        private array $settings = []
+    ) {
+    }
+
+    public function toArray(): array
+    {
+        return $this->settings;
+    }
```
acoulton, line 17: One small thought - do we want to / are we OK with enforcing that the PHP can only eventually return the same types as the YAML (e.g. only scalars, not callables etc)? I think that's probably useful to ensure `.php` / `.yaml` parity.

If so, then it *might* be worth making this class implement `JSONSerializable` instead of `->toArray()` and passing it through `json_decode(json_encode($config), true)` when we read it.

That way any future DTOs for subsections within the config would just have to implement `JSONSerializable` and we'd get recursive `->toArray()` "out of the box" rather than having to build that ourselves or users having to cast the sugar methods / objects back to array in their config files?

behat.php.dist
loic425: done

src/Behat/Behat/ApplicationFactory.php
acoulton: I guess I just think there's limited value to supporting `.php.dist` as a new thing (other than parity with YAML) and it would reduce the complexity and number of cases to test? But I'm not strongly opposed to supporting all three.

features/config.feature
loic425: done

features/config.feature
loic425: yes, good idea

src/Behat/Behat/ApplicationFactory.php
loic425: From userland point of view, I don't think any reason to add .dist.php support.
I think it would be better to deprecate dist.yml instead (in a future PR). Moreover, maybe deprecate "yml" extension instead of "yaml" too.

features/config.feature
loic425: I've added one scenario with alternative PHP configuration file.

src/Behat/Behat/ApplicationFactory.php
carlos-granados: .dist.yml was added on purpose, see the PR https://github.com/Behat/Behat/pull/1464

src/Behat/Config/Config.php
loic425: Do we/you really want to keep yaml configuration forever?
Cause "callables" could be useful in the future even though I don't see any usage for now.

loic425: My proposal is adding a ConfigInterface instead.

Like this one we have on Symfony
https://github.com/symfony/symfony/blob/7.2/src/Symfony/Component/Config/Builder/ConfigBuilderInterface.php

loic425: I've added a commit with that proposal. Let me know what do you think about it.

src/Behat/Behat/ApplicationFactory.php
loic425: hum ok.... so let's go

loic425: done

src/Behat/Config/Config.php
carlos-granados: We do need to use yaml configuration at least until a new major version is launched. There will be many projects out there which will not want to update their config

loic425: Yep, I agree, I a maintainer of Sylius packages which uses a lot Behat and I know that we need a major release for that. I was just suggesting the fact that we may not want to support the both configs forever. That's the reason I'm mitigated about limiting PHP configuration to only scalars.

features/config.feature
```diff
@@ -92,6 +92,131 @@ Feature: Config
...
+      """
+      <?php
+
+      use Behat\Config\ConfigInterface;
+
+      final class CustomConfig implements ConfigInterface
+      {
+        public function toArray(): array
+        {
+          return [];
```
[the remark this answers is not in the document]
loic425: done https://github.com/Behat/Behat/pull/1537/commits/0875046eebc522fb505f1d44a346693b3038593f

src/Behat/Behat/ApplicationFactory.php
stof: I don't think we need to add `behat.php.dist`.
`behat.dist.php` is strictly superior, by providing PHP syntax highlighting by default in any IDE (and in GitHub/Gitlab), while `.php.dist` would not without a special configuration of the IDE (a,d asking the GitHub team to add such special case in their product).

We cannot remove support for `behat.yml.dist` for BC reasons (and even the deprecation is probably not worth the cost for the ecosystem, compared to the maintenance cost it has), but I think we should avoid adding _new_ `*.dist` options.

carlos-granados: @stof mmm, ok, you may be right. @loic425 sorry for the noise, can you please remove the `.php.dist` option?Thanks again

loic425: @carlos-granados agree with @stof ?
That makes a lot of sense to me.

loic425: done https://github.com/Behat/Behat/pull/1537/commits/5fa5aabc265a69cffc9b798b54bb8baeb43d445a

behat.php.dist
```diff
@@ -0,0 +1,15 @@
+<?php
```
loic425, line 1: Should be removed
https://github.com/Behat/Behat/pull/1537#discussion_r1848411572

loic425: done

loic425:
@acoulton @carlos-granados Is there anything left?

acoulton:
@loic425 sorry to disappear, I've been struggling with a nasty flu-ey cough the last couple of days. Think I'm improving now so hopefully back at work and able to catch up with this tomorrow :crossed_fingers:

features/config.feature
```diff
@@ -114,6 +252,17 @@ Feature: Config
       behat.yml
       """
 
+  Scenario: Load custom php config instead of distribution
+    Given a file named "behat.php"
+    Given a file named "behat.php.dist"
```
acoulton, line 257: Needs to be updated for the change to supported filenames
```suggestion
    Given a file named "behat.dist.php"
```

features/config.feature
```diff
@@ -92,6 +92,144 @@ Feature: Config
...
       """
 
+  Scenario: PHP configuration file
+    Given a file named "behat.php" with:
+      """
+      <?php
+
+      use Behat\Config\Config;
+
+      return new Config();
```
acoulton, line 102: I would ideally want to see each of these config examples actually have some config so that it is clear it has been parsed and applied rather than just the file found. 

I realise the existing yaml scenarios in this file don't do that, on the other hand all the other features in our suite use yaml config files so we have a lot of coverage of them being applied elsewhere.

The easiest / least verbose proof would probably be to set it to use a non-default formatter e.g. `progress`.

features/config.feature
```diff
@@ -92,6 +92,144 @@ Feature: Config
...
+      """
+
+  Scenario: Alternative PHP configuration file
+    Given a file named "alternative-behat.php" with:
+      """
+      <?php
+
+      use Behat\Config\Config;
+
+      return new Config();
```
acoulton, line 141: Again ideally this would have some basic config that we can see working.

features/config.feature
```diff
@@ -92,6 +92,144 @@ Feature: Config
...
+        {
+          return [
+            'default' => [
+              'gherkin' => [
+                'filters' => [
+                  'tags' => '@second'
+                ],
+              ],
+            ],
+          ];
```
acoulton, line 192: this is OK as-is, but it does make the example more verbose which slightly hides (for me) what's actually being tested. I would consider proving with a simpler config e.g. formatter (or `testers => [strict => true]]` if we want to use something different to the earlier examples).

src/Behat/Testwork/ServiceContainer/Configuration/ConfigurationLoader.php
```diff
@@ -195,7 +197,20 @@ protected function loadFileConfiguration($configPath, $profile)
         }
 
         $basePath = rtrim(dirname($configPath), DIRECTORY_SEPARATOR);
-        $config = (array) Yaml::parse(file_get_contents($configPath));
+
+        if (
+            \str_ends_with($configPath, '.php')
+            || \str_ends_with($configPath, '.php.dist')
```
acoulton, line 203: I'd prefer a `use str_ends_with;` rather than the inline `\`

src/Behat/Testwork/ServiceContainer/Configuration/ConfigurationLoader.php
```diff
@@ -195,7 +197,20 @@ protected function loadFileConfiguration($configPath, $profile)
...
+        ) {
+            $phpConfig = require $configPath;
+            if (!$phpConfig instanceof ConfigInterface) {
+                throw new ConfigurationLoadingException(sprintf('Configuration file `%s` must implements `%s`.', $configPath, ConfigInterface::class));
```
acoulton, line 207: I think this error message would be slightly clearer (and replaces a typo in the original)
```suggestion
                throw new ConfigurationLoadingException(sprintf('Configuration file `%s` must return an instance of `%s`.', $configPath, ConfigInterface::class));
```

acoulton (changes requested):
@loic425 this is looking good, I have a few comments on the details but it's taking shape well thank you

src/Behat/Config/ConfigInterface.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Behat\Config;
+
+interface ConfigInterface
```
acoulton, line 7: There are a lot of `ConfigInterface` out there already - I realise it's in a namespace, but something more explicit like `BehatConfigInterface` might help people avoid false positives when autocompleting in their IDE.

src/Behat/Config/Config.php
acoulton: The `ConfigInterface` is good, though it doesn't quite solve my thought which was about supporting nested DTOs for partial config (e.g. the `GenericSuite` in your example on #1539, or if an extension wanted to provide one for their own config).

However, that isn't required for this, and on reflection I don't think tying to `JSONSerializable` is a good plan (that was probably the cough medicine talking :laughing:). So I'll carry on that bit of the conversation on #1539

acoulton: On the question of deprecating / dropping yaml support - I think actually if we're careful we could potentially do that quite soon without hurting people, again will pick up on #1539.

src/Behat/Config/ConfigInterface.php
loic425: I'm not sure about this comment, cause we may reuse this ConfigInterface for DTOs support.
like
```php
$config->withExtension(ConfigInterface $extensionConfig);
```

features/config.feature
loic425: done

features/config.feature
loic425: done

src/Behat/Testwork/ServiceContainer/Configuration/ConfigurationLoader.php
loic425: also done

src/Behat/Config/ConfigInterface.php
loic425: Moreover, I think developers are now more careful about that. There are tons of cases like that.

acoulton: To be fair I don't think `BehatConfigInterface` prevents us using it for DTO / nested support - it's still "behat config", we're not saying it's e.g. `BehatConfigFileInterface` or `BehatRootConfigInterface`.

But I am not fixed on this.

I'm just in a mood because all day today my IDE has been forcing me to down-arrow past `Monolog\DateTimeImmutable` (we don't even use monolog, it's just pulled in as an "optional" dependency of another package) and every possible namespaced version of  `InvalidArgumentException` to get to the generic classes I want :frowning_face: 

I think developers are careful, it's just annoying. And like you say there are tons of cases, and really it's probably for the IDE to solve rather than us.

acoulton (approved):
This is looking good to me, thanks @loic425

src/Behat/Testwork/ServiceContainer/Configuration/ConfigurationLoader.php
```diff
@@ -195,7 +198,20 @@ protected function loadFileConfiguration($configPath, $profile)
         }
 
         $basePath = rtrim(dirname($configPath), DIRECTORY_SEPARATOR);
-        $config = (array) Yaml::parse(file_get_contents($configPath));
+
+        if (
+            str_ends_with($configPath, '.php')
+            || str_ends_with($configPath, '.php.dist')
```
carlos-granados, line 204: We no longer support `'.php.dist'` files, this was removed in one of the iterations, so you should remove this

carlos-granados (changes requested):
Looking good, just one small change needed

src/Behat/Testwork/ServiceContainer/Configuration/ConfigurationLoader.php
acoulton: @carlos-granados good catch, I even looked at that line and missed it. Another good reason for dropping that extension, makes this much cleaner.

loic425: good catch indeed. I've fixed.

carlos-granados (approved):
All good now, thanks @loic425 🎉

PR #1538, opened 2024-11-18
Php config imports

loic425:
Referenced issue : #1539 
Based on #1537 

Several proposals for this importing feature.

**Use variadic**
```php
use Behat\Config\Config;

$config = new Config();
$config->import('first_file.php', 'second_file.php', ...);
```

If we have too much files to imports, using multiple lines is still possible.

```php
use Behat\Config\Config;

$config = new Config();
$config->import(
    'first_file.php', 
    'second_file.php', 
    ...,
    ...,
    'another_file.php', 
);
```

**Use array**
```php
use Behat\Config\Config;

$config = new Config();
$config->import(['first_file.php', 'second_file.php', ...]);
```

**[Current state of this PR] Use simple way (and even support array or string at the same time)**
```php
use Behat\Config\Config;

$config = new Config();
$config
    ->import('first_file.php') 
    ->import('second_file.php')
    // ...
;
```

OR

```php
use Behat\Config\Config;

$config = new Config();
$config->import(['first_file.php', 'second_file.php', ...]) ;
```

One of the benefits of this way is that we can add optional arguments later.

```php
public function import(string|array $resource, string $type = null, bool $ignoreErrors = false): self
```

Moreover, we need to think about an implementation for importing using a pattern. 

**Using pattern**
```php
use Behat\Config\Config;

$config = new Config();
$config->import('config/../..php');
```

**Symfony way**

_Dependency Injection_
https://github.com/symfony/symfony/blob/7.3/src/Symfony/Component/DependencyInjection/Loader/Configurator/ContainerConfigurator.php#L64

_Routing_
The routing way is also interesting
https://github.com/symfony/routing/blob/7.1/Loader/Configurator/RoutingConfigurator.php#L37

### DIFF
--- a/src/Behat/Config/Config.php
+++ b/src/Behat/Config/Config.php
@@ -13,6 +13,7 @@ final class Config implements ConfigInterface
     ) {
     }
 
+    /** @param string|string[] $resource **/
     public function import(string|array $resource): self
     {
         $resources = is_string($resource) ? [$resource] : $resource;
--- a/src/Behat/Config/Config.php
+++ b/src/Behat/Config/Config.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace Behat\Config;
 
+use function is_string;
+
 final class Config implements ConfigInterface
 {
     public function __construct(
@@ -11,9 +13,13 @@ final class Config implements ConfigInterface
     ) {
     }
 
-    public function import(string $file): self
+    public function import(string|array $resource): self
     {
-        $this->settings['imports'][] = $file;
+        $resources = is_string($resource) ? [$resource] : $resource;
+
+        foreach ($resources as $resource) {
+            $this->settings['imports'][] = $resource;
+        }
 
         return $this;
     }
--- a/src/Behat/Config/Config.php
+++ b/src/Behat/Config/Config.php
@@ -11,6 +11,13 @@ final class Config implements ConfigInterface
     ) {
     }
 
+    public function import(string $file): self
+    {
+        $this->settings['imports'][] = $file;
+
+        return $this;
+    }
+
     public function toArray(): array
     {
         return $this->settings;
--- a/features/import_suites.feature
+++ b/features/import_suites.feature
@@ -1,0 +1,180 @@
+Feature: Import suites
+  In order to add more suites
+  As a feature writer
+  I need an ability to import external suite configuration files
+
+  Background:
+    Given a file named "features/bootstrap/FirstContext.php" with:
+      """
+      <?php
+
+      class FirstContext implements \Behat\Behat\Context\Context
+      {
+          /** @Given I have :count apple(s) */
+          public function iHaveApples($count) { }
+
+          /** @When I ate :count apple(s) */
+          public function iAteApples($count) { }
+
+          /** @Then I should have :count apple(s) */
+          public function iShouldHaveApples($count) { }
+      }
+      """
+    And a file named "features/bootstrap/SecondContext.php" with:
+      """
+      <?php
+
+      class SecondContext implements \Behat\Behat\Context\Context
+      {
+          /** @Given I have :count apple(s) */
+          public function iHaveApples($count) { }
+
+          /** @When I ate :count apple(s) */
+          public function iAteApples($count) { }
+
+          /** @Then I should have :count apple(s) */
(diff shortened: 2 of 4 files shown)


src/Behat/Config/Config.php
```diff
@@ -0,0 +1,25 @@
...
+namespace Behat\Config;
+
+final class Config
+{
+    public function __construct(
+        private array $settings = []
+    ) {
+    }
+
+    public function import(string $file): self
```
loic425, line 14: Could be interesting...

```suggestion
    public function import(... string $files): self
```

src/Behat/Config/Config.php
```diff
@@ -0,0 +1,27 @@
...
+{
+    public function __construct(
+        private array $settings = []
+    ) {
+    }
+
+    public function import(string ...$files): self
+    {
+        foreach ($files as $file) {
+            $this->settings['imports'][] = $file;
```
acoulton, line 17: I might have missed something, but if the `->import()` method just adds keys/values to the array you are already defining in the file, why wouldn't you just do that inline?

E.g. 
```
return new Config([
  'imports' => [
    __DIR__.'/other-config.php'
  ],
  // Other values 
])
```

acoulton (changes requested):
Obviously we need to look at #1537 first, but probably worth considering the overall feature / plan for future direction

src/Behat/Config/Config.php
```diff
@@ -0,0 +1,27 @@
...
+namespace Behat\Config;
+
+final class Config
+{
+    public function __construct(
+        private array $settings = []
+    ) {
+    }
+
+    public function import(string ...$files): self
```
loic425, line 14: I think this should be better without variadic thing or I should remame the method "imports" instead.
```suggestion
    public function import(string $file): self
```

acoulton:
I think my preference would be:
* For explicit files, a variadic e.g. `import(string ...$imports)`
* For pattern matching, a separate `importMatching(string $pattern)` - if you want multiple patterns, you call it multiple times. Possibly with something like `importMatching(string $pattern, bool $throwIfNoMatches = true)` to give people control / safety over what happens if nothing matches the pattern.

loic425:
> I think my preference would be:
> 
>     * For explicit files, a variadic e.g. `import(string ...$imports)`
> 
>     * For pattern matching, a separate `importMatching(string $pattern)` - if you want multiple patterns, you call it multiple times. Possibly with something like `importMatching(string $pattern, bool $throwIfNoMatches = true)` to give people control / safety over what happens if nothing matches the pattern.

I've removed the two methods way from the proposal and added a simple one based on Symfony routing configurator.

src/Behat/Config/Config.php
```diff
@@ -4,13 +4,26 @@
...
+use function is_string;
+
 final class Config implements ConfigInterface
 {
     public function __construct(
         private array $settings = []
     ) {
     }
 
+    public function import(string|array $resource): self
```
loic425, line 16:
```suggestion
    /** @param string|string[] $resource **/
    public function import(string|array $resource): self
```

carlos-granados (approved):
Looking good to me, thanks @loic425

loic425:
@acoulton Is the current state ok for you.
I've simplified with a single method with string or array of string.

src/Behat/Config/Config.php
```diff
@@ -4,13 +4,27 @@
...
 {
     public function __construct(
         private array $settings = []
     ) {
     }
 
+    /** @param string|string[] $resource **/
+    public function import(string|array $resource): self
+    {
+        $resources = is_string($resource) ? [$resource] : $resource;
```
acoulton, line 19: I would personally use a variadic here, it gives us (and users) strict typing out of the box including protecting against unexpected/unused extra arguments, keeps our own code as thin as possible, and therefore reduces any risk of edge cases.

As a user, the syntax `->import('one.php', 'two.php')` is also cleaner and easier to format IMO than having the redundant array braces.

I know there are other tools out there that do `string|string[]` but my sense is those are interfaces that were designed before variadics were available and so they're tied to them. I don't think we should be, unless there's a reason not to use a variadic that I haven't thought of?

acoulton: Fair enough, makes sense.

acoulton (approved):

Issue #1539, opened 2024-11-19
[RFC] Add Support for PHP File Configuration


loic425:
### Description
Currently, Behat supports configuration through YAML files. While YAML is straightforward for many use cases, offering support for PHP configuration files could bring added flexibility and advantages, such as:

- Enhanced readability for developers familiar with PHP syntax.
- Leveraging PHP logic (e.g., conditions, constants, or reusable configurations).
- Improved IDE support and type checking.

### Proposal
Introduce support for configuring Behat using PHP files alongside the existing YAML configuration. This would involve:

- Parsing configurations from a behat.php file if present.
- Maintaining backward compatibility with YAML configuration files (behat.yml, behat.yaml).
-  Providing a basic example of a PHP configuration file in the documentation.

### Benefits

- Greater flexibility in defining configurations.
-  Ability to dynamically generate configurations based on environment or context.
-  Seamless integration with PHP constants, variables, and classes.

### Example

Here’s an example of what a PHP configuration file (behat.php) could look like:

```php
use Behat\Config\Config;

return new Config([
    'default' => [
        'gherkin' => [
            'filters' => [
                'tags' => '~@php8'
            ],
        ],
    ],
]);
```

### Implementation Notes

- Extend the Behat configuration loader to check for behat.php.
- Update the documentation to explain the new option and its use cases.

### Future helpers

We can do everything with a simple array. But to improve the DX, we can add some helpers to the Config object.

**Importing Configuration Files**

To make configurations modular and reusable, the PHP file should support importing other files. 

It's already possible with the previous proposal:

```php
use Behat\Config\Config;

return new Config([
    'imports' => [
        __DIR__.'/other-config.php'
    ],
])
```

But we need some helpers to improve the DX.

For instance:

```php
use Behat\Config\Config;

$config = new Config();
$config->import('config/suites.php');

return $config;
```

We can import several configuration files with the same "import" method.

```php
use Behat\Config\Config;

$config = new Config();
$config->import([
    'config/extensions.php', 
    'config/suites.php',
]);

return $config;
```

We can also introduce more smart imports.

```php
use Behat\Config\Config;

$config = new Config();
$config->import('config/**/**.php');

return $config;
```

**Adding suites**

Here's is just a proposal, it needs to be discussed.

```php
use App\Tests\Behat\Context\FirstContext;
use App\Tests\Behat\Context\SecondContext;
use Behat\Config\Config;
use Behat\Config\Profile;
use Behat\Config\Suite;

[...]

PR #1540, opened 2024-11-19
Php config suites

loic425:
Referenced issue #1539

Example:
```php
use Behat\Config\Config;
use Behat\Config\Profile;
use Behat\Config\Suite;
use App\Tests\Behat\Context\FirstContext;
use App\Tests\Behat\Context\SecondContext;

$profile = (new Profile('default'))
    ->withSuite(new Suite('first', ['contexts' => [FirstContext::class]]))
    ->withSuite(new Suite('second', ['contexts' => [SecondContext::class]]))
;

```

Or even better
```php
use Behat\Config\Config;
use Behat\Config\Profile;
use Behat\Config\Suite;
use App\Tests\Behat\Context\FirstContext;
use App\Tests\Behat\Context\SecondContext;

$profile = (new Profile('default'))
    ->withSuite((new Suite('first'))->withContexts(FirstContext::class))
    ->withSuite((new Suite('second'))->withContexts(FirstContext::class))
;

```


loic425:
@carlos-granados @acoulton It's now ready for a first review

features/suite.feature
```diff
@@ -48,14 +48,21 @@ Feature: Suites
...
-        suites:
-          first:
-            contexts: [ FirstContext ]
-          second:
-            contexts: [ SecondContext ]
+      <?php
+
+      use Behat\Config\Config;
+      use Behat\Config\Suite;
+      use Behat\Testwork\Suite\GenericSuite;
```
loic425, line 57:
```suggestion
```

src/Behat/Config/Suite.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Behat\Config;
+
+final class Suite implements SuiteConfigInterface
+{
+    public function __construct(
```
carlos-granados, line 9: I don't like that `profile` needs to be passed as a parameter here. This means that if you have several profiles, in any which is not the default you need to be passing the profile name everywhere. I think that it would be best if we had a `withProfile()` function which accepted a `Profile` item (or its corresponding interface) and then this Profile class is the one that would have a `withSuite()` function. Adding the `withProfile` function and the `Profile` class would probably be best done in a separate PR that would be merged before this one

carlos-granados (changes requested):

src/Behat/Config/Suite.php
loic425: Ok, let's go for it.

src/Behat/Config/Suite.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Behat\Config;
+
+final class Suite implements SuiteConfigInterface
```
acoulton, line 7: Once we agree in #1547 whether or not to use interfaces for the DTOs and what to call them this might need to be updated.

acoulton (changes requested):
This LGTM once it's updated based on #1547 and the final discussions there.

loic425:
@carlos-granados @acoulton 
I've updated the PR & the overview in the PR description.

carlos-granados (commented):
Looking good overall, a couple of comments

src/Behat/Config/Suite.php
```diff
@@ -0,0 +1,39 @@
...
+{
+    public function __construct(
+        private string $name,
+        private array $settings = [],
+        private string $profile = 'default',
+    ) {
+    }
+
+    public function withContexts(...$contexts): self
```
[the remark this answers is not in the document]
loic425: Oh, I never used that. Ok, great, I'll add this this morning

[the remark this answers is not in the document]
loic425: done

src/Behat/Config/SuiteConfigInterface.php
```diff
@@ -0,0 +1,12 @@
+<?php
```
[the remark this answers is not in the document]
loic425: indeed, I forgot to remove it. done

src/Behat/Config/Suite.php
```diff
@@ -0,0 +1,48 @@
...
+final class Suite
+{
+    public function __construct(
+        private string $name,
+        private array $settings = [],
+        private string $profile = 'default',
+    ) {
+    }
+
+    public function withContexts(string ...$contexts): self
```
carlos-granados, line 16: The contexts can also be arrays in case you want to pass any parameters to the contexts, for example what we would previously see in a yaml config file:

```
default:
    suites:
        default:
            contexts:
                - MyAwesomeContext:
                    - http://localhost:8080
                    - /var/tmp
```

loic425: array of string?

carlos-granados: In this case we would have:

```
$context = [
    'MyAwesomeContext' => [
        'http://localhost:8080',
        '/var/tmp',
    ]
]

Comment (stof) [reply src/Behat/Config/Suite.php:16]: For the PHP config, I would suggest using a method `->addContext(class-string<Context> $context, array<mixed> $arguments = []): static` instead (of course using valid PHP types and moving the more precise types to phpdoc), and handling the representation of how registering a context with argument looks like inside that method instead of exposing it to the outside. The fact that registering arguments involves an array with a single argument using the class name as key and arguments as values is not something we can properly describe to static analyzers AFAICT.

Comment (acoulton) [reply src/Behat/Config/Suite.php]: TBH, I like the explicit `->addContext(string $context, array $constructorArgs)` both because it hides our implementation detail of how that array is structured, and because it makes more obvious to the user that this is an available feature. I think if you're passing arguments, your config is already quite verbose and the repeated `addContext` calls are less significant.

However, I would **also** keep a simple `->withContexts(string ...$contexts)` for people who just want to provide a list of names with no extra detail.

Comment (carlos-granados) [reply src/Behat/Config/Suite.php]: @acoulton that sounds like a nice compromise. 👍

File: src/Behat/Config/Suite.php [hunk 7]
```diff
@@ -0,0 +1,59 @@
...
+declare(strict_types=1);
+
+namespace Behat\Config;
+
+final class Suite
+{
+    public function __construct(
+        private string $name,
+        private array $settings = [],
+        private string $profile = 'default',
```
carlos-granados, line 12: Do we still want to keep the option of setting the profile name when creating the Suite? I don't think we need this now that we have the `withSuite` function in the Profile class

loic425: Oh... I've just forgot about it. thx

loic425: done

src/Behat/Config/Suite.php
```diff
@@ -0,0 +1,58 @@
...
+
+        return $this;
+    }
+
+    public function name(): string
+    {
+        return $this->name;
+    }
+
+    public function profile(): ?string
```
carlos-granados, line 49: You also need to remove this function

loic425: done

carlos-granados (approved):
Looking good now, thanks @loic425

loic425:
@acoulton Ok for you too?

acoulton (approved):
Great, thanks @loic425

src/Behat/Config/Suite.php
stof: indeed. With an extension resolving contexts in a different way, it could be any string.

loic425: Yes, this is from the friends of behat symfony extension.

PR #1543, opened 2024-11-23
Removes redundant tests for hook failures in junit formatter

carlos-granados:
We are now testing the output of all hook failures in the junit formatter in the `hooks_failures_junit.feature` feature file. It does not make sense to keep the existing tests that we had in the `junit_format.feature` feature for the step failures. This PR removes these redundant tests


acoulton (approved):
Sorry for the slow review @carlos-granados , it's taking me a while to catch up with everything after being off sick last week

Issue #1527, opened 2024-11-09
Use actual files for our internal tests


carlos-granados:
Currently most of out internal tests use files which are generated on the fly with steps like:
```
    Given a file named "features/apples.feature" with:
      """
      .... some content
      """
```
This is quite a pain to work with, since the content is plain text:
- You don't get autocompletion in the IDE
- You don't get syntax highlighting
- Indenting can be a pain, specially if you are mixing php code and feature code

I think we could implement a step like
```
    Given I use file "features/arguments/apples.feature"
```
What this would do is, instead of creating a file on the fly from the feature content, copy this actual file which would be stored in the repository (under the `features\examples` folder, so in this case the file would actually be in `features/examples/features/arguments/apples.feature`) to the temporary folder used to run the tests  

Advantages:
- Autocompletion and syntax highlighting in the IDE for the example files
- Easy indenting in the example files

Disadvantages:
- Makes the tests more difficult to review at a glance
- Could make it more difficult for users who are looking into tests for guidance

In my case I think that the advantages overcome the disadvantages. I think that reviewing can still be easy since you will have the example files in any PR, and that these tests are mostly internal and not really useful for users who are not developing Behat. Good naming of the example files (for example `testSnippetRegexOutputWithSimpleText.feature`) would help a lot

What do you think?
(I don't propose to change all our tests to this new format, at least not initially, this would initially be used only for new tests)

carlos-granados:
Another advantage is that it will probably make our tests more readable. Now a lot of the detail is lost in the details of the generated files

acoulton:
Thanks for raising this - I found this quite painful too on my last PR. I ended up developing with my feature file provisioning a `FeatureContext.php` that just did `require /path/to/my/wipFeatureContext.php` so that I could get autocompletion / autoformatting etc and then only inlined the content back into the feature file at the very end. But that didn't solve the similar issue with the content of the `.feature` inside the `.feature` (and that really upset PHPStorm - if I tried to autoformat it broke the file completely so it took longer than it should have to tidy everything for commit).

I wonder if we should go one step further, and consider provisioning each feature as a standalone directory (e.g. with the FeatureContext, any .feature files, behat.yml etc all in the same place), rather than copying anything - so the behat is bascially just telling it what working directory & command to run, and then making assertions.

Advantages:

* Same as your idea
* Ability to run the "inner" behat process direct from the CLI - making it easier to do things like see colour output, step/trace through with Xdebug for more complex debugging, etc.
[...]

PR #1544, opened 2024-11-24
DX: Use real files instead of generated files in local tests

carlos-granados:
I made a first stab at what we discussed here: https://github.com/Behat/Behat/issues/1527

I choose one example test which used multiple feature and context files. Instead of generating them on the fly these are real files saved in the repository.
I created a new `Fixtures` folder in the `tests` folder to hold these files. To use them you can use a step that will set the folder to run Behat to a specific folder, instead of a temporary one.
Then to choose which features/contexts you need to use for a specific test you use suites defined in the config file. If you need to run several suites you can use a different config file. I used the new Php config file format. Since it is important to show what is specific in a test, I created new steps that let you clearly mark that you are running a specific suite or config file
The tests may need to create temporary files, so I created a `temp` folder in this `fixtures` folder that gets cleared at the beginning and end of each test run.
Naming is very important so that you can easily see which files are being used for each test case, so I tried to give all these files significant names.

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -9,8 +9,10 @@
  */
 
 use Behat\Behat\Context\Context;
-use Behat\Behat\Output\Node\EventListener\JUnit\JUnitDurationListener;
 use Behat\Gherkin\Node\PyStringNode;
+use Behat\Gherkin\Node\TableNode;
+use Behat\Step\Given;
+use Behat\Step\When;
 use PHPUnit\Framework\Assert;
 use Symfony\Component\Process\PhpExecutableFinder;
 use Symfony\Component\Process\Process;
@@ -39,6 +41,10 @@ class FeatureContext implements Context
     /**
      * @var string
      */
+    private $tempDir;
+    /**
+     * @var string
+     */
     private $options = '--format-settings=\'{"timer": false}\' --no-interaction';
     /**
      * @var array
@@ -48,6 +54,8 @@ class FeatureContext implements Context
      * @var string
      */
     private $answerString;
+
+    private bool $workingDirChanged = false;
 
     /**
      * Cleans test folders in the temporary directory.
@@ -73,13 +81,13 @@ class FeatureContext implements Context
             md5(microtime() . rand(0, 10000));
 
         mkdir($dir . '/features/bootstrap/i18n', 0777, true);
-        mkdir($dir . '/junit');
 
         $phpFinder = new PhpExecutableFinder();
         if (false === $php = $phpFinder->find()) {
             throw new \RuntimeException('Unable to find the PHP executable.');
         }
         $this->workingDir = $dir;
+        $this->tempDir = $dir;
         $this->phpBin = $php;
     }
 
@@ -189,6 +197,33 @@ EOL;
     public function iSetBehatParamsEnvironmentVariable(PyStringNode $value)
     {
         $this->env = array('BEHAT_PARAMS' => (string) $value);
+    }
+
+    /**
+     * @When I set the working directory to the :dir fixtures folder
+     */
+    public function iSetTheWorkingDirectoryToTheFixturesFolder($dir): void
+    {
+        $basePath = dirname(__DIR__, 2) . '/tests/Fixtures/';
+        $dir = $basePath . $dir;
+        if (!is_dir($dir)) {
+            throw new RuntimeException(sprintf('The directory "%s" does not exist', $dir));
+        }
+        $this->workingDir = $dir; ;
+        $this->workingDirChanged = true;
+    }
+
+    #[Given('I provide the following options for all behat invocations:')]
+    public function iProvideTheFollowingOptionsForAllBehatInvocations(TableNode $table): void
+    {
+        $this->addBehatOptions($table);
+    }
+
+    #[When('I run behat with the following additional options:')]
+    public function iRunBehatWithTheFollowingAdditionalOptions(TableNode $table): void
+    {
+        $this->addBehatOptions($table);
+        $this->iRunBehat();
     }
 
     /**
@@ -333,7 +368,7 @@ EOL;
     /**
      * Checks whether specified content and structure of the xml is correct without worrying about layout.
      *
-     * @Then /^"([^"]*)" file xml should be like:$/
+     * @Then /^(?:the\s)?"([^"]*)" file xml should be like:$/
      *
      * @param string       $path file path
      * @param PyStringNode $text file content
@@ -341,6 +376,20 @@ EOL;
     public function fileXmlShouldBeLike($path, PyStringNode $text)
     {
         $path = $this->workingDir . '/' . $path;
+        $this->checkXmlFileContents($path, $text);
+    }
+
+    /**
+     * @Then the temp :path file xml should be like:
+     */
+    public function theTempFileFileXmlShouldBeLike($path, PyStringNode $text): void
+    {
+        $path = $this->tempDir . '/' . $path;
+        $this->checkXmlFileContents($path, $text);
+    }
+
+    private function checkXmlFileContents($path, PyStringNode $text)
+    {
         Assert::assertFileExists($path);
 
         $fileContent = trim(file_get_contents($path));
@@ -356,7 +405,6 @@ EOL;
 
         Assert::assertEquals(trim($dom->saveXML(null, LIBXML_NOEMPTYTAG)), $fileContent);
     }
-
 
     /**
      * Checks whether last command output contains provided string.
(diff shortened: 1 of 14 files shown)


features/bootstrap/FeatureContext.php
```diff
@@ -191,6 +192,30 @@ public function iSetBehatParamsEnvironmentVariable(PyStringNode $value)
         $this->env = array('BEHAT_PARAMS' => (string) $value);
     }
 
+    /**
+     * @When I set the working directory to :dir
+     */
+    public function iSetTheWorkingDirectoryTo($dir): void
+    {
+        $this->workingDir = realpath($dir);
```
stof, line 200: instead of resolving the directory based on the current CWD and setting it to `false` in case the directory does not exist, I would suggest making the code more robust:

- treat it explicitly as a a relative path based on `dirname(__DIR, 2)` (or maybe omit the `tests/Fixtures` part of the path by moving that to the base path, to enforce that scenarios uses those fixtures and not something else)
- if we don't support using `..` in that relative path, I don't think we need to use `realpath` (we won't have symlinks in our fixtures anyway, as those don't play well with a git repo aimed at being usable cross-platform)
- add a `is_dir` check to report a proper error if the provided directory does not exist instead of getting weird errors later in the scenario.

features/bootstrap/FeatureContext.php
```diff
@@ -60,6 +60,7 @@ public static function cleanTestFolders()
         if (is_dir($dir = sys_get_temp_dir() . DIRECTORY_SEPARATOR . 'behat')) {
             self::clearDirectory($dir);
         }
+        self::clearDirectory('tests/Fixtures/temp', false);
```
stof, line 63: why do we need `tests/Fixtures/temp` ? Can temporary files be kept in `sys_get_temp_dir()` ?

stof: And if we keep temp files inside `tests/Fixtures`, we really need a `.gitignore` rule to avoid committing them by mistake

stof (commented):
Are you planning to remove entirely the inline generated files ? If no, I suggest tracking a state whether we are running in a fixture-based working directory or a generated one, and throwing an error in steps writing files if they are used in a fixtures-based working directory, to avoid altering fixtures.

stof:
I'm also wondering whether we should rely on a single folder containing multiple suites and running only one of them, or if we should have dedicated fixture folder for each case we want to run (which is what would correspond to the existing scenarios)

carlos-granados:
> Are you planning to remove entirely the inline generated files ? If no, I suggest tracking a state whether we are running in a fixture-based working directory or a generated one, and throwing an error in steps writing files if they are used in a fixtures-based working directory, to avoid altering fixtures.

Yeah, sounds good, let me try to implement this

carlos-granados:
> I'm also wondering whether we should rely on a single folder containing multiple suites and running only one of them, or if we should have dedicated fixture folder for each case we want to run (which is what would correspond to the existing scenarios)

I also wondered about this but decided to use the current structure for two reasons:
- If we add a folder for each scenario I think we would end up with too many levels of folders, making it difficult to review a feature (where you would want to review all its scenarios, not a single scenario)
- In many cases it may be useful to have files which are shared by several scenarios (many features generate files in backgrounds) and having a folder for each scenario would make this more cumbersome

Other tools like Rector have one test file for each case for a feature, all stored in a single folder.

features/bootstrap/FeatureContext.php
carlos-granados: I did this because we had some steps where you had to specify the path of the file to check and I could not use a temp folder, but then I realised that I could add some new steps that used the temp folder, updated

features/junit_format.feature
```diff
@@ -1,96 +1,18 @@
...
-        Examples:
-          | value | result |
-          | 5     | 15     |
-          | 10    | 20     |
-      """
-    When I run "behat --no-colors -f junit -o junit --snippets-for=FeatureContext --snippets-type=regex"
+  Scenario: Run a single feature
+    Given I want to run the suite "single_feature"
+    And I set the output folder to the system temp folder
+    And I run "behat --no-colors -f junit --snippets-for=SingleFeatureContext --snippets-type=regex"
```
acoulton, line 12: I feel this is a bit hard to follow, because the final line looks like it is an explicit `behat` command invocation but in fact it is merged in with the parameters set above. I feel like splitting `I want to run the suite "single_feature"` out to a separate step makes this less obvious, not more.

I think maybe there's a few things that might help:

* Move more of the "common" CLI arguments to the suite config file - e.g. the formatter to use, maybe also the output path?
* Have a step like `Given I will provide the following arguments for all behat invocations:` as a TableNode in e.g. the Background with options that are required to make the output match but not critical to the feature we're testing?
* Change the `I run` to be `I run behat with the following additional options:` as a TableNode so that the individual arguments are on separate lines

So here we might have:

```gherkin
Given I will provide the following arguments for all behat invocations:
  # not sure the best way to represent flags in a table / array we can merge, but {SET}
  # to add it to the list of options and {NOT SET} to override? Or maybe it's safer if
  # we don't allow these values to be overridden later in the feature...
  | --no-colors     | {SET} | 
  | --snippets-type | regex |

# ... 
When I run behat with the following additional options:
  | --snippets-for | SingleFeatureContext |
  | --suite        | single_feature      |
  # not sure about this, but want to avoid a separate "I set the output folder to ... step"
  | --out          | {SYSTEM_TMP_DIR}     |

File: tests/Fixtures/JunitFormat/features/bootstrap/AbortOnPHPErrorContext.php [hunk 4]
```diff
@@ -0,0 +1,26 @@
...
+
+class Foo {}
+class AbortOnPHPErrorContext implements Context
+{
+    private $value;
+
+    #[Given('/I have entered (\d+)/')]
+    public function iHaveEntered($num) {
+        $this->value = $num;
+    }
```
acoulton, line 15: There's a lot of duplication in these step implementations between suites, which made sense when we were writing separate files as the only file for the scenario, but now I think adds to the noise and makes it harder to see what's actually triggering the behaviour we're testing. Maybe it would make sense to pull them all into a shared context?

tests/Fixtures/JunitFormat/features/abort_on_php_error.feature
```diff
@@ -0,0 +1,11 @@
...
+  In order to see correct information
+  As a features developer
+  I want to see the right output in JUnit when I abort due to a PHP error
+
+  Background:
+    Given I have entered 10
+
+  Scenario: Failed
+    When I add 4
+    Then I must have 14
```
acoulton, line 11: Obviously in real life the point is that any step could cause an error, but in the tests I think the indirection "I must have 14" => trigger an error again makes this harder to follow especially now that the steps and features are separated from the feature file.

Maybe we should just be explicit: `Then I trigger a PHP error`?

This might also again help with combining step definitions into a smaller number of contexts, to make it easier to follow what's going on?

tests/Fixtures/JunitFormat/features/bootstrap/SkippedTestCasesContext.php
```diff
@@ -0,0 +1,22 @@
...
+use Behat\Behat\Context\Context;
+use Behat\Hook\BeforeScenario;
+use Behat\Step\Given;
+use Behat\Step\Then;
+
+class SkippedTestCasesContext implements Context
+{
+    private $value;
+
+    #[BeforeScenario]
```
acoulton, line 12: This might be a step too far, but again it *could* be possible to implement this as a tagged hook, so that it is always present in the context but only triggered for the feature that actually needs it?

acoulton (commented):
@carlos-granados sorry for the slow reply, I've started looking at this a few times since you opened it but not quite had long enough to get my head round it / work out what I think.

As we expected, moving the files out of the features makes it quite a bit harder to follow what implementation is being tested by each feature - particularly if every scenario in the feature file has a different `.feature` and a different `Context`. Even with very clear naming, by the time you get to the contexts it is already hard to remember which scenario is using it and what it's expecting it to do (in a PR diff).

There might be some Behat features where we need that level of complexity /  isolation.  But I think for most things - e.g. formatters - we should try to keep the implementations as simple and concise as possible *across the whole feature/PR* rather than prioritising keeping each individual scenario's files small. So thinking more like writing a "normal" behat suite than the file-per-scenario pattern we had when everything was inline,

I've made a couple of suggestions, would be interested to hear what you think?

acoulton:
An afterthought: there are a lot of features that need very similar implementations (e.g. all the different formatters need to be able to have some passing / failing / skipped steps).

Maybe it would be worth consolidating all of them to use the same Context(s) rather than restating that everywhere : again this might help to focus in on what is actually being tested...?

carlos-granados:
@acoulton thanks for your helpful comments, I will work on this and see how it can be improved with your ideas, it may take a few days

acoulton:
@carlos-granados glad it was helpful. I'll probably not be around much now till early January but will pick up in due course.

Have a great Christmas and New Year!

features/junit_format.feature
carlos-granados: I implemented the suggestion of adding these two step definitions, can you take a look and see if it looks better with them? I will continue with your other suggestions another day

acoulton (commented):
I think this is quite a lot easier to follow now - taking out the repetition really helps to focus on what's actually being tested.

It's slightly artificial looking at a diff between the old tests and the new pattern. I guess we won't really know for sure until we're reviewing a PR for a new / changed feature - but it feels to me like this approach will work well and be easier to review than the current inline format.

features/bootstrap/FeatureContext.php
```diff
@@ -191,6 +199,49 @@ public function iSetBehatParamsEnvironmentVariable(PyStringNode $value)
...
+        $this->options .= ' --suite=' . $suite;
+    }
+
+    /**
+     * @Given I want to run the config :config
+     */
+    public function iWantToRunTheconfig($config): void
+    {
+        $this->options .= ' --config=' . $config;
+    }
```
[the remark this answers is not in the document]
carlos-granados: Good spot

carlos-granados:
@acoulton I did a full review of all changes, squashed them and rebased the branch, this should be ready for the final review. Perhaps instead of looking at the diff, which might be difficult to follow, maybe check out the branch and take a look, it will probably be easier to see the final result

Issue #1542, opened 2024-11-21
Add support for Transform as an Attribute


carlos-granados:
Currently all existing annotations are supported as attributes except `@Transform`. Since we want to start using attributes everywhere we need to add this missing attribute

PR #1545, opened 2024-11-26
feat: adds the `Transform` attribute, equivalent to the `@Transform` annotation

carlos-granados:
Adds the Attribute equivalent to the `@Transform` annotation, completing the set of attributes needed to start replacing annotations with attributes everywhere.

Closes https://github.com/Behat/Behat/issues/1542


src/Behat/Behat/Context/Reader/AttributeContextReader.php
```diff
@@ -42,10 +42,6 @@ public function registerAttributeReader(AttributeReader $reader)
      */
     public function readContextCallees(ContextEnvironment $environment, $contextClass)
     {
-        if (\PHP_MAJOR_VERSION < 8) {
```
carlos-granados, line 45: These checks are no longer needed

features/definitions_transformations_attribute.feature
```diff
@@ -0,0 +1,188 @@
...
+  As a feature writer
+  I need to use transformation functions using PHP attributes
+
+  Background:
+    Given a file named "features/bootstrap/User.php" with:
+      """
+      <?php
+      class User
+      {
+          private $username;
```
loic425, line 12: You are in php 8.1 now, using typed properties would be better :)
using constructors with promoted properties can be possible too.

src/Behat/Behat/Transformation/Context/Attribute/TrasnformationAttributeReader.php
```diff
@@ -0,0 +1,70 @@
...
+     * @var DocBlockHelper
+     */
+    private $docBlockHelper;
+
+    /**
+     * Initializes reader.
+     *
+     * @param DocBlockHelper $docBlockHelper
+     */
+    public function __construct(DocBlockHelper $docBlockHelper)
```
loic425, line 36: promoted properties in constructor

src/Behat/Behat/Transformation/Context/Attribute/TrasnformationAttributeReader.php
```diff
@@ -0,0 +1,70 @@
...
+{
+    /**
+     * @var DocBlockHelper
+     */
+    private $docBlockHelper;
+
+    /**
+     * Initializes reader.
+     *
+     * @param DocBlockHelper $docBlockHelper
```
loic425, line 34: not needed

src/Behat/Transformation/Transform.php
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Represents an Attribute for a Transform transformation
+ */
+#[\Attribute(\Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]
+final class Transform implements Transformation
+{
+    public ?string $pattern;
+
+    public function __construct(?string $pattern = null)
```
loic425, line 21: constructor promotion

src/Behat/Behat/Transformation/Context/Attribute/TrasnformationAttributeReader.php
```diff
@@ -0,0 +1,70 @@
...
+use Behat\Behat\Transformation\Context\Annotation\TransformationAnnotationReader;
+use Behat\Transformation as Attribute;
+use ReflectionMethod;
+
+/**
+ * Reads transformation Attributes from the context class.
+ *
+ * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ */
+final class TrasnformationAttributeReader extends TransformationAnnotationReader implements AttributeReader
```
stof, line 24:
```suggestion
final class TransformationAttributeReader extends TransformationAnnotationReader implements AttributeReader
```

stof, line 24: Why extending the annotation reader ? Apparently, you are *calling* the public API of the annotation reader as part of your logic, so this could use delegation instead of inheritance.
Or maybe it should not reuse the TransformationAnnotationReader, avoiding to concatenate the pattern just to extract it again (and making it easier to deprecate annotations in the future if their implementation is not a dependency of the attribute reader)

src/Behat/Behat/Transformation/Context/Attribute/TrasnformationAttributeReader.php
```diff
@@ -0,0 +1,70 @@
...
+{
+    /**
+     * @var DocBlockHelper
+     */
+    private $docBlockHelper;
+
+    /**
+     * Initializes reader.
```
stof, line 32: this description is not needed either (any `__construct` method is about initializing the instance, that's how PHP works)

src/Behat/Behat/Transformation/Context/Attribute/TrasnformationAttributeReader.php
```diff
@@ -0,0 +1,70 @@
...
+     * @{inheritdoc}
+     */
+    public function readCallees(string $contextClass, ReflectionMethod $method)
+    {
+        $attributes = $method->getAttributes(Attribute\Transformation::class, \ReflectionAttribute::IS_INSTANCEOF);
+
+        $callees = [];
+        foreach ($attributes as $attribute) {
+            $docLine = '@Transform';
+            $pattern = $attribute->newInstance()->pattern;;
```
stof, line 51: the `Transformation` interface does not guarantee the existence of such property. If your code supports only the `Transform` class, why not using that class directly when loading attributes, instead of adding the internal interface that works only for that class ?

features/definitions_transformations_attribute.feature
carlos-granados: @loic425 most of this code comes from copying existing code but you are right that we should use the opportunity to use modern PHP features when possible, I have updated my PR

src/Behat/Behat/Transformation/Context/Attribute/TrasnformationAttributeReader.php
carlos-granados: I had to read this twice to find the difference 🤣 Good spot!

carlos-granados: My initial implementation was much more tied to the AnnotationReader but you are right that the final implementation does not need inheritance. What I have done is move the common functionality to another utility class. Thanks for the hint

src/Behat/Behat/Transformation/Context/Attribute/TrasnformationAttributeReader.php
carlos-granados: I followed the structure used for other attributes which they all define a common interface for each class. But you are right that if there is only one attribute of this class the interface is not really needed, removed

src/Behat/Behat/Transformation/Context/Attribute/TransformationAttributeReader.php
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * @return Transformation[]
+     */
+    public function readCallees(string $contextClass, ReflectionMethod $method): array
+    {
+        $attributes = $method->getAttributes(Attribute\Transform::class);
+
+        $callees = [];
+        foreach ($attributes as $attribute) {
+            $pattern = $attribute->newInstance()->pattern ?? '';
```
stof, line 41: should we even make the pattern nullable in the attribute ? Can't we use the empty string as default value instead, and a non nullable property ?

src/Behat/Behat/Transformation/Context/Factory/TransformationCalleeFactory.php
```diff
@@ -0,0 +1,59 @@
...
+use Behat\Behat\Transformation\Transformation\TokenNameAndReturnTypeTransformation;
+use Behat\Behat\Transformation\Transformation\TokenNameTransformation;
+use ReflectionMethod;
+
+/**
+ * Generates the callee for a transformation
+ *
+ * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ */
+class TransformationCalleeFactory
```
stof, line 28: I suggest marking this class a `@internal`. I don't think we want to make it covered by BC.

src/Behat/Behat/Transformation/Context/Attribute/TransformationAttributeReader.php
carlos-granados: Yeah, makes sense

src/Behat/Transformation/Transform.php
```diff
@@ -0,0 +1,23 @@
...
+namespace Behat\Transformation;
+
+/**
+ * Represents an Attribute for a Transform transformation
+ */
+#[\Attribute(\Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]
+final class Transform
+{
+    public function __construct(
+        public string $pattern = ''
```
stof, line 20: should we make it readonly as it is public ? I don't think we should allow mutating the attribute instance.

carlos-granados: I did not add `readonly` because I thought that was PHP 8.2. Now I realise that PHP 8.2 introduced readonly classes, readonly properties were available in PHP 8.1. It's hard to keep track of all these changes! 🤣 Updated

acoulton (approved):
LGTM, thanks

PR #1546, opened 2024-11-26
PHP Config extensions

loic425:
Referenced issue #1539

Example:
```php
use Behat\Config\Config;
use Behat\Config\Profile;
use Behat\Config\Suite;
use Behat\MinkExtension;
use FriendsOfBehat\MinkDebugExtension;

$profile = (new Profile())
    ->withExtension(new Extension(MinkExtension::class, [
        'base_url' => 'https://127.0.0.1:8080',
    ]))
    ->withExtension(new Extension(MinkDebugExtension::class, [
        'directory' => 'etc/build',
    ]))
;

return (new Config)->withProfile($profile);
```

### DIFF
--- a/src/Behat/Config/Extension.php
+++ b/src/Behat/Config/Extension.php
@@ -1,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Behat\Config;
+
+final class Extension implements ExtensionConfigInterface
+{
+    public function __construct(
+        private string $name,
+        private array $settings = [],
+    ) {
+    }
+
+    public function name(): string
+    {
+        return $this->name;
+    }
+
+    public function toArray(): array
+    {
+        return $this->settings;
+    }
+}
--- a/src/Behat/Config/Profile.php
+++ b/src/Behat/Config/Profile.php
@@ -32,6 +32,10 @@ final class Profile
 
     public function withExtension(ExtensionConfigInterface $extension): self
     {
+        if (array_key_exists($extension->name(), $this->settings['extensions'] ?? [])) {
+            throw new ConfigurationLoadingException(sprintf('The extension "%s" already exists.', $extension->name()));
+        }
+
         $this->settings['extensions'][$extension->name()] = $extension->toArray();
 
         return $this;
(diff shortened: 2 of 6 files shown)


src/Behat/Config/Extension.php
```diff
@@ -0,0 +1,30 @@
...
+declare(strict_types=1);
+
+namespace Behat\Config;
+
+final class Extension implements ExtensionInterface
+{
+    public function __construct(
+        private string $name,
+        private array $settings = [],
+        private string $profile = 'default',
```
acoulton, line 12: I think this is just left over from the original?

src/Behat/Config/ExtensionInterface.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Behat\Config;
+
+interface ExtensionInterface extends ConfigInterface
```
acoulton, line 7: As with #1547 I would prefer `ExtensionConfigInterface` - especially for this one, as I hope extension developers will implement it themselves to provide custom DTOs / helpers, so we want to make it really clear to them / in their codebase what it is.

acoulton (commented):
Apart from the naming question to resolve and a small fix now that `Profile` is separate, it LGTM, thanks.

src/Behat/Config/ExtensionInterface.php
loic425: done

loic425:
@acoulton I've rebased the PR to fix conflicts
@carlos-granados Ok for you too?

src/Behat/Config/Profile.php
```diff
@@ -30,6 +30,13 @@ public function name(): string
         return $this->name;
     }
 
+    public function withExtension(ExtensionConfigInterface $extension): self
+    {
+        $this->settings['extensions'][$extension->name()] = $extension->toArray();
```
carlos-granados, line 35: Shouldn't we check here that there is not already a config for this extension in the same way that we checked that there was not a duplicate config for a suite?

carlos-granados (commented):
Just one question

src/Behat/Config/Profile.php
loic425: Good catch!

loic425: done

carlos-granados (approved):
All good now, thanks @loic425

PR #1547, opened 2024-11-27
PHP Config profiles

loic425:
Referenced issue #1539

Example:
```php
use Behat\Config\Config;
use Behat\Config\Profile;

$config = new Config();
$config->withProfile(new Profile('default', [
    'extensions' => [
        'some_extension' => [],
    ],
]));

return $config;
```

```php
use Behat\Config\Config;
use Behat\Config\Profile;

$config = new Config(['imports' => ['pretty.yml']]);
$config
    ->withProfile(new Profile('default', [
        'formatters' => [
            'pretty' => false,
            'progress' => null,
        ],
    ]))
    ->withProfile(new Profile('pretty_without_paths', [
        'formatters' => [
            'progress' => false,
            'pretty' => [
                'paths' => false,
            ],
        ],
    ]))
;
return $config;
```

### DIFF
--- a/features/extensions.feature
+++ b/features/extensions.feature
@@ -4,13 +4,26 @@ Feature: Extensions
   I need to be able to write simple extensions
 
   Background:
-    Given a file named "behat.yml" with:
+    Given a file named "behat.php" with:
       """
-      default:
-        extensions:
-          custom_extension.php:
-            param1: val1
-            param2: val2
+      <?php
+
+      use Behat\Config\Config;
+      use Behat\Config\Profile;
+
+      $profile = new Profile('default', [
+        'extensions' => [
+          'custom_extension.php' => [
+            'param1' => 'val1',
+            'param2' => 'val2',
+          ],
+        ],
+      ]);
+
+      $config = new Config();
+      $config->withProfile($profile);
+
+      return $config;
       """
     And a file named "features/bootstrap/FeatureContext.php" with:
       """
--- a/src/Behat/Config/Config.php
+++ b/src/Behat/Config/Config.php
@@ -25,6 +25,13 @@ final class Config implements ConfigInterface
         return $this;
     }
 
+    public function withProfile(Profile $profile): self
+    {
+        $this->settings[$profile->name()] = $profile->toArray();
+
+        return $this;
+    }
+
     public function toArray(): array
     {
         return $this->settings;
--- a/src/Behat/Config/Config.php
+++ b/src/Behat/Config/Config.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Behat\Config;
 
+use Behat\Testwork\ServiceContainer\Exception\ConfigurationLoadingException;
 use function is_string;
 
 final class Config implements ConfigInterface
@@ -27,6 +28,10 @@ final class Config implements ConfigInterface
 
     public function withProfile(Profile $profile): self
     {
+        if (array_key_exists($profile->name(), $this->settings)) {
+            throw new ConfigurationLoadingException(sprintf('The profile "%s" already exists.', $profile->name()));
+        }
+
         $this->settings[$profile->name()] = $profile->toArray();
 
         return $this;
--- a/features/profiles.feature
+++ b/features/profiles.feature
@@ -80,21 +80,32 @@ Feature: Profiles
           progress: false
           pretty: ~
       """
-    And a file named "behat.yml" with:
+    And a file named "behat.php" with:
       """
-      default:
-        formatters:
-          pretty:   false
-          progress: ~
+      <?php
 
-      pretty_without_paths:
-        formatters:
-          progress: false
-          pretty:
-            paths: false
+      use Behat\Config\Config;
+      use Behat\Config\Profile;
 
-      imports:
-        - pretty.yml
+      $config = new Config(['imports' => ['pretty.yml']]);
+      $config
+        ->withProfile(new Profile('default', [
+          'formatters' => [
+            'pretty' => false,
+            'progress' => null,
+          ],
+        ]))
+        ->withProfile(new Profile('pretty_without_paths', [
+          'formatters' => [
+            'progress' => false,
+            'pretty' => [
+              'paths' => false,
+            ],
+          ],
+        ]))
+      ;
+
+      return $config;
       """
 
   Scenario:
--- a/src/Behat/Config/Profile.php
+++ b/src/Behat/Config/Profile.php
@@ -1,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Behat\Config;
+
+final class Profile
+{
+    public function __construct(
+        private string $name,
+        private array $settings = []
+    ) {
+    }
+
+    public function name(): string
+    {
+        return $this->name;
+    }
+
+    public function toArray(): array
+    {
+        return $this->settings;
+    }
+}
--- a/tests/Behat/Tests/Config/ConfigTest.php
+++ b/tests/Behat/Tests/Config/ConfigTest.php
@@ -6,6 +6,7 @@ namespace Behat\Tests\Config;
 
 use Behat\Config\Config;
 use Behat\Config\Profile;
+use Behat\Testwork\ServiceContainer\Exception\ConfigurationLoadingException;
 use PHPUnit\Framework\TestCase;
 
 final class ConfigTest extends TestCase
@@ -67,5 +68,21 @@ final class ConfigTest extends TestCase
                 ],
             ],
         ], $config->toArray());
+    }
+
+    public function testItThrowsAnExceptionWhenAddingExistingProfile(): void
+    {
+        $config = new Config();
+
+        $config->withProfile(new Profile('default'));
+
+        $this->expectException(ConfigurationLoadingException::class);
+        $this->expectExceptionMessage('The profile "default" already exists.');
+
+        $config->withProfile(new Profile('default', [
+            'extensions' => [
+                'some_extension' => [],
+            ],
+        ]));
     }
 }
--- a/tests/Behat/Tests/Config/ConfigTest.php
+++ b/tests/Behat/Tests/Config/ConfigTest.php
@@ -1,8 +1,11 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Behat\Tests\Config;
 
 use Behat\Config\Config;
+use Behat\Config\Profile;
 use PHPUnit\Framework\TestCase;
 
 final class ConfigTest extends TestCase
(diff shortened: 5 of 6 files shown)


loic425:
@carlos-granados I didn't add any Behat tests, do you think we'll need some? cause it will be used via withExtension and withSuite in the Behat tests later.

carlos-granados:
> @carlos-granados I didn't add any Behat tests, do you think we'll need some? cause it will be used via withExtension and withSuite in the Behat tests later.

I think that, since this is going to be merged independently, it should have its own tests

loic425:
> > @carlos-granados I didn't add any Behat tests, do you think we'll need some? cause it will be used via withExtension and withSuite in the Behat tests later.
> 
> I think that, since this is going to be merged independently, it should have its own tests

done :)

loic425:
> LGTM, thanks @loic425

Thx, I've detected an issue, I will submit it on this PR.

src/Behat/Testwork/ServiceContainer/Configuration/ConfigurationLoader.php
```diff
@@ -200,19 +200,25 @@ protected function loadFileConfiguration($configPath, $profile)
...
-            $config = $phpConfig->toArray();
+            $config = $this->getPHPConfigObject($configPath)->toArray();
         } else {
             $config = (array) Yaml::parse(file_get_contents($configPath));
         }
 
         return $this->loadConfigs($basePath, $config, $profile);
     }
 
+    private function getPHPConfigObject(string $configPath): ConfigInterface
```
loic425, line 211: @carlos-granados It prevents var collisions.
Now, the only possible collision is with "$configPath" in the throw exception.

carlos-granados: @loic425 I really don't understand why this change is needed, can you please explain? 🤔

features/extensions.feature
```diff
@@ -4,13 +4,26 @@ Feature: Extensions
...
-        extensions:
-          custom_extension.php:
-            param1: val1
-            param2: val2
+      <?php
+
+      use Behat\Config\Config;
+      use Behat\Config\Profile;
+
+      $profile = new Profile('default', [
```
loic425, line 14: I've added a `$profile` var here that proves the vars collision is fixed.

src/Behat/Testwork/ServiceContainer/Configuration/ConfigurationLoader.php
loic425: yeah indeed...

When using the "require" method. We override existing arguments from the current method (here the $profile var)

carlos-granados: Ahh, I get it now. Thanks!

loic425:
@acoulton Could you please make a review on this one too?
We're going on fast on the whole PHP config, that's great. Thanks a lot to both of you.

src/Behat/Testwork/ServiceContainer/Configuration/ConfigurationLoader.php
stof: if you want to reduce conflict even more, you could make this private method a pure wrapper isolating `require` by returning `mixed`, and validating the ConfigInterface type in the caller.

Also, this should probably be a static method to avoid giving access to `$this` in the required file. Or maybe even stricter by using a bound closure to also remove access to `self` (and so also potential access to private state of the current class): https://github.com/composer/composer/blob/5eeba719d3f1388f68773c28e3971a7f441ee919/src/Composer/Autoload/ClassLoader.php#L575-L577

loic425: @stof I've made the change, I hope this is exactly why you have described it.

src/Behat/Config/ProfileInterface.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Behat\Config;
+
+interface ProfileInterface extends ConfigInterface
```
acoulton, line 7: I think we should call all of these `XXXConfigInterface` (like you have `SuiteConfigInterface` in #1540). Otherwise for things like Suite, Extension, etc there is potential for confusion - at least if just glancing at the code - about whether you're expected to pass an actual instance of the extension rather than just a config object.

Equally, I am wondering if we definitely need separate interfaces and final classes for all of these, if the interface  for most of them is only going to have `name()` and `toArray()`. If the sugar methods are only defined / implemented on the final class then it will be harder for people to find those, and the interfaces add extra noise.

Since you have to use the classes we provide to get the sugar methods, and you always have the option to pass an array (or a class of your own that doesn't satisfy any interface) into the `$settings` of the container object, there are still plenty of ways you can customise the config without needing to be able to implement an interface?

The exception to this might be for extensions and formatters, where by definition we're not in control of the available config options so interfaces would be useful to allow extensions to define a custom config object of their own.

src/Behat/Config/Config.php
```diff
@@ -11,6 +11,13 @@ public function __construct(
     ) {
     }
 
+    public function withProfile(ProfileInterface $profile): self
+    {
+        $this->settings[$profile->name()] = $profile->toArray();
```
acoulton, line 16: What if they have already defined a profile with that name? Should we perhaps throw rather than silently overwriting (bearing in mind we're in the context of a single config here, obviously imports are merged together later).

The examples in the feature files already suggest that the PHP config files might become a lot more verbose / have more variables etc than the YAML ones - I think that increases the risk of accidentally having a duplicate key, and I wonder if we should protect against that?

acoulton (changes requested):
I'm finding it a bit hard to picture what the end-state of this will look like - based on the PRs open so far I have a slight concern that expressing everything with chained sugar methods (as opposed to named constructor arguments, for example) will make the config files become a lot more verbose and hard to follow.

Although I appreciate that sugar methods have the major advantage that it's very clear whether you've provided a value, which can't really be achieved even with optional method arguments.

I realise that routers, DI containers etc follow the sugar pattern, but the key difference is that they are part of the tested code, so a confusion over config should show up in CI. Whereas if the config for the test framework is not applied as the user expects, they could end up with CI giving false positives without realising.

I wonder if it's worth updating #1539 with a full worked example of a config with multiple profiles, suites, formatters, extensions etc so we can check we're happy with the overall shape of the API? 

It may well be this is the right path, but I'm worried that things might look OK when dealing with the initial simple objects and we only realise later that we've backed ourselves into a corner.

Particularly because we will want to keep releasing from master regularly, which means very soon we'll have to guarantee BC for any of the incremental PRs that get merged.

src/Behat/Config/ProfileInterface.php
loic425: @acoulton For the `XXXConfigInterface` part, I'm mitigated on that. This was an error on #1540.
If we use this suffix for the interface, we need a Config suffix on implementations SuiteConfig, ProfileConfig...

But maybe that's better... To be honest, naming is never easy :+1:

loic425: For the final classes, they contains default helpers/sugar methods, and users can define its own ones. But maybe that's too much control.

src/Behat/Config/Config.php
loic425: We could add an optional arguments `bool $overwrite = false`
But cause you are able to redefine the same profile configuration via another file, it will not work.

src/Behat/Testwork/ServiceContainer/Configuration/ConfigurationLoader.php
loic425: @carlos-granados Cause the PR is going to be harder than expected, maybe should I open another PR for this specific fix?

src/Behat/Config/Config.php
stof: No you cannot. Defining the same key twice in a Yaml config file is a Yaml syntax error.

loic425: I mean, in multiple yaml configuration files,  you are able to redefine a suite with same key as defined in another file.

src/Behat/Config/ProfileInterface.php
acoulton: > If we use this suffix for the interface, we need a Config suffix on implementations SuiteConfig, ProfileConfig...

I don't think we automatically *have to*. IMO they have different usecases:

* The interface is for when you're looking at the method you're calling to see what it accepts. Here, the priority is to make very clear to the end-user that `withExtension` takes an object with extension configuration, not e.g. an instance of an extension (which you might assume, if new).
* The concrete class is generally for when you're looking at / working in a config file. Here, it is already clear that you have config objects, and probably the priority is to keep the naming as concise as possible so that users are mostly looking at the actual config, rather than the boilerplate PHP structure.

So if we have interfaces, personally I would have `SuiteConfigInterface`, `ProfileConfigInterface`, `ExtensionConfigInterface` and then `SuiteConfig`, `ProfileConfig`, `ExtensionConfig` - or even just `Suite`, `Profile`, `Extension` (leaving the Config in the namespace).

acoulton: > For the final classes, they contains default helpers/sugar methods, and users can define its own ones. But maybe that's too much control.

I think it adds complexity (and a lot of very similar interfaces) for not much benefit. These objects (especially the bits that are defined by the interface) are just DTOs, and I wouldn't usually implement an interface for that.

In reality, if users want to use any of our helper methods, they'll mostly have to use our DTOs, because it's only the DTOs that have the helpers. If they don't, they can just do something like:

```
return new Behat\Config\Config([
  'default' => (new MyCustomProfileConfigObject(...))->toArray()
])
```

On the other hand, for the more common case of a user wanting to use the default helpers, it would be much easier to see that `->withProfile` takes a `Profile` or `ProfileConfig` object, and click directly through to that in their IDE rather than clicking first to an interface and then see what implements it.

> But if we define interfaces for extensions, I think it would be better to have the same control on Profiles, Suites etc.

Extensions and formatters are special cases, because unlike Profiles, Suites, etc we have no knowledge or control of what config schema they take. So it would be very common that an extension author would want to provide a config object that you could pass to `->withExtension()`.

acoulton (approved):
Thanks for all the work on this @loic425, LGTM. I'll ask @carlos-granados to re-review as it's changed a bit since he looked.

src/Behat/Config/ProfileInterface.php
loic425: I removed the interface.

carlos-granados (approved):
Goog job @loic425 🎉

PR #1548, opened 2024-11-28
Update to Psalm 5

carlos-granados:
- Updates Psalm to version 5, which is compatible with all our supported Symfony versions.
- Adds some fixes for some issues found by this new version

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -162,9 +162,6 @@ jobs:
           ini-values: "phar.readonly=0,zend.exception_ignore_args=Off"
           coverage: none
 
-      - name: Install dependencies
-        run: composer install --no-dev ${{ matrix.php == '8.4' && '--ignore-platform-req=php+' || '' }}
-
       - uses: actions/download-artifact@v4
         with:
           name: behat.phar
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -89,9 +89,29 @@ jobs:
         if: matrix.php >= 8.0
         run: ./bin/behat -fprogress --strict --tags=@php8
 
+  static-analysis:
+    name: Static analysis
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        #runs on the lowest and highest supported versions (change this to 8.4 once psalm supports it)
+        php: [8.1, 8.3]
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: "${{ matrix.php }}"
+          ini-values: "zend.exception_ignore_args=Off"
+          coverage: none
+
+      - name: Install dependencies
+        run: composer update
+
       - name: Run Psalm
-        if: matrix.php != 8.4 && matrix.composer-mode != 'lowest'
-        run: ./vendor/bin/psalm --output-format=github
+        run: ./vendor/bin/psalm --output-format=github --php-version=${{ matrix.php }}
 
   build-phar:
     name: Build PHAR file
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -69,11 +69,6 @@ jobs:
           composer config --global --no-plugins allow-plugins.symfony/flex true &&
           composer global require symfony/flex
 
-      # until psalm is updated to v5 which is compatible with Symfony 7
-      - name: Remove vimeo/psalm
-        if: matrix.symfony-version == '7.0'
-        run: composer remove vimeo/psalm --no-update --dev
-
       - name: Install latest dependencies
         if: matrix.composer-mode == 'update'
         env:
@@ -94,31 +89,8 @@ jobs:
         if: matrix.php >= 8.0
         run: ./bin/behat -fprogress --strict --tags=@php8
 
-  static-analysis:
-    name: Static analysis
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-
-      - name: Setup PHP
-        uses: shivammathur/setup-php@v2
-        with:
-          php-version: 8.1
-          ini-values: "zend.exception_ignore_args=Off"
-          coverage: none
-
-      # until psalm is updated to v5 which is compatible with Symfony 7
-      - name: Force symfony version for psalm compatibility
-        run: |
-          composer config --global --no-plugins allow-plugins.symfony/flex true &&
-          composer global require symfony/flex
-
-      - name: Install dependencies
-        env:
-          SYMFONY_REQUIRE: 5.4.*
-        run: composer update
-
       - name: Run Psalm
+        if: matrix.php != 8.4 && matrix.composer-mode != 'lowest'
         run: ./vendor/bin/psalm --output-format=github
 
   build-phar:
@@ -171,7 +143,7 @@ jobs:
           coverage: none
 
       - name: Install dependencies
-        run: composer install
+        run: composer install --no-dev ${{ matrix.php == '8.4' && '--ignore-platform-req=php+' || '' }}
 
       - uses: actions/download-artifact@v4
         with:
--- a/psalm.xml
+++ b/psalm.xml
@@ -2,15 +2,14 @@
 <psalm
     errorLevel="7"
     resolveFromConfigFile="true"
+    findUnusedBaselineEntry="false"
+    findUnusedCode="false"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xmlns="https://getpsalm.org/schema/config"
     xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
 >
     <projectFiles>
         <directory name="src" />
-        <ignoreFiles>
-            <directory name="vendor" />
-        </ignoreFiles>
     </projectFiles>
 
     <issueHandlers>
@@ -23,21 +22,6 @@
                 <file name="src/Behat/Behat/HelperContainer/Exception/NotFoundException.php" />
             </errorLevel>
         </UnrecognizedStatement>
-
-        <!-- Suppress errors due to PSR exceptions not implementing Throwable -->
-        <!-- https://github.com/vimeo/psalm/issues/1856 -->
-        <InvalidThrow>
-            <errorLevel type="suppress">
-                <referencedClass name="Psr\Container\ContainerExceptionInterface" />
-            </errorLevel>
-        </InvalidThrow>
-
-        <!-- suppress errors related to Symfony return types including enums -->
-        <UndefinedDocblockClass>
-            <errorLevel type="suppress">
-                <referencedClass name="UnitEnum" />
-            </errorLevel>
-        </UndefinedDocblockClass>
 
         <!-- Suppress errors related to legacy PHPUnit classes -->
         <UndefinedClass>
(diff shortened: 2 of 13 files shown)


.github/workflows/build.yml
```diff
@@ -69,11 +69,6 @@ jobs:
           composer config --global --no-plugins allow-plugins.symfony/flex true &&
           composer global require symfony/flex
 
-      # until psalm is updated to v5 which is compatible with Symfony 7
```
carlos-granados, line 72: This is not needed any longer

.github/workflows/build.yml
```diff
@@ -94,31 +89,8 @@ jobs:
         if: matrix.php >= 8.0
         run: ./bin/behat -fprogress --strict --tags=@php8
 
-  static-analysis:
```
carlos-granados, line 97: Since we now run Psalm for all Symfony versions, we run it as part of the tests instead of running as a separate job

.github/workflows/build.yml
```diff
@@ -94,31 +89,8 @@ jobs:
...
-          composer config --global --no-plugins allow-plugins.symfony/flex true &&
-          composer global require symfony/flex
-
-      - name: Install dependencies
-        env:
-          SYMFONY_REQUIRE: 5.4.*
-        run: composer update
-
       - name: Run Psalm
+        if: matrix.php != 8.4 && matrix.composer-mode != 'lowest'
```
carlos-granados, line 93: We can't run Psalm on PHP 8.4 because it is not yet compatible with this PHP version. Also, we don't want to run it with the lower version of dependencies because it shows some errors which have been fixed in later versions

.github/workflows/build.yml
```diff
@@ -171,7 +143,7 @@ jobs:
           coverage: none
 
       - name: Install dependencies
-        run: composer install
+        run: composer install --no-dev ${{ matrix.php == '8.4' && '--ignore-platform-req=php+' || '' }}
```
carlos-granados, line 146: When installing the dependencies to test the Phar we don't need the dev dependencies. And even if it they are not installed, composer does check them so we need to tell it to ignore the php version requirement or it will fail due to PHPStan 5

psalm.xml
```diff
@@ -2,15 +2,14 @@
 <psalm
     errorLevel="7"
     resolveFromConfigFile="true"
+    findUnusedBaselineEntry="false"
```
carlos-granados, line 5: These params now need to be set explicitly

psalm.xml
```diff
@@ -2,15 +2,14 @@
...
+    findUnusedCode="false"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xmlns="https://getpsalm.org/schema/config"
     xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
 >
     <projectFiles>
         <directory name="src" />
-        <ignoreFiles>
```
carlos-granados, line 11: We don't need to ignore `vendor` here because there is no such folder under `src`

psalm.xml
```diff
@@ -24,21 +23,6 @@
             </errorLevel>
         </UnrecognizedStatement>
 
-        <!-- Suppress errors due to PSR exceptions not implementing Throwable -->
```
carlos-granados, line 27: These errors are no longer shown

psalm.xml
```diff
@@ -56,7 +40,16 @@
         <MethodSignatureMismatch>
             <errorLevel type="suppress">
                 <file name="src/Behat/Testwork/EventDispatcher/TestworkEventDispatcherSymfonyLegacy.php" />
+                <file name="src/Behat/Testwork/EventDispatcher/TestworkEventDispatcherSymfony5.php" />
```
carlos-granados, line 43: The new PHPStan also finds this problem in this file

psalm.xml
```diff
@@ -56,7 +40,16 @@
...
             </errorLevel>
         </MethodSignatureMismatch>
+
+        <!-- Psalm does not correctly understand the spread operator being applied on array_merge -->
```
carlos-granados, line 47: New error which is incorrectly thrown by the new version

src/Behat/Behat/Context/Environment/Handler/ContextEnvironmentHandler.php
```diff
@@ -155,7 +155,7 @@ function ($context) {
      *
      * @param Suite $suite
      *
-     * @return string[]
+     * @return array<string|array>
```
carlos-granados, line 158: Incorrect tag, the array can also be an array of arrays

src/Behat/Testwork/Call/CallResults.php
```diff
@@ -18,6 +18,8 @@
  * Aggregates multiple call results into a collection and provides an informational API on top of that.
  *
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ *
```
carlos-granados, line 21: The `IteratorAggregate` interface is defined as a template, so we need to specify the type here when implementing it

src/Behat/Testwork/Call/RuntimeCallee.php
```diff
@@ -42,21 +42,13 @@ class RuntimeCallee implements Callee
...
      *
-     * @param callable    $callable
+     * @param callable $callable
      * @param null|string $description
      *
      * @throws BadCallbackException If invalid callback provided
      */
     public function __construct($callable, $description = null)
     {
-        if (!is_array($callable) && !is_callable($callable)) {
```
carlos-granados, line 52: If it is a `callable` then by definition it cannot be anything other than these two options, so this check is redundant. We should not be trying to check the types of arguments at run time, this should be done by static analysis tools or tests

src/Behat/Testwork/Specification/SpecificationArrayIterator.php
```diff
@@ -19,6 +19,7 @@
  * Return instance of this class from locator if specifications cannot be searched lazily.
  *
  * @author Christophe Coevoet <stof@notk.org>
+ * @extends ArrayIterator<int, mixed>
```
carlos-granados, line 22: `ArrayIterator` is defined as a template, so we need to specify the concrete type when extending it

.github/workflows/build.yml
stof: In the current setup, you run Psalm only if test succeed, which prevents getting the full feedback. I'd rather keep a separate job. It makes the feedback more understandable and allows getting the full feedback without iterating until we manage to get a green CI.

src/Behat/Behat/Context/Environment/Handler/ContextEnvironmentHandler.php
stof: can we describe the shape of this array ?

src/Behat/Testwork/Call/RuntimeCallee.php
stof: The proper fix is to use a `callable` parameter type (which was not possible due to the PHP requirement in older Behat versions).

src/Behat/Behat/Context/Environment/Handler/ContextEnvironmentHandler.php
carlos-granados: Not really sure how to best describe this shape because the arrays are also arrays of strings or other arrays and so on recursively, not sure how to best express this

.github/workflows/build.yml
stof: I think we don't need to install the dependencies at all. We don't use them

psalm.xml
stof: This is not really related to new Psalm. It is related to the fact that you removed the part forcing to use Symfony 5 in the static analysis job

carlos-granados: Ah, yes, you are right. Nevertheless, we need to add this to be able to run it with other Symfony versions

psalm.xml
carlos-granados: So, we have this code:
```
        return array_unique(
            array_merge(
                ...array_map(
                    function (Snippet $snippet) {
                        if (!$snippet instanceof ContextSnippet) {
                            return array();
                        }

                        return $snippet->getUsedClasses();
                    },
                    $this->snippets
                )
            )
        );
```

And PHPStan complains with:
`ERROR: NamedArgumentNotAllowed - src/Behat/Behat/Snippet/AggregateSnippet.php:116:20 - Method array_merge called with named unpacked array non-empty-array<array-key, array<array-key, string>> (array with string keys)`

So the issue seems to be that it understands that the unpack operator produces an `array-key` indexed array, which I think is wrong as it should be indexed by integers, as far as I know and this is what seems to be causing the issue. I tried several other approachs like saving the array result in another variable and type hinting it but could not get it to work. If you can think of a solution, suggestions are welcome

src/Behat/Testwork/Call/RuntimeCallee.php
carlos-granados: Needs to be `callable|array` because the arrays that we receive are not the new modern callables (object, string) but instead the old callables (string, string) and php throws an error if you just use `callable` for this case

.github/workflows/build.yml
carlos-granados: Absolutely right, removed

src/Behat/Testwork/Call/RuntimeCallee.php
stof: This is not true. `callable` is really any callable.

Unless we support passing a class name instead of an object for the first array item of callables meant to call a non-static method on an instance that will be created later. But then it means that the phpdoc is wrong.

psalm.xml
stof: `$this->snippets` is documented as `Snippet[]` which is a shortcut notation for `array<array-key, Snippet>`, not for `array<int, Snippet>`. So Psalm is actually right in reporting it right now.

src/Behat/Testwork/Call/RuntimeCallee.php
carlos-granados: > Unless we support passing a class name instead of an object for the first array item of callables meant to call a non-static method on an instance 

Yes, that is exactly what I meant, sorry if I did not explain it correctly.

psalm.xml
carlos-granados: Not sure why, but I thought that `Snippet[]` was a `list` of snippets, i.e. the keys had to be ints. You never stop learning! I modified the code and now the error suppression is not needed

.github/workflows/build.yml
```diff
@@ -97,29 +92,26 @@ jobs:
   static-analysis:
     name: Static analysis
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        #runs on the lowest and highest supported versions (change this to 8.4 once psalm supports it)
+        php: [8.1, 8.3]
```
acoulton, line 99: Just to confirm, I think this means we're now running with symfony 6 (on 8.1) and symfony 7 (on 8.3) but not with symfony 5. Is it worth adding symfony5 to the matrix as well, or is that not worth it for the cost / benefit?

src/Behat/Behat/Context/Environment/Handler/ContextEnvironmentHandler.php
acoulton: So the array case is when the user has specified constructor arguments for their context, I think? In which case the shape is entirely under user control (including that they can optionally specify named params, so the keys could be strings or ints).

It looks to me like these examples from the docs:
```yaml
contexts:
  - MyAwesomeContext:
        - http://localhost:8080
        - /var/tmp
  - MyAwesomeContext:
        baseUrl: http://localhost:8080
        tempPath: /var/tmp
```

stay exactly in that structure here:
```
[
  ['MyAwesomeContext' => ['http://localhost:8080', '/var/tmp'],
  ['MyAwesomeContext' => ['tempPath' => '/var/tmp', 'baseUrl' => 'http://localhost:8080'],
]
```

So I think the best we could do would be `list<string|array<class-string<Context>,array>>`? - unless there's a way to mark that the array version will have exactly **one** `class-string<Context>` key

src/Behat/Testwork/Call/RuntimeCallee.php
acoulton: Maybe we should deprecate calling with an old-style `[string,string]`?

acoulton (approved):
LGTM as-is, a couple of suggestions if you'd like to pursue them.

.github/workflows/build.yml
carlos-granados: Symfony 5.4 is now security fixes only, so I think it is not worth it

src/Behat/Testwork/Call/RuntimeCallee.php
carlos-granados: I think that before deprecating it we should probably change our code to use [object, string] arrays as currently we use [string, string] everywhere

acoulton: Ah, I hadn't realised we did - yes, definitely.

Issue #1379, opened 2021-11-03
Generate snippets with Attributes instead of Annotations


rpkamp:
Behat has full support for PHP 8 Attributes since version 3.10.0.

For example, this means that instead of 

```
/**
 * @Given something
 */
public function something() {}
```

it is now possible to write

```
#[Given('something')]
public function something() {}
```

and the same goes for `When`, `Then` and all hooks (`BeforeScenario` et al).

However, when generating snippets<sup>1</sup> Behat will use Annotations instead of Attributes. 

Behat should be able to generate snippets using Attributes instead of Annotions when applicable (ie on PHP >= 8). Note that when generating a snippet with an Attribute that Attribute should also be `use`d in the class where it's added, otherwise it won't work.

Since Behat supports PHP7.2+ at the time of writing it still needs to support generating snippets with Annotations, but it could default to Attributes when PHP >=8 is detected for example. This could also be an option, but I'm unsure if that is worth it. It feels like something that would be OK te enforce, since the entire ecosystem is moving away from Annotations to Attributes anyway. Curently `features/snippets.feature` tests for output with Annotations. Those scenarios could be run with a new tag `@php7` similar to the already existing `@php8` to prevent them from being checked on PHP8. And then the same scenarios but updated using attributes could be added with `@php8` tags. 

<sup>1</sup> A snippet can be generated when at the end of a run Behat has detected undefined steps. It can add a skeleton of the step to a feature context.

ciaranmcnulty:
There is an existing `--snippets-type` CLI option that hints what type of snippets to generate. We could easily add Annotations as a supported type, the question would be how to handle defaulting

rpkamp:
I looked into that. `--snippets-type` is for regex vs turnip. Annotations and attributes seem orthogonal to those options.

Maybe `--snippets-syntax` with options `annotations` and `attributes`? And when not supplied, default depends on PHP version, i.e. `attributes` for PHP >= 8 and `annotations` for PHP < 8?

rpkamp:
I think the smart thing for now would be to hold off on implementing this until PHPStorm supports Attributes for Behat.

Right now the DX is absolutely terrible, all the nice things you could do with Behat (find usages, ctrl+click to jump to definition from gherkin, etc) don't work with Attributes.

I did create an issue for this in Youtrack: https://youtrack.jetbrains.com/issue/WI-63963

pieterocp:
[...]

PR #1549, opened 2024-11-28
Add possibility to use attributes for Context snippets

fmatsos:
### Description
An old issue (cf. #1379) proposed to generate snippets with attributes in place of annotations.
I worked on it last days and finally came up with a working solution.

### Proposal
For now, I didn't switch all annotations to attributes.
I add a new CLI option to use attributes if we want, and an associated config key also.

So you can now do:

`behat -f progress --no-colors --snippets-for=FeatureContext --snippets-use-attributes` for example.

or add the new config key in `behat.yml` to use attributes by default:
```yaml
default:
  contexts:
    snippets:
      use_attributes: true
``` 

Although attributes are now widely supported, I think it's better to act in two phases: first, this PR, to implement attributes as an option. And later to invert and use attributes by default.

### DIFF
--- a/features/append_snippets.feature
+++ b/features/append_snippets.feature
@@ -123,6 +123,8 @@ Feature: Append snippets option
       """
       <?php
 
+      use Behat\Step\Given;
+      use Behat\Step\Then;
       use Behat\Behat\Context\Context,
           Behat\Behat\Tester\Exception\PendingException;
       use Behat\Gherkin\Node\PyStringNode,
@@ -182,41 +184,31 @@ Feature: Append snippets option
 
           private function doSomethingUndefinedWith() {}
 
-          /**
-           * @Then /^do something undefined with \$$/
-           */
+          #[Then('/^do something undefined with \$$/')]
           public function doSomethingUndefinedWith2(): void
           {
               throw new PendingException();
           }
 
-          /**
-           * @Then /^do something undefined with \\(\d+)$/
-           */
+          #[Then('/^do something undefined with \\\\(\d+)$/')]
           public function doSomethingUndefinedWith3($arg1): void
           {
               throw new PendingException();
           }
 
-          /**
-           * @Given /^pystring:$/
-           */
+          #[Given('/^pystring:$/')]
           public function pystring(PyStringNode $string): void
           {
               throw new PendingException();
           }
 
-          /**
-           * @Given /^pystring (\d+):$/
-           */
+          #[Given('/^pystring (\d+):$/')]
           public function pystring2($arg1, PyStringNode $string): void
           {
               throw new PendingException();
           }
 
-          /**
-           * @Given /^table:$/
-           */
+          #[Given('/^table:$/')]
           public function table(TableNode $table): void
           {
               throw new PendingException();
@@ -293,6 +285,8 @@ Feature: Append snippets option
       """
       <?php
 
+      use Behat\Step\Given;
+      use Behat\Step\Then;
       use Behat\Behat\Tester\Exception\PendingException;
       use Behat\Behat\Context\Context;
       use Behat\Gherkin\Node\PyStringNode,
@@ -352,41 +346,31 @@ Feature: Append snippets option
 
           private function doSomethingUndefinedWith() {}
 
-          /**
-           * @Then /^do something undefined with \$$/
-           */
+          #[Then('/^do something undefined with \$$/')]
           public function doSomethingUndefinedWith2(): void
           {
               throw new PendingException();
           }
 
-          /**
-           * @Then /^do something undefined with \\(\d+)$/
-           */
+          #[Then('/^do something undefined with \\\\(\d+)$/')]
           public function doSomethingUndefinedWith3($arg1): void
           {
               throw new PendingException();
           }
 
-          /**
-           * @Given /^pystring:$/
-           */
+          #[Given('/^pystring:$/')]
           public function pystring(PyStringNode $string): void
           {
               throw new PendingException();
           }
 
-          /**
-           * @Given /^pystring (\d+):$/
-           */
+          #[Given('/^pystring (\d+):$/')]
           public function pystring2($arg1, PyStringNode $string): void
           {
               throw new PendingException();
           }
 
-          /**
-           * @Given /^table:$/
-           */
+          #[Given('/^table:$/')]
           public function table(TableNode $table): void
           {
               throw new PendingException();
@@ -463,6 +447,8 @@ Feature: Append snippets option
 
       use Behat\Gherkin\Node\TableNode;
       use Behat\Gherkin\Node\PyStringNode;
+      use Behat\Step\Given;
+      use Behat\Step\Then;
       use Behat\Behat\Tester\Exception\PendingException;
       use Behat\Behat\Context\Context;
 
@@ -520,41 +506,31 @@ Feature: Append snippets option
 
           private function doSomethingUndefinedWith() {}
 
-          /**
-           * @Then /^do something undefined with \$$/
-           */
+          #[Then('/^do something undefined with \$$/')]
           public function doSomethingUndefinedWith2(): void
           {
               throw new PendingException();
           }
 
-          /**
-           * @Then /^do something undefined with \\(\d+)$/
-           */
+          #[Then('/^do something undefined with \\\\(\d+)$/')]
           public function doSomethingUndefinedWith3($arg1): void
           {
               throw new PendingException();
           }
 
-          /**
-           * @Given /^pystring:$/
-           */
+          #[Given('/^pystring:$/')]
           public function pystring(PyStringNode $string): void
           {
               throw new PendingException();
           }
 
-          /**
-           * @Given /^pystring (\d+):$/
-           */
+          #[Given('/^pystring (\d+):$/')]
           public function pystring2($arg1, PyStringNode $string): void
           {
               throw new PendingException();
           }
 
-          /**
-           * @Given /^table:$/
-           */
+          #[Given('/^table:$/')]
           public function table(TableNode $table): void
           {
               throw new PendingException();
@@ -641,6 +617,9 @@ Feature: Append snippets option
       """
       <?php
 
+      use Behat\Step\Then;
+      use Behat\Step\When;
+      use Behat\Step\Given;
       use Behat\Behat\Tester\Exception\PendingException;
       use Behat\Behat\Context\Context;
       use Behat\Gherkin\Node\TableNode;
@@ -649,81 +628,61 @@ Feature: Append snippets option
       class FirstContext implements Context
       {
 
-          /**
-           * @Given /^I have (\d+) apples$/
-           */
+          #[Given('/^I have (\d+) apples$/')]
           public function iHaveApples($arg1): void
           {
               throw new PendingException();
           }
 
-          /**
-           * @When /^I ate (\d+) apple$/
-           */
+          #[When('/^I ate (\d+) apple$/')]
           public function iAteApple($arg1): void
           {
               throw new PendingException();
           }
 
-          /**
-           * @Then /^I should have (\d+) apples$/
-           */
+          #[Then('/^I should have (\d+) apples$/')]
           public function iShouldHaveApples($arg1): void
           {
               throw new PendingException();
           }
 
-          /**
-           * @When /^I found (\d+) apples$/
-           */
+          #[When('/^I found (\d+) apples$/')]
           public function iFoundApples($arg1): void
           {
               throw new PendingException();
           }
 
-          /**
-           * @Then /^do something undefined with \$$/
-           */
+          #[Then('/^do something undefined with \$$/')]
           public function doSomethingUndefinedWith(): void
           {
               throw new PendingException();
           }
 
-          /**
-           * @When /^I ate (\d+) apples$/
-           */
+          #[When('/^I ate (\d+) apples$/')]
           public function iAteApples($arg1): void
           {
               throw new PendingException();
           }
 
-          /**
-           * @Then /^do something undefined with \\(\d+)$/
-           */
+          #[Then('/^do something undefined with \\\\(\d+)$/')]
           public function doSomethingUndefinedWith2($arg1): void
           {
               throw new PendingException();
           }
 
-          /**
-           * @Given /^pystring:$/
-           */
+          #[Given('/^pystring:$/')]
           public function pystring(PyStringNode $string): void
           {
               throw new PendingException();
           }
 
-          /**
-           * @Given /^pystring (\d+):$/
-           */
+          #[Given('/^pystring (\d+):$/')]
           public function pystring2($arg1, PyStringNode $string): void
           {
               throw new PendingException();
           }
 
-          /**
-           * @Given /^table:$/
-           */
+          #[Given('/^table:$/')]
           public function table(TableNode $table): void
           {
               throw new PendingException();
--- a/features/snippets.feature
+++ b/features/snippets.feature
@@ -112,6 +112,14 @@ Feature: Snippets generation and addition
           {
               throw new PendingException();
           }
+
+      --- Don't forget these 5 use statements:
+
+          use Behat\Behat\Tester\Exception\PendingException;
+          use Behat\Step\Given;
+          use Behat\Step\When;
+          use Behat\Step\Then;
+          use Behat\Gherkin\Node\PyStringNode;
       """
 
   Scenario: Appending regex snippets to a particular context
@@ -203,6 +211,14 @@ Feature: Snippets generation and addition
           {
               throw new PendingException();
           }
+
+      --- Don't forget these 5 use statements:
+
+          use Behat\Behat\Tester\Exception\PendingException;
+          use Behat\Step\Given;
+          use Behat\Step\When;
+          use Behat\Step\Then;
+          use Behat\Gherkin\Node\PyStringNode;
       """
 
   Scenario: Appending turnip snippets to a particular context
@@ -298,6 +314,11 @@ Feature: Snippets generation and addition
           {
               throw new PendingException();
           }
+
+      --- Don't forget these 2 use statements:
+
+          use Behat\Behat\Tester\Exception\PendingException;
+          use Behat\Step\Given;
       """
 
   Scenario: Generating snippets for steps with slashes
@@ -330,6 +351,11 @@ Feature: Snippets generation and addition
           {
               throw new PendingException();
           }
+
+      --- Don't forget these 2 use statements:
+
+          use Behat\Behat\Tester\Exception\PendingException;
+          use Behat\Step\Then;
       """
 
   Scenario: Generating snippets using interactive --snippets-for
@@ -385,6 +411,14 @@ Feature: Snippets generation and addition
           {
               throw new PendingException();
           }
+
+      --- Don't forget these 5 use statements:
+
+          use Behat\Behat\Tester\Exception\PendingException;
+          use Behat\Step\Given;
+          use Behat\Step\When;
+          use Behat\Step\Then;
+          use Behat\Gherkin\Node\PyStringNode;
       """
 
   Scenario: Generating snippets for steps with apostrophes
@@ -431,4 +465,11 @@ Feature: Snippets generation and addition
           {
               throw new PendingException();
           }
+
+      --- Don't forget these 4 use statements:
+
+          use Behat\Behat\Tester\Exception\PendingException;
+          use Behat\Step\Given;
+          use Behat\Step\When;
+          use Behat\Step\Then;
       """
--- a/features/snippets.feature
+++ b/features/snippets.feature
@@ -59,73 +59,55 @@ Feature: Snippets generation and addition
 
       --- FeatureContext has missing steps. Define them with these snippets:
 
-          /**
-           * @Given /^I have magically created (\d+)\$$/
-           */
+          #[Given('/^I have magically created (\d+)\$$/')]
           public function iHaveMagicallyCreated($arg1): void
           {
               throw new PendingException();
           }
 
-          /**
-           * @When /^I have chosen '([^']*)' in coffee machine$/
-           */
+          #[When('/^I have chosen \'([^\']*)\' in coffee machine$/')]
           public function iHaveChosenInCoffeeMachine($arg1): void
           {
               throw new PendingException();
           }
 
-          /**
-           * @Then /^I should have '([^']*)'$/
-           */
+          #[Then('/^I should have \'([^\']*)\'$/')]
           public function iShouldHave($arg1): void
           {
               throw new PendingException();
           }
 
-          /**
-           * @Then /^I should get a '([^']*)':$/
-           */
+          #[Then('/^I should get a \'([^\']*)\':$/')]
           public function iShouldGetA($arg1, PyStringNode $string): void
           {
               throw new PendingException();
           }
 
-          /**
-           * @Then /^I should get a simple string:$/
-           */
+          #[Then('/^I should get a simple string:$/')]
           public function iShouldGetASimpleString(PyStringNode $string): void
           {
               throw new PendingException();
           }
 
-          /**
-           * @When /^I have chosen "([^"]*)" in coffee machine$/
-           */
+          #[When('/^I have chosen "([^"]*)" in coffee machine$/')]
           public function iHaveChosenInCoffeeMachine2($arg1): void
           {
               throw new PendingException();
           }
 
-          /**
-           * @When /^do something undefined with \\(\d+)$/
-           */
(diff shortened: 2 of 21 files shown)


fmatsos:
PR is in draft since I need to fix certain tests yet. But you can already review the code and let me know :)

stof:
I would rather migrate to attributes directly, to avoid adding a new CLI flag and configuration setting that must then be covered by backward compatibility

fmatsos:
Okay, I'm agree with that. I was very conservative on this point, I see the problem with potentials BCs after. So I can switch to attributes by default of course, and update tests accordingly.

carlos-granados:
@fmatsos yes, as we commented on the issue you linked, we prefer to just move to generating attributes directly. In the next days I plan to work on updating our documentation to use attributes and just mention annotations as an option that will be deprecated and dropped in the future

features/append_snippets.feature
```diff
@@ -22,44 +22,32 @@ Feature: Append snippets option
               $this->parameters = $parameters;
           }
 
-          /**
-           * @Given /^I have (\d+) apples?$/
-           */
+          #[Given('/^I have (\d+) apples?$/')]
```
carlos-granados, line 25: You need to add the `use` statements for all these attributes, for example 

`use Behat\Step\Given;`

src/Behat/Behat/Context/Snippet/Generator/ContextSnippetGenerator.php
```diff
@@ -28,72 +30,53 @@
  */
 final class ContextSnippetGenerator implements SnippetGenerator
 {
-    /**
```
carlos-granados, line 31: Please keep this comment as it provides info about the array shape for this property

src/Behat/Behat/Context/Snippet/Generator/ContextSnippetGenerator.php
```diff
@@ -174,52 +148,46 @@ private function getMethodArguments(StepNode $step, $tokenCount)
...
 
         return $usedClasses;
     }
 
     /**
      * Generates snippet template using regex, method name and arguments.
-     *
-     * @param string   $pattern
-     * @param string   $methodName
-     * @param string[] $methodArguments
```
carlos-granados, line 201: Please keep the info about the $methodArguments parameter as it provides info about the array shape

src/Behat/Behat/Context/Snippet/Generator/ContextSnippetGenerator.php
```diff
@@ -317,43 +262,30 @@ private function getMethodNameNotProposedEarlier($contextClass, $stepPattern, $n
 
     /**
      * Returns already proposed method names.
-     *
-     * @param string $contextClass
-     *
-     * @return string[]
```
carlos-granados, line 323: Please keep the info about the return type

src/Behat/Behat/Context/Snippet/Generator/ContextSnippetGenerator.php
```diff
@@ -317,43 +262,30 @@ private function getMethodNameNotProposedEarlier($contextClass, $stepPattern, $n
...
-    private function getMethodArgument($argument)
+    private function getMethodArgument(mixed $argument): string
     {
-        $arg = '__unknown__';
-        if ($argument instanceof PyStringNode) {
-            $arg = 'PyStringNode $string';
-        } elseif ($argument instanceof TableNode) {
-            $arg = 'TableNode $table';
-        }
+        $arg = match (true) {
```
carlos-granados, line 284: You can get rid of this temporary variable and just return the result of the match

src/Behat/Behat/Snippet/Generator/SnippetGenerator.php
```diff
@@ -26,21 +26,11 @@ interface SnippetGenerator
...
     /**
      * Checks if generator supports search query.
-     *
-     * @param Environment $environment
-     * @param StepNode    $step
-     *
-     * @return bool
      */
-    public function supportsEnvironmentAndStep(Environment $environment, StepNode $step);
+    public function supportsEnvironmentAndStep(Environment $environment, StepNode $step): bool;
```
carlos-granados, line 30: Adding these return types in the interface is a BC break. If an external extension implements this interface and they have not added these return types, php will throw an error because the declaration of these functions in their classes does not match the declaration in the interface, so please remove them
It is ok to leave the return types in the `ContextSnippetGenerator` class because adding more types in a derived or implementing class is OK and the class is final, so there can't be any other class out there extending it

carlos-granados (changes requested):
Many thanks for this work @fmatsos There are a few changes needed, please see my comments

src/Behat/Behat/Context/Snippet/Generator/ContextSnippetGenerator.php
acoulton: Although I think `string[string]` is not correct - is that even valid? Shouldn't it be `array<string,array<string,string>>`?

acoulton (changes requested):
As you will need to work on the feature files anyway to address @carlos-granados feedback, could I also please ask you to split your final commit `test: update tests to check features in place of annotations` into two:

* A first one that *only* updates the features / expectations that are to do with generating snippets, where the output has changed because of the code changes in this branch.
* A separate one (if you wish) to convert existing features to use attributes instead of annotations - these aren't strictly required for this PR and it would be easier to see the diff for the feature itself, and to see why the tests are now failing, if they were kept in a separate commit.

Thanks!

fmatsos:
Your feedback has been processed, and I've also made a few additional changes to add required uses for attributes. I've rollback the test updates to separate them into two commits as suggested by Robert.

The next commit for tests, which will replace all annotations with attributes, will come when I've solved two remaining problems. I have two scenarios that no longer work (locally):.

```
    features/legacy_snippets.feature:119
    features/snippets.feature:117
```

I'm looking why I get these errors now.

Thanks again for your feedback!

features/append_snippets.feature
```diff
@@ -123,6 +123,8 @@ Feature: Append snippets option
       """
       <?php
 
+      use Behat\Step\Given;
+      use Behat\Step\Then;
       use Behat\Behat\Context\Context,
           Behat\Behat\Tester\Exception\PendingException;
       use Behat\Gherkin\Node\PyStringNode,
```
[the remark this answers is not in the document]
carlos-granados: I would not bother with that. The code is correct without ordering and if the user wants to order them, there are other tools out there that can do this easily

fmatsos:
> The next commit for tests, which will replace all annotations with attributes, will come when I've solved two remaining problems. I have two scenarios that no longer work (locally):.
> 
> ```
>     features/legacy_snippets.feature:119
>     features/snippets.feature:117
> ```

I finally found the problem. On some regex patterns, like `\(\d+)`, backslashe before paranthesis is not escaped as required. So regex is not valid.

src/Behat/Behat/Definition/Call/Given.php
```diff
@@ -17,6 +17,8 @@
  */
 final class Given extends RuntimeDefinition
 {
+    public static string $keyword = 'Given';
```
acoulton, line 20: Can these be `const` rather than static variables?

acoulton (commented):
This is taking shape well.

How hard would it be to output the additional use statements when displaying the "define them with these snippets" message to the user? Or to generate (or convert) snippets for display as FQCN.

If not, could we add a generic "You may also need to add use statements for the `Behat\Step\(Given/When/Then)` attributes" that always appears?

I'm just thinking for things like PendingException/TableNode they will get a runtime error if the class isn't imported. But by design if an attribute is undefined/not imported it will just be ignored, which could cause confusion if the user copies and pastes the snippet without realising.

fmatsos:
> How hard would it be to output the additional use statements when displaying the "define them with these snippets" message to the user? Or to generate (or convert) snippets for display as FQCN.

Tell me if I understand correctly @acoulton: you ask if it'll not be better to remove the use statements from snippets for attributes, and just display their FQCN (or generic message) to user instead?
In which case I can't give a response now but I can take a look.

carlos-granados:
@fmatsos there are two cases where we use the snippets:
- When they are directly added to the context class
- When they are displayed and the user needs to take care of adding them to the class.

In the first case, the ideal would be that the use statements were added directly without user intervention. In the second case it would be good if we could print the use statements as part of the snippet output so that the user could also copy them

I think that what @acoulton meant was that if these two options were not possible, we should look into using the FQN in the snippets, but only if we could not add the use statements as that is a better solution.

@acoulton please confirm

acoulton:
@fmatsos sorry that I wasn't clear.

Yes, exactly as @carlos-granados says, in the second case where we display snippets for the user to add I think in order of preference:

* We should print the use statements (if any) as well as the snippets;
* Or we should print snippets with their FQCN for the user to import when they paste into their file;
* Or worst case we should just always print a fixed message to remind them that they might need to add use statements.

fmatsos:
Hi!

I'm always stuck with this error:
`The regex '/^do something undefined with \(\d+)$/' is invalid: preg_match(): Compilation failed: unmatched closing parenthesis at offset 34`

Occurred in files:
```
    features/legacy_snippets.feature:119
    features/snippets.feature:117
```

I did some research, set-by-step debugging, and still not find how to fix it...
I don't understand why backslash before the opening bracket is not escaped. The same is escaped rightly for other tests.

carlos-granados:
I'll try to debug this later today

carlos-granados:
I added a fix, you need to convert the `\\` that is used in the expression to represent a backlash to `\\\\`. This is read by php as `\\` and the regular expression interpreter reads it as `\`. The beauty of PHP and regular expressions 😆 🤯

fmatsos:
Thanks a lot @carlos-granados!

I began to be crazy about this lol
Since all the other tests passed, I didn't think at all it was the expression that needed fixing. You save me from a big headache!

acoulton:
One thing I've found that can be helpful with `\\` hell is to switch to a nowdoc (`<<<'REGEX'`) even temporarily for debugging, as PHP reads them literally without parsing/escaping any `\` characters

fmatsos:
> One thing I've found that can be helpful with `\\` hell is to switch to a nowdoc (`<<<'REGEX'`) even temporarily for debugging, as PHP reads them literally without parsing/escaping any `\` characters

Nice to know this trick!

FIY I'm on your request about use statements messages. Once it's done, this PR will be okay for final review. I'll do another PR to update the rest features from annotations to attributes (I think it'll be better for review and better to have two separate PRs for that) :)

acoulton:
@fmatsos perfect, that sounds like a good plan.

i18n.php
```diff
@@ -20,6 +21,7 @@
     'bg'    => array(
         'snippet_context_choice'  => 'В сет <snippet_undefined><snippet_keyword>%count%</snippet_keyword> има недекларирани стъпки. Изберете в кой Context да бъдат създадени:</snippet_undefined>',
         'snippet_proposal_title'  => '<snippet_undefined><snippet_keyword>%count%</snippet_keyword> има липсващи стъпки. Можете да ги създадете чрез:</snippet_undefined>',
+        'snippet_proposal_use'    => 'Не забравяйте тези %count% use изрази:',
```
fmatsos, line 24: I didn't know if I should translate to other languages or not. So I used a translation tool to get a generic translated sentence for each language (except for English and French, which use plural correctly).

features/snippets.feature
```diff
@@ -59,77 +59,67 @@ Feature: Snippets generation and addition
...
               throw new PendingException();
           }
+
+      --- Don't forget these 5 use statements:
+
+          use Behat\Behat\Tester\Exception\PendingException;
+          use Behat\Step\Given;
+          use Behat\Step\When;
+          use Behat\Step\Then;
+          use Behat\Gherkin\Node\PyStringNode;
```
fmatsos, line 122: I only test the output for use statements in the snippets feature. Let me know if you want to test the output in other features (I think it's not really necessary, but I may be wrong).

i18n.php
acoulton: Brilliant, thanks - I can't really check any of the languages but the English & French look right and the German looks convincing :laughing:

features/snippets.feature
acoulton: I agree, we don't need it everywhere. The only other place I guess might be the i18n feature to prove your translation is coming through OK?

It's really helpful to have the full list with the PendingException, PyStringNode etc as well - thank you.

src/Behat/Behat/Snippet/Printer/ConsoleSnippetPrinter.php
```diff
@@ -57,13 +57,20 @@ public function __construct(OutputInterface $output, TranslatorInterface $transl
...
+        $usedClasses = [];
         foreach ($snippets as $snippet) {
+            foreach ($snippet->getUsedClasses() as $usedClass) {
+                $usedClasses[$usedClass] = true;
+            }
+
             $this->output->writeln(sprintf('<snippet_undefined>%s</snippet_undefined>', $snippet->getSnippet()) . PHP_EOL);
         }
+
+        $this->outputClassesUsesStatements($usedClasses);
```
acoulton, line 73: Slightly cleaner IMO and solves that I think the PHPDoc typehint ` * @param array<string> $usedClasses` below is not currently correct as it's actually getting `array<string,true>` at the moment?

```suggestion
        $this->outputClassesUsesStatements(array_keys($usedClasses));
```

src/Behat/Behat/Snippet/Printer/ConsoleSnippetPrinter.php
```diff
@@ -84,4 +91,24 @@ public function printUndefinedSteps($suiteName, array $steps)
...
+
+        if (0 === $classCount) {
+            return;
+        }
+
+        $message = $this->translator->trans('snippet_proposal_use', ['%count%' => $classCount], 'output');
+
+        $this->output->writeln('--- '.$message.PHP_EOL);
+
+        foreach (array_keys($usedClasses) as $usedClass) {
```
acoulton, line 110: `array_keys` moved to where this method is called.

```suggestion
        foreach ($usedClasses as $usedClass) {
```

acoulton (changes requested):
One tiny fix, but otherwise this is looking great thank you :tada:

features/snippets.feature
fmatsos: I've also included the output test in the i18n feature file 😄

acoulton (approved):
Brilliant, thanks @fmatsos

loic425:
Congratulations @fmatsos ;) This will be really great to use.

i18n.php
```diff
@@ -218,6 +231,7 @@
     'pt'    => array(
         'snippet_proposal_title' => '<snippet_undefined><snippet_keyword>%count%</snippet_keyword> contém definições em falta. Defina-as com estes exemplos:</snippet_undefined>',
         'snippet_missing_title'  => '<snippet_undefined>Os exemplos para as seguintes definições da suite <snippet_keyword>%count%</snippet_keyword> não foram gerados (verifique a configuração):</snippet_undefined>',
+        'snippet_proposal_use'   => 'Não te esqueças destas %count% declarações de uso:',
```
carlos-granados, line 234: Should be `de use` (`use` should not be translated)

i18n.php
```diff
@@ -2,6 +2,7 @@
     'en'    => array(
         'snippet_context_choice'  => '<snippet_undefined><snippet_keyword>%count%</snippet_keyword> suite has undefined steps. Please choose the context to generate snippets:</snippet_undefined>',
         'snippet_proposal_title'  => '<snippet_undefined><snippet_keyword>%count%</snippet_keyword> has missing steps. Define them with these snippets:</snippet_undefined>',
+        'snippet_proposal_use'    => '{1} Don\'t forget this use statement:|]1,Inf]Don\'t forget these %count% use statements:',
```
carlos-granados, line 5: I don't think the singular is really needed. We will always be printing at least 2 use statements, the one for the current step and the PendingException

i18n.php
```diff
@@ -234,6 +248,7 @@
     'pt-BR' => array(
         'snippet_proposal_title' => '<snippet_undefined><snippet_keyword>%count%</snippet_keyword> possue etapas faltando. Defina elas com esse(s) trecho(s) de código:</snippet_undefined>',
         'snippet_missing_title'  => '<snippet_undefined>Trecho de códigos para as seguintes etapas em <snippet_keyword>%count%</snippet_keyword> suite não foram geradas (verique sua configuração):</snippet_undefined>',
+        'snippet_proposal_use'   => 'Não se esqueça destas %count% declarações de uso:',
```
carlos-granados, line 251: Same here

i18n.php
```diff
@@ -250,6 +265,7 @@
     'ro'    => array(
         'snippet_proposal_title'  => '<snippet_undefined><snippet_keyword>%count%</snippet_keyword> are pași lipsa. Puteți implementa pașii cu ajutorul acestor fragmente de cod:</snippet_undefined>',
         'snippet_missing_title'   => '<snippet_undefined>Fragmentele de cod pentru urmatorii pași din suita <snippet_keyword>%count%</snippet_keyword> nu au fost generate (contextul tau implementeaza interfata SnippetAcceptingContext?):</snippet_undefined>',
+        'snippet_proposal_use'    => 'Nu uitați aceste %count% instrucțiuni de utilizare:',
```
carlos-granados, line 268: Should be `de use`

src/Behat/Behat/Context/Snippet/ContextSnippet.php
```diff
@@ -20,32 +20,23 @@
...
     /**
      * Initializes definition snippet.
      *
-     * @param StepNode $step
-     * @param string   $template
-     * @param string   $contextClass
      * @param string[] $usedClasses
      */
-    public function __construct(StepNode $step, $template, $contextClass, array $usedClasses = array())
+    public function __construct(StepNode $step, string $template, string $contextClass, array $usedClasses = [])
```
carlos-granados, line 39: Since we are modernizing this code, maybe use constructor property promotion?

src/Behat/Behat/Context/Snippet/Generator/FixedContextIdentifier.php
```diff
@@ -19,25 +19,20 @@
...
-    private $contextClass;
+    private ?string $contextClass;
 
     /**
      * Initialises identifier.
-     *
-     * @param string $contextClass
      */
-    public function __construct($contextClass)
+    public function __construct(?string $contextClass = null)
```
carlos-granados, line 27: Maybe use constructor property promotion here as well?

src/Behat/Behat/Context/Snippet/Generator/FixedPatternIdentifier.php
```diff
@@ -10,34 +10,27 @@
...
-    private $patternType;
+    private ?string $patternType;
 
     /**
      * Initialises identifier.
-     *
-     * @param string $patternType
      */
-    public function __construct($patternType)
+    public function __construct(?string $patternType = null)
```
carlos-granados, line 25: And here too

src/Behat/Behat/Snippet/Printer/ConsoleSnippetPrinter.php
```diff
@@ -84,4 +91,24 @@ public function printUndefinedSteps($suiteName, array $steps)
 
         $this->output->writeln('');
     }
+
+    /**
+     * @param array<string> $usedClasses
+     */
+    public function outputClassesUsesStatements(array $usedClasses): void
+    {
+        $classCount = \count($usedClasses);
```
carlos-granados, line 100: I prefer to use `if ($usedClasses === [])`, faster and shorter, feel free to ignore

carlos-granados (changes requested):
Great job overall @fmatsos, just a few small comments

i18n.php
fmatsos: Correct!

src/Behat/Behat/Context/Snippet/ContextSnippet.php
fmatsos: I agree :)

I was thinking of proposing this kind of modification in another PR, across the entire codebase.

src/Behat/Behat/Snippet/Printer/ConsoleSnippetPrinter.php
fmatsos: Oh! That's an oversight, I prefer this syntax too 😄

carlos-granados (approved):
All good now @fmatsos Do you think you might rebase it to the latest master? Also you may want to squash some of the commits

fmatsos:
Hi @carlos-granados! 
I squashed into 3 commits. The branch was also already up to date with master.
Everything should be ok now 😃 

Have a good evening!

src/Behat/Behat/Context/Snippet/Generator/ContextSnippetGenerator.php
```diff
@@ -197,30 +180,29 @@ private function getUsedClasses(StepNode $step)
...
+            $this->preparePattern($pattern),
             $methodName,
             implode(', ', $methodArguments)
         );
     }
 
+    private function preparePattern(string $pattern): string
+    {
+        $pattern = str_replace('%', '%%', $pattern);
+        $pattern = str_replace('\\\\', '\\\\\\\\', $pattern);
```
stof, line 198: instead of those replacements, I suggest removing the quotes from the template and using `var_export()` to make PHP generate the proper literal representation of the string with any needed escaping.

src/Behat/Behat/Context/Snippet/Generator/ContextSnippetGenerator.php
```diff
@@ -319,36 +278,27 @@ private function getMethodNameNotProposedEarlier($contextClass, $stepPattern, $n
     /**
      * Returns already proposed method names.
      *
-     * @param string $contextClass
-     *
-     * @return string[]
+     * @return array<string, array<string, string>
```
stof, line 281: This type is wrong. It does not return an array of arrays as it returns the proposed methods for a single context class. the proper return type is `array<string, string>`

Btw, the current annotation is even totally broken because it misses a `>` to be a valid type.

src/Behat/Behat/Context/Snippet/Generator/ContextSnippetGenerator.php
fmatsos: You suggest something like that?

```php
        $pattern = str_replace('%', '%%', $pattern);

        return var_export($pattern, true);
```

4 tests from snippets.feature fail with this change, since some `\` are now escaped too. I presume it's normal and expected.

src/Behat/Behat/Context/Snippet/Generator/ContextSnippetGenerator.php
acoulton: Well spotted @stof, sorry I missed that. @fmatsos do you have time to fix that? Thanks

src/Behat/Behat/Context/Snippet/Generator/ContextSnippetGenerator.php
acoulton: @fmatsos I'm guessing that is what @stof meant, it looks ok to me. Could you clarify about the tests - is it that snippets.feature is currently valid and using var_export gives the wrong result? Or is it that the current expectations in snippets.feature are not correctly escaped and using var_export fixes that? Thanks

src/Behat/Behat/Context/Snippet/Generator/ContextSnippetGenerator.php
fmatsos: Yes, I already have the fix on a new branch with the next proposal ;)

src/Behat/Behat/Context/Snippet/Generator/ContextSnippetGenerator.php
fmatsos: Of course! I'll test the final regex to check if new pattern is valid or not.

fmatsos: Hi @stof 

I check deeply your suggestion, but it seems I can't use `var_export` in this case. If I use it, I have `\` from regex keywords that are escaped too. So the pattern is not valid anymore.

Example:
- expected: `#[Then('/^do something undefined with \\(\d+)$/')]`
- result: `#[Then(''/^do something undefined with \\\\(\\d+)$/'')]`

src/Behat/Behat/Context/Snippet/Generator/ContextSnippetGenerator.php
acoulton: Since the var_export question is resolved, I've just pushed a fix for this in #1569

PR #1551, opened 2024-12-01
chore: remove mentions to php 8

carlos-granados:
There were some places in our code where we were taking specific actions for PHP 8. This is not needed any longer as we don't support a lower version any more. This PR removes those special cases

### DIFF
--- a/behat.yml.dist
+++ b/behat.yml.dist
@@ -1,4 +1,1 @@
 default:
-  gherkin:
-    filters:
-      tags: ~@php8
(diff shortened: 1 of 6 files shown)


behat.yml.dist
```diff
@@ -1,4 +1 @@
 default:
-  gherkin:
```
carlos-granados, line 2: Without this config, this file is not actually needed but I thought it would be good to keep it as a reference for anyone looking at the code and trying to find where we configure things

tests/Behat/Tests/Config/ConfigTest.php
```diff
@@ -20,7 +20,7 @@ public function testItReturnsSettings(): void
             'default' => [
                 'gherkin' => [
                     'filters' => [
-                        'tags' => '~@php8'
+                        'tags' => '~@test'
```
carlos-granados, line 23: This change is not strictly necessary but I thought it would be better to remove any mention to PHP 8

acoulton (approved):

PR #1554, opened 2024-12-03
chore: first step towards Psalm level 6

carlos-granados:
We should be increasing our Psalm level. Moving to error level 6 produces more than 100 errors, so, to make this more manageable, we can introduce the level 6 rules in small batches. This PR introduces one of these rules with the corresponding needed changes

### DIFF
--- a/src/Behat/Behat/Output/Node/EventListener/AST/OutlineListener.php
+++ b/src/Behat/Behat/Output/Node/EventListener/AST/OutlineListener.php
@@ -25,7 +25,6 @@ use Behat\Gherkin\Node\ExampleNode;
 use Behat\Testwork\Event\Event;
 use Behat\Testwork\Output\Formatter;
 use Behat\Testwork\Output\Node\EventListener\EventListener;
-use phpDocumentor\Reflection\DocBlock\Tags\Example;
 
 /**
  * Listens to expanded outline events and calls appropriate printers.
--- a/src/Behat/Behat/Output/Node/EventListener/AST/OutlineListener.php
+++ b/src/Behat/Behat/Output/Node/EventListener/AST/OutlineListener.php
@@ -139,9 +139,8 @@ final class OutlineListener implements EventListener
         $this->exampleSetupPrinter->printSetup($formatter, $event->getSetup());
 
         $this->example = $event->getScenario();
-        if ($this->example instanceof ExampleNode) {
-            $this->examplePrinter->printHeader($formatter, $event->getFeature(), $this->example);
-        }
+        assert($this->example instanceof ExampleNode);
+        $this->examplePrinter->printHeader($formatter, $event->getFeature(), $this->example);
     }
 
     /**
--- a/src/Behat/Behat/Output/Node/EventListener/AST/OutlineListener.php
+++ b/src/Behat/Behat/Output/Node/EventListener/AST/OutlineListener.php
@@ -25,6 +25,7 @@ use Behat\Gherkin\Node\ExampleNode;
 use Behat\Testwork\Event\Event;
 use Behat\Testwork\Output\Formatter;
 use Behat\Testwork\Output\Node\EventListener\EventListener;
+use phpDocumentor\Reflection\DocBlock\Tags\Example;
 
 /**
  * Listens to expanded outline events and calls appropriate printers.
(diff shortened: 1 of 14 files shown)


src/Behat/Behat/Output/Node/Printer/JUnit/JUnitFeaturePrinter.php
```diff
@@ -69,7 +69,7 @@ public function printFooter(Formatter $formatter, TestResult $result)
         if (0 === count($stats)) {
             $totalCount = 0;
         } else {
-            $totalCount = array_sum($stats);
+            $totalCount = (int) array_sum($stats);
```
carlos-granados, line 72: `$totalCount` needs to be an int

src/Behat/Behat/Output/Node/EventListener/AST/OutlineListener.php
```diff
@@ -25,6 +25,7 @@
 use Behat\Testwork\Event\Event;
 use Behat\Testwork\Output\Formatter;
 use Behat\Testwork\Output\Node\EventListener\EventListener;
+use phpDocumentor\Reflection\DocBlock\Tags\Example;
```
stof, line 28: this use statement looks wrong.

src/Behat/Behat/Output/Statistics/TotalStatistics.php
```diff
@@ -175,9 +175,9 @@ public function registerHookStat(HookStat $stat)
     /**
      * Returns counters for different scenario result codes.
      *
-     * @return array[]
+     * @return int[]
```
stof, line 178: I suggest documenting this as `array<TestResults::*, int>` to be more precise about the array keys.

src/Behat/Behat/Output/Node/Printer/Pretty/PrettySetupPrinter.php
```diff
@@ -148,14 +149,15 @@ private function printTeardownHookCallResult(OutputPrinter $printer, CallResult
         $style = $this->resultConverter->convertResultCodeToString($resultCode);
         $hook = $callResult->getCall()->getCallee();
         $path = $hook->getPath();
+        $hookName = (string)$hook;
```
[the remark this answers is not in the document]
carlos-granados: Updated, thanks!!

src/Behat/Behat/Output/Node/EventListener/AST/OutlineListener.php
```diff
@@ -136,10 +137,12 @@ private function printExampleHeaderOnBeforeExampleEvent(Formatter $formatter, Ev
             return;
         }
 
-        $this->example = $event->getScenario();
-
         $this->exampleSetupPrinter->printSetup($formatter, $event->getSetup());
-        $this->examplePrinter->printHeader($formatter, $event->getFeature(), $this->example);
+
+        $this->example = $event->getScenario();
```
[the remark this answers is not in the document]
acoulton: The structure of the way the event listeners is registered is somewhat complex, but I **think** the assumption is that events only reach this listener if we're in the course of running an Outline and therefore the scenario **should** be an ExampleNode?

If so, getting an unexpected type of ScenarioNode here would suggest either there's a problem with our event binding, or with our assumptions about how the nodes / events are structured. 

In which case it might be better / safer to throw an exception (and hopefully fail CI) rather than silently not print anything? This would be the intentional version of the current behaviour where it will hit a TypeError.

src/Behat/Behat/Output/Node/EventListener/AST/OutlineTableListener.php
```diff
@@ -215,7 +216,9 @@ private function printExampleRowOnAfterExampleEvent(Formatter $formatter, Event
             $this->stepSetupPrinter->printSetup($formatter, $beforeEvent->getSetup());
         }
 
-        $this->exampleRowPrinter->printExampleRow($formatter, $this->outline, $example, $this->stepAfterTestedEvents);
+        if ($example instanceof ExampleNode) {
+            $this->exampleRowPrinter->printExampleRow($formatter, $this->outline, $example, $this->stepAfterTestedEvents);
+        }
```
acoulton, line 221: Same observation as the other Outline-related listener above, I think perhaps we should throw?

acoulton (commented):
Good work @carlos-granados just a question about how best to handle unexpected types

src/Behat/Behat/Output/Node/EventListener/AST/OutlineTableListener.php
carlos-granados: I've updated the code to use assertions, which is what PHPStan recommends for this kind of type narrowing. What do you think?

acoulton: That looks like a good solution to me

acoulton (approved):

PR #1555, opened 2024-12-03
dx: add composer scripts for testing tools

carlos-granados:
While developing for Behat I found myself running the commands that run the tests quite often. To make this easier this PR introduces some composer scripts, including one that will run all your tests. The script that runs all the tests runs behat in progress mode, while the one that runs it individually runs it in pretty mode

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -70,7 +70,7 @@
             "@phpunit",
             "@psalm"
         ],
-        "behat": "LANG=C bin/behat",
+        "behat": "LANG=C bin/behat  --rerun",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -19,13 +19,13 @@
         "behat/gherkin": "^4.10.0",
         "behat/transliterator": "^1.5",
         "composer-runtime-api": "^2.2",
-        "symfony/console": "^5.4 || ^6.4 || ^7.0",
+        "psr/container": "^1.0 || ^2.0",
```
carlos-granados, line 22: Just re-ordered these dependencies

PR #1556, opened 2024-12-03
PHP Config - Add Profile filters

loic425:
Referenced issue #1539

```php

use Behat\Config\Config;
use Behat\Config\Profile;
use Behat\Config\Filter\NameFilter;
use Behat\Config\Filter\TagFilter;

return (new Config())
    ->withProfile(
        (new Profile('default'))
            ->withFilter(new NameFilter('name1'))
            ->withFilter(new TagFilter('tag2'))
  )
;
```


loic425:
@acoulton @carlos-granados 
I think this is now ready to review. I prefer my today implementation.

src/Behat/Config/Gherkin/Filter/FilterInterface.php
```diff
@@ -0,0 +1,12 @@
...
+
+declare(strict_types=1);
+
+namespace Behat\Config\Gherkin\Filter;
+
+interface FilterInterface
+{
+    public function name(): string;
+
+    public function value(): mixed;
```
loic425, line 11: the interface value returns a mixed but the other current filter return a string.
I'm not aware if some other filters can have other value types, that's why, I've specified mixed on the interface.

src/Behat/Config/Gherkin/Filter/FilterInterface.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Behat\Config\Gherkin\Filter;
```
carlos-granados, line 5: The filters can also be applied on the suite level, so perhaps it would be better to remove the `Gherkin` part of the namespace? I know that the files that implement the filtering are in this Gherkin namespace but I don't think that forces us to use the same here

src/Behat/Config/Gherkin/Filter/FilterInterface.php
carlos-granados: Right now the types of filters is limited and there is no way to add new filters, so I would say this can be a string

src/Behat/Config/Gherkin/SimpleFilter.php
```diff
@@ -0,0 +1,26 @@
+<?php
```
carlos-granados, line 1: I don't think this is needed. If you want to create one of the existing filters just use the corresponding class. And any new filters that you add will not be handled, Behat will throw an exception

carlos-granados: Unless perhaps all previous filters extend this SimpleFilter with most of the functionality delegated to it

carlos-granados (commented):
Some comments

src/Behat/Config/Gherkin/SimpleFilter.php
loic425: Yes, I forgot to remove it. I think we can start simple without it at all.

src/Behat/Config/Gherkin/Filter/FilterInterface.php
loic425: done

src/Behat/Config/Gherkin/Filter/FilterInterface.php
loic425: done

src/Behat/Config/Gherkin/SimpleFilter.php
loic425: I've just removed it.

carlos-granados (approved):
LGTM, thanks @loic425

acoulton (approved):
Great, thanks @loic425

PR #1557, opened 2024-12-04
PHP Config - suite filters

loic425:
I'm wondering if we should implement `withFilters(FilterInterface ...$filters) `instead.
But for now, we have implemented them like this on Profile.

```php
use Behat\Config\Config;
use Behat\Config\Profile;
use Behat\Config\Suite;
use Behat\Config\Filter\RoleFilter;

$profile = (new Profile('default'))
    ->withSuite(
        (new Suite('little_kid'))
            ->withContexts(LittleKidContext::class)
            ->withFilter(new RoleFilter('little kid'))
        )
    ->withSuite(
        (new Suite('big_brother'))
            ->withContexts(BigBrotherContext::class)
            ->withFilter(new RoleFilter('big brother'))
        )
;

```

### DIFF
--- a/tests/Behat/Tests/Config/SuiteTest.php
+++ b/tests/Behat/Tests/Config/SuiteTest.php
@@ -7,6 +7,7 @@ namespace Behat\Tests\Config;
 use Behat\Config\Filter\NameFilter;
 use Behat\Config\Filter\TagFilter;
 use Behat\Config\Suite;
+use Behat\Testwork\ServiceContainer\Exception\ConfigurationLoadingException;
 use PHPUnit\Framework\TestCase;
 
 final class SuiteTest extends TestCase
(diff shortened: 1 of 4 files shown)


loic425:
@acoulton @carlos-granados 
easy to review now we have the previous one.

acoulton (approved):
LGTM. 

I think I'd stick with `->withFilter()` - there will often only be one, or at most a very short list, and in time we might want to add helpers like e.g. `(new TagFilter())->include('ui')->exclude('php84')` which would be neatest if each filter was in a separate method call.

carlos-granados:
> LGTM.
> 
> I think I'd stick with `->withFilter()` - there will often only be one, or at most a very short list, and in time we might want to add helpers like e.g. `(new TagFilter())->include('ui')->exclude('php84')` which would be neatest if each filter was in a separate method call.

Agreed

tests/Behat/Tests/Config/SuiteTest.php
```diff
@@ -61,4 +65,18 @@ public function testAddingPaths(): void
             ]
         ], $config->toArray());
     }
+
+    public function testAddingFilters(): void
```
carlos-granados, line 69: You are missing the test to confirm that you cannot repeat a filter

carlos-granados (commented):
Looks good, just one comment

tests/Behat/Tests/Config/SuiteTest.php
loic425: done

carlos-granados (approved):
Thanks @loic425

PR #1558, opened 2024-12-05
PHP Config - profile formatters

loic425:
```php
use Behat\Config\Config;
use Behat\Config\Profile;
use Behat\Config\Formatter\PrettyFormatter;
use Behat\Config\Formatter\ProgressFormatter;

$profile = (new Profile('pretty_without_paths'))
    ->disableFormatter(ProgressFormatter::NAME)
    ->withFormatter(new PrettyFormatter(paths: false))
;

``

### DIFF
--- a/features/preferred_profile.feature
+++ b/features/preferred_profile.feature
@@ -73,13 +73,21 @@ Feature: Preferred Profiles
             | 5       | 3       | 8      |
             | 5       | 5       | 10     |
       """
-    And a file named "pretty.yml" with:
+    And a file named "pretty.php" with:
       """
-      pretty_without_paths:
-        formatters:
-          progress: false
-          pretty:
-            paths: false
+      <?php
+
+      use Behat\Config\Config;
+      use Behat\Config\Profile;
+      use Behat\Config\Formatter\PrettyFormatter;
+      use Behat\Config\Formatter\ProgressFormatter;
+
+      $profile = (new Profile('pretty_without_paths'))
+        ->disableFormatter(ProgressFormatter::NAME)
+        ->withFormatter(new PrettyFormatter(paths: false))
+      ;
+
+      return (new Config())->withProfile($profile);
 
       """
     And a file named "behat.yml" with:
@@ -95,7 +103,7 @@ Feature: Preferred Profiles
         progress
 
       imports:
-        - pretty.yml
+        - pretty.php
       """
 
   Scenario:
--- a/src/Behat/Behat/Output/ServiceContainer/Formatter/JUnitFormatterFactory.php
+++ b/src/Behat/Behat/Output/ServiceContainer/Formatter/JUnitFormatterFactory.php
@@ -10,6 +10,7 @@
 
 namespace Behat\Behat\Output\ServiceContainer\Formatter;
 
+use Behat\Config\Formatter\JUnitFormatter;
 use Behat\Testwork\Exception\ServiceContainer\ExceptionExtension;
 use Behat\Testwork\Output\ServiceContainer\Formatter\FormatterFactory;
 use Behat\Testwork\Output\ServiceContainer\OutputExtension;
@@ -143,11 +144,9 @@ final class JUnitFormatterFactory implements FormatterFactory
         $container->setDefinition('output.junit.statistics', $definition);
 
         $definition = new Definition('Behat\Testwork\Output\NodeEventListeningFormatter', array(
-            'junit',
+            JUnitFormatter::NAME,
             'Outputs the failures in JUnit compatible files.',
-            array(
-                'timer' => true,
-            ),
+            JUnitFormatter::defaults(),
             $this->createOutputPrinterDefinition(),
             new Definition('Behat\Testwork\Output\Node\EventListener\ChainEventListener', array(
                 array(
--- a/src/Behat/Behat/Output/ServiceContainer/Formatter/PrettyFormatterFactory.php
+++ b/src/Behat/Behat/Output/ServiceContainer/Formatter/PrettyFormatterFactory.php
@@ -14,6 +14,7 @@ use Behat\Behat\Definition\ServiceContainer\DefinitionExtension;
 use Behat\Behat\EventDispatcher\Event\BackgroundTested;
 use Behat\Behat\EventDispatcher\Event\OutlineTested;
 use Behat\Behat\EventDispatcher\Event\ScenarioTested;
+use Behat\Config\Formatter\PrettyFormatter;
 use Behat\Testwork\Exception\ServiceContainer\ExceptionExtension;
 use Behat\Testwork\Output\ServiceContainer\Formatter\FormatterFactory;
 use Behat\Testwork\Output\ServiceContainer\OutputExtension;
@@ -171,14 +172,9 @@ class PrettyFormatterFactory implements FormatterFactory
         $container->setDefinition('output.pretty.statistics', $definition);
 
         $definition = new Definition('Behat\Testwork\Output\NodeEventListeningFormatter', array(
-            'pretty',
+            PrettyFormatter::NAME,
             'Prints the feature as is.',
-            array(
-                'timer'     => true,
-                'expand'    => false,
-                'paths'     => true,
-                'multiline' => true,
-            ),
+            PrettyFormatter::defaults(),
             $this->createOutputPrinterDefinition(),
             new Definition('Behat\Testwork\Output\Node\EventListener\ChainEventListener', array(
                     array(
--- a/src/Behat/Behat/Output/ServiceContainer/Formatter/ProgressFormatterFactory.php
+++ b/src/Behat/Behat/Output/ServiceContainer/Formatter/ProgressFormatterFactory.php
@@ -10,6 +10,7 @@
 
 namespace Behat\Behat\Output\ServiceContainer\Formatter;
 
+use Behat\Config\Formatter\ProgressFormatter;
 use Behat\Testwork\Exception\ServiceContainer\ExceptionExtension;
 use Behat\Testwork\Output\ServiceContainer\Formatter\FormatterFactory;
 use Behat\Testwork\Output\ServiceContainer\OutputExtension;
@@ -139,11 +140,9 @@ class ProgressFormatterFactory implements FormatterFactory
         $container->setDefinition('output.progress.statistics', $definition);
 
         $definition = new Definition('Behat\Testwork\Output\NodeEventListeningFormatter', array(
-            'progress',
+            ProgressFormatter::NAME,
             'Prints one character per step.',
-            array(
-                'timer' => true
-            ),
+            ProgressFormatter::defaults(),
             $this->createOutputPrinterDefinition(),
             new Definition('Behat\Testwork\Output\Node\EventListener\ChainEventListener', array(
                     array(
--- a/src/Behat/Config/Formatter/Formatter.php
+++ b/src/Behat/Config/Formatter/Formatter.php
@@ -1,0 +1,62 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Behat\Config\Formatter;
+
+use Behat\Testwork\Output\Printer\Factory\OutputFactory;
+
+class Formatter implements FormatterConfigInterface
+{
+    public function __construct(
+        private readonly string $name,
+        private array $settings = [],
+    ) {
+    }
+
+    /**
+     * @param int $level use OutputFactory::VERBOSITY_*
+     */
+    public function withOutputVerbosity(int $level): self
+    {
+        $this->settings['output_verbosity'] = $level;
+
+        return $this;
+    }
+
+    public function withOutputPath(string $path): self
+    {
+        $this->settings['output_path'] = $path;
+
+        return $this;
+    }
+
+    public function withOutputDecorated(bool $decorated = true): self
+    {
+        $this->settings['output_decorate'] = $decorated;
+
+        return $this;
+    }
+
+    public function withOutputStyles(array $styles): self
+    {
+        $this->settings['output_styles'] = $styles;
+
+        return $this;
+    }
+
+    public function name(): string
+    {
+        return $this->name;
+    }
+
+    public function toArray(): array
+    {
+        return $this->settings;
+    }
+
+    public static function defaults(): array
+    {
+        return [];
+    }
+}
--- a/src/Behat/Config/Formatter/FormatterConfigInterface.php
+++ b/src/Behat/Config/Formatter/FormatterConfigInterface.php
@@ -1,0 +1,14 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Behat\Config\Formatter;
+
+use Behat\Config\ConfigInterface;
+
+interface FormatterConfigInterface extends ConfigInterface
+{
+    public function name(): string;
+
+    public static function defaults(): array;
+}
--- a/src/Behat/Config/Formatter/JUnitFormatter.php
+++ b/src/Behat/Config/Formatter/JUnitFormatter.php
@@ -1,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Behat\Config\Formatter;
+
+final class JUnitFormatter extends Formatter
+{
+    public const NAME = 'junit';
+
+    public function __construct()
+    {
+        parent::__construct(name: self::NAME);
+    }
+}
--- a/src/Behat/Config/Formatter/PrettyFormatter.php
+++ b/src/Behat/Config/Formatter/PrettyFormatter.php
@@ -1,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Behat\Config\Formatter;
+
+final class PrettyFormatter extends Formatter
+{
+    public const NAME = 'pretty';
+
+    /**
+     * @param bool $timer show time and memory usage at the end of the test run
+     * @param bool $expand print each example of a scenario outline separately
+     * @param bool $paths display the file path and line number for each scenario
+     *                    and the context file and method for each step
+     * @param bool $multiline print out PyStrings and TableNodes in full
+     */
+    public function __construct(
+        bool $timer = true,
+        bool $expand = false,
+        bool $paths = true,
+        bool $multiline = true,
+    ) {
+        parent::__construct(name: self::NAME, settings: [
+            'timer' => $timer,
+            'expand' => $expand,
+            'paths' => $paths,
+            'multiline' => $multiline,
+        ]);
+    }
+
+    public static function defaults(): array
+    {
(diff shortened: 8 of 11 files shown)

### CODE REVIEW

**Comment by loic425** [comment]
@carlos-granados I'm not that familiar with filters feature.
Could you explain me the different way and options we have to use them?

The documentation is not great about them.

**Comment by acoulton** [comment]
@loic425 it's essentially very flexible - each formatter can define its own options.

There is a work-in-progress cookbook for adding a custom formatter at https://github.com/Behat/docs/pull/149/ - https://behat--149.org.readthedocs.build/en/149/cookbooks/custom_formatter.html#create-the-extension which adds a bit more detail on how formatters work / can be configured.

**Comment by acoulton** [comment]
In addition to the list @carlos-granados provided, I think **every** formatter array can also include `output_path`, `output_verbosity`, `output_decorate`, `output_styles` options (though how much effect they have will vary by formatter). These are processed by the [OutputManager](https://github.com/Behat/Behat/blob/master/src/Behat/Testwork/Output/OutputManager.php#L144) before any remaining formatter-specific parameters are passed through to the formatter implementation.

For the pretty formatter, I believe:
* `expand: true` causes it to print a full scenario with all steps for each example of an `Outline: ...` instead of the default which looks more like the feature file with the steps printed only once and the results shown in / between rows of the table (like this https://github.com/Behat/Behat/blob/master/features/pretty_format.feature#L116).
* `multiline:false` means it renders `...` instead of any PyStringNode / TableNode or similar arguments - by default they're rendered in full https://github.com/Behat/Behat/blob/master/src/Behat/Behat/Output/Node/Printer/Pretty/PrettyStepPrinter.php#L129

I'd suggest perhaps instead of `withFormatters(...)` we have:
* individual `withFormatter()` to keep it neat if the object has chained helpers
* `disableFormatter(string $name)` - which translates to `['formatters'][$name] = false`

I am *fairly* sure that enabling a formatter in YAML (e.g. providing `progress: true`) just means `configure it with the default arguments` (rather than e.g. `enable it but merge config from the default profile`) - but I'm not 100% on that. If I'm right then I don't think we need a separate `enableFormatter` method.

And then probably provide an implementation of `FormatterConfigInterface` for each of our standard formatters with the specific options they take?

I *might* be tempted to implement these as named constructor arguments with the defaults matching the ones that are defined in the `XXXFormatterFactory`. This would avoid too much verbosity and mean the defaults are obvious to the user rather than buried away. 

If we did that I'd suggest then having the XXXFormatterFactory reference those class defaults instead of redefining them, to keep them consistent over time.

So we'd have something like:
```php
class PrettyFormatter implements FormatterConfigInterface {

   public function __construct(
       bool $timer = true,
       bool $expand = false,
       bool $paths = true,
       bool $multiline = true
   ) {/*...*/}
  
  public function withOutputPath(string $path) {
    // Probably need withOutputPath, withOutputVerbosity, etc to be chained helpers as we don't want to specify values here by default
   // We could perhaps implement these on a base `FormatterConfig` object that individual formatters extend from, as
   // third-party formatters would also need to support these...
  }
}
```

And then [this in PrettyFormatterFactory](https://github.com/Behat/Behat/blob/master/src/Behat/Behat/Output/ServiceContainer/Formatter/PrettyFormatterFactory.php#L173):

```php
 $definition = new Definition('Behat\Testwork\Output\NodeEventListeningFormatter', array(
            'pretty',
            'Prints the feature as is.',
            array(
                'timer'     => true,
                'expand'    => false,
                'paths'     => true,
                'multiline' => true,
            ),
```

Would become something more like:
```php
 $definition = new Definition('Behat\Testwork\Output\NodeEventListeningFormatter', array(
            PrettyFormatter::NAME, // probably need this as a const to make it easy to pass to ->disableFormatter()
            'Prints the feature as is.',
            (new PrettyFormatter())->toArray()
```

What do you and @carlos-granados think?

loic425:
@acoulton I didn't see your previous comment before this new commit.

loic425:
@acoulton I've updated the PR with your suggested changes.

src/Behat/Config/Formatter/Formatter.php
```diff
@@ -0,0 +1,54 @@
...
+    public function withOutputPath(string $path): self
+    {
+        $this->settings['output_path'] = $path;
+
+        return $this;
+    }
+
+    public function withOutputDecorated(bool $decorated = true): self
+    {
+        $this->settings['output_decorated'] = $decorated;
```
acoulton, line 33: Should be `output_decorate` I think

acoulton (commented):
This is looking good to me.

I think I'd still ideally like to see the FormatterFactory classes now taking their defaults from these classes, to avoid the risk a parameter gets added / changed in one place but not the other.

src/Behat/Config/Formatter/Formatter.php
```diff
@@ -0,0 +1,54 @@
...
+
+class Formatter implements FormatterConfigInterface
+{
+    public function __construct(
+        private readonly string $name,
+        private array $settings = [],
+    ) {
+    }
+
+    public function withOutputVerbosity(int $level): self
```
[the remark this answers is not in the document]
loic425: yes, but those constants are deprecated... oh no, you're right, I was searching for the new ones.

src/Behat/Config/Formatter/JUnitFormatter.php
```diff
@@ -0,0 +1,18 @@
...
+declare(strict_types=1);
+
+namespace Behat\Config\Formatter;
+
+final class JUnitFormatter extends Formatter
+{
+    public const NAME = 'junit';
+
+    public function __construct(
+        bool $timer = true,
```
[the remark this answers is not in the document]
loic425: You're right, the option does not work, but it seems like a bug.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites name="frontend_managing_account_organization">
  <testsuite name="Editing an organization" tests="1" skipped="0" failures="0" errors="0" time="0.999">
    <testcase name="Renaming an organization" classname="Editing an organization" status="passed" time="0.981" file="features/frontend/account/organization/editing.feature"></testcase>
  </testsuite>
  <testsuite name="Switching organizations" tests="1" skipped="0" failures="0" errors="0" time="1.079">
    <testcase name="Switching to a different organization" classname="Switching organizations" status="passed" time="1.067" file="features/frontend/account/organization/switching_organizations.feature"></testcase>
  </testsuite>
  <testsuite name="Editing an organization" tests="5" skipped="0" failures="0" errors="0" time="2.082">
    <testcase name="Postal code must be at least 5 characters" classname="Editing an organization" status="passed" time="0.405" file="features/frontend/account/organization/validation.feature"></testcase>
    <testcase name="Postal code must contains only numbers" classname="Editing an organization" status="passed" time="0.402" file="features/frontend/account/organization/validation.feature"></testcase>
    <testcase name="City must contains only text" classname="Editing an organization" status="passed" time="0.368" file="features/frontend/account/organization/validation.feature"></testcase>
    <testcase name="Phone number must be at least 10 characters" classname="Editing an organization" status="passed" time="0.369" file="features/frontend/account/organization/validation.feature"></testcase>
    <testcase name="Email must be valid" classname="Editing an organization" status="passed" time="0.480" file="features/frontend/account/organization/validation.feature"></testcase>
  </testsuite>
</testsuites>
```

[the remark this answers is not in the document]
loic425: I've removed the option for now.

src/Behat/Config/Formatter/Formatter.php
loic425: fixed

src/Behat/Config/Formatter/Formatter.php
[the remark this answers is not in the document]
loic425: done

src/Behat/Config/Formatter/JUnitFormatter.php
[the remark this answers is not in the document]
acoulton: The `time` attribute is a required part of the junit XML schema I believe, so probably correct that we are always filling it.

In the other formatters AFAIK the `timer` option controls whether to print statistics on time / step counts after the run - to be honest I'm not sure why you'd need or want to turn that off but the option exists so...

But for JUnit it wouldn't make sense to print anything that's not part of the XML so I think maybe this was just added by mistake / because maybe at one point all formatters took a `timer` or something.

I think we can definitely remove from the object. We maybe need to leave it in the JUnitFormatterFactory options for BC in case people are currently passing it - not sure if those parameters get validated in any way.

src/Behat/Config/Formatter/PrettyFormatter.php
```diff
@@ -0,0 +1,47 @@
...
+            $settings['expand'] = $expand;
+        }
+
+        if (null !== $paths) {
+            $settings['paths'] = $paths;
+        }
+
+        if (null !== $multiline) {
+            $settings['multiline'] = $multiline;
+        }
```
loic425, line 33:
```suggestion
        $settings['timer'] = $timer ?? $settings['timer'];
        $settings['expand'] = $expand ?? $settings['expand'];
        $settings['paths'] = $paths ?? $settings['paths'];
        $settings['multiline'] = $multiline ?? $settings['multiline'];
```

src/Behat/Config/Formatter/ProgressFormatter.php
```diff
@@ -0,0 +1,27 @@
...
+    public const NAME = 'progress';
+
+    public function __construct(
+        ?bool $timer = null,
+    ) {
+        $settings = self::defaults();
+
+        if (null !== $timer) {
+            $settings['timer'] = $timer;
+        }
```
loic425, line 18:
```suggestion
        $settings['timer'] = $timer ?? $settings['timer'];
```

src/Behat/Behat/Output/ServiceContainer/Formatter/JUnitFormatterFactory.php
```diff
@@ -145,9 +146,7 @@ private function loadFormatter(ContainerBuilder $container)
         $definition = new Definition('Behat\Testwork\Output\NodeEventListeningFormatter', array(
             'junit',
```
acoulton, line 147:
```suggestion
            JUnitFormatter::NAME,
```

src/Behat/Behat/Output/ServiceContainer/Formatter/PrettyFormatterFactory.php
```diff
@@ -173,12 +174,7 @@ protected function loadFormatter(ContainerBuilder $container)
         $definition = new Definition('Behat\Testwork\Output\NodeEventListeningFormatter', array(
             'pretty',
```
acoulton, line 175:
```suggestion
            PrettyFormatter::NAME,
```

src/Behat/Behat/Output/ServiceContainer/Formatter/ProgressFormatterFactory.php
```diff
@@ -141,9 +142,7 @@ protected function loadFormatter(ContainerBuilder $container)
         $definition = new Definition('Behat\Testwork\Output\NodeEventListeningFormatter', array(
             'progress',
```
acoulton, line 143:
```suggestion
            ProgressFormatter::NAME,
```

src/Behat/Config/Formatter/ProgressFormatter.php
```diff
@@ -0,0 +1,25 @@
...
+final class ProgressFormatter extends Formatter
+{
+    public const NAME = 'progress';
+
+    public function __construct(
+        ?bool $timer = null,
+    ) {
+        $settings = self::defaults();
+
+        $settings['timer'] = $timer ?? $settings['timer'];
```
acoulton, line 16: I think I would do this the other way round - leave it as:
```suggestion
        bool $timer = true,
    ) {
        $settings = ['timer' => true]
```

And then implement something like this:

```php
public static function defaults(): array {
  return (new self())->toArray()
}
```

That way there's less scope for errors, it's easy for the user to see what default will be applied, and we can keep nullable typehints for options that are actually nullable?

acoulton (commented):
Thanks, a couple of suggestions

acoulton:
One thing I've realised having now had a chance to test this is that with YAML, it actually is possible to do this:

```yaml
default:
  formatters:
    pretty:
      expand: true
      multiline: false

ci:
  formatters:
    pretty:
      multiline: true
      # This block is merged with the block from the `default` profile above
      # So because expand is not specified here, it will inherit the customised `true` value,
      # not the formatter default of `false`
```

This is not possible with the current PHP implementation.

Personally I think that's OK. I think it's much clearer to specify all the non-default formatter options in each separate profile, rather than relying on them being deep-merged implicitly. Especially because most other parts of the config are overwritten rather than merged.

If people really want to avoid typing the parameters twice, since they're in PHP they have quite a lot of options to use variables or similar to explicitly share values between different profiles in their config file.

I also think using constructor parameters for these is much easier for users to read and to discover the available / default options, and the benefit of that outweighs the loss.

But just thought worth highlighting in case others think the internal deep-merge is an important feature.

src/Behat/Config/Formatter/ProgressFormatter.php
loic425: Yes that seems smart. I'll try it soon, thx 👍

loic425: done

src/Behat/Behat/Output/ServiceContainer/Formatter/JUnitFormatterFactory.php
loic425: done

src/Behat/Behat/Output/ServiceContainer/Formatter/PrettyFormatterFactory.php
loic425: done

src/Behat/Behat/Output/ServiceContainer/Formatter/ProgressFormatterFactory.php
loic425: done

acoulton (approved):
@loic425 great, thanks!

If you have time to look at adding a short documentation for the parameters as https://github.com/Behat/Behat/pull/1558#discussion_r1873376230 that would be great. If not I think we can merge this and I can pick that up

loic425:
> If you have time to look at adding a short documentation for the parameters as [#1558 (comment)](https://github.com/Behat/Behat/pull/1558#discussion_r1873376230) that would be great. If not I think we can merge this and I can pick that up

Yes, I'll work on it on monday.

features/preferred_profile.feature
```diff
@@ -95,7 +103,7 @@ Feature: Preferred Profiles
         progress
 
       imports:
-        - pretty.yml
+        - pretty.php
```
carlos-granados, line 106: This is nice, I did not realize that you could import php files from yaml files 😄

src/Behat/Config/Formatter/Formatter.php
```diff
@@ -0,0 +1,63 @@
...
+class Formatter implements FormatterConfigInterface
+{
+    public function __construct(
+        private readonly string $name,
+        private array $settings = [],
+    ) {
+    }
+
+    /**
+     * @param int $level use Behat\Testwork\Output\Printer\Factory\OutputFactory::VERBOSITY_*
```
carlos-granados, line 19: I think that since you are already importing this name, you can use its short form `OutputFactory::VERBOSITY_*` here

src/Behat/Config/Formatter/FormatterConfigInterface.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Behat\Config\Formatter;
+
+use Behat\Config\ConfigInterface;
+
+interface FormatterConfigInterface extends ConfigInterface
```
carlos-granados, line 9: Should this interface include the `toArray()` function as well?

src/Behat/Config/Formatter/PrettyFormatter.php
```diff
@@ -0,0 +1,24 @@
...
+
+final class PrettyFormatter extends Formatter
+{
+    public const NAME = 'pretty';
+
+    public function __construct(
+        bool $timer = true,
+        bool $expand = false,
+        bool $paths = true,
+        bool $multiline = true,
```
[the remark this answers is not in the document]
carlos-granados: Agree, @loic425 do you think you could add that?

carlos-granados (commented):
Overall looking great, nice job @loic425 just a few small questions

src/Behat/Config/Formatter/FormatterConfigInterface.php
loic425: The toArray function comes from the ConfigInterface, so yes.

features/preferred_profile.feature
loic425: AHAH, and you can import yaml files from php config :)

src/Behat/Config/Formatter/PrettyFormatter.php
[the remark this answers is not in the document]
loic425: hum, this will not be easy for me to write a good short sentence. Should you help me on that?

src/Behat/Config/Formatter/Formatter.php
loic425: done

src/Behat/Config/Formatter/PrettyFormatter.php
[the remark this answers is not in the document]
carlos-granados: My suggestion:
```php
/**
 * @param bool $timer show time and memory usage at the end of the test run
 * @param bool $expand print each example of an scenario outline separately
 * @param bool $paths display the file path and line number for each scenario
 *                    and the context file and method for each step
 * @param bool $multiline print out PyStrings and TableNodes in full
 */
 ```
 @acoulton what do you think
 
 @loic425 remember to also add a comment for the `timer` param in the progress formatter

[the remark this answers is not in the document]
loic425: Yes, I'll do that tomorrow morning 👍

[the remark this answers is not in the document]
loic425: added on PrettyFormatter and timer comment on ProgressFormatter.

acoulton:
@loic425 this is looking good to me. Would you be able to rebase onto latest master and potentially squash the commits? Thanks :)

loic425:
> @loic425 this is looking good to me. Would you be able to rebase onto latest master and potentially squash the commits? Thanks :)

Yes, squashed & rebased.

carlos-granados (approved):
All good now, thanks @loic425

PR #1559, opened 2024-12-06
fix: allow using the narrative filter on the command line

carlos-granados:
When looking at the PR that added the filters to the PHP config I noticed that the narrative filter had not been added as the filters that could be run from the command line. It was also missing from one error message. 
I also found that there was no test for running the role filter from the command line

This PR addresses all these issues

### DIFF
--- a/src/Behat/Behat/Gherkin/Cli/FilterController.php
+++ b/src/Behat/Behat/Gherkin/Cli/FilterController.php
@@ -11,6 +11,7 @@
 namespace Behat\Behat\Gherkin\Cli;
 
 use Behat\Gherkin\Filter\NameFilter;
+use Behat\Gherkin\Filter\NarrativeFilter;
 use Behat\Gherkin\Filter\RoleFilter;
 use Behat\Gherkin\Filter\TagFilter;
 use Behat\Gherkin\Gherkin;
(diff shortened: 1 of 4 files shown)


src/Behat/Behat/Gherkin/Cli/FilterController.php
```diff
@@ -64,7 +65,13 @@ public function configure(Command $command)
                 '--role', null, InputOption::VALUE_REQUIRED,
                 "Only executeCall the features with actor role matching" . PHP_EOL .
                 "a wildcard."
-            );
+            )
+            ->addOption(
+                '--narrative', null, InputOption::VALUE_REQUIRED,
+                "Only executeCall the features with actor description" . PHP_EOL .
```
acoulton, line 71: Is `Only executeCall the...` here intentional (I realise it's in the existing descriptions too)? On first read it looks odd, should it just be `Only execute the...`?

acoulton (commented):
Great, just a question on option descriptions

src/Behat/Behat/Gherkin/Cli/FilterController.php
carlos-granados: I just copied and pasted without paying much attention to the text, but you are right that this reads weird, updated

acoulton (approved):
Thanks!

PR #1560, opened 2024-12-08
DX: disable Xdebug by default and provide a CLI option to activate it

carlos-granados:
If you have been debugging your code with Xdebug and forget to turn it off, when running Behat it can be 2-3 times slower, even if you don't set any breakpoints or actually use Xdebug. Following what other tools like PHPStan or Rector do, with this PR by default we disable Xdebug when running Behat. We add a new `--xdebug` command line option that allows you to run Behat without disabling Xdebug. 

This uses the https://github.com/composer/xdebug-handler library, also used by PHPStan and Rector to provide this same functionality

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,6 +19,7 @@
         "behat/gherkin": "^4.10.0",
         "behat/transliterator": "^1.5",
         "composer-runtime-api": "^2.2",
+        "composer/xdebug-handler": "^3.0",
         "psr/container": "^1.0 || ^2.0",
         "symfony/config": "^5.4 || ^6.4 || ^7.0",
         "symfony/console": "^5.4 || ^6.4 || ^7.0",
--- a/src/Behat/Testwork/Cli/Application.php
+++ b/src/Behat/Testwork/Cli/Application.php
@@ -14,6 +14,7 @@ use Behat\Testwork\ServiceContainer\Configuration\ConfigurationLoader;
 use Behat\Testwork\ServiceContainer\ContainerLoader;
 use Behat\Testwork\ServiceContainer\Exception\ConfigurationLoadingException;
 use Behat\Testwork\ServiceContainer\ExtensionManager;
+use Composer\XdebugHandler\XdebugHandler;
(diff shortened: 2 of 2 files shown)


acoulton (approved):
That is super helpful, thanks :heart:

ostrolucky:
This broke phpstorm's behat debug run. You should consider reverting it.

carlos-granados:
@ostrolucky can you provide more details?

carlos-granados:
@ostrolucky You should be able to edit the Behat run configuration to add the `--xdebug` flag in the test runner options. I just tested this in my PHPStorm installation and it works correctly, please see this screenshot:
Screenshot 2024-12-23 at 14 16 33

ostrolucky:
Yeah that seems better than having to define BEHAT_ALLOW_XDEBUG=1 env variable. However, this stuff needs to be done each time you are debugging, or define it as default for behat configuration (which is going to defeat the purpose of making this optional). In any case it's quite non-transparent and different from other test runners like phpunit. I think behat should do it the same way as phpunit. Otherwise we need to reach out to jetbrains to update their IDE.

carlos-granados:
Thanks @ostrolucky if you feel that there is a better way to handle this, feel free to open an issue in this repository with your suggestion and we will consider it. Or, even better, submit a PR 😄

cooperaj:
Spent most of yesterday fighting my environment, getting highly annoyed that debugging had broken for us. I was at point of scouring the Jetbrains issue tracker and manually debugging their behat helper scripts.

Then I found that our renovate run had brought in a point release of behat that has this (workflow breaking) change in it.

Our PHPStorm test configs only ever enabled debugging if you requested it so we didn't see any unnecessary slowdown. We've now implemented the test runner flag in the configs and are back to normal.

It's a good feature, I just wish it hadn't appeared as it did and been made the default operation.

acoulton:
@cooperaj apologies for the hassle - it's clear that we hadn't properly understood the impact this change would have. This is fixed by #1581 which will be in the 3.18.0 release, coming this evening or tomorrow.

PR #1562, opened 2024-12-09
Psalm level 6 2

carlos-granados:
Applies three new rules for PSalm level 6

### DIFF
--- a/src/Behat/Behat/Hook/Call/RuntimeFeatureHook.php
+++ b/src/Behat/Behat/Hook/Call/RuntimeFeatureHook.php
@@ -17,7 +17,6 @@ use Behat\Gherkin\Node\FeatureNode;
 use Behat\Testwork\Call\Exception\BadCallbackException;
 use Behat\Testwork\Hook\Call\RuntimeFilterableHook;
 use Behat\Testwork\Hook\Scope\HookScope;
-use ReflectionFunction;
 use ReflectionMethod;
 
 /**
@@ -32,7 +31,7 @@ abstract class RuntimeFeatureHook extends RuntimeFilterableHook
      *
      * @param string      $scopeName
      * @param null|string $filterString
-     * @param callable|array $callable
+     * @param callable|array{class-string, string} $callable
      * @param null|string $description
      *
      * @throws BadCallbackException If callback is method, but not a static one
--- a/src/Behat/Behat/Hook/Call/RuntimeFeatureHook.php
+++ b/src/Behat/Behat/Hook/Call/RuntimeFeatureHook.php
@@ -17,6 +17,8 @@ use Behat\Gherkin\Node\FeatureNode;
 use Behat\Testwork\Call\Exception\BadCallbackException;
 use Behat\Testwork\Hook\Call\RuntimeFilterableHook;
 use Behat\Testwork\Hook\Scope\HookScope;
+use ReflectionFunction;
+use ReflectionMethod;
 
 /**
  * Represents a feature hook.
@@ -30,7 +32,7 @@ abstract class RuntimeFeatureHook extends RuntimeFilterableHook
      *
      * @param string      $scopeName
      * @param null|string $filterString
-     * @param callable    $callable
+     * @param callable|array $callable
      * @param null|string $description
      *
      * @throws BadCallbackException If callback is method, but not a static one
@@ -40,10 +42,19 @@ abstract class RuntimeFeatureHook extends RuntimeFilterableHook
         parent::__construct($scopeName, $filterString, $callable, $description);
 
         if ($this->isAnInstanceMethod()) {
+            if (is_array($callable)) {
+                $className = $callable[0];
+                $methodName = $callable[1];
+            } else {
+                $reflection = new ReflectionMethod($callable);
+                $className = $reflection->getDeclaringClass()->getShortName();
+                $methodName = $reflection->getName();
+            }
+
             throw new BadCallbackException(sprintf(
                 'Feature hook callback: %s::%s() must be a static method',
-                $callable[0],
-                $callable[1]
+                $className,
+                $methodName
             ), $callable);
         }
     }
--- a/src/Behat/Behat/Hook/Call/RuntimeFeatureHook.php
+++ b/src/Behat/Behat/Hook/Call/RuntimeFeatureHook.php
@@ -40,22 +40,7 @@ abstract class RuntimeFeatureHook extends RuntimeFilterableHook
     {
         parent::__construct($scopeName, $filterString, $callable, $description);
 
-        if ($this->isAnInstanceMethod()) {
-            if (is_array($callable)) {
-                $className = $callable[0];
-                $methodName = $callable[1];
-            } else {
-                $reflection = new ReflectionMethod($callable);
-                $className = $reflection->getDeclaringClass()->getShortName();
-                $methodName = $reflection->getName();
-            }
-
-            throw new BadCallbackException(sprintf(
-                'Feature hook callback: %s::%s() must be a static method',
-                $className,
-                $methodName
-            ), $callable);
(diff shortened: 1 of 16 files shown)


src/Behat/Behat/Hook/Call/RuntimeFeatureHook.php
```diff
@@ -30,7 +32,7 @@ abstract class RuntimeFeatureHook extends RuntimeFilterableHook
      *
      * @param string      $scopeName
      * @param null|string $filterString
-     * @param callable    $callable
+     * @param callable|array $callable
```
stof, line 35: I suggest using `array{class-string, string}` to restrict the shape of the accepted array for the case of context instance methods.

src/Behat/Behat/Hook/Call/RuntimeFeatureHook.php
```diff
@@ -40,10 +42,19 @@ public function __construct($scopeName, $filterString, $callable, $description =
         parent::__construct($scopeName, $filterString, $callable, $description);
 
         if ($this->isAnInstanceMethod()) {
+            if (is_array($callable)) {
```
carlos-granados, line 45: Depending on the type of callable that we have we get the class name and method name using one way or another

src/Behat/Behat/Output/Node/EventListener/Flow/FirstBackgroundFiresFirstListener.php
```diff
@@ -38,7 +38,7 @@ class FirstBackgroundFiresFirstListener implements EventListener
      */
     private $firstBackgroundEnded = false;
     /**
-     * @var Event[]
+     * @var array<array{0: Event, 1: string}>
```
carlos-granados, line 41: Wrong doc comment

src/Behat/Behat/Output/Statistics/PhaseStatistics.php
```diff
@@ -111,7 +111,7 @@ public function registerHookStat(HookStat $stat)
     /**
      * Returns counters for different scenario result codes.
      *
-     * @return array<TestResult::*, int>
+     * @return array<StepResult::*, int>
```
carlos-granados, line 114: This includes one case that is defined in StepResult, so we use the derived class

stof, line 114: this change is wrong. Scenario stats are not indexed by step result

src/Behat/Behat/Output/Statistics/TotalStatistics.php
```diff
@@ -32,11 +32,11 @@ final class TotalStatistics implements Statistics
      */
     private $memory;
     /**
-     * @var array
+     * @var array<StepResult::*, int>
```
stof, line 35: This is wrong. The scenario counters are indexed by TestResult constants

stof (dismissed):

src/Behat/Behat/Hook/Call/RuntimeFeatureHook.php
```diff
@@ -40,10 +42,19 @@ public function __construct($scopeName, $filterString, $callable, $description =
...
+                $className = $callable[0];
+                $methodName = $callable[1];
+            } else {
+                $reflection = new ReflectionMethod($callable);
+                $className = $reflection->getDeclaringClass()->getShortName();
```
stof, line 50: why the shortname (which is not unique) ?

src/Behat/Behat/Output/Statistics/PhaseStatistics.php
carlos-granados: Actually some scenario results were indexed by the StepResult::UNDEFINED constant, that's why I introduced this change. But thinking more about this, it does not make sense for this constant to be defined in `StepResult` if it is not only used by steps, so I moved it to the base `TestResult` class and now scenario stats are only indexes by `TestResult` constants

src/Behat/Behat/Hook/Call/RuntimeFeatureHook.php
carlos-granados: This is only used to print out an error message. Using the short class name makes the error message clearer and easier to read and I think that the case of having duplicate context class names and method names would be so rare that I don't think we should worry about using the full name

src/Behat/Testwork/Hook/Call/RuntimeSuiteHook.php
```diff
@@ -37,10 +38,19 @@ public function __construct($scopeName, $filterString, $callable, $description =
...
         if ($this->isAnInstanceMethod()) {
+            if (is_array($callable)) {
+                $className = $callable[0];
+                $methodName = $callable[1];
+            } else {
+                $reflection = new ReflectionMethod($callable);
+                $className = $reflection->getDeclaringClass()->getShortName();
+                $methodName = $reflection->getName();
+            }
+
```
acoulton, line 49: Now that this code is more complex and duplicated in RuntimeFeatureHook, might it be worth pulling it up to a parent class as something like `$this->throwIfInstanceMethod($callable)`?

src/Behat/Behat/Hook/Call/RuntimeFeatureHook.php
acoulton: Agreed, I think short name is preferable here

acoulton (approved):
Looks good, just a small question/opportunity to extract some duplicate code if you'd like to do that

src/Behat/Testwork/Hook/Call/RuntimeSuiteHook.php
carlos-granados: Good shout, done!

acoulton:
@carlos-granados it doesn't look like the branch is updated?

carlos-granados:
@acoulton argh, I committed the code but forgot to push it, done

src/Behat/Testwork/Call/RuntimeCallee.php
```diff
@@ -116,4 +116,25 @@ public function isAnInstanceMethod()
         return $this->reflection instanceof ReflectionMethod
             && !$this->reflection->isStatic();
     }
+
+    public function throwIfInstanceMethod(callable|array $callable, string $hookType): void
```
acoulton, line 120: Should perhaps have the full $callable type definition and be a protected, rather than public, method?

```suggestion
    /**
     * @param callable|array{class-string, string} $callable
     */
    protected function throwIfInstanceMethod(callable|array $callable, string $hookType): void
```

carlos-granados: Ah, yes of course, it needs to be protected, not public 🤦  And adding the full type definition makes sense, updated

acoulton (approved):
Great :)

carlos-granados:
@stof happy with this now? I would like to merge it to continue with this upgrade. Thanks!

src/Behat/Behat/Output/Statistics/TotalStatistics.php
```diff
@@ -76,7 +76,7 @@ public function resetAllCounters()
         $this->scenarioCounters = $this->stepCounters = array(
             TestResult::PASSED    => 0,
             TestResult::FAILED    => 0,
-            StepResult::UNDEFINED => 0,
+            TestResult::UNDEFINED => 0,
             TestResult::PENDING   => 0,
             TestResult::SKIPPED   => 0
```
fmatsos, line 81: It might be interesting to perhaps use Enums (not int this PR, of course)?

stof: I don't think it is worth the BC break, especially given that our stats are using those test results as array keys, which is not possible with enums (as they are objects, which are invalid array keys)

Issue #1563, opened 2024-12-10
Migrate features from annotations to attributes


fmatsos:
With #1549 we started to migrate some annotations to their attributes equivalents. We have restricted migration to the only features and scenarios that are required in this PR.
Since changes are now merged, we can now migrate the rest to attributes too.

First concerns are Step annotations. But we can also migrate the other types of annotations (like Transform, beforeStep, etc.)

carlos-granados:
Completed via https://github.com/Behat/Behat/pull/1564

PR #1564, opened 2024-12-10
feat: migrate annotations to attributes for features

fmatsos:
With https://github.com/Behat/Behat/pull/1549 we started to migrate some annotations to their attributes equivalents. We have restricted migration to the only features and scenarios that are required in this PR.
Since changes are now merged, we can now migrate the rest to attributes too.

First concerns are Step annotations. But we can also migrate the other types of annotations (like Transform, beforeStep, etc.)

<hr />

**Issue #1563**


features/tag_filters.feature
```diff
@@ -12,19 +12,13 @@ Feature: Tags
 
       class FeatureContext implements Context
       {
-          /**
-           * @Given /^Some slow step N(\d+)$/
-           */
+          #[Given('/^Some slow step N(\d+)$/')]
```
stof, line 15: you forgot adding the use statement for the attributes in those files, which breaks things

features/definitions_patterns.feature
```diff
@@ -237,61 +230,17 @@ Feature: Step Definition Pattern
       Token name should not exceed 32 characters, but `too12345678911234567890123456789012` was used.
       """
 
-  Scenario: Multiline definitions
```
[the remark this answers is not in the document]
fmatsos: Fixed in local!

carlos-granados (approved):
LGTM, thanks @fmatsos

fmatsos:
@acoulton is it okay for you too?

features/definitions_transformations.feature
```diff
@@ -31,17 +31,20 @@ Feature: Step Arguments Transformations
...
+      use Behat\Step\Given;
+      use Behat\Step\Then;
+      use Behat\Transformation\Transform;
 
       class FeatureContext implements Context
       {
           private $user;
 
-          /** @Transform /"([^\ "]+)(?: - (\d+))?" user/ */
+          #[Transform('/"([^\ "]+)(?: - (\d+))?" user/')]
```
acoulton, line 42: I'm not sure we should be changing this feature, instead we should rename it to `definitions_transformations_annotations.feature` (we already have `definitions_transformations_attributes.feature`)?

features/hooks.feature
```diff
@@ -20,108 +20,86 @@ Feature: hooks
...
 
       class FeatureContext implements Context
       {
           private $number;
           private $scenarioFilter;
 
-          /**
-           * @BeforeFeature
-           */
+          #[BeforeFeature]
```
acoulton, line 35: We might need separate `hooks_attributes.feature` and `hooks_annotations.feature`?

features/i18n.feature
```diff
@@ -12,35 +12,30 @@ Feature: I18n
...
+      use Behat\Step\When;
 
       class FeatureContext implements Context
       {
           private $value = 0;
 
-          /**
-           * @Given /Я ввел (\d+)/
-           */
+          #[Given('/Я ввел (\d+)/')]
```
acoulton, line 23: I'm not certain, but it *might* be worth at least one simple test to prove that the russian is also translated to match e.g. the `@Given` annotation? I think at the moment the parsing/i18n is implemented so that this happens before it looks for definitions (so we can be pretty sure that annotations and attributes will behave the same), but there's no guarantee that would stay the same in future.

acoulton (changes requested):
It's tricky from the diff to see what feature coverage we have left that is using annotations, but I'm nervous that this might move too much over.

Certainly everything that is testing formatters, result handling, CLI options, configuration etc can be tested with attributes-only (the steps only exist to trigger the expected output / behaviour).

But we definitely still need feature coverage of annotations for:

* Step definitions themselves, probably with at least one example of step arguments using regex and one with turnip (and one with no arguments) to prove that these are still parsed as expected
* Hooks
* Transformations
* Probably at least one basic `i18n` example.
* Perhaps at least one example of a FeatureContext using **both** attributes and annotations, to prove that people can safely migrate older code in stages.

The easiest for maintenance is probably to have explicitly named `XXX_attributes.feature` and `XXX_annotations` feature files for the features specific to matching gherkin to PHP implementations. That way, it will be easy to make sure everything is covered, and easy to remove the annotation-only versions if/when we remove annotation support in future.

fmatsos:
> It's tricky from the diff to see what feature coverage we have left that is using annotations, but I'm nervous that this might move too much over.
> 
> Certainly everything that is testing formatters, result handling, CLI options, configuration etc can be tested with attributes-only (the steps only exist to trigger the expected output / behaviour).
> 
> But we definitely still need feature coverage of annotations for:
> 
> * Step definitions themselves, probably with at least one example of step arguments using regex and one with turnip (and one with no arguments) to prove that these are still parsed as expected
> * Hooks
> * Transformations
> * Probably at least one basic `i18n` example.
> * Perhaps at least one example of a FeatureContext using **both** attributes and annotations, to prove that people can safely migrate older code in stages.
> 
> The easiest for maintenance is probably to have explicitly named `XXX_attributes.feature` and `XXX_annotations` feature files for the features specific to matching gherkin to PHP implementations. That way, it will be easy to make sure everything is covered, and easy to remove the annotation-only versions if/when we remove annotation support in future.

Looking your comment and the PR/code, I'm totally agree to "double" these cases.
I'll do the required changes for that!

fmatsos:
@acoulton I rollback on some features, as you mentioned, to keep annotations resting for the moment.
Meanwhile, I copied these tests to their attributes equivalent.

I fix a method visibility too, after @stof pointed it here: https://github.com/Behat/Behat/pull/1567#discussion_r1890246045

Hope it'll be fine now 😄

carlos-granados (approved):
Looking good, many thanks for this very thorough job, @fmatsos 🎉

acoulton (approved):
Very thorough indeed, thanks @fmatsos :)

rpkamp:
Awesome PR @fmatsos :clap:

PR #1566, opened 2024-12-12
Psalm level 6 3

carlos-granados:
Apply several new rules. Most don't need any changes in the code

### DIFF
--- a/src/Behat/Behat/Context/Environment/InitializedContextEnvironment.php
+++ b/src/Behat/Behat/Context/Environment/InitializedContextEnvironment.php
@@ -28,7 +28,7 @@ use Psr\Container\ContainerInterface;
 final class InitializedContextEnvironment implements ContextEnvironment, ServiceContainerEnvironment
 {
     /**
-     * @var string
+     * @var Suite
      */
     private $suite;
     /**
--- a/src/Behat/Behat/Output/Node/EventListener/AST/OutlineListener.php
+++ b/src/Behat/Behat/Output/Node/EventListener/AST/OutlineListener.php
@@ -138,8 +138,9 @@ final class OutlineListener implements EventListener
 
         $this->exampleSetupPrinter->printSetup($formatter, $event->getSetup());
 
-        $this->example = $event->getScenario();
-        assert($this->example instanceof ExampleNode);
+        $scenario = $event->getScenario();
+        assert($scenario instanceof ExampleNode);
+        $this->example = $scenario;
         $this->examplePrinter->printHeader($formatter, $event->getFeature(), $this->example);
     }
 
--- a/src/Behat/Behat/Output/Node/EventListener/Statistics/HookStatsListener.php
+++ b/src/Behat/Behat/Output/Node/EventListener/Statistics/HookStatsListener.php
@@ -111,7 +111,7 @@ final class HookStatsListener implements EventListener
     {
         $call = $hookCallResult->getCall();
         $callee = $call->getCallee();
-        $hook = (string) $callee;
+        $hook = $callee->__toString();
         $scope = $call->getScope();
         $path = $callee->getPath();
         $stdOut = $hookCallResult->getStdOut();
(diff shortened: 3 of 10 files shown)


src/Behat/Behat/Context/Environment/InitializedContextEnvironment.php
```diff
@@ -28,7 +28,7 @@
 final class InitializedContextEnvironment implements ContextEnvironment, ServiceContainerEnvironment
 {
     /**
-     * @var string
+     * @var Suite
```
carlos-granados, line 31: Comment was wrong

src/Behat/Behat/Output/Node/EventListener/AST/OutlineListener.php
```diff
@@ -138,8 +138,9 @@ private function printExampleHeaderOnBeforeExampleEvent(Formatter $formatter, Ev
 
         $this->exampleSetupPrinter->printSetup($formatter, $event->getSetup());
 
-        $this->example = $event->getScenario();
-        assert($this->example instanceof ExampleNode);
+        $scenario = $event->getScenario();
```
carlos-granados, line 141: We need to assert that the scenario is an ExampleNode before assigning to the property as the property has the `ExampleNode` type

src/Behat/Behat/Output/Node/EventListener/Statistics/HookStatsListener.php
```diff
@@ -111,7 +111,7 @@ private function captureHookStat(CallResult $hookCallResult)
     {
         $call = $hookCallResult->getCall();
         $callee = $call->getCallee();
-        $hook = (string) $callee;
+        $hook = $callee->__toString();
```
carlos-granados, line 114: PSalm is not happy with the cast and requires a explicit call to __toString()

src/Behat/Behat/Tester/Cli/RerunController.php
```diff
@@ -39,7 +39,7 @@ final class RerunController implements Controller
      */
     private $key;
     /**
-     * @var string[]
+     * @var array<string, string[]>
```
carlos-granados, line 42: doc-block was wrong

src/Behat/Testwork/Counter/Timer.php
```diff
@@ -51,7 +51,7 @@ public function stop()
     }
 
     /**
-     * @return null|float
+     * @return float
```
carlos-granados, line 54: This function never returns null

src/Behat/Testwork/Exception/TestworkException.php
```diff
@@ -10,11 +10,13 @@
...
 
+use Throwable;
+
 /**
  * All testwork exceptions implement this interface.
  *
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
  */
-interface TestworkException
+interface TestworkException extends Throwable
```
carlos-granados, line 20: This exception interface needs to extend the Throwable interface so that Psalm accepts that any exception which implements it can be thrown

src/Behat/Behat/Output/Node/EventListener/Statistics/HookStatsListener.php
carlos-granados: I ran this with the full level 6 enabled and did not get any complains, so maybe it is detected at another level. Let me try to see which is the best way to handle this

carlos-granados: I have narrowed the types before applying the conversion to string

src/Behat/Behat/Output/Node/Printer/Pretty/PrettySkippedStepPrinter.php
```diff
@@ -144,8 +145,10 @@ private function getArgumentString(ArgumentInterface $argument, $collapse = fals
 
             return $text;
         }
-
-        return (string) $argument;
+        if ($argument instanceof TableNode) {
+            return $argument;
```
stof, line 149: This should still cast the argument, otherwise it does not return a string.

src/Behat/Behat/Output/Node/Printer/Pretty/PrettyStepPrinter.php
```diff
@@ -195,8 +196,10 @@ private function getArgumentString(ArgumentInterface $argument, $collapse = fals
 
             return $text;
         }
-
-        return (string) $argument;
+        if ($argument instanceof TableNode) {
+            return $argument;
```
stof, line 200: same here

src/Behat/Behat/Output/Node/Printer/Pretty/PrettySkippedStepPrinter.php
carlos-granados: Updated

acoulton (approved):

PR #1567, opened 2024-12-13
fix: Inherit step definition attributes on methods extended from parent Context

### LINKED ISSUE #1565: Definition override does not work with attributes
With migration in #1563, from annotations to attributes in features, a problem occurred with `definitions_override.feature`.

When we migrate the annotation from parent context to attribute, the tests don’t work anymore. But they pass when we rollback to annotation.
Of course, required use statements have been added to use the equivalent attribute.

fmatsos:
As mentioned by @stof here: https://github.com/Behat/Behat/issues/1565#issuecomment-2535430466, `AttributeContextReader` was incomplete compared to its annotation equivalent.

We add here the missing code to retrieve parent callees, like annotations do.

Closes #1565

### DIFF
--- a/features/definitions_override_attributes.feature
+++ b/features/definitions_override_attributes.feature
@@ -1,20 +1,19 @@
-Feature: Step Definitions Override
+Feature: Step Definitions Override Attributes
   In order to fine-tune definitions defined in parent classes
   As a step definitions developer
-  I need to be able to override definition methods using annotations
+  I need to be able to override definition methods using step definition attributes
 
-  Scenario: Overridden method without own annotation will inherit parent pattern
+  Scenario: Overridden method without own attribute will inherit parent pattern
     Given a file named "features/bootstrap/FeatureContext.php" with:
       """
       <?php
 
       use Behat\Behat\Context\Context;
+      use Behat\Step\Then;
 
       class ParentContext
       {
-          /**
-           * @Then :token should be :value
-           */
+          #[Then(':token should be :value')]
           public function shouldBe($token, $value) {}
       }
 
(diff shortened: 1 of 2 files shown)


src/Behat/Behat/Context/Reader/AttributeContextReader.php
```diff
@@ -57,8 +59,34 @@ public function readContextCallees(ContextEnvironment $environment, $contextClas
...
     {
         $callees = [];
+
         foreach ($this->readers as $reader) {
-            $callees = array_merge($callees, $reader->readCallees($contextClass, $method));
+            $callees = array_merge(
+                $this->readParentCallees($contextClass, $method, $reader, $callees),
+                $reader->readCallees($contextClass, $method)
+            );
+
```
loic425, line 68:
```suggestion
            );
```

loic425, line 68:
```suggestion
```

fmatsos:
@carlos-granados @acoulton when you have some time to check this one 😄

features/definitions_override.feature
```diff
@@ -9,33 +9,40 @@ Feature: Step Definitions Override
...
+      use Behat\Step\Then;
 
       class ParentContext
       {
           /**
            * @Then :token should be :value
            */
           public function shouldBe($token, $value) {}
+
+          #[Then(':token should not be :value')]
```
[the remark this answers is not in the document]
fmatsos: Done!

features/definitions_override.feature
```diff
@@ -1,4 +1,5 @@
 Feature: Step Definitions Override
+  Using step definitions as annotations
```
acoulton, line 2: I would add this to the `I need to be able to override definition methods using annotations` to keep the `In order to .. as a ... I need to...` three-line structure

features/definitions_override_attributes.feature
```diff
@@ -0,0 +1,75 @@
+Feature: Step Definitions Override Attributes
+  Using step definitions as attributes
+  In order to fine-tune definitions defined in parent classes
+  As a step definitions developer
+  I need to be able to override definition methods
```
acoulton, line 5:
```suggestion
  In order to fine-tune definitions defined in parent classes
  As a step definitions developer
  I need to be able to override definition methods using step definition attributes
```

features/definitions_override_attributes.feature
```diff
@@ -0,0 +1,75 @@
...
+          Then 5 should be 10
+      """
+    When I run "behat -f progress --no-colors"
+    Then it should pass with:
+      """
+      ..
+
+      1 scenario (1 passed)
+      2 steps (2 passed)
+      """
```
acoulton, line 75: We should probably also have a case for what happens if the parent has an annotation and the child has an attribute (which might occur if a project migrates to attributes but is still using a third-party library that uses annotations)?

I think so long as that still ends up with both definitions attached, we can probably skip the case for parent attribute / child annotation which will presumably behave the same.

acoulton (changes requested):
The implementation LGTM thanks, just a couple of comments on the tests.

fmatsos:
> Great, thanks

You're welcome! 😄 

And sorry for the confusion with the title of PR. My English is not always good and understandable ahah!

carlos-granados (approved):
Looking good, thanks @fmatsos

src/Behat/Behat/Context/Reader/AttributeContextReader.php
```diff
@@ -57,8 +59,33 @@ public function readContextCallees(ContextEnvironment $environment, $contextClas
...
+            );
+        }
+
+        return $callees;
+    }
+
+    /**
+     * @return array<int, Callee>
+     */
+    public function readParentCallees(
```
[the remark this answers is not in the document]
fmatsos: I'll fix it here: https://github.com/Behat/Behat/pull/1564 👍🏻

PR #1570, opened 2024-12-18
Psalm level 6 5

carlos-granados:
Applies all the remaining rules needed to reach Psalm level 6

### DIFF
--- a/src/Behat/Testwork/Argument/MixedArgumentOrganiser.php
+++ b/src/Behat/Testwork/Argument/MixedArgumentOrganiser.php
@@ -12,8 +12,10 @@ namespace Behat\Testwork\Argument;
 
 use ReflectionFunctionAbstract;
 use ReflectionClass;
+use ReflectionNamedType;
 use ReflectionParameter;
 use ReflectionException;
+use ReflectionUnionType;
 
 /**
  * Organises function arguments using its reflection.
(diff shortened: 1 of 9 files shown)


src/Behat/Testwork/Argument/MixedArgumentOrganiser.php
```diff
@@ -241,13 +243,10 @@ private function getReflectionClassesFromParameter(\ReflectionParameter $paramet
 
         $type = $parameter->getType();
 
-        /**
```
carlos-granados, line 244: This is not needed not that we are in PHP 8

src/Behat/Testwork/Autoloader/ServiceContainer/AutoloaderExtension.php
```diff
@@ -78,7 +78,9 @@ public function configure(ArrayNodeDefinition $builder)
             ->treatTrueLike($this->defaultPaths)
             ->treatNullLike(array())
             ->treatFalseLike(array())
-
+            ;
```
carlos-granados, line 81: The return value of `treatFalseLike()` is a NodeDefinition, which does not have the `prototype()` function, so we need to narrow down the actual type

src/Behat/Testwork/Ordering/Cli/OrderController.php
```diff
@@ -77,14 +77,15 @@ public function execute(InputInterface $input, OutputInterface $output)
         $orderer = $input->getOption('order');
 
         if (!$orderer) {
-            return;
+            return null;
```
carlos-granados, line 80: This function needs to return a value

src/Behat/Testwork/Output/Printer/JUnitOutputPrinter.php
```diff
@@ -141,8 +141,9 @@ public function setFileName($fileName, $extension = 'xml')
         if ('.'.$extension !== substr($fileName, strlen($extension) + 1)) {
             $fileName .= '.'.$extension;
         }
-
-        $this->getOutputFactory()->setFileName($fileName);
+        $outputFactory = $this->getOutputFactory();
```
carlos-granados, line 144: Again need to narrow down the actual type

acoulton (approved):
Great :)

Issue #1519, opened 2024-11-06
Small refactors and removals for next major version


acoulton:
This is just a placeholder issue for us to track things that come up that would create too much noise as individual issues but need to be done in the next major to avoid a BC break.

* [ ] Remove unused JUnitOutlineStoreListener $outlineListener argument from JUnitScenarioPrinter constructor (#1459)
* [ ] Remove ApplicationFactory::VERSION (discussed [here ](https://github.com/Behat/Behat/discussions/1484#discussioncomment-11141724 )so far, will be deprecated in a soon-to-come PR)

PR #1571, opened 2024-12-18
DX: initial install of PHPStan

carlos-granados:
We aim to replace Psalm with PHPStan as the static analysis tool of choice for this repository. This PR adds this new tool and adds the changes needed to reach level 1.

For now we keep Psalm, we don't want to remove it until we have reached at least PHPStan level 3, which could be the equivalent of the current level 6 in Psalm

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -109,6 +109,9 @@ jobs:
       - name: Run Psalm
         run: ./vendor/bin/psalm --output-format=github --php-version=${{ matrix.php }}
 
+      - name: Run PHPStan
+        run: ./vendor/bin/phpstan
+
   build-phar:
     name: Build PHAR file
     runs-on: ubuntu-latest
--- a/composer.json
+++ b/composer.json
@@ -31,6 +31,7 @@
 
     "require-dev": {
         "herrera-io/box": "~1.6.1",
+        "phpstan/phpstan": "^2.0",
         "phpunit/phpunit": "^9.6",
         "sebastian/diff": "^4.0",
         "symfony/polyfill-php84": "^1.31",
@@ -69,10 +70,12 @@
         "all-tests": [
             "@behat-progress",
             "@phpunit",
-            "@psalm"
+            "@psalm",
+            "@phpstan"
         ],
         "behat": "LANG=C bin/behat  --rerun",
         "behat-progress": "LANG=C bin/behat --format=progress",
+        "phpstan": "phpstan",
         "phpunit": "phpunit",
         "psalm": "psalm"
     },
--- a/phpstan.dist.neon
+++ b/phpstan.dist.neon
@@ -1,0 +1,16 @@
+parameters:
+    level: 1
+    paths:
+        - src
+    excludePaths:
+        - src/Behat/Testwork/EventDispatcher/TestworkEventDispatcherSymfonyLegacy.php
+        - src/Behat/Testwork/EventDispatcher/TestworkEventDispatcherSymfony5.php
+    ignoreErrors:
+        -
+            identifier: class.notFound
+            paths:
+                - src/Behat/Testwork/Exception/Stringer/PHPUnitExceptionStringer.php
+        -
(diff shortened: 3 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -109,6 +109,9 @@ jobs:
       - name: Run Psalm
         run: ./vendor/bin/psalm --output-format=github --php-version=${{ matrix.php }}
 
+      - name: Run PHPStan
```
carlos-granados, line 112: We run both Psalm and PHPStan in CI

composer.json
```diff
@@ -69,10 +70,12 @@
         "all-tests": [
             "@behat-progress",
             "@phpunit",
-            "@psalm"
+            "@psalm",
+            "@phpstan"
         ],
         "behat": "LANG=C bin/behat  --rerun",
         "behat-progress": "LANG=C bin/behat --format=progress",
+        "phpstan": "phpstan",
```
carlos-granados, line 78: Add new composer script

phpstan.dist.neon
```diff
@@ -0,0 +1,16 @@
+parameters:
+    level: 1
+    paths:
+        - src
+    excludePaths:
+        - src/Behat/Testwork/EventDispatcher/TestworkEventDispatcherSymfonyLegacy.php
```
carlos-granados, line 6: We exclude these two files as they generate a lot of false positives depending on the Symfony version

phpstan.dist.neon
```diff
@@ -0,0 +1,16 @@
...
+        - src/Behat/Testwork/EventDispatcher/TestworkEventDispatcherSymfony5.php
+    ignoreErrors:
+        -
+            identifier: class.notFound
```
carlos-granados, line 10: We ignore this error in this class as it refers to some PHPUnit classes that may not be available depending on the PHPUnit version

phpstan.dist.neon
```diff
@@ -0,0 +1,16 @@
...
+            paths:
+                - src/Behat/Testwork/Exception/Stringer/PHPUnitExceptionStringer.php
+        -
+            identifier: constructor.unusedParameter
```
carlos-granados, line 14: This class has a constructor parameter that is not used but it needs to be present because it is expected that all `SimpleArgumentTransformation` classes have it

src/Behat/Behat/Output/Node/Printer/Pretty/PrettyScenarioPrinter.php
```diff
@@ -106,7 +106,7 @@ private function printKeyword(OutputPrinter $printer, $keyword)
      * Prints scenario title (first line of long title).
      *
      * @param OutputPrinter $printer
-     * @param string        $longTitle
+     * @param string|null   $longTitle
```
carlos-granados, line 109: The parameter can be null

src/Behat/Testwork/Output/Exception/FormatterNotFoundException.php
```diff
@@ -33,6 +33,7 @@ class FormatterNotFoundException extends InvalidArgumentException implements Out
     public function __construct($message, $name)
     {
         parent::__construct($message);
+        $this->name = $name;
```
carlos-granados, line 36: This constructor parameter was not used but it actually needs to be used to set our private property

carlos-granados:
We can probably wait to merge this until the new release is out

stof (approved):

phpstan.dist.neon
carlos-granados: I have added a task for this to https://github.com/Behat/Behat/issues/1519, the issue that we have to track what we want to implement for the next major

src/Behat/Testwork/Output/Exception/FormatterNotFoundException.php
acoulton: Is it worth just making that a promoted property at this point?

acoulton (approved):

PR #1573, opened 2024-12-19
chore: update code to PHPStan level 2

carlos-granados:
Adds the necessary changes to update our codebase to PHPStan level 2

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ behat.yml
 vendor
 composer.lock
 .phpunit.result.cache
+phpstan.neon
--- a/src/Behat/Behat/Context/Cli/ContextSnippetsController.php
+++ b/src/Behat/Behat/Context/Cli/ContextSnippetsController.php
@@ -87,5 +87,6 @@ final class ContextSnippetsController implements Controller
                 new FixedPatternIdentifier($input->getOption('snippets-type'))
             ))
         );
+        return null;
     }
 }
--- a/src/Behat/Behat/Definition/Pattern/Policy/RegexPatternPolicy.php
+++ b/src/Behat/Behat/Definition/Pattern/Policy/RegexPatternPolicy.php
@@ -22,7 +22,7 @@ use Behat\Transliterator\Transliterator;
 final class RegexPatternPolicy implements PatternPolicy
 {
     /**
-     * @var string[string]
+     * @var array<string,string>
      */
     private static $replacePatterns = array(
         "/(?<=\W|^)\\\'(?:((?!\\').)*)\\\'(?=\W|$)/" => "'([^']*)'", // Single quoted strings
--- a/src/Behat/Behat/EventDispatcher/Cli/StopOnFailureController.php
+++ b/src/Behat/Behat/EventDispatcher/Cli/StopOnFailureController.php
@@ -81,12 +81,7 @@ final class StopOnFailureController implements Controller
     }
 
     /**
-     * Executes controller.
-     *
-     * @param InputInterface  $input
-     * @param OutputInterface $output
-     *
-     * @return null|integer
+     * {@inheritdoc}
      */
     public function execute(InputInterface $input, OutputInterface $output)
     {
@@ -95,6 +90,7 @@ final class StopOnFailureController implements Controller
         }
 
         $this->stopOnFailureHandler->registerListeners();
+        return null;
     }
 
 }
--- a/src/Behat/Behat/EventDispatcher/Cli/StopOnFailureController.php
+++ b/src/Behat/Behat/EventDispatcher/Cli/StopOnFailureController.php
@@ -80,9 +80,6 @@ final class StopOnFailureController implements Controller
         );
     }
 
-    /**
-     * {@inheritdoc}
-     */
     public function execute(InputInterface $input, OutputInterface $output)
     {
         if (!$input->getOption('stop-on-failure')) {
--- a/src/Behat/Behat/Gherkin/ServiceContainer/GherkinExtension.php
+++ b/src/Behat/Behat/Gherkin/ServiceContainer/GherkinExtension.php
@@ -79,25 +79,26 @@ final class GherkinExtension implements Extension
      */
     public function configure(ArrayNodeDefinition $builder)
     {
-        $builder
+        $childrenBuilder = $builder
             ->addDefaultsIfNotSet()
             ->children()
-                ->scalarNode('cache')
-                    ->info('Sets the gherkin parser cache folder')
-                    ->defaultValue(
-                        is_writable(sys_get_temp_dir())
-                            ? sys_get_temp_dir() . DIRECTORY_SEPARATOR . 'behat_gherkin_cache'
-                            : null
-                    )
-                ->end()
-                ->arrayNode('filters')
-                    ->info('Sets the gherkin filters (overridable by CLI options)')
-                    ->performNoDeepMerging()
-                    ->defaultValue(array())
-                    ->useAttributeAsKey('name')
-                    ->prototype('scalar')->end()
-                ->end()
-            ->end()
+        ;
+        $childrenBuilder
+            ->scalarNode('cache')
+                ->info('Sets the gherkin parser cache folder')
+                ->defaultValue(
+                    is_writable(sys_get_temp_dir())
+                        ? sys_get_temp_dir() . DIRECTORY_SEPARATOR . 'behat_gherkin_cache'
+                        : null
+                )
+        ;
+        $childrenBuilder
+            ->arrayNode('filters')
+                ->info('Sets the gherkin filters (overridable by CLI options)')
+                ->performNoDeepMerging()
+                ->defaultValue(array())
+                ->useAttributeAsKey('name')
+                ->prototype('scalar')->end()
         ;
     }
 
--- a/src/Behat/Behat/HelperContainer/BuiltInServiceContainer.php
+++ b/src/Behat/Behat/HelperContainer/BuiltInServiceContainer.php
@@ -115,13 +115,8 @@ final class BuiltInServiceContainer implements PsrContainerInterface
 
     /**
      * Gets and validates a class from schema.
-     *
-     * @param string       $id
-     * @param string|array $schema
-     *
-     * @return string
      */
-    private function getAndValidateClass($id, array $schema)
+    private function getAndValidateClass(string $id, array $schema): string
     {
         if (!isset($schema['class'])) {
             $schema['class'] = $id;
--- a/src/Behat/Behat/Output/Node/EventListener/AST/ScenarioNodeListener.php
+++ b/src/Behat/Behat/Output/Node/EventListener/AST/ScenarioNodeListener.php
@@ -78,14 +78,10 @@ final class ScenarioNodeListener implements EventListener
 
     /**
      * Prints scenario/background header on BEFORE event.
-     *
-     * @param Formatter                     $formatter
-     * @param ScenarioLikeTested|AfterSetup $event
-     * @param string                        $eventName
      */
-    private function printHeaderOnBeforeEvent(Formatter $formatter, ScenarioLikeTested|AfterSetup $event, $eventName)
+    private function printHeaderOnBeforeEvent(Formatter $formatter, ScenarioLikeTested $event, string $eventName)
     {
-        if ($this->beforeEventName !== $eventName || !$event instanceof AfterSetup || !$event instanceof ScenarioLikeTested) {
+        if ($this->beforeEventName !== $eventName || !$event instanceof AfterSetup) {
             return;
         }
 
@@ -98,12 +94,8 @@ final class ScenarioNodeListener implements EventListener
 
     /**
      * Prints scenario/background footer on AFTER event.
-     *
-     * @param Formatter                      $formatter
-     * @param ScenarioLikeTested|AfterTested $event
-     * @param string                         $eventName
      */
-    private function printFooterOnAfterEvent(Formatter $formatter, ScenarioLikeTested|AfterTested $event, $eventName)
+    private function printFooterOnAfterEvent(Formatter $formatter, ScenarioLikeTested $event, string $eventName)
     {
         if ($this->afterEventName !== $eventName || !$event instanceof AfterTested) {
             return;
--- a/src/Behat/Behat/Output/Node/EventListener/AST/ScenarioNodeListener.php
+++ b/src/Behat/Behat/Output/Node/EventListener/AST/ScenarioNodeListener.php
@@ -83,9 +83,9 @@ final class ScenarioNodeListener implements EventListener
      * @param ScenarioLikeTested|AfterSetup $event
      * @param string                        $eventName
      */
-    private function printHeaderOnBeforeEvent(Formatter $formatter, ScenarioLikeTested $event, $eventName)
+    private function printHeaderOnBeforeEvent(Formatter $formatter, ScenarioLikeTested|AfterSetup $event, $eventName)
     {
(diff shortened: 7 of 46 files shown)


.gitignore
```diff
@@ -5,3 +5,4 @@ behat.yml
 vendor
 composer.lock
 .phpunit.result.cache
+phpstan.neon
```
carlos-granados, line 8: Allows having a local phpstan.neon files that overrides some rules or modifies some settings

src/Behat/Behat/Context/Cli/ContextSnippetsController.php
```diff
@@ -87,5 +87,6 @@ public function execute(InputInterface $input, OutputInterface $output)
                 new FixedPatternIdentifier($input->getOption('snippets-type'))
             ))
         );
+        return null;
```
carlos-granados, line 90: These controller functions need to return a value

src/Behat/Behat/Definition/Pattern/Policy/RegexPatternPolicy.php
```diff
@@ -22,7 +22,7 @@
 final class RegexPatternPolicy implements PatternPolicy
 {
     /**
-     * @var string[string]
+     * @var array<string,string>
```
carlos-granados, line 25: Annotation was wrong

src/Behat/Behat/EventDispatcher/Cli/StopOnFailureController.php
```diff
@@ -81,12 +81,7 @@ public function configure(Command $command)
     }
 
     /**
-     * Executes controller.
-     *
-     * @param InputInterface  $input
-     * @param OutputInterface $output
-     *
-     * @return null|integer
+     * {@inheritdoc}
```
carlos-granados, line 84: Prefer to use the base annotation for all controller classes

src/Behat/Behat/Gherkin/ServiceContainer/GherkinExtension.php
```diff
@@ -79,25 +79,26 @@ public function initialize(ExtensionManager $extensionManager)
      */
     public function configure(ArrayNodeDefinition $builder)
     {
-        $builder
+        $childrenBuilder = $builder
```
carlos-granados, line 82: In some cases the `end()` function will return a type that cannot be used to chain other calls to other config functions because these other functions are not defined in the returned type. We split the chain of calls so that the functions are always defined in the returned types

src/Behat/Behat/HelperContainer/BuiltInServiceContainer.php
```diff
@@ -115,13 +115,8 @@ private function getAndValidateServiceSchema($id)
...
     /**
      * Gets and validates a class from schema.
-     *
-     * @param string       $id
-     * @param string|array $schema
-     *
-     * @return string
      */
-    private function getAndValidateClass($id, array $schema)
+    private function getAndValidateClass(string $id, array $schema): string
```
carlos-granados, line 119: The annotation for the $schema parameter was wrong. I changed this to use proper php types everywhere

src/Behat/Behat/Output/Node/EventListener/AST/ScenarioNodeListener.php
```diff
@@ -83,9 +83,9 @@ public function listenEvent(Formatter $formatter, Event $event, $eventName)
      * @param ScenarioLikeTested|AfterSetup $event
      * @param string                        $eventName
      */
-    private function printHeaderOnBeforeEvent(Formatter $formatter, ScenarioLikeTested $event, $eventName)
+    private function printHeaderOnBeforeEvent(Formatter $formatter, ScenarioLikeTested|AfterSetup $event, $eventName)
```
carlos-granados, line 86: The event can implement any of these interfaces

src/Behat/Behat/Output/Node/EventListener/AST/ScenarioNodeListener.php
```diff
@@ -83,9 +83,9 @@ public function listenEvent(Formatter $formatter, Event $event, $eventName)
...
     {
-        if ($this->beforeEventName !== $eventName || !$event instanceof AfterSetup) {
+        if ($this->beforeEventName !== $eventName || !$event instanceof AfterSetup || !$event instanceof ScenarioLikeTested) {
```
carlos-granados, line 88: But we only want to deal with it when it implements both

src/Behat/Behat/Output/Node/EventListener/Statistics/HookStatsListener.php
```diff
@@ -111,6 +112,7 @@ private function captureAfterHookStats(HookedTeardown $teardown)
     private function captureHookStat(CallResult $hookCallResult)
     {
         $call = $hookCallResult->getCall();
+        assert($call instanceof HookCall);
```
carlos-granados, line 115: type narrowing

src/Behat/Behat/Output/Node/Printer/JUnit/JUnitSetupPrinter.php
```diff
@@ -52,23 +54,20 @@ public function printTeardown(Formatter $formatter, Teardown $teardown)
         }
     }
 
-    /**
```
carlos-granados, line 55: Unnecessary annotations

src/Behat/Behat/Output/Node/Printer/JUnit/JUnitSetupPrinter.php
```diff
@@ -52,23 +54,20 @@ public function printTeardown(Formatter $formatter, Teardown $teardown)
...
-     * @param Formatter $formatter
-     * @param CallResults $results
-     * @param string $messageType
-     */
-    private function handleHookCalls(Formatter $formatter, CallResults $results, $messageType)
+    private function handleHookCalls(Formatter $formatter, CallResults $results, string $messageType): void
     {
-        /** @var CallResult $hookCallResult */
         foreach ($results as $hookCallResult) {
+            assert($hookCallResult instanceof CallResult);
```
carlos-granados, line 60: Prefer to use assertions to @var annotations

src/Behat/Behat/Snippet/SnippetRegistry.php
```diff
@@ -55,7 +55,7 @@ public function registerSnippetGenerator(SnippetGenerator $generator)
      * @param Environment $environment
      * @param StepNode    $step
      *
-     * @return null|Snippet
+     * @return void
```
carlos-granados, line 58: Function does not return anything

src/Behat/Behat/Tester/ServiceContainer/TesterExtension.php
```diff
@@ -81,8 +81,6 @@ public function configure(ArrayNodeDefinition $builder)
                             ? sys_get_temp_dir() . DIRECTORY_SEPARATOR . 'behat_rerun_cache'
                             : null
                     )
-                ->end()
```
carlos-granados, line 84: These `end()` calls are not really needed because they just return a value which is not used and they return a type which cannot be chained with another end call, so I just removed them

src/Behat/Hook/Hook.php
```diff
@@ -18,4 +18,5 @@
  */
 interface Hook
 {
+    public function getFilterString(): ?string;
```
carlos-granados, line 21: Adding this function call is a BC break but I think we can allow it. Otherwise we cannot guarantee that a class implementing this interface will have a filterString property. I don't expect anyone to have defined other classes that implement this interface which is applied to Hook attributes

src/Behat/Step/Definition.php
```diff
@@ -18,4 +18,5 @@
  */
 interface Definition
 {
+    public function getPattern(): ?string;
```
carlos-granados, line 21: Same here for access to the pattern property, again it is a BC break but I don't expect anyone to have defined other classes that implement this interface

src/Behat/Testwork/Specification/SpecificationIterator.php
```diff
@@ -18,7 +18,7 @@
  *
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
  *
- * @extends Iterator<int, Suite>
+ * @extends Iterator<int, mixed>
```
carlos-granados, line 21: This annotation was wrong, my mistake, looking at its definition I thought it would implement an iterator of Suites but it actually implements an iterator of specifications which are mixed

.gitignore
stof: I always failed to see the use case for that. Running phpstan locally should use the same config than the CI IMO

src/Behat/Behat/EventDispatcher/Cli/StopOnFailureController.php
stof: The inline tag `{@inheritdoc}` is about inheriting the parent description, which is useful only to merge it with more description. It has nothing to do with inheriting tags like `@param` (which are inherited automatically anyway in phpdoc tools).
You should either remove the phpdoc entirely or use the `@inheritdoc` tag as an explicit marker meaning "the phpdoc is inherited from the parent, I did not forgot to add it".
My preference is to remove it entirely.

src/Behat/Behat/Output/Node/EventListener/AST/ScenarioNodeListener.php
stof: no need to change this native type. The check in `listenEvent` ensures that we always get a ScenarioLikeTested

src/Behat/Behat/Output/Node/EventListener/AST/ScenarioNodeListener.php
```diff
@@ -103,7 +103,7 @@ private function printHeaderOnBeforeEvent(Formatter $formatter, ScenarioLikeTest
      * @param ScenarioLikeTested|AfterTested $event
      * @param string                         $eventName
      */
-    private function printFooterOnAfterEvent(Formatter $formatter, ScenarioLikeTested $event, $eventName)
+    private function printFooterOnAfterEvent(Formatter $formatter, ScenarioLikeTested|AfterTested $event, $eventName)
```
stof, line 106: same here

.gitignore
carlos-granados: I have this configuration in this file:
```
includes:
	- phpstan.dist.neon

parameters:
	editorUrl: 'phpstorm://open?file=%%file%%&line=%%line%%'
```
This allows me to open phpstan found errors in phpstorm, but obviously this config is local to me

src/Behat/Behat/Output/Node/Printer/JUnit/JUnitSetupPrinter.php
stof: The right solution here is to add `@template-implements \IteratorAggregate<array_key, CallResult>` on the `CallResults` class, removing the need for local type refinements.

I agree that assertions are better than `@var` because wrong assertions will make our CI fail (assuming we configured GHA to use the development ini config in setup-php), but proper type inference is even better from a type safety point of view, as it is better to fail during type checking than at runtime.

.gitignore
stof: ah, I did not know that phpstan had this parameter. This is indeed a valid use case.

src/Behat/Testwork/Specification/SpecificationIterator.php
stof: It should actually be generic instead of using `mixed`, allowing better types for implementations of this interface. For instance, `LazyFeatureIterator` is an iterator over `FeatureNode`, not over `mixed`.
Basically, it should forward the (covariant) template type from `Iterator` for the value type.

`NoSpecificationsIterator` will be a `SpecificationIterator<never>` and `SpecificationArrayIterator` will be generic (inferred from the constructor argument passing the array). `GroupedSpecificationIterator` will also be generic (and its static `group` method as well).

carlos-granados: You lost me here. What would be your proposed solution?

src/Behat/Behat/EventDispatcher/Cli/StopOnFailureController.php
carlos-granados: OK, understood, I also prefer to remove it

src/Behat/Behat/Output/Node/EventListener/AST/ScenarioNodeListener.php
carlos-granados: Yeah, my solution was wrong, I actually needed to fix the doc-block annotation but in fact it is even better to just remove it

src/Behat/Testwork/Specification/SpecificationIterator.php
stof:
```php
/**
 * @template-covariant T
 * @extends Iterator<int, T>
 */
```

And then updating child classes to define the type of specification they work with.

src/Behat/Behat/Output/Node/Printer/JUnit/JUnitSetupPrinter.php
carlos-granados: The `CallResults` class already had the `@implements` annotation so this assert was not actually needed (and neither a couple of other asserts). I had converted the @var annotations to asserts without checking if they were actually needed, now they have been removed

src/Behat/Testwork/Specification/SpecificationIterator.php
carlos-granados: This surpasses my knowledge/understanding of generics, I am in a bit of a loss to what would actually need to go in the child classes, if you can provide some code samples (or a new commit), I would be grateful

stof: I started adding generic types for `SpecificationIterator`, and this is quite far-reaching (which is mostly related to the fact that `Behat/Testwork` is written to be agnostic of Gherkin and so does not hardcode that the specification being used is FeatureNode). I'll submit that as a separate PR instead.

stof: I submitted it as #1575

.gitignore
acoulton: Very useful tip, I didn't know that either - thanks!

src/Behat/Behat/Gherkin/ServiceContainer/GherkinExtension.php
acoulton: IMO this is also more readable so 👍

src/Behat/Behat/Output/Node/Printer/JUnit/JUnitSetupPrinter.php
```diff
@@ -52,23 +54,17 @@ public function printTeardown(Formatter $formatter, Teardown $teardown)
...
-                /** @var HookCall $call */
                 $call = $hookCallResult->getCall();
                 $scope = $call->getScope();
-                /** @var JUnitOutputPrinter $outputPrinter */
                 $outputPrinter = $formatter->getOutputPrinter();
+                assert($outputPrinter instanceof JUnitOutputPrinter);
 
-                $message = $call->getCallee()->getName();
+                $callee = $call->getCallee();
+                $message = $callee->getName();
```
acoulton, line 67: Not sure why this change?

src/Behat/Hook/Hook.php
acoulton: We should remember to mention this (and Definition) in the changelog. If we hit similar things in future we should probably pull to a separate PR for visibility, but I think it's OK here.

acoulton (approved):
Thanks @carlos-granados :)

PR #1575, opened 2024-12-21
Add generic types for the specification type for SpecificationIterator

stof:
in hook and event objects, I decided to use `SpecificationIterator<mixed>` as type instead of making event classes generic, to avoid projects using hooks or events to have to specify that they work with a `BeforeSuiteHook<FeatureNode>` in Behat.
I expect the `getSpecificationIterator` or `getSpecificationIterators` methods of those objects to be used only for rare use cases, so the additional type safety for those rare use cases is not worth annoying everyone for the simple use cases.


acoulton (approved):
This is great, thanks for this @stof - at the edges of my generics experience here!

carlos-granados (approved):
Thanks @stof !

PR #1576, opened 2024-12-21
feat: adds a formatter option that allows you to set if you want stdout displayed as part of the formatter output

carlos-granados:
When running your tests, the stdout output from the tests will not be printed as part of the formatter output if you use the progress formatter and will be printed if you use the pretty formatter. These are sensible defaults but in some cases you may want to see the test output as part of the progress output or not see the output as part of the pretty output.

This PR adds a new `show_output` option to the formatters options which can be used to enable or disable stdout for any formatter (only applies to pretty and progress formatters, not the junit formatter where it does not make sense to include the output)

This option can be set to `yes`, `no` or `on-fail` for the pretty formatter. It defaults to `yes`
This option can be set to `yes`, `no`, `on-fail` or `in-summary` for the progress formatter. It defaults to `in-summary`. This option will show the test output in the summary at the end of the test run if there are any failed steps

### DIFF
--- a/src/Behat/Behat/Output/Node/Printer/ListPrinter.php
+++ b/src/Behat/Behat/Output/Node/Printer/ListPrinter.php
@@ -102,13 +102,17 @@ final class ListPrinter
     /**
      * Prints step list.
      *
-     * @param OutputPrinter $printer
      * @param string        $intro
      * @param integer       $resultCode
      * @param StepStat[]    $stepStats
      */
-    public function printStepList(OutputPrinter $printer, $intro, $resultCode, array $stepStats)
-    {
+    public function printStepList(
+        OutputPrinter $printer,
+        $intro,
+        $resultCode,
+        array $stepStats,
+        string $showOutput = "in-summary"
+    ) {
         if (!count($stepStats)) {
             return;
         }
@@ -120,9 +124,17 @@ final class ListPrinter
 
         foreach ($stepStats as $num => $stepStat) {
             if ($stepStat instanceof StepStatV2) {
-                $this->printStepStat($printer, $num + 1, $stepStat, $style);
+                $this->printStepStat($printer, $num + 1, $stepStat, $style, $showOutput);
             } elseif ($stepStat instanceof StepStat) {
-                $this->printStat($printer, $stepStat->getText(), $stepStat->getPath(), $style, $stepStat->getStdOut(), $stepStat->getError());
+                $this->printStat(
+                    $printer,
+                    $stepStat->getText(),
+                    $stepStat->getPath(),
+                    $style,
+                    $stepStat->getStdOut(),
+                    $stepStat->getError(),
+                    $showOutput
+                );
             }
         }
     }
@@ -157,7 +169,6 @@ final class ListPrinter
     /**
      * Prints hook stat.
      *
-     * @param OutputPrinter $printer
      * @param string        $name
      * @param string        $path
      * @param string        $style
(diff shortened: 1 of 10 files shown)


src/Behat/Behat/Output/Node/Printer/ListPrinter.php
```diff
@@ -102,13 +102,17 @@ public function printScenariosList(OutputPrinter $printer, $intro, $resultCode,
...
      * @param StepStat[]    $stepStats
      */
-    public function printStepList(OutputPrinter $printer, $intro, $resultCode, array $stepStats)
-    {
+    public function printStepList(
+        OutputPrinter $printer,
+        $intro,
+        $resultCode,
+        array $stepStats,
+        string $showOutput = "in-summary"
```
carlos-granados, line 114: The default behaviour is to only print the test output in the summary

src/Behat/Behat/Output/Node/Printer/Progress/ProgressStepPrinter.php
```diff
@@ -52,6 +56,10 @@ public function printStep(Formatter $formatter, Scenario $scenario, StepNode $st
         $printer = $formatter->getOutputPrinter();
         $style = $this->resultConverter->convertResultToString($result);
 
+        if ($this->hasPrintedOutput) {
```
carlos-granados, line 59: After printing any output, we need to print a new line before continuing to print the "dots" of the progress

carlos-granados:
Replaces https://github.com/Behat/Behat/pull/1471

src/Behat/Behat/Output/Node/Printer/Progress/ProgressStepPrinter.php
acoulton: I would suggest perhaps putting this as a comment in the code for future reference?

acoulton: Also, I might be missing it but how does this flag get set back to false if the next step does *not* have any output?

src/Behat/Config/Formatter/ProgressFormatter.php
```diff
@@ -10,12 +10,15 @@ final class ProgressFormatter extends Formatter
 
     /**
      * @param bool $timer show time and memory usage at the end of the test run
+     * @param string $showOutput show the test stdout output as part of the formatter output (yes, no, on-fail, in-summary)
```
acoulton, line 13: A question - would it be worth using a `BackedEnum` for these? (I realise might have to be two, one for Pretty & one for  Progress, since they take different options). 

I guess would have to convert that to the string value in `->toArray` for parity with YAML / CLI options but we could maybe then cast it back to the enum when we read `->formatter->getParameter()` and use it internally?

It would give us validation of the supported values (and help users to pick the right ones with the PHP config interface) and type/value-safety internally. It might also be cleaner to move the duplicated `if ($showOutput === 'yes' || ($showOutput === 'on-fail' && !$result->isPassed())) {` to the enum implementation(s) in case of any future new modes (though perhaps not, I can't at the moment think what other modes we might need).

I'd equally be very happy with this as-is if you think enums could cause issues / would be too hard to support well - just thought worth considering for new features now we are >=php8 and have the PHP config interface.

acoulton (commented):
Thanks for doing this @carlos-granados

src/Behat/Behat/Output/Node/Printer/Progress/ProgressStepPrinter.php
carlos-granados: Comment added. Regarding setting the flag back to false, I made a last minute change to the code and when doing so, I accidentally removed this. Thanks for noticing it!

src/Behat/Config/Formatter/ProgressFormatter.php
acoulton: Makes sense, thanks

acoulton (approved):
Great, thanks :)

AlexSkrypnyk:
This just broke Behat that is using a custom printer that does not have `show_option` set in config.

In `printStatistics()`, there is `$showOutput = $formatter->getParameter(ShowOutputOption::OPTION_NAME);` that returns `NULL` id the `show_option` is not provided. That NULL is then passed into `printStepList()`.

```
TypeError: Behat\Behat\Output\Node\Printer\ListPrinter::printStepList(): Argument #5 ($showOutput) must be of type Behat\Config\Formatter\ShowOutputOption, null given, called in 
```

There should be a check added for the `$showOutput` value and a default should be used instead.

AlexSkrypnyk:
Also, this PR did not add any documentation about using this new option.

PR #1578, opened 2024-12-23
chore: update code to PHPStan level 3

carlos-granados:
Adds the changes needed to reach PHPStan level 3

I have seen that it should not be too difficult to reach level 5, so I will wait until then before attempting to remove Psalm

### DIFF
--- a/src/Behat/Behat/EventDispatcher/Event/AfterScenarioSetup.php
+++ b/src/Behat/Behat/EventDispatcher/Event/AfterScenarioSetup.php
@@ -65,9 +65,7 @@ final class AfterScenarioSetup extends ScenarioTested implements AfterSetup
     }
 
     /**
-     * Returns scenario node.
-     *
-     * @return ScenarioNode
+     * @return Scenario
      */
     public function getScenario()
     {
--- a/src/Behat/Behat/Output/Node/EventListener/AST/OutlineListener.php
+++ b/src/Behat/Behat/Output/Node/EventListener/AST/OutlineListener.php
@@ -33,7 +33,7 @@ use Behat\Testwork\Output\Node\EventListener\EventListener;
  */
 final class OutlineListener implements EventListener
 {
-    private ?ExampleNode $example;
+    private ?ExampleNode $example = null;
 
     public function __construct(
         private OutlinePrinter $outlinePrinter,
--- a/src/Behat/Behat/Output/Node/EventListener/AST/OutlineListener.php
+++ b/src/Behat/Behat/Output/Node/EventListener/AST/OutlineListener.php
@@ -33,52 +33,15 @@ use Behat\Testwork\Output\Node\EventListener\EventListener;
  */
 final class OutlineListener implements EventListener
 {
-    /**
-     * @var OutlinePrinter
-     */
-    private $outlinePrinter;
-    /**
-     * @var ExamplePrinter
-     */
-    private $examplePrinter;
-    /**
-     * @var StepPrinter
-     */
-    private $stepPrinter;
-    /**
-     * @var SetupPrinter
-     */
-    private $stepSetupPrinter;
-    /**
-     * @var SetupPrinter
-     */
-    private $exampleSetupPrinter;
-    /**
-     * @var ExampleNode
-     */
-    private $example;
+    private ?ExampleNode $example;
 
-    /**
-     * Initializes listener.
-     *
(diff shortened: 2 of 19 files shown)


src/Behat/Behat/EventDispatcher/Event/AfterScenarioSetup.php
```diff
@@ -65,9 +65,7 @@ public function getFeature()
     }
 
     /**
-     * Returns scenario node.
-     *
-     * @return ScenarioNode
+     * @return Scenario
```
carlos-granados, line 68: annotation was wrong and I used the opportunity a comment which does not provide any useful information

src/Behat/Behat/Output/Node/EventListener/AST/OutlineListener.php
```diff
@@ -33,52 +33,15 @@
...
-    private $stepSetupPrinter;
-    /**
-     * @var SetupPrinter
-     */
-    private $exampleSetupPrinter;
-    /**
-     * @var ExampleNode
-     */
-    private $example;
+    private ?ExampleNode $example;
```
carlos-granados, line 36: Some properties need to be defined as nullable. I used the opportunity to also change several properties to use constructor promotion

src/Behat/Behat/Output/Node/EventListener/AST/OutlineTableListener.php
```diff
@@ -123,7 +90,7 @@ private function captureStepEvent(StepTested $event)
     {
         if ($event instanceof AfterStepSetup) {
             $this->stepBeforeTestedEvents[$event->getStep()->getLine()] = $event;
-        } else {
+        } elseif ($event instanceof AfterStepTested) {
```
carlos-granados, line 93: The following code should only apply if the event is of this type

src/Behat/Behat/Output/Node/EventListener/Statistics/StepStatsListener.php
```diff
@@ -160,12 +140,8 @@ private function captureStepStatsOnAfterEvent(Event $event)
 
     /**
      * Gets exception from the step test results.
-     *
-     * @param StepResult $result
-     *
-     * @return null|Exception
      */
-    private function getStepException(StepResult $result)
+    private function getStepException(StepResult $result): ?Throwable
```
carlos-granados, line 144: getException is now marked as returning a Throwable

src/Behat/Behat/Output/Statistics/ScenarioStat.php
```diff
@@ -18,30 +20,13 @@
...
-     * @var string
-     */
-    private $title;
-    /**
-     * @var string
-     */
-    private $path;
-    /**
-     * @var integer
+     * @param TestResult::* $resultCode
```
carlos-granados, line 23: We add more specific types for some parameters and return values

src/Behat/Behat/Transformation/Transformation/ReturnTypeTransformation.php
```diff
@@ -107,7 +107,7 @@ public function getPriority()
      */
     public function getPattern()
     {
-        return null;
+        return '';
```
carlos-granados, line 110: This function needs to return a string

src/Behat/Behat/Output/Node/EventListener/AST/OutlineTableListener.php
```diff
@@ -37,61 +37,28 @@
...
-    private $outline;
-    /**
-     * @var Setup
-     */
-    private $exampleSetup;
-    /**
-     * @var bool
-     */
-    private $headerPrinted = false;
+    private ?OutlineNode $outline;
```
stof, line 40: I would add `= null` to avoid an uninitialized state for those properties (both phpstan and psalm are bad at dealing with potentially-uninitialized properties)

stof: same for lots of properties for which you added property types btw.

carlos-granados: Good shout, updated

carlos-granados:
I have created an issue in the docs repo to update the formatters documentation https://github.com/Behat/docs/issues/185

acoulton (approved):
Great, thanks :)

Issue #1577, opened 2024-12-23
Add config/env to always enable xdebug


asgrim:
I found out that 3.17 disables Xdebug by default (ref #1560). Since I very _very_ often use Xdebug to help write the contexts, I need step debugging to be always enabled when working locally with it. In CI, it would usually make sense to disable Xdebug, but not for local development.

I'd like to propose either reverting the default behaviour of disabling Xdebug, or providing some way of always not disabling Xdebug without having to specify `--xdebug` every time, e.g. with an environment variable, maybe something in `behat.php` or similar. It makes sense for tools like phpstan/rector/composer to disable Xdebug, but IMO not in Behat; we are writing test code with Behat, and just like writing tests with phpunit/Codeception, devs should be able to step through that test code. I personally would rather see something like a CI option to improve performance, e.g. `vendor/bin/behat --no-xdebug` or similar.

ostrolucky:
At the moment, you can set `BEHAT_ALLOW_XDEBUG=1` as env variable. This way you can set it globally for all your project, since you can set env variables for your unix user by default too.

stof:
phpunit has a nice trick in its phpt runner where it launches subprocesses: it does not disable XDebug when there is a debugger session attached (without requiring any opt-in from the CLI). I'm wondering whether something similar could be done in Behat.

carlos-granados:
@stof that's a really good suggestion, I implemented it in this PR https://github.com/Behat/Behat/pull/1581

magnetik:
I've got bitten by this too. Looking forward for the fix !

carlos-granados:
Closed via https://github.com/Behat/Behat/pull/1581

PR #1581, opened 2024-12-24
Do not disable xdebug if there is an active debugging connection

carlos-granados:
Following @stof 's suggestion in https://github.com/Behat/Behat/issues/1577 in this PR we check if there is an active Xdebug session, and if there is, we don't disable it, even if no `--xdebug` flag has been passed

### DIFF
--- a/src/Behat/Testwork/Cli/Application.php
+++ b/src/Behat/Testwork/Cli/Application.php
@@ -99,7 +99,9 @@ final class Application extends BaseApplication
      */
     public function doRun(InputInterface $input, OutputInterface $output): int
     {
-        $isXdebugAllowed = $input->hasParameterOption('--xdebug');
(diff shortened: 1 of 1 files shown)


acoulton (approved):
Great solution, thanks

ostrolucky:
Tested, fixes my issue

carlos-granados:
@acoulton given that some users have been affected by this, I thought that perhaps we could release this change as a patch release with only this fix, instead of waiting for a regular release. What do you think?

acoulton:
@carlos-granados we could do - but looking at https://github.com/Behat/Behat/compare/v3.17.0...master I don't think there's really anything controversial there to specifically require a separate 3.17.1 with just this change. So maybe we just merge this and do a 3.18.0 (which would be simplest)?

carlos-granados:
@acoulton ok, let's go ahead with a 3.18 release. Can you prepare it?

Issue #1405, opened 2022-07-13
Do not trigger CI build when only changelog file is updated


pamil:

carlos-granados:
Seems this could be achieved using the `paths-ignore` option in the workflow. There may be other files that we might want to ignore as well, like the README

codisart:
Here is my proposal #1582, I added to the ignore paths, CHANGELOG, CONTRIBUTING, README.

carlos-granados:
Fixed in https://github.com/Behat/Behat/pull/1582

PR #1582, opened 2024-12-28
ci: #1405 do not run ci builds for some specific files

codisart:
fix latest php version for matrix strategy

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,7 +3,11 @@ name: Build
 on:
   push:
     branches: [master]
-  pull_request: ~
+    paths-ignore:
+      - '**.md'
+  pull_request:
+    paths-ignore:
+      - '**.md'
   release:
     types: [created]
   workflow_dispatch: ~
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -3,7 +3,15 @@ name: Build
 on:
   push:
     branches: [master]
-  pull_request: ~
+    paths-ignore:
+      - 'CHANGELOG.md'
```
carlos-granados, line 7: Perhaps we could just use '- '**.md' to cover all markdown files?

.github/workflows/build.yml
```diff
@@ -30,19 +38,19 @@ jobs:
...
           # Windows on latest PHP only
-          - php: 8.2
+          - php: 8.4
             os: windows-latest
             composer-mode: update
             symfony-version: ''
 
           # Symfony LTS jobs:
-          - php: 8.2
+          - php: 8.4
```
carlos-granados, line 53: I think it is best if this job remains with PHP 8.2

carlos-granados (changes requested):

.github/workflows/build.yml
codisart: better idea ! 

done.

.github/workflows/build.yml
codisart: done

acoulton (approved):
LGTM, thanks!

carlos-granados (approved):
Thanks @codisart

PR #1380, opened 2021-12-25
Add support for psr/container 2.0, remove container-interop

wouterj:
This adds support for psr/container 2.0 (on top of 1.0, this is why the parameter types aren't introduced).

Container-interop is abandoned since psr/container 1.0 and does not support 2.0. As users using container-interop have been warned about this for more than 2 years now while running composer install, I think it's safe to remove the integration of this package.

PR #1583, opened 2025-01-03
chore: update code to PHPStan level 4

carlos-granados:


phpstan.dist.neon
```diff
@@ -13,4 +13,24 @@ parameters:
         -
             identifier: constructor.unusedParameter
             paths:
+                - src/Behat/Behat/Definition/Printer/ConsoleDefinitionPrinter.php
```
carlos-granados, line 16: All these classes have an unused parameter in the constructor. We cannot remove it because they are handled in a generic way that provides this parameter when constructing them so we need to ignore the error

phpstan.dist.neon
```diff
@@ -13,4 +13,24 @@ parameters:
...
+                - src/Behat/Behat/EventDispatcher/Cli/StopOnFailureController.php
+                - src/Behat/Behat/Output/Node/Printer/ListPrinter.php
+                - src/Behat/Behat/Tester/Runtime/RuntimeFeatureTester.php
                 - src/Behat/Behat/Transformation/Transformation/ReturnTypeTransformation.php
+                - src/Behat/Testwork/Ordering/Cli/OrderController.php
+        -
+            identifier: function.alreadyNarrowedType
```
carlos-granados, line 23: This error and the next one happen in the event dispatcher that needs to work with both versions of the event dispatcher which change the order of the parameters in the dispatch function, so we need to ignore them

phpstan.dist.neon
```diff
@@ -13,4 +13,24 @@ parameters:
...
+            paths:
+                - src/Behat/Testwork/EventDispatcher/TestworkEventDispatcher.php
+        -
+            identifier: deadCode.unreachable
+            paths:
+                - src/Behat/Testwork/EventDispatcher/TestworkEventDispatcher.php
+
+        -
+            identifier: if.alwaysFalse
```
carlos-granados, line 32: These errors happen because PHPStan cannot understand the `if(false)` construction that we use to define some interfaces

src/Behat/Behat/Context/Snippet/Generator/ContextSnippetGenerator.php
```diff
@@ -161,17 +161,18 @@ private function getUsedClasses(StepNode $step): array
     {
         $usedClasses = [PendingException::class];
 
-        $usedClasses[] = match ($step->getKeywordType()) {
+        $definitionClass = match ($step->getKeywordType()) {
```
carlos-granados, line 164: The match statement did not contemplate the case where the $definitionClass could be something else

src/Behat/Behat/Context/Snippet/Generator/ContextSnippetGenerator.php
```diff
@@ -161,17 +161,18 @@ private function getUsedClasses(StepNode $step): array
...
+        if ($definitionClass) {
+            $usedClasses[] = $definitionClass;
+        }
 
         foreach ($step->getArguments() as $argument) {
-            $usedClasses[] = match (true) {
-                $argument instanceof TableNode => TableNode::class,
-                $argument instanceof PyStringNode => PyStringNode::class,
-            };
+            $usedClasses[] = $argument::class;
```
carlos-granados, line 175: $argument can only be a TableNode or a PyStringNode

src/Behat/Behat/Definition/Exception/AmbiguousMatchException.php
```diff
@@ -23,15 +23,6 @@
  */
 final class AmbiguousMatchException extends RuntimeException implements SearchException
 {
-    /**
-     * @var string
-     */
-    private $text;
```
carlos-granados, line 29: These were not used

src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitDurationListener.php
```diff
@@ -75,11 +75,9 @@ private function captureAfterScenarioEvent(Event $event): void
 
         $key = $this->getHash($event->getScenario());
         $timer = $this->scenarioTimerStore[$key];
-        if ($timer instanceof Timer) {
-            $timer->stop();
-            $this->resultStore[$key] = $timer->getTime();
-            unset($this->scenarioTimerStore[$key]);
-        }
+        $timer->stop();
```
carlos-granados, line 78: $timer is always a Timer

src/Behat/Behat/Output/Node/Printer/ListPrinter.php
```diff
@@ -316,7 +311,8 @@ private function getLocationFromScope(?HookScope $scope): ?string
                     ':' . $scope->getScenario()->getLine(),
                 $scope instanceof BeforeStepScope, $scope instanceof AfterStepScope =>
                     $this->relativizePaths($scope->getFeature()->getFile()) .
-                    ':' . $scope->getStep()->getLine()
+                    ':' . $scope->getStep()->getLine(),
+                default => null
```
carlos-granados, line 315: The match statement needs to cover all possible values

src/Behat/Testwork/Call/CallCenter.php
```diff
@@ -181,7 +181,7 @@ private function handleException($exception)
             $exception = $handler->handleException($exception);
         }
 
-        if ($exception instanceof Throwable) {
+        if (!$exception instanceof Exception) {
```
carlos-granados, line 184: We convert throwables which are not exceptions into exceptions

src/Behat/Testwork/ServiceContainer/Configuration/ConfigurationLoader.php
```diff
@@ -190,6 +190,8 @@ protected function loadEnvironmentConfiguration()
      * @return array
      *
      * @throws ConfigurationLoadingException If config file is not found
+     *
+     * @phpstan-impure
```
carlos-granados, line 194: This needs to be added so that PHPStan knows that this function can modify the $profileFound variable

src/Behat/Testwork/Tester/Result/TestResult.php
```diff
@@ -23,6 +23,8 @@ interface TestResult
     public const UNDEFINED = 30;
     public const FAILED = 99;
 
+    public const NO_TESTS = -100;
```
carlos-granados, line 26: We move this constant to this class so that it can be part of the constants associated to the TestResult class

phpstan.dist.neon
acoulton: Would it be better / more obvious to introduce a new interface(s) with the expected constructor signature for these classes?

phpstan.dist.neon
acoulton: It looks to me like these interface definitions may no longer be required since https://github.com/behat/behat/commit/0c6b45fd10198976941c059a3089ddd901228c17 in #1380 in 2021

I can't find any reference to the `Behat\\Behat\\HelperContainer\\ContainerInterface` class and I think we could definitely remove that safely - perhaps keeping the class_alias in that file just in case anyone externally is using it.

The `Behat\\Behat\\HelperContainer\\Exception\\ContainerException` and `Behat\\Behat\\HelperContainer\\ContainerInterface\\NotFoundException` are still used and theoretically could be class_aliased to the exceptions from `container-interop` if end-users have that installed, but that actually feels wrong if our container is using the interfaces from psr/container. And container-interop has been deprecated for a long time.

Removing our custom exception interfaces would be a BC break if anyone is trying to catch them. That seems unlikely, but to be safe we could just make them standard interfaces that directly extend the PSR interface rather than the class_alias and if (false) magic? And then add them to the list to remove in the next major?

src/Behat/Behat/EventDispatcher/Cli/StopOnFailureController.php
```diff
@@ -54,10 +48,10 @@ final class StopOnFailureController implements Controller
      * Initializes controller.
      *
      * @param EventDispatcherInterface $eventDispatcher
+     * @deprecated $eventDispatcher events are now dispatched in the StopOnFailureHandler
```
acoulton, line 51: Doesn't this mark the whole constructor method as deprecated?

acoulton (commented):
@carlos-granados thanks for this, mostly LGTM with a couple of suggestions

src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitDurationListener.php
stof: is it guaranteed that it is a timer, or could it be `null` with a notice about the key not being defined ?

src/Behat/Testwork/Tester/Result/TestResult.php
stof: I'm not sure this is the right solution. Places dealing with the result of a single test would never need to care about the `NO_TESTS` result.

phpstan.dist.neon
carlos-granados: Actually, looking into this in more detail I can see that almost all these classes are never constructed directly other than as part of the container. Should we just remove these unused constructor parameters? There is the tiny possibility that someone out there is constructing them directly but my guess is that all code will be retrieving them from the container so removing the unused parameter should have no effect. What do you think?

The exception is the `ReturnTypeTransformation` class, which is not constructed as part of the container and which is constructed in the `TransformationCalleeFactory` `create()` method. The constructor for these transformations are called in a generic way assuming that they implement the `SimpleArgumentTransformation` interface. We could maybe add the constructor signature to this interface but this would not help in removing the fact that the constructor for this particular class does not use one of the parameters, so I think we need to keep the error ignore

stof: once an interface enforces the signature, phpstan should not be complaining about the unused parameter as it will know that the interface requires it.

src/Behat/Behat/EventDispatcher/Cli/StopOnFailureController.php
stof: this is indeed how this tag will be interpreted.

src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitDurationListener.php
carlos-granados: Ah, yes, it could be that it is not defined. I implemented another fix

phpstan.dist.neon
carlos-granados: @stof I tried this but after adding the constructor signature to the interface, PHPStan is still complaining about the unused constructor parameter and in fact this is listed as a bug in PHPStan 😞 
https://github.com/phpstan/phpstan/issues/11454

src/Behat/Behat/EventDispatcher/Cli/StopOnFailureController.php
carlos-granados: OK, updated

phpstan.dist.neon
carlos-granados: I created a PR in PHPStan to fix this issue https://github.com/phpstan/phpstan-src/pull/3777 😄

phpstan.dist.neon
carlos-granados: Code updated after merging your proposal

phpstan.dist.neon
acoulton: > Actually, looking into this in more detail I can see that almost all these classes are never constructed directly other than as part of the container. Should we just remove these unused constructor parameters? There is the tiny possibility that someone out there is constructing them directly but my guess is that all code will be retrieving them from the container so removing the unused parameter should have no effect. What do you think?

Hmmm - I think you are probably right that it *should* be safe to remove them, I guess even if someone has an extension that is wrapping/decorating them they should really be configuring that through the container as well. But it does make me a little nervous...

acoulton (approved):
Looking good, thanks @carlos-granados

phpstan.dist.neon
carlos-granados: OK, so I left the PHPStan error exclusions and added a deprecated warning to these parameters. I also added this to the list of things to do for our next major release. Does everything look OK now?

src/Behat/Behat/Context/Snippet/Generator/ContextSnippetGenerator.php
stof: this should be solved in behat/gherkin by defining a more precise type for this getter (there is only 3 valid keyword types)

acoulton (approved):
This looks good to me generally, there's just @stof's comment on the ContextSnippetGenerator changes

src/Behat/Behat/Context/Snippet/Generator/ContextSnippetGenerator.php
carlos-granados: OK, until this is solved at that level, I added an assertion here

PR #1584, opened 2025-01-06
change: Remove internal wrappers for PSR Container interfaces

acoulton:
These were previously provided to support either container-interop or psr/container at runtime. container-interop has been deprecated since 2017 and Behat removed support for it in #1380.

@carlos-granados this is a proposal for my comment on #1583 - however in implementing it I've realised the classes were actually all already marked as `@internal` so we could probably justify removing them completely in a minor release and just updating our exception classes to extend the PSR interfaces directly. That would obviously be cleaner.

Looking at the exceptions this is all very low-level and I think it's very unlikely this could affect end-users.


src/Behat/Behat/HelperContainer/Exception/NotFoundException.php
```diff
@@ -10,18 +10,12 @@
 
 namespace Behat\Behat\HelperContainer\Exception;
 
-class_alias(
-    interface_exists('Interop\\Container\\Exception\\NotFoundException')
-        ? 'Interop\\Container\\Exception\\NotFoundException'
-        : 'Psr\\Container\\NotFoundExceptionInterface',
-    'Behat\\Behat\\HelperContainer\\Exception\\NotFoundException'
```
acoulton, line 17: If we wanted to be *certain* that this wouldn't affect end users, we could update our composer.json to conflict with all versions of `container-interop` before 1.2.0.

That was the [final release of container-interop](https://github.com/container-interop/container-interop/pull/91) in Feb 2017, and in that release the interfaces we're aliasing just extend the PSR interfaces so this extra hop is redundant.

Equally I think it's very unlikely anyone still has an older version of container-interop installed at this point (1.1.0 was in January 2015).

carlos-granados:
@acoulton given that these classes were marked as internal and that container-interop was deprecated in 2017 I think we should just remove these classes.

carlos-granados (approved):
Good job, thanks @acoulton

PR #1585, opened 2025-01-09
Remove legacy Symfony event dispatchers

carlos-granados:
While looking at the changes needed to reach PHPStan level 4 I realised that there are a couple of classes which were not used and which we could probably just remove.

These two classes (`TestworkEventDispatcherSymfony5` and `TestworkEventDispatcherSymfonyLegacy`) were used by a previous version of the `TestworkEventDispatcher` class to provide different functionality for Symfony versions <5 and >= 5. This is not used any longer and, since we now only support Symfony 5+ it does not makes sense to maintain these classes. There is a very minor risk that someone out there may be using these classes but a quick search through GitHub does not reveal any usage and these classes were just an internal implementation detail so I don't believe they will be directly used by anyone. And they concentrate the vast majority of ignored errors in Psalm and PHPStan.

I also modified the `TestworkEventDispatcher` because now we only need to support the parameters supported by Symfony 5+, so we don't need the previous bc functionality

I would like to merge this before I merge the code to upgrade to PHPStan level 4, @acoulton please let me know what you think


acoulton (approved):
LGTM, agreed these should be long-gone and shouldn't be being used by end-users.

PR #1587, opened 2025-01-10
Fix ShowOutputOption handling to allow for the case when a formatter does not define it

carlos-granados:
Fixes the problem detected by @AlexSkrypnyk and reported here https://github.com/Behat/Behat/pull/1576


acoulton (approved):

AlexSkrypnyk:
Thank you so much for fixing this so fast!

PR #1589, opened 2025-01-16
chore: changes to achieve PHPStan level 5

carlos-granados:
This is the last PHPStan level that we can achieve easily. Level 6 requires a lot more changes. Once this PR is merged I will work to remove PSalm from our codebase

### DIFF
--- a/src/Behat/Behat/Context/Cli/InteractiveContextIdentifier.php
+++ b/src/Behat/Behat/Context/Cli/InteractiveContextIdentifier.php
@@ -69,7 +69,7 @@ final class InteractiveContextIdentifier implements TargetContextIdentifier
         }
 
         $message = $this->translator->trans('snippet_context_choice', array('%count%' => $suiteName), 'output');
-        $choices = array_values(array_merge(array('None'), $contextClasses));
+        $choices = array_merge(array('None'), $contextClasses);
         $default = '1';
 
         $answer = $this->askQuestion('>> ' . $message, $choices, $default);
--- a/src/Behat/Behat/Definition/Call/DefinitionCall.php
+++ b/src/Behat/Behat/Definition/Call/DefinitionCall.php
@@ -20,6 +20,8 @@ use Behat\Testwork\Environment\Environment;
  * Enhances environment call with definition information.
  *
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * @method Definition getCallee()
  */
 final class DefinitionCall extends EnvironmentCall
 {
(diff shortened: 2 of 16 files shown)


src/Behat/Behat/Context/Cli/InteractiveContextIdentifier.php
```diff
@@ -69,7 +69,7 @@ public function guessTargetContextClass(ContextEnvironment $environment)
         }
 
         $message = $this->translator->trans('snippet_context_choice', array('%count%' => $suiteName), 'output');
-        $choices = array_values(array_merge(array('None'), $contextClasses));
+        $choices = array_merge(array('None'), $contextClasses);
```
carlos-granados, line 72: The merged array is already an integers indexed list, `array_values()` does not do anything here

src/Behat/Behat/Definition/Call/DefinitionCall.php
```diff
@@ -20,6 +20,8 @@
  * Enhances environment call with definition information.
  *
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ *
```
carlos-granados, line 23: For this class, the inherited `getCallee(()` function will return a `Definition`, we can specify this here so that PHPStan is aware of this

src/Behat/Behat/HelperContainer/Call/Filter/ServicesResolver.php
```diff
@@ -152,15 +152,6 @@ private function repackageDefinitionCall(DefinitionCall $call, array $newArgumen
     {
         $definition = $call->getCallee();
 
-        if (!$definition instanceof Definition) {
```
carlos-granados, line 155: This will always be a `Definition`

src/Behat/Behat/Tester/Result/ExecutedStepResult.php
```diff
@@ -88,7 +88,7 @@ public function getException()
     }
 
     /**
-     * {@inheritdoc}
+     * @return self::PENDING | self::FAILED | self::PASSED
```
carlos-granados, line 91: We need to be more strict about the range of TestResult values that can be returned or accepted by the different functions

src/Behat/Testwork/Exception/ExceptionPresenter.php
```diff
@@ -78,16 +80,15 @@ public function setDefaultVerbosity($defaultVerbosity)
...
-     * @param integer   $verbosity
-     *
-     * @return string
      */
-    public function presentException(Exception $exception, $verbosity = null)
+    public function presentException(Throwable $exception, int $verbosity = null): string
     {
         $verbosity = $verbosity ?: $this->defaultVerbosity;
 
+        if (!$exception instanceof Exception) {
```
carlos-granados, line 88: If the Throwable is not an Exception, we need to convert it

carlos-granados:
Need to fix some code for PHP 8.4

carlos-granados:
> Need to fix some code for PHP 8.4

fixed

acoulton (approved):
Great work @carlos-granados

PR #1590, opened 2025-01-21
chore: remove Psalm

carlos-granados:
Now that we have reached PHPStan level 5 there is enough static analysis, so we can remove Psalm from our project. This PR removes this tool and changes any usages in our code


.github/workflows/build.yml
```diff
@@ -95,8 +95,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        #runs on the lowest and highest supported versions (change this to 8.4 once psalm supports it)
-        php: [8.1, 8.3]
+        php: [8.1, 8.4]
```
carlos-granados, line 98: We can now run in PHP 8.4 as we are not using Psalm which does not support this version

.github/workflows/build.yml
```diff
@@ -110,9 +109,6 @@ jobs:
       - name: Install dependencies
         run: composer update
 
-      - name: Run Psalm
```
carlos-granados, line 113: Remove Psalm from CI

composer.json
```diff
@@ -34,8 +34,7 @@
         "phpunit/phpunit": "^9.6",
         "sebastian/diff": "^4.0",
         "symfony/polyfill-php84": "^1.31",
-        "symfony/process": "^5.4 || ^6.4 || ^7.0",
-        "vimeo/psalm": "^5.0"
```
carlos-granados, line 38: Remove Psalm from our dependencies

composer.json
```diff
@@ -69,14 +68,12 @@
         "all-tests": [
             "@behat-progress",
             "@phpunit",
-            "@psalm",
```
carlos-granados, line 72: And from our test commands

src/Behat/Behat/Context/Environment/InitializedContextEnvironment.php
```diff
@@ -24,6 +24,8 @@
  * @see ContextEnvironmentHandler
  *
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * @template T of Context
```
carlos-granados, line 28: We define T globally for the whole class as we are going to use it in a couple of places

src/Behat/Behat/Context/Environment/InitializedContextEnvironment.php
```diff
@@ -35,8 +37,7 @@ final class InitializedContextEnvironment implements ContextEnvironment, Service
     private ?ContainerInterface $serviceContainer = null;
 
     /**
-     * @var array<class-string<Context>, Context>
-     * @psalm-var class-string-map<T as Context, T>
+     * @var array<class-string<T>, T>
```
carlos-granados, line 40: This is the equivalent of what we were achieving previously with the Psalm annotation: declaring that this is an array where the keys are class strings of Context classes and the values are instances of the Context class defined in the key

src/Behat/Behat/Context/Environment/InitializedContextEnvironment.php
```diff
@@ -86,7 +87,6 @@ public function hasContexts()
 
     /**
      * {@inheritdoc}
-     * @psalm-suppress InvalidReturnStatement, InvalidReturnType
```
carlos-granados, line 89: Some Psalm suppressions are not needed any longer

src/Behat/Behat/Context/Environment/InitializedContextEnvironment.php
```diff
@@ -115,8 +114,6 @@ public function getContexts()
     /**
      * Returns registered context by its class name.
      *
-     * @template T of Context
```
carlos-granados, line 118: This was declared at the class level

src/Behat/Testwork/Call/CallResult.php
```diff
@@ -75,8 +75,8 @@ public function getReturn()
     /**
      * Check if call thrown exception.
      *
-     * @psalm-assert-if-true Exception $this->exception
-     * @psalm-assert-if-true Exception $this->getException()
+     * @phpstan-assert-if-true Exception $this->exception
```
carlos-granados, line 78: We use the PHPStan equivalent annotations

src/Behat/Behat/Context/Environment/InitializedContextEnvironment.php
stof: This is not the same than before. The `class-string-map` was saying that each value of the array had to be an instance of the class used as key. The type you use here says that all keys must be class names extending the base type `T` and and values must be instances of that `T`, without any relation between keys and values.

src/Behat/Behat/Context/Environment/InitializedContextEnvironment.php
stof, line 28: making the `InitializedContextEnvironment` class generic does not make sense here

src/Behat/Behat/Context/Environment/InitializedContextEnvironment.php
stof: this is totally wrong. A class-level template or a method-level template are totally different behaviors.

The old signature was saying that anytime you pass a class string to that `getContext` class, you get an instance of the requested class.
the new signature means that you have to pass a class name being a subtype of the generic type of the environment instance (which cannot have any proper type more precise than `Context` as its generic type due to the way Behat works) and you get an instance of that generic type (i.e. an instance of `Context`), with no guarantee that it is an instance of the requested class.

PR #1591, opened 2025-01-22
Fix types for InitializedContextEnvironment

stof:
This class has no reason to become generic. And its `getContext` method must keep using a method-level generic type.

This fixes my comments from https://github.com/Behat/Behat/pull/1590
There is no errors to ignore for phpstan due to the missing support for the `class-string-map` type because of the level we use right now. This error does exists (see the `phpstan.return` error reported in https://phpstan.org/r/a8603171-7f14-42b8-90af-e83e409b2650) but it belongs to level 7.


carlos-granados (approved):
Thanks!

PR #1592, opened 2025-01-22
DX: PHP CS Fixer PSR2

carlos-granados:
I have included initial configuration to use PHP CS Fixer in our project and the changes needed to conform to PSR2 as an initial Coding Standard step. Later on we can move this up to PSR12, PER-CS and add the Symfony variant in sucesive PRs

I have split the changes in several commits: 
- The first commit adds the initial configuration, including PSR1 standard, for which we don't need any changes in the code
- The next commits introduce the rules in related groups (spacing, braces, etc...) to make them easier to review
- The last commit adds all the missing rules to reach PSR2 and adds a config to run this in CI

Once we are happy with this, I will add a blame ignore file with the commits with code changes so that these changes are ignored by git blame


.github/workflows/build.yml
```diff
@@ -112,6 +112,30 @@ jobs:
...
         run: ./vendor/bin/phpstan
 
+  coding-standard:
+    name: Coding standards
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        # PHP CS Fixer does not support PHP 8.4 yet
+        php: [8.1, 8.3]
```
stof, line 122: why would we need to check coding standards on multiple PHP versions ? The formatting of source code won't change depending on the version of PHP being used, so the 2 jobs would report exactly the same thing (and that's a good thing as we cannot change the formatting in different ways if they report different things)

acoulton (approved):
This looks great, thanks for breaking it up so clearly.

carlos-granados:
@acoulton I have added the git blame ignore file, so this is now fully ready to be merged. If you think it is OK, can you please merge it? Thanks!

acoulton (approved):
Thanks :)

PR #1594, opened 2025-01-27
feature: Add option to print unused definitions

carlos-granados:
In large codebases, step definitions can become obsolete and we can end up with a number of unused definitions and we would probably want to get rid of them in order to remove dead code. This PR adds a new `--print-unused-definitions` option to the Behat runner which will print a list of unused definitions at the end of the run.

### DIFF
--- a/i18n.php
+++ b/i18n.php
@@ -17,6 +17,7 @@
         'pending_count'           => '[1,Inf] %count% pending',
         'undefined_count'         => '[1,Inf] %count% undefined',
         'skipped_count'           => '[1,Inf] %count% skipped',
+        'unused_definitions'      => '{0} No unused definitions|{1} 1 unused definition:|]1,Inf] %count% unused definitions:',
     ),
     'bg'    => array(
         'snippet_context_choice'  => 'В сет <snippet_undefined><snippet_keyword>%count%</snippet_keyword> има недекларирани стъпки. Изберете в кой Context да бъдат създадени:</snippet_undefined>',
@@ -36,6 +37,7 @@
         'pending_count'           => '[1,Inf] %count% изчакващи',
         'undefined_count'         => '[1,Inf] %count% неопределени',
         'skipped_count'           => '[1,Inf] %count% пропуснати',
+        'unused_definitions'      => '{0} Няма неизползвани дефиниции|{1} 1 неизползвана дефиниция:|]1,Inf] %count% неизползвани дефиниции:',
     ),
     'cs'    => array(
         'snippet_proposal_title' => '<snippet_undefined><snippet_keyword>%count%</snippet_keyword> obsahuje chybné kroky. Definujte je za použití následujícího kódu:</snippet_undefined>',
@@ -53,6 +55,7 @@
         'pending_count'          => '{1} %count% čeká|{2,3,4} %count% čekají|]4,Inf] %count% čeká',
         'undefined_count'        => '{1} %count% nedefinován|{2,3,4} %count% nedefinovány|]4,Inf] %count% nedefinováno',
         'skipped_count'          => '{1} %count% přeskočen|{2,3,4} %count% přeskočeny|]4,Inf] %count% přeskočeno',
+        'unused_definitions'     => '{0} Žádné nepoužité definice|{1} 1 nepoužitá definice:|]1,Inf] %count% nepoužitých definic:',
     ),
     'de'    => array(
         'snippet_proposal_title' => '<snippet_undefined><snippet_keyword>%count%</snippet_keyword> hat fehlende Schritte. Definiere diese mit den folgenden Snippets:</snippet_undefined>',
@@ -70,6 +73,7 @@
         'pending_count'          => '[1,Inf] %count% ausstehend',
         'undefined_count'        => '[1,Inf] %count% nicht definiert',
         'skipped_count'          => '[1,Inf] %count% übersprungen',
+        'unused_definitions'     => '{0} Keine unbenutzten Definitionen|{1} 1 unbenutzte Definition:|]1,Inf] %count% unbenutzte Definitionen:',
     ),
     'es'    => array(
         'snippet_proposal_title' => '<snippet_undefined>A <snippet_keyword>%count%</snippet_keyword> le faltan pasos. Defínelos con estos pasos:</snippet_undefined>',
@@ -87,6 +91,7 @@
         'pending_count'          => '[1,Inf] %count% pendientes',
         'undefined_count'        => '[1,Inf] %count% por definir',
         'skipped_count'          => '[1,Inf] %count% saltadas',
+        'unused_definitions'     => '{0} No hay definiciones sin usar|{1} 1 definición sin usar:|]1,Inf] %count% definiciones sin usar:',
     ),
     'fr'    => array(
         'snippet_proposal_title' => '<snippet_undefined><snippet_keyword>%count%</snippet_keyword> a des étapes manquantes. Définissez-les avec les modèles suivants :</snippet_undefined>',
@@ -104,6 +109,7 @@
         'pending_count'          => '[1,Inf] %count% en attente',
         'undefined_count'        => '[1,Inf] %count% indéfinis',
         'skipped_count'          => '[1,Inf] %count% ignorés',
+        'unused_definitions'     => '{0} Aucune définition inutilisée|{1} 1 définition inutilisée:|]1,Inf] %count% définitions inutilisées:',
     ),
     'hu'    => array(
         'snippet_context_choice'  => '<snippet_undefined><snippet_keyword>%count%</snippet_keyword> sorozat meghatározatlan lépéseket tartalmaz. Válaszd ki a környezetet kódrészlet készítéséhez:</snippet_undefined>',
@@ -123,6 +129,7 @@
         'pending_count'           => '[1,Inf] %count% elintézendő',
         'undefined_count'         => '[1,Inf] %count% meghatározatlan',
         'skipped_count'           => '[1,Inf] %count% kihagyott',
+        'unused_definitions'      => '{0} Nincsenek nem használt definíciók|{1} nem használt definíció:|]1,Inf] %count% nem használt definíciók:',
     ),
     'it'    => array(
         'snippet_proposal_title' => '<snippet_undefined><snippet_keyword>%count%</snippet_keyword> ha dei passaggi mancanti. Definiscili con questi snippet:</snippet_undefined>',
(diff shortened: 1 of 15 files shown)


i18n.php
```diff
@@ -36,6 +37,7 @@
         'pending_count'           => '[1,Inf] %count% изчакващи',
         'undefined_count'         => '[1,Inf] %count% неопределени',
         'skipped_count'           => '[1,Inf] %count% пропуснати',
+        'unused_definitions'      => '{0} Няма неизползвани дефиниции|{1} 1 неизползвана дефиниция:|]1,Inf] %count% неизползвани дефиниции:',
```
carlos-granados, line 40: I used ChatGPT to generate these translations. They are OK for the languages which are familiar to me like Spanish, Italian, Portuguese or French. They look ok in other non-latin languages like German or Dutch. No idea about translations in other character sets like Russian or Japanese, hope they are OK

src/Behat/Behat/Definition/Call/RuntimeDefinition.php
```diff
@@ -29,6 +29,8 @@ abstract class RuntimeDefinition extends RuntimeCallee implements Definition
      */
     private $pattern;
 
+    private bool $used = false;
```
carlos-granados, line 32: We define a flag that we use to check if a definition has been used or not

src/Behat/Behat/Definition/Cli/UnusedDefinitionsController.php
```diff
@@ -0,0 +1,91 @@
...
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+
+/**
+ * Prints unused definitions
+ *
+ * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ */
+final class UnusedDefinitionsController implements Controller
```
carlos-granados, line 31: This is the controller that records the definitions and prints those which are unused

src/Behat/Behat/Definition/Cli/UnusedDefinitionsController.php
```diff
@@ -0,0 +1,91 @@
...
+        if ($input->getOption('print-unused-definitions')) {
+            $this->eventDispatcher->addListener(SuiteTested::AFTER, array($this, 'registerDefinitionUsages'), -999);
+            $this->eventDispatcher->addListener(ExerciseCompleted::AFTER, array($this, 'printUnusedDefinitions'), -999);
+        }
+        return null;
+    }
+
+    public function registerDefinitionUsages(AfterSuiteTested $event): void
+    {
+        $environmentDefinitions = $this->definitionRepository->getEnvironmentDefinitions($event->getEnvironment());
```
carlos-granados, line 66: This listener is called after each suite is tested. We get a list of all found definitions and save them in an array, noting if they have been used or not

src/Behat/Behat/Definition/Cli/UnusedDefinitionsController.php
```diff
@@ -0,0 +1,91 @@
...
+        foreach ($environmentDefinitions as $definition) {
+            if ($definition instanceof RuntimeDefinition) {
+                $path = $definition->getPath();
+                if (!isset($this->definitionUsage[$path])) {
+                    $this->definitionUsage[$path] = [
+                        'definition' => $definition,
+                        'used' => $definition->hasBeenUsed(),
+                    ];
+                } elseif ($definition->hasBeenUsed()) {
+                    $this->definitionUsage[$path]['used'] = true;
```
carlos-granados, line 76: If a definition is used in any suite, we count it as used

src/Behat/Behat/Definition/Cli/UnusedDefinitionsController.php
```diff
@@ -0,0 +1,91 @@
...
+                }
+            }
+        }
+    }
+
+    public function printUnusedDefinitions(): void
+    {
+        $unusedDefinitions = array_filter($this->definitionUsage, function ($definition) {
```
carlos-granados, line 84: We filter out those definitions which have not been used and print them

src/Behat/Behat/Definition/Printer/ConsoleDefinitionInformationPrinter.php
```diff
@@ -39,12 +39,37 @@ public function setSearchCriterion($criterion)
      * {@inheritdoc}
      */
     public function printDefinitions(Suite $suite, $definitions)
+    {
+        $this->printDefinitionsWithOptionalSuite($definitions, $suite);
```
carlos-granados, line 43: This printer assumed that a Suite was available. We modify it so that a suite becomes optional

src/Behat/Behat/Definition/Printer/ConsoleDefinitionInformationPrinter.php
```diff
@@ -39,12 +39,37 @@ public function setSearchCriterion($criterion)
...
+     * @param Definition[] $definitions
+     */
+    private function printDefinitionsWithOptionalSuite(array $definitions, ?Suite $suite = null): void
     {
         $search = $this->searchCriterion;
         $output = array();
 
         foreach ($definitions as $definition) {
-            $definition = $this->translateDefinition($suite, $definition);
+            if ($suite) {
```
carlos-granados, line 70: If we don't have a suite, we can't translate the definitions. This is not a problem for the unused definitions functionality, this will be used to find them in the code, so it is best to print them in the original language used in the code

src/Behat/Behat/Definition/Printer/ConsoleDefinitionInformationPrinter.php
```diff
@@ -66,18 +91,16 @@ public function printDefinitions(Suite $suite, $definitions)
...
-     * @param Definition $definition
-     *
      * @return string[]
      */
-    private function extractHeader(Suite $suite, Definition $definition)
+    private function extractHeader(?Suite $suite, Definition $definition): array
     {
         $pattern = $definition->getPattern();
         $lines = array();
+        $indent = $suite ? '{suite} <def_dimmed>|</def_dimmed> ' : '';
```
carlos-granados, line 100: If there is no suite, we don't indent the output text with the name of the suite

src/Behat/Behat/Definition/Printer/ConsoleDefinitionPrinter.php
```diff
@@ -67,8 +67,11 @@ public function __construct(
      *
      * @param string $text
      */
-    final protected function write($text)
+    final protected function write($text, bool $addEmptyLine = false)
     {
+        if ($addEmptyLine) {
```
carlos-granados, line 72: We add the possibility to print an empty line before the text so that we can separate it from the previous output

src/Behat/Behat/Definition/Printer/ConsoleDefinitionPrinter.php
```diff
@@ -101,6 +104,11 @@ final protected function translateDefinition(Suite $suite, Definition $definitio
         return $this->translator->translateDefinition($suite, $definition);
     }
 
+    final protected function translateInfoText(string $infoText, array $parameters): string
```
carlos-granados, line 107: This function is used to translate the header info that is printed before the list of unused definitions

src/Behat/Behat/Tester/Runtime/RuntimeStepTester.php
```diff
@@ -117,6 +118,11 @@ private function testDefinition(Environment $env, FeatureNode $feature, StepNode
             return new UndefinedStepResult();
         }
 
+        $definition = $search->getMatchedDefinition();
```
carlos-granados, line 121: If a definition has been found, we mark it as used even if it may be skipped, as we want to count skipped definitions as used

src/Behat/Behat/Definition/Cli/UnusedDefinitionsController.php
```diff
@@ -0,0 +1,91 @@
...
+            return $definition['used'] === false;
+        });
```
acoulton, line 86: Maybe neater as a one-liner?

```suggestion
        $unusedDefinitions = array_filter($this->definitionUsage, fn ($definition) => ! $definition['used']);
```

src/Behat/Behat/Definition/Printer/ConsoleDefinitionPrinter.php
acoulton: To me, `add` feels like it's going to append it - maybe `prependEmptyLine` or `lineBreakBefore` (which is really what it's doing - it doesn't necessarily add an *empty* line if something else has written characters without a linebreak...)?

src/Behat/Behat/Tester/Runtime/RuntimeStepTester.php
acoulton: Maybe worth having this as a code comment to remind us why it's in this sequence in future?

src/Behat/Behat/Definition/Cli/UnusedDefinitionsController.php
```diff
@@ -0,0 +1,91 @@
...
+        private EventDispatcherInterface $eventDispatcher,
+        private ConsoleDefinitionInformationPrinter $printer,
+    ) {
+    }
+
+    public function configure(Command $command): void
+    {
+        $command
+            ->addOption(
+                '--print-unused-definitions', null, InputOption::VALUE_NONE,
```
acoulton, line 49: Would it be feasible to have this as a config option instead / as well as the CLI flag so that we could enable it by default for specific profiles?

acoulton (commented):
Thanks for this, this will be a **really** useful feature :)

Just a couple of small comments.

carlos-granados:
@acoulton I added three new commits:
- In the first one I add some small changes following your suggestions
- In the second one I fix some coding standard issues found by our new linter
- In the third one I add the possibility to trigger this option using the yaml or php configurations

acoulton (approved):
Looks great, thanks :)

src/Behat/Behat/Definition/Call/RuntimeDefinition.php
```diff
@@ -68,4 +70,14 @@ public function __toString()
     {
         return $this->getType() . ' ' . $this->getPattern();
     }
+
+    public function markAsUsed(): void
```
stof, line 74: should this be tagged as `@internal` ?

TomasVotruba:
:clap:  Thanks Carlos, can't wait to try this (once we upgrade all the blocker extensions)

carlos-granados:
@TomasVotruba this has been released https://github.com/Behat/Behat/releases/tag/v3.19.0 🎉

carlos-granados:
@TomasVotruba documentation has also been updated with this option https://github.com/Behat/docs/pull/187

vever001:
Hello, I just noticed this new feature, thanks.
We needed this for a large project a while ago and decided to use https://github.com/nicwortel/behat-unused-step-definitions-extension which has some very useful extra options (include/exclude filters, and ignorePatternAliases).
It would be great to see those options included here.

carlos-granados:
Had no idea this extension existed!! 😆 I'll take a look

Issue #1106, opened 2017-11-23
Table transformation with umlauts


UFOMelkor:
I tried to make a table transformation with column names that contains german umlauts.
Sadly my transformation are not applied.

As a modified example from the features:
```php
/** @Transform table:username,äge */
```
would not be applied for this table:
```gherkin
          Given I am user:
            | username | äge |
            | ever.zet | 22  |
```
It would work if the modifier `u` is added to the [pattern regex](https://github.com/Behat/Behat/blob/ccb728f6d4dc156c38c19ce7b7a7613c175e86af/src/Behat/Behat/Transformation/Transformation/ColumnBasedTableTransformation.php#L28): `/^table\:(?:\*|[[:print:]]+)$/u` but I'm not deep enough on topic to know about possible side effects …

everzet:
Thanks for reporting!

mnofresno:
Hi! I've created [this PR](https://github.com/mnofresno/Behat/pull/1) to solve this issue, but I was unable to make the tests fail, the new added test always passing regardless the age number matches or not.

Please help me I want to start contribuiting to this project.

Thanks!

ciaranmcnulty:
IT looks like your Pr is being overlooked because it's not to this repo

mnofresno:
I know it's not to this repo, I've asked a question I need to solve before creating the definitive PR.

PR #1595, opened 2025-01-29
Allow unicode characters in table transformations

carlos-granados:
Currently if you try to use any non-ascii unicode characters in column or row names for table transformations it won't work. This PR fixes this issue by adding the `u` modifier to the regex expressions used to check the patterns, this allows unicode characters when applying the regex

Fixes https://github.com/Behat/Behat/issues/1106


acoulton (approved):
Brilliant, thank you :)

PR #1597, opened 2025-02-07
fix: add some fixes to the print unused definitions feature

carlos-granados:
Following @stof 's comments on this PR https://github.com/Behat/Behat/pull/1594 I have added some fixes:

- Mark some methods as internal
- Add an interface for the printer that prints unused definitions
- Fix the case where the definitions were translated, previously it would not have counted them as used. 
- Added some tests for this case


carlos-granados:
Added documentation for this feature in this PR https://github.com/Behat/docs/pull/187

PR #1601, opened 2025-02-21
fix(junit): Add file path in output

klausi:
Junit support for the file path is missing, adding it here.

### DIFF
--- a/src/Behat/Behat/Output/Node/Printer/JUnit/JUnitFeaturePrinter.php
+++ b/src/Behat/Behat/Output/Node/Printer/JUnit/JUnitFeaturePrinter.php
@@ -68,6 +68,7 @@ final class JUnitFeaturePrinter implements FeaturePrinter
             'failures' => $stats[TestResult::FAILED],
             'errors' => $stats[TestResult::PENDING] + $stats[TestResult::UNDEFINED],
             'time' => $this->durationListener ? $this->durationListener->getFeatureDuration($this->currentFeature) : '',
+            'file' => $this->currentFeature->getFile(),
         ));
 
         $this->statistics->reset();


klausi:
Upload stable diff file for composer patches.

[behat-file-path-1601.diff.txt](https://github.com/user-attachments/files/18905591/behat-file-path-1601.diff.txt)

src/Behat/Behat/Output/Node/Printer/JUnit/JUnitFeaturePrinter.php
```diff
@@ -68,6 +68,7 @@ public function printFooter(Formatter $formatter, TestResult $result)
             'failures' => $stats[TestResult::FAILED],
             'errors' => $stats[TestResult::PENDING] + $stats[TestResult::UNDEFINED],
             'time' => $this->durationListener ? $this->durationListener->getFeatureDuration($this->currentFeature) : '',
+            'file' => $this->currentFeature->getFile(),
```
acoulton, line 71: I think for consistency this would need to be a relative path the same as the `file` attribute on the `<testcase` node.

Currently that is calculated internal to the JUnitScenarioPrinter https://github.com/Behat/Behat/blob/610fa32937bf52bd79d812103464600ef4649e36/src/Behat/Behat/Output/Node/Printer/JUnit/JUnitScenarioPrinter.php#L86-L91 but it would probably be better to extract that rather than duplicating it...

Note also that it appears Gherkin does not guarantee the file is defined - `->getFile()` can be `null|string`. I have to say I'm not sure why/when it could be missing... The scenario printer therefore only adds a `file` attribute if there is a file to report.

acoulton (changes requested):
Thanks for the contribution @klausi 

I've checked at https://github.com/testmoapp/junitxml?tab=readme-ov-file#example (closest thing there is to a JUnit spec) and can see the `file` attribute is supported on `<testcase>` so we'd be happy to add this.

I think it does need a slightly different implementation, and the example outputs in https://github.com/Behat/Behat/blob/master/features/junit_format.feature also need to be updated with the new attribute to pass the tests.

If you have time to work on this, that would be great :)

klausi:
You are absolutely correct, thank you for the pointers!

I would probably just copy the file handling code. You know the golden rule of software engineering: duplicating simple code once is fine and avoids abstraction complexity. The third time you want to copy - then you do the refactor :-)

I won't have time to work on this right now as I'm busy with other things. Anyone feel free to pick this up!

acoulton:
You're absolutely right about early abstractions, I was just conscious that we also "relativise" paths for the pretty and progress formatters and suspected that might also be duplicated (or abstracted).

Looking at them now, I see they actually have slightly different logic. They show paths relative to the behat `%paths.base%` configuration - this is the path to the `behat.yml` / `behat.php` config file, which is usually (but not always) the same as the `cwd()` that the JunitFeaturePrinter uses.

I think it would *probably* be better if the JUnit `file` attributes were consistent with the other formatters, and guaranteed to be relative to the base directory regardless of `cwd()`? But I'm not sure if that would pose a BC issue (we don't use JUnit much)?

klausi:
Thanks for incorporating some version of the changes!

Unfortunately relative file paths in Junit XML do not work on CircleCI. We need absolute file paths so that CircleCI can split up tests for parallel runs.

CircleCI error: "Error autodetecting timing type, falling back to weighting by name. Autodetect no matching filename or classname.  If file names are used, double check paths for absolute vs relative."

I will check if I can make a new pull request to use absolute file names as it was in my original pull request.

acoulton:
@klausi I haven't double-checked but I'm pretty sure the junit should now follow the behaviour of all other formatters - it should print absolute paths if you configure it like that using the $profile->withPathOptions() method in your behat.php.

klausi:
After some more digging I found that we can alter our circleci config to also use relative paths now, so probably no change in Behat needed. Sorry for the noise!

PR #1602, opened 2025-02-24
Fix incorrect getLibPath() method in GherkinExtension.php after Gherkin package restructuring

VadymHrechukha:
This PR fixes an issue in Behat\Behat\Gherkin\ServiceContainer\GherkinExtension::getLibPath(), where the method incorrectly assumes the path to the Gherkin library using '/../../../'.

After [Behat/Gherkin#288](https://github.com/Behat/Gherkin/pull/288), the CachedArrayKeywords.php file was relocated, making the hardcoded path incorrect. This caused issues with loading Gherkin keyword data, leading to errors like:
```
PHP Fatal error:  Uncaught TypeError: Behat\Gherkin\Keywords\ArrayKeywords::__construct(): 
Argument #1 ($keywords) must be of type array, false given...
```
Changes:

- Updated getLibPath() to use '/../' instead of '/../../../' to correctly resolve the library path.

Impact:

- Fixes Behat execution errors caused by the incorrect path resolution.
- Ensures compatibility with the latest Behat/Gherkin restructuring.

PR #1603, opened 2025-02-25
Fix getLibPath()

carlos-granados:
Alternate solution to the problem described here https://github.com/Behat/Behat/pull/1602

I think this solution is more robust and future proof

### DIFF
--- a/src/Behat/Behat/Gherkin/ServiceContainer/GherkinExtension.php
+++ b/src/Behat/Behat/Gherkin/ServiceContainer/GherkinExtension.php
@@ -19,6 +19,7 @@ use Behat\Testwork\ServiceContainer\ServiceProcessor;
 use Behat\Testwork\Specification\ServiceContainer\SpecificationExtension;
 use Behat\Testwork\Suite\ServiceContainer\SuiteExtension;
 use Behat\Testwork\Translator\ServiceContainer\TranslatorExtension;
+use Composer\InstalledVersions;
 use ReflectionClass;
 use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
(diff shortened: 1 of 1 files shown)


carlos-granados:
Mmm, this does not seem to work when building the PHAR, will try to find out why

carlos-granados:
> Mmm, this does not seem to work when building the PHAR, will try to find out why

Ah, of course, you cannot use `realpath()` with files within PHARs, updated

carlos-granados:
> I think we should add a method in the Gherkin package returning the path to the i18n file instead of relying on the composer runtime API to access the install path and still making some assumption about the location of that file in the package.

mmm, you may be right. Where would you add it? In the `Behat\Gherkin\Gherkin` class?

carlos-granados:
> > I think we should add a method in the Gherkin package returning the path to the i18n file instead of relying on the composer runtime API to access the install path and still making some assumption about the location of that file in the package.
> 
> mmm, you may be right. Where would you add it? In the `Behat\Gherkin\Gherkin` class?

In fact, it would probably make more sense to add a function that returned the array of keywords instead of a file name, don't you think?

carlos-granados:
> > > I think we should add a method in the Gherkin package returning the path to the i18n file instead of relying on the composer runtime API to access the install path and still making some assumption about the location of that file in the package.
> > 
> > 
> > mmm, you may be right. Where would you add it? In the `Behat\Gherkin\Gherkin` class?
> 
> In fact, it would probably make more sense to add a function that returned the array of keywords instead of a file name, don't you think?

Just realised that the `CachedArrayKeywords` class that uses this path is located in the Gherkin package. Why would it need to be told the path by the Behat package? We should probably just calculate the path in that class without having to have it provided by Behat

stof:
`CachedArrayKeywords` takes a path as input

carlos-granados:
Proposed an alternative approach here: https://github.com/Behat/Gherkin/pull/290

carlos-granados:
Closed in favour of https://github.com/Behat/Behat/pull/1604

PR #1604, opened 2025-02-27
Remove dependency on file location in Gherkin package

carlos-granados:
Fixes the issue described in https://github.com/Behat/Behat/pull/1602 by using the new named constructor for `CachedArrayKeywords` instead of relying on the specific location of a file within the Gherkin package


stof (approved):

acoulton (approved):
Thanks @carlos-granados

PR #1605, opened 2025-02-28
feat: add new option that allows you to convert the yaml config to php

carlos-granados:
We add a new command which is invoked via the new `--convert-config` cli option. This command will load the current config for the Behat runner and, if it is in the yaml format, convert it to the php format and replace it. If you want to convert a file which is not the default config file, you can load it using the `--config` cli option and it will convert that file instead.

The configuration is loaded in the usual way before being converted which means that it is validated. In the converter we can assume that we have a valid config and this makes this code simpler as we don't need to check for the validity of the config that needs to be converted.

To convert the yaml configuration, we use the capabilities of the [PHP-Parser](https://github.com/nikic/PHP-Parser) library. From the yaml configuration which gets converted into an array, we build a number of `Config` objects and then each of these config objects has a new `toPhpExpr()` function which they use to convert their settings to an `Expr` node, using the `BuilderFactory` provided by the PHP-Parser. When doing this conversion we generate other `Config` objects which in turn add to the expression.

When we have the final expression, we use the printer provided by the PHP-Parser to convert it into PHP code that we export to a file.

If the configuration has imports, we also process each of these imports in turn and also convert them to the php format as needed.


features/convert_config.feature
```diff
@@ -0,0 +1,381 @@
...
+  As a developer
+  I need to be able to convert this configuration to the new PHP format
+
+  Background:
+    Given I set the working directory to the "ConvertConfig" fixtures folder
+    And I clear the default behat options
+    And I provide the following options for all behat invocations:
+      | option           | value |
+      | --no-colors      |       |
+      | --convert-config |       |
```
carlos-granados, line 12: In all these cases we are calling the new --convert-config option and use the --config option to specify which file to convert. Previously we copy one or more files to be converted to the temp folder and then we check the converted file

src/Behat/Config/Config.php
```diff
@@ -5,14 +5,27 @@
...
 
 use Behat\Testwork\ServiceContainer\Exception\ConfigurationLoadingException;
+use PhpParser\BuilderFactory;
+use PhpParser\Node;
+use PhpParser\Node\Name\FullyQualified;
 
 use function is_string;
 
-final class Config implements ConfigInterface
+final class Config implements ConfigInterface, ConfigConverterInterface
```
carlos-granados, line 14: Each Config object that knows how to convert itself to a php expression implements this new interface

src/Behat/Config/Config.php
```diff
@@ -5,14 +5,27 @@
...
 {
+    public const IMPORTS_SETTING = 'imports';
```
carlos-granados, line 16: I have converted several string literals into named constants to make the code more meaningful

src/Behat/Config/Config.php
```diff
@@ -38,8 +51,47 @@ public function withProfile(Profile $profile): self
         return $this;
     }
 
+    public function withPreferredProfile(string $profileName): self
```
carlos-granados, line 54: There was no function to add this specific setting programmatically, so I added it

src/Behat/Config/Config.php
```diff
@@ -38,8 +51,47 @@ public function withProfile(Profile $profile): self
...
+    }
+
     public function toArray(): array
     {
         return $this->settings;
     }
+
+    public function toPhpExpr(): Node\Expr
+    {
+        $configObject =  $this->builderFactory->new(new FullyQualified(self::class));
```
carlos-granados, line 68: Each of these toPhpExpr() work in a similar way: first we use the builder factory to build an expr that builds an instance of this particular config object (an equivalent of new Config() or similar). Then we look at the settings of this object to see if can call any functions to add any setting (for example `withProfile() or similar). Finally we create the list of parameters for this object and add them, which may include any setting which has not been set via a function

src/Behat/Config/Config.php
```diff
@@ -38,8 +51,47 @@ public function withProfile(Profile $profile): self
...
+        $expr = $configObject;
+
+        foreach ($this->settings as $settingsName => $settings) {
+            if ($settingsName === self::PREFERRED_PROFILE_NAME_SETTING) {
+                $args = $this->builderFactory->args([$settings]);
+                $expr = $this->builderFactory->methodCall($expr, self::PREFERRED_PROFILE_FUNCTION, $args);
+            } elseif ($settingsName === self::IMPORTS_SETTING) {
+                if (count($settings) === 1) {
```
carlos-granados, line 76: For imports we call the import function with a single string or an array of strings depending on the number of imports

src/Behat/Config/Config.php
```diff
@@ -38,8 +51,47 @@ public function withProfile(Profile $profile): self
...
+                    $args = $this->builderFactory->args([$settings[0]]);
+                } else {
+                    $args = $this->builderFactory->args([$settings]);
+                }
+                $expr = $this->builderFactory->methodCall($expr, self::IMPORT_FUNCTION, $args);
+            } else {
+                $profile = new Profile($settingsName, $settings ?? []);
```
carlos-granados, line 83: In many cases we will build a new Config object (in this case a Profile object) and call its toPhpExpr to build an expression which can be used on a function call on the current object (in this case a withProfile() call

src/Behat/Config/Converter/CustomPrettyPrinter.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace Behat\Config\Converter;
+
+use PhpParser\PrettyPrinter\Standard;
+use PhpParser\Node\Expr\Array_;
+use PhpParser\Node\Expr\MethodCall;
+
+class CustomPrettyPrinter extends Standard
```
carlos-granados, line 9: We override a couple of methods in the standard printer provided by the php parser library to customize the php output

src/Behat/Config/Converter/CustomPrettyPrinter.php
```diff
@@ -0,0 +1,33 @@
...
+{
+    protected function pExpr_Array(Array_ $node): string
+    {
+        $printedArray = '[';
```
carlos-granados, line 13: We make sure that all arrays are printed in multiple lines

src/Behat/Config/Converter/CustomPrettyPrinter.php
```diff
@@ -0,0 +1,33 @@
...
+        $printedArray .= $this->pCommaSeparatedMultiline($node->items, true);
+
+        return $printedArray . ($this->nl . ']');
+    }
+
+    protected function pExpr_MethodCall(MethodCall $node): string
+    {
+        $result =  $this->pDereferenceLhs($node->var);
```
carlos-granados, line 21: When we do a method call, we indent the call one indent, this makes it much easier to see to which object each method call belongs

src/Behat/Config/Converter/CustomPrettyPrinter.php
```diff
@@ -0,0 +1,33 @@
...
+        $this->indent();
+        $result .=  $this->nl . '->' . $this->pObjectProperty($node->name) . '(';
+        if (count($node->args) > 1) {
```
carlos-granados, line 24: If a method call has more than one argument we make sure that they are printed in multiple lines, this is used for lists of contexts or paths

src/Behat/Config/Converter/UseStatementAdder.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+namespace Behat\Config\Converter;
+
+use PhpParser\Node;
+use PhpParser\NodeVisitorAbstract;
+
+class UseStatementAdder extends NodeVisitorAbstract
```
carlos-granados, line 8: This class, inspired by similar functionality in Rector is used to traverse all nodes in the php code looking for objects that we define using their FQN. When they are found, we replace this with the short name and make a note of this usage. This will be later used to add a list of use statements at the top of the file

src/Behat/Config/Filter/Filter.php
```diff
@@ -0,0 +1,42 @@
...
+declare(strict_types=1);
+
+namespace Behat\Config\Filter;
+
+use Behat\Config\ConfigConverterInterface;
+use PhpParser\BuilderFactory;
+use PhpParser\Node\Expr;
+use PhpParser\Node\Name\FullyQualified;
+
+class Filter implements FilterInterface, ConfigConverterInterface
```
carlos-granados, line 12: I created a base class for all named filters which implements most of the functionality of these filters

src/Behat/Config/Filter/NameFilter.php
```diff
@@ -4,20 +4,13 @@
 
 namespace Behat\Config\Filter;
 
-final class NameFilter implements FilterInterface
+final class NameFilter extends Filter
```
carlos-granados, line 7: The named filters now just provide a name to the parent class and leave the rest of the implementation to this class

src/Behat/Config/Formatter/Formatter.php
```diff
@@ -55,8 +78,63 @@ public function toArray(): array
...
+    }
+
     public static function defaults(): array
     {
         return [];
     }
+
+    public function applyBaseSettings(Expr $expr): Expr
+    {
+        foreach ($this->settings as $settingName => $setting) {
```
carlos-granados, line 104: This function is used to apply the base options like output_path by calling the corresponding function (withOutputPath() for example)

src/Behat/Config/Formatter/Formatter.php
```diff
@@ -55,8 +78,63 @@ public function toArray(): array
...
+                $expr = $this->builderFactory->methodCall($expr, $functionName, $args);
+                unset($this->settings[$settingName]);
+            }
+        }
+        return $expr;
+    }
+
+    protected function toPhpExprForNamedFormatter(): Expr
+    {
+        $formatterObject =  $this->builderFactory->new(new FullyQualified(static::class));
```
carlos-granados, line 117: This function provides the functionality to convert to a PHP Expr for all named formatters

src/Behat/Config/Formatter/Formatter.php
```diff
@@ -55,8 +78,63 @@ public function toArray(): array
...
+
+        $expr = $this->applyBaseSettings($formatterObject);
+
+        $defaults = static::defaults();
+        $argValues = [];
+        foreach ($defaults as $name => $defaultValue) {
```
carlos-granados, line 123: We look through each of the default values for the formatter and only if the value passed does not match the default, we set the corresponding parameter in the constructor of the formatter

src/Behat/Config/Formatter/JUnitFormatter.php
```diff
@@ -4,12 +4,19 @@
...
 namespace Behat\Config\Formatter;
 
+use PhpParser\Node\Expr;
+
 final class JUnitFormatter extends Formatter
 {
     public const NAME = 'junit';
 
-    public function __construct()
+    public function __construct(...$baseOptions)
```
carlos-granados, line 13: Previously these named formatters would not accept any settings as array values. This would not allow us to pass any setting other than those defined in the constructors unless we used functions. In all other cases using functions is optional, so I added the capability for these formatters to accept further options. This is also needed to be able to build them when we are building them to generate the PHP expr

src/Behat/Config/Profile.php
```diff
@@ -79,4 +110,125 @@ public function toArray(): array
     {
         return $this->settings;
     }
+
+    public function toPhpExpr(): Node\Expr
+    {
+        $profileObject =  $this->builderFactory->new(new FullyQualified(self::class));
+        $expr = $profileObject;
+
+        $this->addFormattersToExpr($expr);
```
carlos-granados, line 119: We process each of the possible settings for a profile

src/Behat/Config/Profile.php
```diff
@@ -79,4 +110,125 @@ public function toArray(): array
...
+        $profileObject->args = $args;
+
+        return $expr;
+    }
+
+    private function addFormattersToExpr(Expr &$expr): void
+    {
+        if (isset($this->settings[self::FORMATTERS_SETTING])) {
+            foreach ($this->settings[self::FORMATTERS_SETTING] as $name => $formatterSettings) {
+                if ($formatterSettings === false) {
```
carlos-granados, line 139: If a formatter is set to false, we use the disableFormatter function

src/Behat/Config/Profile.php
```diff
@@ -79,4 +110,125 @@ public function toArray(): array
...
+                    $args = $this->builderFactory->args([$name]);
+                    $expr = $this->builderFactory->methodCall($expr, self::DISABLE_FORMATTER_FUNCTION, $args);
+                } else {
+                    if ($formatterSettings === null) {
```
carlos-granados, line 143: If it is set to null (for example with pretty: ~ we consider it the same as true

src/Behat/Config/Profile.php
```diff
@@ -79,4 +110,125 @@ public function toArray(): array
...
+                        $formatterSettings = true;
+                    }
+                    if (isset($formatterSettings[ShowOutputOption::OPTION_NAME])) {
+                        $formatterSettings[ShowOutputOption::PARAMETER_NAME] =
+                            ShowOutputOption::from($formatterSettings[ShowOutputOption::OPTION_NAME]);
+                        unset($formatterSettings[ShowOutputOption::OPTION_NAME]);
+                    }
+
+                    $formatter = match($name) {
+                        PrettyFormatter::NAME => $formatterSettings === true ? new PrettyFormatter() : new PrettyFormatter(...$formatterSettings),
```
carlos-granados, line 153: We create a new formatter object with all passed options, including base options like output_path

src/Behat/Config/Profile.php
```diff
@@ -79,4 +110,125 @@ public function toArray(): array
...
+                        ProgressFormatter::NAME => $formatterSettings === true ? new ProgressFormatter() : new ProgressFormatter(...$formatterSettings),
+                        JUnitFormatter::NAME => new JUnitFormatter(),
+                        default => $formatterSettings === true ? new Formatter($name) : new Formatter($name, $formatterSettings),
+                    };
+                    $formatterExpr = $formatter->toPhpExpr();
+                    $args = $this->builderFactory->args([$formatterExpr]);
```
carlos-granados, line 159: And then generate the appropriate Expr to generate it

src/Behat/Testwork/Cli/Application.php
```diff
@@ -146,6 +147,7 @@ protected function getDefaultCommands(): array
 
         $commands[] = new DumpReferenceCommand($this->extensionManager);
         $commands[] = new DebugCommand($this, $this->configurationLoader, $this->extensionManager);
+        $commands[] = new ConvertConfigCommand($this->configurationLoader);
```
carlos-granados, line 150: We add a new command that can do this conversion

src/Behat/Testwork/Cli/Application.php
```diff
@@ -236,6 +238,10 @@ protected function getCommandName(InputInterface $input): string
             return 'debug';
         }
 
+        if ($input->hasParameterOption(array('--convert-config'))) {
```
carlos-granados, line 241: If the --convert-config option was passed, we use this new command

src/Behat/Testwork/Cli/ConvertConfigCommand.php
```diff
@@ -0,0 +1,111 @@
...
+
+        $this->output->writeln(['', 'Starting conversion']);
+        $this->convertFile($configPath);
+        $this->output->writeln(['Conversion finished', '']);
+        return 0;
+    }
+
+    private function convertFile(string $filePath): string
+    {
+        $this->output->writeln('Converting configuration file: ' . $filePath);
```
carlos-granados, line 53: This function is used both to convert the main file and any imported files

src/Behat/Testwork/Cli/ConvertConfigCommand.php
```diff
@@ -0,0 +1,111 @@
...
+
+        $yamlConfig = (array) Yaml::parse(file_get_contents($filePath));
```
carlos-granados, line 55: We load the yaml file and first of all convert any imported file by calling this function recursively

src/Behat/Testwork/Cli/ConvertConfigCommand.php
```diff
@@ -0,0 +1,111 @@
...
+
+            $newImports = [];
+            foreach ($imports as $import) {
+                $import = $basePath . $import;
+                $newImports[] = substr($this->convertFile($import), strlen($basePath));
+            }
+            $yamlConfig[Config::IMPORTS_SETTING] = $newImports;
+        }
+
+        $config = new Config($yamlConfig);
```
carlos-granados, line 69: Once this is done we create a new Config object and get the expr that builds it (it will include internal calls to other expressions)

src/Behat/Testwork/Cli/ConvertConfigCommand.php
```diff
@@ -0,0 +1,111 @@
...
+
+        $factory = new BuilderFactory();
+
+        $expr = $config->toPhpExpr();
+
+        $configStms = [new Return_($expr)];
```
carlos-granados, line 75: The final code will be a return statement with the expression that creates the Config

src/Behat/Testwork/Cli/ConvertConfigCommand.php
```diff
@@ -0,0 +1,111 @@
...
+
+        $traverser = new NodeTraverser();
```
carlos-granados, line 77: We traverse all the created code to look for Fully Qualified Names that we want to convert into short names

src/Behat/Testwork/Cli/ConvertConfigCommand.php
```diff
@@ -0,0 +1,111 @@
...
+        $useStatementAdder = new UseStatementAdder();
+        $traverser->addVisitor($useStatementAdder);
+        $configStms = $traverser->traverse($configStms);
+
+        $useStmts = [];
+
+        $useStatements = $useStatementAdder->getUseStatements();
```
carlos-granados, line 84: Then we add all the use statements needed for all the FQNs that we have replaced

src/Behat/Testwork/Cli/ConvertConfigCommand.php
```diff
@@ -0,0 +1,111 @@
...
+        sort($useStatements);
+        foreach ($useStatements as $useStatement) {
+            $useStmts[] = $factory->use($useStatement)->getNode();
+        }
+        if (count($useStmts) !== 0) {
```
carlos-granados, line 89: If we added any use statements we need to add a blank like after them

src/Behat/Testwork/Cli/ConvertConfigCommand.php
```diff
@@ -0,0 +1,111 @@
...
+            $useStmts[] = new Nop();
+        }
+
+        $stmts = array_merge($useStmts, $configStms);
+        $stmts[] = new Nop();
```
carlos-granados, line 94: We add a blank like at the end of the file

src/Behat/Testwork/Cli/ConvertConfigCommand.php
```diff
@@ -0,0 +1,111 @@
...
+
+        // Find the name of the output file
+        $outputFileName = $filePath;
```
carlos-granados, line 97: we use some regular expressions to find the name of the file that will replace the config file

src/Behat/Testwork/Cli/ConvertConfigCommand.php
```diff
@@ -0,0 +1,111 @@
...
+        // First, handle cases where `.dist` is part of the extension
+        $outputFileName = preg_replace('/(\.ya?ml)\.dist$|\.dist\.(ya?ml)$/', '.dist.php', $outputFileName);
+        // Then, handle regular `.yaml` or `.yml` files
+        $outputFileName = preg_replace('/\.(ya?ml)$/', '.php', $outputFileName);
+
+        $printer = new CustomPrettyPrinter();
```
carlos-granados, line 103: Then we use our custom printer (derived from the PHP-parser printer) to output all the generated statements as PHP code in the output file

src/Behat/Testwork/Cli/ConvertConfigCommand.php
```diff
@@ -0,0 +1,111 @@
...
+
+        file_put_contents($outputFileName, $printer->prettyPrintFile($stmts));
+
+        unlink($filePath);
```
carlos-granados, line 107: And we remove the old config file

tests/Fixtures/ConvertConfig/custom_extension.php
```diff
@@ -0,0 +1,81 @@
...
+use Behat\Testwork\Output\Printer\OutputPrinter;
+use Behat\Testwork\Output\Printer\StreamOutputPrinter;
+use Behat\Testwork\ServiceContainer\Extension;
+use Behat\Testwork\Output\ServiceContainer\OutputExtension;
+use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Reference;
+
+class CustomExtension implements Extension
```
carlos-granados, line 14: This file is used in some tests. When we define an extension, we need to have a real extension that can be loaded. Also, if we define a custom formatter, this formatter needs to exist

tests/Fixtures/ConvertConfig/extensions.yaml
```diff
@@ -0,0 +1,8 @@
+default:
```
carlos-granados, line 1: We have a number of yaml files that we use to test all the different cases that we want to test

tests/Fixtures/ConvertConfig/features/one.feature
```diff
@@ -0,0 +1 @@
+Feature: test
```
carlos-granados, line 1: When testing the paths settings, these need to point to existing files, so I added a couple of test feature files to be able to test it

tests/Fixtures/ConvertConfig/full_configuration.yaml
```diff
@@ -0,0 +1,36 @@
+imports:
```
carlos-granados, line 1: This is the most complete configuration which is used to test that all config options work together correctly

src/Behat/Config/ConfigConverterInterface.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Behat\Config;
+
+use PhpParser\Node\Expr;
+
+interface ConfigConverterInterface
```
stof, line 9: Should we tag this interface and all the implementation methods in our config objects as `@internal` ? I'm not sure we want to make the config converter part of our public API for extensions

carlos-granados: Mmm, I would probably mark as internal the methods but maybe not the interface itself as maybe some extension may want to add their own config objects which can also be converted?? What do you think?

stof: good idea

stof: would this actually work though ? How would the yaml config be turned into the extension config object ?

carlos-granados: I'd probably need to add some new code for this, let me give this some thought

carlos-granados: OK, I have implemented something that will allow us to do this. I have added a new ConfigurableExtensionInterface and if an extension implements it, it should be able to return a config object when the function defined in this interface is called. Then when processing extensions in the config converter, we check if the extension implements this interface and if it does we call the function to get the config object instead of creating it ourselves.

The part that I don't like much is that in order to be able to instantiate the Extension to check if it implements this interface we need the ExtensionManager. And since the Profile class is not a service or have access to the service container I solved this issue by adding a static property which can be set with this service. Feels like a hack but could not think of a better solution, open to suggestions.

Also added the ConfigConverterInterface in a couple of places where I had forgotten to add it, added some @internal tags and fixed the visibility of one function

src/Behat/Config/Converter/UseStatementAdder.php
```diff
@@ -0,0 +1,35 @@
...
+    {
+        if ($node instanceof Node\Name\FullyQualified) {
+            $className = $node->toString();
+            $shortName = $node->getLast();
+
+            $this->uses[$shortName] = $className;
+
+            return new Node\Name($shortName);
+        }
+        return null;
```
fmatsos, line 25: Perhaps use an early return here?

```suggestion
        if (!$node instanceof Node\Name\FullyQualified) {
            return null;
        }
        
        $className = $node->toString();
        $shortName = $node->getLast();

        $this->uses[$shortName] = $className;

         return new Node\Name($shortName);
```

src/Behat/Config/Formatter/Formatter.php
```diff
@@ -59,4 +82,65 @@ public static function defaults(): array
...
+            if ($this->settings[$name] !== $defaultValue) {
+                $value = $this->settings[$name];
+                if ($name === ShowOutputOption::OPTION_NAME) {
+                    $value = ShowOutputOption::from($value);
+                    $name = ShowOutputOption::PARAMETER_NAME;
+                }
+                $argValues[$name] = $value;
+            }
+        }
```
fmatsos, line 125: It's a personal preference, but here I would also use an early return to avoid nested conditions.

```suggestion
        foreach ($defaults as $name => $defaultValue) {
            if ($this->settings[$name] === $defaultValue) {
                continue;
            }
            
            $value = $this->settings[$name];
            if ($name === ShowOutputOption::OPTION_NAME) {
                $value = ShowOutputOption::from($value);
                $name = ShowOutputOption::PARAMETER_NAME;
            }
            
            $argValues[$name] = $value;
        }
```

src/Behat/Config/Profile.php
```diff
@@ -79,4 +113,142 @@ public function toArray(): array
...
+        if (isset($this->settings[self::DEFINITIONS_SETTING][self::PRINT_UNUSED_DEFINITIONS_SETTING])) {
+            $args = $this->builderFactory->args([
+                $this->settings[self::DEFINITIONS_SETTING][self::PRINT_UNUSED_DEFINITIONS_SETTING]
+            ]);
+            $expr = $this->builderFactory->methodCall($expr, self::UNUSED_DEFINITIONS_FUNCTION, $args);
+            unset($this->settings[self::DEFINITIONS_SETTING][self::PRINT_UNUSED_DEFINITIONS_SETTING]);
+            if ($this->settings[self::DEFINITIONS_SETTING] === []) {
+                unset($this->settings[self::DEFINITIONS_SETTING]);
+            }
+        }
```
fmatsos, line 218:
```suggestion
        if (!isset($this->settings[self::DEFINITIONS_SETTING][self::PRINT_UNUSED_DEFINITIONS_SETTING])) {
            return;
        }
        
        $args = $this->builderFactory->args([
            $this->settings[self::DEFINITIONS_SETTING][self::PRINT_UNUSED_DEFINITIONS_SETTING]
        ]);
        
        $expr = $this->builderFactory->methodCall($expr, self::UNUSED_DEFINITIONS_FUNCTION, $args);
        unset($this->settings[self::DEFINITIONS_SETTING][self::PRINT_UNUSED_DEFINITIONS_SETTING]);
        
        if ($this->settings[self::DEFINITIONS_SETTING] === []) {
            unset($this->settings[self::DEFINITIONS_SETTING]);
        }
        
```

src/Behat/Config/Profile.php
```diff
@@ -79,4 +113,142 @@ public function toArray(): array
...
+    {
+        if (isset($this->settings[self::SUITES_SETTING])) {
+            foreach ($this->settings[self::SUITES_SETTING] as $name => $suiteSettings) {
+                $suiteObject = new Suite($name, $suiteSettings ?? []);
+                $args = $this->builderFactory->args([$suiteObject->toPhpExpr()]);
+                $expr = $this->builderFactory->methodCall($expr, self::SUITE_FUNCTION, $args);
+                unset($this->settings[self::SUITES_SETTING][$name]);
+            }
+            unset($this->settings[self::SUITES_SETTING]);
+        }
```
fmatsos, line 252:
```suggestion
        if (isset($this->settings[self::SUITES_SETTING])) {
            return;
        }
        
        foreach ($this->settings[self::SUITES_SETTING] as $name => $suiteSettings) {
            $suiteObject = new Suite($name, $suiteSettings ?? []);
            $args = $this->builderFactory->args([$suiteObject->toPhpExpr()]);
            $expr = $this->builderFactory->methodCall($expr, self::SUITE_FUNCTION, $args);
            unset($this->settings[self::SUITES_SETTING][$name]);
        }

        unset($this->settings[self::SUITES_SETTING]);
```

src/Behat/Config/Profile.php
```diff
@@ -79,4 +113,142 @@ public function toArray(): array
...
+                }
+
+                $args = $this->builderFactory->args([$extensionObject->toPhpExpr()]);
+                $expr = $this->builderFactory->methodCall($expr, self::EXTENSION_FUNCTION, $args);
+                unset($this->settings[self::EXTENSIONS_SETTING][$name]);
+            }
+            if ($this->settings[self::EXTENSIONS_SETTING] === []) {
+                unset($this->settings[self::EXTENSIONS_SETTING]);
+            }
+        }
```
fmatsos, line 239:
```suggestion
        if (!isset($this->settings[self::EXTENSIONS_SETTING])) {
            return;
        }
        
        foreach ($this->settings[self::EXTENSIONS_SETTING] as $name => $extensionSettings) {
            $extension = self::$extensionManager->activateExtension($name);
            if ($extension instanceof ConfigurableExtensionInterface) {
                $extensionObject = $extension->getExtensionConfigObject($name, $extensionSettings);
            } else {
                $extensionObject = new Extension($name, $extensionSettings ?? []);
            }

            $args = $this->builderFactory->args([$extensionObject->toPhpExpr()]);
            $expr = $this->builderFactory->methodCall($expr, self::EXTENSION_FUNCTION, $args);
            unset($this->settings[self::EXTENSIONS_SETTING][$name]);
        }
        
        if ($this->settings[self::EXTENSIONS_SETTING] === []) {
            unset($this->settings[self::EXTENSIONS_SETTING]);
        }
```

fmatsos (commented):
A first review pass :)

Globally, I would avoid nested conditions and/or logic in conditions (like `if (isset(...))`). I personally prefer to do early returns as much as possible and keep conditions logics simple.

src/Behat/Config/Converter/UseStatementAdder.php
carlos-granados: Updated

src/Behat/Config/Filter/Filter.php
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * @internal
+     */
+    public function toPhpExpr(): Expr
+    {
+        $filterObject =  $this->builderFactory->new(new FullyQualified(static::class));
+
+        $args = $this->builderFactory->args([$this->value]);
+        $filterObject->args = $args;
```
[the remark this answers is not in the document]
carlos-granados: Yeah, this code was initially more complex, hence the variable, but now it is not needed

src/Behat/Config/Formatter/Formatter.php
carlos-granados: Nice one, updated

src/Behat/Config/Profile.php
carlos-granados: Yup, will update all cases

carlos-granados:
@fmatsos thanks for your suggestions, I have updated the code

src/Behat/Testwork/Cli/ConvertConfigCommand.php
```diff
@@ -0,0 +1,115 @@
...
+        $outputFileName = $filePath;
+        // First, handle cases where `.dist` is part of the extension
+        $outputFileName = preg_replace('/(\.ya?ml)\.dist$|\.dist\.(ya?ml)$/', '.dist.php', $outputFileName);
+        // Then, handle regular `.yaml` or `.yml` files
+        $outputFileName = preg_replace('/\.(ya?ml)$/', '.php', $outputFileName);
+
+        $printer = new CustomPrettyPrinter();
+
+        file_put_contents($outputFileName, $printer->prettyPrintFile($stmts));
+
```
acoulton, line 110: Would it be worth / would it be hard to re-read the generated PHP config, call the `$config->toArray()` and assert that the result is identical to the values from the YAML, as a final sanity/safety check that there aren't any odd userland edge cases that we've missed and accidentally removed from their files?

acoulton (commented):
@carlos-granados thanks for all your work on this, this is going to be extremely useful :)

I have a few thoughts on some of the details, be interested to hear your opinions.

My overall sense is that it's a little hard to see easily whether every known option is converted, and therefore may also take a little work to keep up to date with future changes.

That said, if the tool doesn't migrate every single option to a method call it's probably not the end of the world - your fallback logic to include it in the general settings array should be safe and if we do 99% of the work users can always tweak the files themselves.

My other thought is that we want to avoid confusing end-users when they come to use these config objects themselves.

I've made a couple of suggestions of rough ideas to keep the code generation more concise / declarative and contained in the `toPhpExpr` with some of the detail extracted to an external class. I think that might help? It would also make it easier to eventually remove this tooling if / when we eventually drop support for the YAML files.

features/bootstrap/FeatureContext.php
```diff
@@ -617,8 +650,8 @@ private function addBehatOptions(TableNode $table): void
             $option = $row['option'];
             $value = $row['value'];
             if ($value !== '') {
-                if ($value === '{SYSTEM_TMP_DIR}') {
-                    $value = $this->tempDir;
+                if (str_contains($value, '{SYSTEM_TMP_DIR}')) {
```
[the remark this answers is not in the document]
carlos-granados: Good idea, updated

src/Behat/Config/ConfigConverterInterface.php
carlos-granados: @acoulton thinking about this, I think that in the end you are right and since to implementing this capability we need to do weird things with the code , it is best to just remove it. As you said, we can convert to the simpler array format and then the users may convert to extension config if available. So I have undone this addition and returned to the previous version of the code.

features/convert_config.feature
```diff
@@ -0,0 +1,408 @@
...
+
+  Scenario: Convert empty file
+    When I copy the "empty.yaml" file to the temp folder
+    When I run behat with the following additional options:
+      | option   | value                          |
+      | --config | {SYSTEM_TMP_DIR}/empty.yaml |
```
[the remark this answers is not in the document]
carlos-granados: Updated

src/Behat/Config/Config.php
```diff
@@ -5,14 +5,27 @@
...
+    private const IMPORT_FUNCTION = 'import';
+    private const PROFILE_FUNCTION = 'withProfile';
+    private const PREFERRED_PROFILE_FUNCTION = 'withPreferredProfile';
+
+    private BuilderFactory $builderFactory;
+
     public function __construct(
         private array $settings = []
     ) {
+        $this->builderFactory = new BuilderFactory();
```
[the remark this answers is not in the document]
carlos-granados: Updated so that the BuilderFactory is only created once and only when we want to do this conversion

src/Behat/Config/Config.php
```diff
@@ -38,8 +51,50 @@ public function withProfile(Profile $profile): self
...
+                $expr = $this->builderFactory->methodCall($expr, self::IMPORT_FUNCTION, $args);
+            } else {
+                $profile = new Profile($settingsName, $settings ?? []);
+                $args = $this->builderFactory->args([$profile->toPhpExpr()]);
+                $expr = $this->builderFactory->methodCall($expr, self::PROFILE_FUNCTION, $args);
+            }
+            unset($this->settings[$settingsName]);
+        }
+
+        if (count($this->settings) !== 0) {
```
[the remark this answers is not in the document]
carlos-granados: Yeah, you are right, removed

src/Behat/Config/Extension.php
```diff
@@ -4,12 +4,19 @@
 
 namespace Behat\Config;
 
-final class Extension implements ExtensionConfigInterface
```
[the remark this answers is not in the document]
carlos-granados: Does not apply any more since I undid these changes

src/Behat/Config/Formatter/PrettyFormatter.php
```diff
@@ -25,23 +31,34 @@ public function __construct(
...
-        parent::__construct(name: self::NAME, settings: [
-            'timer' => $timer,
-            'expand' => $expand,
-            'paths' => $paths,
-            'multiline' => $multiline,
+        $settings = [
+            self::TIMER_SETTING => $timer,
+            self::EXPAND_SETTING => $expand,
+            self::PATHS_SETTING => $paths,
+            self::MULTILINE_SETTING => $multiline,
```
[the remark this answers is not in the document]
carlos-granados: Didn't know about that `get_defined_vars()` function, great to learn about it. Updated

src/Behat/Config/Profile.php
```diff
@@ -79,4 +113,147 @@ public function toArray(): array
...
+                    $formatterSettings[ShowOutputOption::PARAMETER_NAME] =
+                        ShowOutputOption::from($formatterSettings[ShowOutputOption::OPTION_NAME]);
+                    unset($formatterSettings[ShowOutputOption::OPTION_NAME]);
+                }
+
+                $formatter = match($name) {
+                    PrettyFormatter::NAME => $formatterSettings === true ? new PrettyFormatter() : new PrettyFormatter(...$formatterSettings),
+                    ProgressFormatter::NAME => $formatterSettings === true ? new ProgressFormatter() : new ProgressFormatter(...$formatterSettings),
+                    JUnitFormatter::NAME => new JUnitFormatter(),
+                    default => $formatterSettings === true ? new Formatter($name) : new Formatter($name, $formatterSettings),
```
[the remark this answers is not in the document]
carlos-granados: Yeah, agree, let's leave it as is

src/Behat/Config/Profile.php
```diff
@@ -79,4 +113,147 @@ public function toArray(): array
...
+        $profileObject =  $this->builderFactory->new(new FullyQualified(self::class));
+        $expr = $profileObject;
+
+        $this->addFormattersToExpr($expr);
+        $this->addFiltersToExpr($expr);
+        $this->addUnusedDefinitionsToExpr($expr);
+        $this->addExtensionsToExpr($expr);
+        $this->addSuitesToExpr($expr);
+
+        if (count($this->settings) === 0) {
```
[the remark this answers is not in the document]
carlos-granados: In this case, I don't agree. I think it is OK that the internal methods work with `$this->settings`. Having to pass them around would make the code just more cumbersome

src/Behat/Config/Profile.php
```diff
@@ -79,4 +113,147 @@ public function toArray(): array
...
+                $extensionObject = $extension->getExtensionConfigObject($name, $extensionSettings);
+            } else {
+                $extensionObject = new Extension($name, $extensionSettings ?? []);
+            }
+
+            $args = $this->builderFactory->args([$extensionObject->toPhpExpr()]);
+            $expr = $this->builderFactory->methodCall($expr, self::EXTENSION_FUNCTION, $args);
+            unset($this->settings[self::EXTENSIONS_SETTING][$name]);
+        }
+        if ($this->settings[self::EXTENSIONS_SETTING] === []) {
```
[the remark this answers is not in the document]
carlos-granados: Yes, you are right, updated

src/Behat/Config/Profile.php
```diff
@@ -79,4 +113,147 @@ public function toArray(): array
...
+    private function addSuitesToExpr(Expr &$expr): void
+    {
+        if (!isset($this->settings[self::SUITES_SETTING])) {
+            return;
+        }
+        foreach ($this->settings[self::SUITES_SETTING] as $name => $suiteSettings) {
+            $suiteObject = new Suite($name, $suiteSettings ?? []);
+            $args = $this->builderFactory->args([$suiteObject->toPhpExpr()]);
+            $expr = $this->builderFactory->methodCall($expr, self::SUITE_FUNCTION, $args);
+            unset($this->settings[self::SUITES_SETTING][$name]);
```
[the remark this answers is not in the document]
carlos-granados: Updated

carlos-granados:
@acoulton I've added some changes following your suggestions. I will continue tomorrow with more changes

src/Behat/Config/Suite.php
```diff
@@ -64,4 +82,68 @@ public function toArray(): array
...
+    }
+
+    private function addPathsToExpr(Expr &$expr): void
+    {
+        if (isset($this->settings[self::PATHS_SETTING])) {
+            $args = $this->builderFactory->args($this->settings[self::PATHS_SETTING]);
+            $expr = $this->builderFactory->methodCall($expr, self::PATHS_FUNCTION, $args);
+            unset($this->settings[self::PATHS_SETTING]);
+        }
+    }
```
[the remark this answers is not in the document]
carlos-granados: Done, I think this is a really good change and the best thing is that it removes the need to be passing the BuilderFactory around everywhere

src/Behat/Testwork/Cli/ConvertConfigCommand.php
carlos-granados: I tried to implement this but found out that there are some cases where we implement default values differently, so for example if you have a yaml file with just `default:`, this will be transformed into an array `['default' => null]` when reading the yaml but for the php config we end up with `['default' => []]`. The final functionality is the same but trying to deal with these differences when checking the conversion is complex so I decided not to add this check

carlos-granados:
@acoulton this should be now ready for another review

src/Behat/Config/ConfigConverterInterface.php
acoulton: @carlos-granados I think that's a good call. In which case maybe we should after all make this interfaces and methods `@internal`?

acoulton (commented):
@carlos-granados this is looking great, thanks for spending so much time on it.

I think apart from the question about marking things as `@internal` this is all looking good to me :)

carlos-granados:
@acoulton added the `@internal` tags and squashed all commits. Once this is merged I will work to add some documentation to the docs

src/Behat/Config/ConfigConverterInterface.php
stof: if we don't convert extension config, I would also keep those internal.

acoulton (approved):
Great work, thanks @carlos-granados

PR #1606, opened 2025-03-10
Php-CS-fixer PER-CS2.0

carlos-granados:
Updates our coding standard to PER-CS2.0. A few of the setting will be overridden when we update to the Symfony standard so instead of applying some changes now and then undoing them when upgrade to the Symfony standard, I have applied the overridden rules now.

I have split the code in several commits to make it easier to review.

Again, once everyone is happy with these changes I will add a new commit with the blame ignore data


acoulton (approved):
This is `['brilliant', 'awesome', 'great']` thanks :laughing:

carlos-granados:
@acoulton I have rebased the code and added the changes to the git blame ignore file. If you feel everything is OK, can you please merge this?

PR #1608, opened 2025-03-18
Add step line in JUnit formatter

magnetik:
Hi,

From https://github.com/testmoapp/junitxml, I've seen that the `<testcase>` tag can support a line number. It can helps pinpointing a failing test quickly. 

Thanks !

### DIFF
--- a/src/Behat/Testwork/Output/Printer/JUnitOutputPrinter.php
+++ b/src/Behat/Testwork/Output/Printer/JUnitOutputPrinter.php
@@ -99,8 +99,6 @@ final class JUnitOutputPrinter extends StreamOutputPrinter
 
     /**
      * Adds attributes to the current <testcase> node.
-     *
-     * @param array $testcaseAttributes
      */
     public function addCurrentTestCaseAttributes(array $testcaseAttributes)
     {
--- a/src/Behat/Testwork/Output/Printer/JUnitOutputPrinter.php
+++ b/src/Behat/Testwork/Output/Printer/JUnitOutputPrinter.php
@@ -98,6 +98,16 @@ final class JUnitOutputPrinter extends StreamOutputPrinter
     }
 
     /**
+     * Adds attributes to the current <testcase> node.
+     *
+     * @param array $testcaseAttributes
+     */
+    public function addCurrentTestCaseAttributes(array $testcaseAttributes)
+    {
+        $this->addAttributesToNode($this->currentTestcase, $testcaseAttributes);
+    }
+
+    /**
      * Add a testcase child element.
      *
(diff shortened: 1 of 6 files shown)


acoulton:
@magnetik thanks for the contribution, I agree this could be useful. However it looks to me like the `line` attribute is expected to be [on the `testcase` element](https://github.com/testmoapp/junitxml/blob/1bcb6fb8b620906ec001a2c8e7a7c8e7c563a64b/examples/junit-complete.xml#L75-L78)?

Your code appears to be putting it on the `failure` / `error` element where I think it's not supported?

It would be great if you could fix that, and update the JUnit examples in the failing tests to include the new attribute? Thanks :)

magnetik:
oh yep my bad. I will correct it

magnetik:
Hey @acoulton. Do you have any input or things I need to change? Thanks :)

acoulton (changes requested):
@magnetik apologies that I didn't get a chance to look at this sooner. 

It looks like the junit examples in the `hook_failures_junit_(annotations|attributes).feature` files need to be updated to add the new attribute there as well.

The implementation looks good to me - in a perfect world I'd rather have kept the code alongside writing the other attributes (e.g. in JunitScenarioPrinter) but from a quick look I think that would be challenging with the existing structure of the listener / printer / output classes. So I think you've found a good solution to add this to what we have already.

> made me wonder if I should report the line only for failed steps because for valid steps, it will always return the last one

I think that's OK, it's probably better for end-users to consistently have the attribute present with "the line number of the step that ran" than to have to work out what it means / work around it only being present sometimes.

magnetik:
It was my turn to be late to the party. I fixed both files.

Thanks for your input :+1:

src/Behat/Testwork/Output/Printer/JUnitOutputPrinter.php
```diff
@@ -97,6 +97,16 @@ public function addTestcase(array $testcaseAttributes = [])
         $this->addAttributesToNode($this->currentTestcase, $testcaseAttributes);
     }
 
+    /**
+     * Adds attributes to the current <testcase> node.
+     *
+     * @param array $testcaseAttributes
```
acoulton, line 103: Just removing the redundant phpdoc parameter to meet our phpcs rules.

```suggestion
```

acoulton (approved):
@magnetik thanks very much for updating this, there's one tiny change to pass the standards check -  I'll commit that for you now (once this is merged, the CI will run immediately for any future contributions from you, so it'll be much easier to get feedback).

carlos-granados (approved):
Thanks @magnetik

PR #1611, opened 2025-03-29
feat: support providing multiple paths

### LINKED ISSUE #834: Running multiple *.feature files
Currently there is no way to run behat with multiple *.feature files in one execution. 
Use case is running suite in parallel. I want to run half my features on one server and another half on the other.

For example
```bash
bin/behat $(find ./features -name "*.feature"  | sed -n '0~2p'  | tr '\n' ' ')
```
but currently it throws RuntimeException "Too many arguments."

marmichalski:
Resolves #834.

This PR adds support for providing multiple paths to behat cli: `behat one.feature two.feature nested/features/`

There is a room for improvement to deduplicate the features, but I do not consider this a critical requirement, as this can be considered user error for now. The example of this would be: `behat features/a/one.feature features/a`


composer.json
```diff
@@ -75,7 +75,7 @@
         ],
         "behat": "LANG=C bin/behat  --rerun",
         "behat-progress": "LANG=C bin/behat --format=progress",
-        "phpstan": "phpstan",
+        "phpstan": "phpstan --memory-limit=-1",
```
marmichalski, line 78: was failing for me locally, can revert as unrelated

acoulton (commented):
@marmichalski this looks good to me, and very helpful, thanks.

I agree that it's probably not necessary to try and detect / warn about overlapping paths.

Only question is about the feature coverage - I think it would probably also be worth specifying behaviour when providing directories, and also when specifying specific scenarios e.g. `features/feature1.feature:7`? I had a quick look and I can't see existing features that cover these so it would be good to make sure we're not breaking anything.

I'd suggest restructuring from the current "everything in one .feature file" to the newer approach with a fixtures directory that @carlos-granados used in e.g. #1593  - that would make it easier to have more than 2 feature files in nested directories and reduce the verbosity of the setup in the `.feature` file itself.

Are you happy to take a look at that?

marmichalski:
> @marmichalski this looks good to me, and very helpful, thanks.
> 
> Only question is about the feature coverage - I think it would probably also be worth specifying behaviour when providing directories, and also when specifying specific scenarios e.g. `features/feature1.feature:7`? I had a quick look and I can't see existing features that cover these so it would be good to make sure we're not breaking anything.

This has been covered, also covered directories filtering.

> I'd suggest restructuring from the current "everything in one .feature file" to the newer approach with a fixtures directory that @carlos-granados used in e.g. #1593 - that would make it easier to have more than 2 feature files in nested directories and reduce the verbosity of the setup in the `.feature` file itself.
> 

I've had a look at what you are proposing here, but all of the scenarios in this file require quite the same setup - we assert that the features were executed only from properly filtered paths I am not sure if it's worth splitting them out to separate feature files? I believe this setup adds more confidence in the filtering too, eg when you provide `features/a` as path that `features/b` is not picked up

marmichalski:
> @marmichalski thanks for this - it looks good to me. The test failures are due to #1635 - we'll get that figured out and then we will be able to see this green.
> 
> > > I'd suggest restructuring from the current "everything in one .feature file" to the newer approach with a fixtures directory...
> 
> > I've had a look at what you are proposing here, but all of the scenarios in this file require quite the same setup...
> 
> ...
> I probably wasn't very clear. I was suggesting that instead of the current "Given a file named ..." in the background, you would have those as actual files in a structure like:
> This makes it easier to see the actual directory structure of the files in the fixtures directory. It also reduces the noise from the top of the feature and the "when I run behat" steps (things like the FeatureContext implementation, or the `-f pretty`, all of which are required but not directly relevant to what we're testing here).
> If you'd like to have a go at that that would be great, if not I'd be happy with this as-is and we can migrate it to the newer feature / fixtures approach in due course.

Ah sorry! Added fixtures.

The CI failure seems unrelated - the translations are different when you run the tests with latest dependencies 😢

tests/Fixtures/PathFilters/features/a/feature1.feature
```diff
@@ -0,0 +1,8 @@
+Feature: First Feature
+
+  Scenario: First Scenario
+    Given Some slow step N12
```
carlos-granados, line 4: What are these NX numbers? Do they serve any purpose? They seem weird

tests/Fixtures/PathFilters/features/bootstrap/FeatureContext.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+use Behat\Behat\Context\Context;
+use Behat\Step\Given;
+
+class FeatureContext implements Context
+{
+    #[Given('/^Some slow step N(\d+)$/')]
+    public function someSlowStepN($num)
```
carlos-granados, line 9: I understand that the name of these steps does not actually matter, but naming them slow, normal and fast may induce users to think that they have something to do with the speed of the test, when it doesn't. I would probable name them something more neutral like `firstStep`, `secondSet` or something like that

carlos-granados (commented):
Overall the functionality seems correct and well tested, I just added a couple of small comments/questions

tests/Fixtures/PathFilters/features/a/feature1.feature
marmichalski: I just copied the features you guys use:
https://github.com/search?q=repo%3ABehat%2FBehat%20N14&type=code

tests/Fixtures/PathFilters/features/bootstrap/FeatureContext.php
marmichalski: https://github.com/search?q=repo%3ABehat%2FBehat%20someSlowStepN&type=code

Happy to change it to whatever though...

tests/Fixtures/PathFilters/features/a/feature1.feature
carlos-granados: Ah, yeah 🤣 then ignore me

carlos-granados (approved):
Looks good to me as well, thanks @marmichalski

PR #1614, opened 2025-04-02
Throw an exception if a given parameter is a PyStringNode or a TableNode, but this argument is not expected in the context

### LINKED ISSUE #1609: Proposition: throw exception if a step as an unexpected TableNode
Hi,

Imagine the following step
```gherkin
        Given I send a GET request to "/v1/endpoint/1"
        Then the response status code should be 200
        And the JSON nodes should be equal to:
            | id | 1 |
```

If I forget the following line

```diff
        Given I send a GET request to "/v1/endpoint/1"
        Then the response status code should be 200
-       And the JSON nodes should be equal to:
            | id | 1 |
```

Then the step will pass, even if the id is not 1 (in fact the TableNode is completly ignored, as it is linked to the "response status code" step).

Do you think it will be a good idea to trigger an error if a step has a TableNode in the feature file, but the same step does not accept a TableNode has parameter ?

Thanks

jdeniau:
Fixes #1609 according to [@acoulton comment](https://github.com/Behat/Behat/issues/1609#issuecomment-2769851124)

### DIFF
--- a/features/arguments.feature
+++ b/features/arguments.feature
@@ -33,8 +33,18 @@ Feature: Step Arguments
               $this->input = $string;
           }
 
+          #[Given('/^an untyped pystring:$/')]
+          public function anUntypedPystring(PyStringNode $string) {
+              $this->input = $string;
+          }
+
           #[Given('/^a table:$/')]
           public function aTable(TableNode $table) {
+              $this->input = $table;
+          }
+
+          #[Given('/^an untyped table:$/')]
+          public function anUntypedTable($table) {
               $this->input = $table;
           }
 
@@ -52,6 +62,10 @@ Feature: Step Arguments
           public function iHaveNumberAndNumber($number1, $number2) {
               \PHPUnit\Framework\Assert::assertEquals(13, intval($number1));
               \PHPUnit\Framework\Assert::assertEquals(243, intval($number2));
+          }
+
+          #[Given('/^a step with no argument$/')]
+          public function aStepWithNoArgument(): void {
           }
       }
       """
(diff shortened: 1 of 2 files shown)


features/arguments.feature
```diff
@@ -33,11 +33,21 @@ Feature: Step Arguments
               $this->input = $string;
           }
 
+          #[Given('/^an untyped pystring:$/')]
+          public function anUntypedPystring(PyStringNode $string) {
```
stof, line 37: this does not like an untyped one

src/Behat/Testwork/Argument/MixedArgumentOrganiser.php
```diff
@@ -53,15 +55,40 @@ private function prepareArguments(array $parameters, array $arguments)
...
+        }
+
         return $this->reorderArguments($parameters, $arguments);
     }
 
+    /**
+     * return true if the arguments array contain a TableNode or PyStringNode
+     * @param array<mixed> $arguments
+     */
+    private function hasKnownedArgument(array $arguments): bool
```
stof, line 81: `Knowned` is not a valid English word. I think you meant `known` instead.

src/Behat/Testwork/Argument/MixedArgumentOrganiser.php
```diff
@@ -53,15 +55,40 @@ private function prepareArguments(array $parameters, array $arguments)
 
         list($named, $typehinted, $numbered) = $this->splitArguments($parameters, $arguments);
 
+        $someArgumentsAreKnown = $this->hasKnownArgument($numbered);
```
acoulton, line 58: I'm not sure about using "known" for these methods / variables. I might think that meant e.g. an argument that was "known" because it matches a variable name or typehint where really these are the opposite of that.

Cucumber tends to call tables and strings ["multiline step arguments"](https://cucumber.io/docs/gherkin/reference/#examples) and Behat also uses `multiline` to mean tables and strings for things like [the formatter option for whether to display them in full](https://github.com/Behat/Behat/blob/master/src/Behat/Config/Formatter/PrettyFormatter.php#L24).

So maybe `$wasMultilineProvided = $this->hasMultilineArgument($numbered)` and `if ($wasMultilineProvided && ! $this->hasMultilineArgument($arguments))`?

src/Behat/Testwork/Argument/MixedArgumentOrganiser.php
```diff
@@ -53,15 +55,40 @@ private function prepareArguments(array $parameters, array $arguments)
...
+
         $arguments =
             $this->prepareNamedArguments($parameters, $named) +
             $this->prepareTypehintedArguments($parameters, $typehinted) +
             $this->prepareNumberedArguments($parameters, $numbered) +
             $this->prepareDefaultArguments($parameters);
 
+        // If a parameter was a TableNode or a PystringNode, but has been throwned out of the arguments,
```
acoulton, line 66:
```suggestion
        // If a parameter was a TableNode or a PystringNode, but has been removed from the final arguments,
```

src/Behat/Testwork/Argument/MixedArgumentOrganiser.php
```diff
@@ -53,15 +55,40 @@ private function prepareArguments(array $parameters, array $arguments)
...
+            throw new \InvalidArgumentException(
+                'You have passed a TableNode or PystringNode as an argument, but it was not used in the function. This is probably an error in your feature file.'
+            );
+        }
+
         return $this->reorderArguments($parameters, $arguments);
     }
 
+    /**
+     * return true if the arguments array contain a TableNode or PyStringNode
```
acoulton, line 78: I think this comment is redundant (especially if the method is renamed)?

acoulton (changes requested):
@jdeniau this looks great, thanks, just a couple of small comments on the implementation.

One thing this highlights though is that you won't now be able to define a step to throw a PendingException with no arguments if the feature file includes a TableNode / PyString.

For example like this in the test that's failing: https://github.com/Behat/Behat/blob/edb265a32329d514e3a5ec44924646e0e02d95ee/features/pretty_format.feature#L417-L420

We can obviously fix that test (I see why it was trying to keep the context file small, but I think it'd be clearer to make the step definitions explicit anyway).

And I *think* it's also OK to enforce this on end users:

* if they're using our snippets or an IDE to generate pending steps from an existing feature then the TableNode / PyString argument would usually be included in the snippet anyway so it won't affect them.
* if they've added function themselves without arguments, it will now generate an error instead of being marked as pending - but it's unlikely they've committed / pushed that to a main branch so at worst they'll need to update some existing work-in-progress.

That seems like a reasonable trade-off for making the behaviour stricter and reducing the risk of false positives in future.

@carlos-granados @stof what do you think?

src/Behat/Testwork/Argument/MixedArgumentOrganiser.php
acoulton: Oh I know, sometimes I wish instead of an AI that completed the code based on my variable/method names, I could have one where I write the code and the AI just gives everything a good name 🤣

carlos-granados:
> 
> That seems like a reasonable trade-off for making the behaviour stricter and reducing the risk of false positives in future.
> 
> @carlos-granados @stof what do you think?

Agree that this sounds reasonable and I think this is OK to go forward as soon as @jdeniau makes these small adjustments

stof:
I also agree it is a reasonable tradeoff

acoulton:
@jdeniau I think this one is actually very close to being finished, do you have time to resolve the small naming / wording changes? If not I can pick this up. Thanks :)

jdeniau:
You are right, I will finish this in the next days 👍

PR #1615, opened 2025-04-02
Display line number of failing test in pretty mode

### LINKED ISSUE #1612: Idea: Display the exact line of the issue in "pretty" mode
When the format is "progress", we do have two informations about tests that fail:

```
--- Failed steps:

001 Scenario: We can do something # features/some_test.feature:6
      And the assertion should pass                                      # features/some_test.feature:38
```

But on "pretty" mode, we do only have the first line displayed:

```
--- Failed scenarios:

    features/some_test.feature:6
```

In pretty mode, we do have all the trace before, but we might have a lot of skipped tests after the exception, and we have to scroll all the way up to the error.

Would it be acceptable to have the same kind or return ?

Either with the same error as in "progress" mode, or something like

```
--- Failed scenarios:

    features/some_test.feature:6-38
```
(this way we can just copy paste to run the test from line 6 to 38)

Thanks

jdeniau:
Fixes #1612 

### Other possible unused option

I hesitated to use the `printStepList` used in the Progress printer, but it would have display two time the same informations.

### Implementation

I am not really fan of my implementation, because the failing scenario is not linked to the failing steps.
I did assume that, as the rest of the steps in the scenario are just skipped, there would be a match between the `failingScenariosStats` and the `failingStepsStats`, but I would have been far more confident if I could do something like `$failingStep->getScenario()` to find the failing step(s?) of a scenario.

### I18n

I do not know what are the translation process, so I did only translate in english and in french, but it will probably display something wrong in other languages (unless there is a fallback on english ?)

### DIFF
--- a/src/Behat/Behat/Output/Node/Printer/ListPrinter.php
+++ b/src/Behat/Behat/Output/Node/Printer/ListPrinter.php
@@ -73,7 +73,7 @@ final class ListPrinter
      * @param integer        $resultCode
      * @param ScenarioStat[] $scenarioStats
      */
-    public function printScenariosList(OutputPrinter $printer, $intro, $resultCode, array $scenarioStats)
+    public function printScenariosList(OutputPrinter $printer, $intro, $resultCode, array $scenarioStats, ?array $stepStats = null)
     {
         if (!count($scenarioStats)) {
             return;
@@ -83,9 +83,21 @@ final class ListPrinter
         $intro = $this->translator->trans($intro, [], 'output');
 
         $printer->writeln(sprintf('--- {+%s}%s{-%s}' . PHP_EOL, $style, $intro, $style));
-        foreach ($scenarioStats as $stat) {
+        foreach ($scenarioStats as $key => $stat) {
             $path = $this->relativizePaths((string) $stat);
-            $printer->writeln(sprintf('    {+%s}%s{-%s}', $style, $path, $style));
+
+            $stepStat = isset($stepStats[$key]) ? $stepStats[$key] : null;
+            $stepLine = $stepStat ? $this->extractLineNumber((string) $stepStat) : null;
+
+            if ($stepLine !== null) {
+                $lineNumber = $this->translator->trans('on_line_number', ['%line%' => $stepLine], 'output');
+                
+                $lineHelper = ' (' . $lineNumber . ')';
+            } else {
+                $lineHelper = '';
+            }
+
+            $printer->writeln(sprintf('    {+%s}%s%s{-%s}', $style, $path, $lineHelper, $style));
         }
 
         $printer->writeln();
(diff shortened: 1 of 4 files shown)


src/Behat/Behat/Output/Node/Printer/ListPrinter.php
```diff
@@ -83,9 +83,21 @@ public function printScenariosList(OutputPrinter $printer, $intro, $resultCode,
...
+            $stepStat = isset($stepStats[$key]) ? $stepStats[$key] : null;
+            $stepLine = $stepStat ? $this->extractLineNumber((string) $stepStat) : null;
+
+            if ($stepLine !== null) {
+                $lineNumber = $this->translator->trans('on_line_number', ['%line%' => $stepLine], 'output');
+                
+                $lineHelper = ' (' . $lineNumber . ')';
+            } else {
+                $lineHelper = '';
+            }
```
acoulton, line 98: I might structure this more like

```
$path = $this->relativizePaths((string) $stat);
$path = $this->appendFailingStepText($stat, $stepStats, $path);
$printer->writeln(sprintf('    {+%s}%s%s{-%s}', $style, $path, $lineHelper, $style));
```

Partly to reduce the diff in this method, and also I think then it probably wouldn't need as many new variables (the ->appendStepText function could just early `return $path` if there's nothing to append)?

acoulton (commented):
@jdeniau I see what you mean about the implementation and trying to match the scenario and the step. I agree it feels a bit uncomfortable just matching on the array index.

AFAICS, although `Statistics::getFailedSteps()` is typed as returning `StepStat[]` for BC reasons, in practice we only ever create / register instances of `StepStatV2`. And that has a `getScenarioPath()` which I think *should* always match the `getPath()` of `ScenarioStat`.

So maybe it's possible to search through the `$stepStats` array looking for one that is an instanceof `StepStatV2` and has a matching `->getScenarioPath()`? That is possibly cleaner / more robust than matching on the array indices?

On the translations, I believe that @carlos-granados has asked ChatGPT to generate the translations for new strings he's added recently. They seemed to be plausible, for languages we could speak / vaguely recognise 😆

carlos-granados:
> So maybe it's possible to search through the `$stepStats` array looking for one that is an instanceof `StepStatV2` and has a matching `->getScenarioPath()`? That is possibly cleaner / more robust than matching on the array indices?
> 
> On the translations, I believe that @carlos-granados has asked ChatGPT to generate the translations for new strings he's added recently. They seemed to be plausible, for languages we could speak / vaguely recognise 😆

Yes, what @acoulton proposes seems more robust, please explore that. And he is right that I asked ChatGPT for recent translations, so I think we can do the same in this case.

acoulton:
@jdeniau do you think you'll have time to get back to this? If not I'd be happy to pick it up from here - would be good to get it merged.

jdeniau:
To be honest, I don't think when I could get back to this and I don't know remember if those are easy changes.
If you want to continue, that's not a problem. If you don't, I'd tell you if I work on it back.

jdeniau:
@acoulton Everything should be OK 🎉

jdeniau:
@carlos-granados yes I saw that! Nice catch. I will fix that soon.

jdeniau:
Everything should be fine now. 

About the reported lines before the scenario, it's because we are on a Scenario Outline with examples, so the step is before the reported line, which is OK.

PR #1619, opened 2025-04-10
Fix & improve conversion of Context and Extension YAML configs

acoulton:
This fixes a bug when converting YAML configuration that contains Contexts with constructor arguments. 

It also adds features to reference Contexts and Extensions as `Context::class` rather than `'Context'` in the generated PHP.

I have combined into one PR as they are quite related, but can split the bugfix from the features if you'd prefer.

### DIFF
--- a/features/convert_config.feature
+++ b/features/convert_config.feature
@@ -403,6 +403,7 @@ Feature: Convert config
   Scenario: Full configuration
     When I copy the "full_configuration.yaml" file to the temp folder
     And I copy the "imported.yaml" file to the temp folder
+    And the "MY_SECRET_PASSWORD" environment variable is set to "sesame"
     When I run behat with the following additional options:
       | option   | value                                    |
       | --config | {SYSTEM_TMP_DIR}/full_configuration.yaml |
@@ -437,7 +438,12 @@ Feature: Convert config
               ->withPrintUnusedDefinitions(true)
               ->withExtension(new Extension('custom_extension.php'))
               ->withSuite((new Suite('my_suite'))
-                  ->withContexts('MyContext')
+                  ->addContext(
+                      'MyContext',
+                      [
+                          'password' => '%env(MY_SECRET_PASSWORD)%',
+                      ]
+                  )
                   ->withPaths('one.feature')
                   ->withFilter(new TagFilter('@run'))))
           ->withProfile((new Profile('other'))
--- a/features/convert_config.feature
+++ b/features/convert_config.feature
@@ -182,6 +182,41 @@ Feature: Convert config
       """
     And the temp "suite_contexts.yaml" file should have been removed
 
+  Scenario: Suite contexts with arguments
+    When I copy the "suite_contexts_with_args.yaml" file to the temp folder
+    When I run behat with the following additional options:
+      | option   | value                                          |
+      | --config | {SYSTEM_TMP_DIR}/suite_contexts_with_args.yaml |
+    Then the temp "suite_contexts_with_args.php" file should be like:
+      """
+      <?php
+
+      use Behat\Config\Config;
+      use Behat\Config\Profile;
+      use Behat\Config\Suite;
+
+      return (new Config())
+          ->withProfile((new Profile('default'))
+              ->withSuite((new Suite('my_suite'))
+                  ->addContext('MyContext')
+                  ->addContext(
+                      'App\AContextWithPositionalArgs',
+                      [
+                          'First Arg',
+                          'Second Arg',
+                      ]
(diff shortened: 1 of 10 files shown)


src/Behat/Testwork/ServiceContainer/ExtensionManager.php
```diff
@@ -131,11 +131,9 @@ public function debugInformation()
     /**
      * Attempts to guess full extension class from relative.
      *
-     * @param string $locator
-     *
-     * @return string
+     * @internal
      */
-    private function getFullExtensionClass($locator)
+    public static function guessFullExtensionClassName(string $locator): string
```
acoulton, line 136: This is a pure function, it needs to be public for the "convert shorthand to FQCN extension class" feature. I've kept that as a separate commit in case it needs further discussion.

tests/Fixtures/ConvertConfig/class_references_loader.php
```diff
@@ -0,0 +1,56 @@
...
+// Hence using an ugly extension file to run this code.
+class_alias(CreateClassNamesExtension::class, 'Some\Behat\Extension\ExplicitlyReferencedExtension');
+class_alias(CreateClassNamesExtension::class, 'Some\ShorthandExtension\ServiceContainer\ShorthandExtension');
+$context = new class implements Context {
+};
+class_alias($context::class, 'MyContext');
+class_alias($context::class, 'test\MyApp\Contexts\MyFirstContext');
+class_alias($context::class, 'test\MyApp\Contexts\MySecondContext');
+
+return new CreateClassNamesExtension();
```
acoulton, line 56: I think this is a bit ugly, but all the other ways I could think of doing it (given the complexity of configuring an autoloader when the YAML (and therefore `%paths.base%` are in a different directory to the other fixture files) were at least as ugly and a lot more verbose.

src/Behat/Config/Converter/UsedClassesCollector.php
```diff
@@ -14,9 +14,28 @@ class UsedClassesCollector extends NodeVisitorAbstract
...
+            // E.g. a `MyClass::class` argument
+            return new Node\Expr\ClassConstFetch(
+                $this->importClass($node->class),
+                $node->name,
+                $node->getAttributes(),
+            );
+        }
+
+        if ($node instanceof Node\Expr\ClassConstFetch) {
+            var_dump($node->class::class);
```
carlos-granados, line 31: I need to take a deeper look but in the meantime you probably want to remove this 😉

acoulton: Indeed I do 🤦‍♂️😆

acoulton: Fixed, thanks :)

features/convert_config.feature
```diff
@@ -253,6 +288,50 @@ Feature: Convert config
       """
     And the temp "extensions.yaml" file should have been removed
 
+  Scenario: Class references for known extensions and contexts
+    Given I copy the "class_references.yaml" file to the temp folder
+    When  I run behat with the following additional options:
+      | option   | value                                  |
+      | --config | {SYSTEM_TMP_DIR}/class_references.yaml |
+      | -vvv     |                                        |
```
carlos-granados, line 296: You probably need to remove this debugging option

tests/Fixtures/ConvertConfig/class_references.yaml
```diff
@@ -0,0 +1,25 @@
+default:
+    extensions:
+        # Loads the custom classes we need for this feature
+        class_references_loader.php: ~
+        # Is referenced exactly by its class name, and the class is defined, so will be imported & referenced
+        Some\Behat\Extension\ExplicitlyReferencedExtension: ~
+        # Uses the "automagic" Some\ShorthandExtension => Some\ShorthandExtension\ServiceContainer\ShorthandExtension
+        # Is not converted to a class reference
```
carlos-granados, line 8: This is actually converted to a class reference so this comment does not seem correct

carlos-granados (commented):
@acoulton this is mostly OK, great job, just a couple of tiny comments

tests/Fixtures/ConvertConfig/class_references.yaml
acoulton: Ah yes, correct - I had added like that on the first "convert class refs but not shorthand" commit and forgot to change it in the second "also convert shorthand" commit. Fixed now.

carlos-granados (approved):
All good now @acoulton 🎉

PR #1620, opened 2025-04-13
Add option to print all paths as absolute paths

carlos-granados:
Currently we have several places in our code where we handle paths to relativize them. This PR replaces them with a single service that can print paths and which is configurable. And we add a new option which can be used to tell Behat to print all paths as absolute instead of relative paths. This option is available on the command line and also in the Php and Yaml configs

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -457,6 +457,24 @@ EOL;
         );
     }
 
+    /**
+     * Checks whether last command output contains provided string including absolute paths
+     *
+     * @Then the output with absolute paths should contain:
+     *
+     * @param PyStringNode $text PyString text instance
+     */
+    public function theOutputWithAbsolutePathsShouldContain(PyStringNode $text)
+    {
+        if (str_contains($this->getOutput(false), $this->getExpectedOutput($text))) {
+            return;
+        }
+
+        throw new UnexpectedValueException(
+            $this->getOutputDiff($text, false)
+        );
+    }
+
     private function getExpectedOutput(PyStringNode $expectedText)
     {
         $text = strtr($expectedText, [
@@ -559,7 +577,7 @@ EOL;
         return $this->process->getExitCode();
     }
 
-    private function getOutput()
+    private function getOutput(bool $removeAbsolutePaths = true): string
     {
         $output = $this->process->getErrorOutput() . $this->process->getOutput();
 
@@ -569,7 +587,11 @@ EOL;
         }
 
         // Remove location of the project
-        $output = str_replace(realpath(dirname(dirname(__DIR__))) . DIRECTORY_SEPARATOR, '', $output);
+        $output = str_replace(
+            realpath(dirname(dirname(__DIR__))) . DIRECTORY_SEPARATOR,
+            $removeAbsolutePaths ? '' : 'BASE_PATH/',
+            $output
+        );
 
         // Replace wrong warning message of HHVM
         $output = str_replace('Notice: Undefined index: ', 'Notice: Undefined offset: ', $output);
@@ -620,11 +642,11 @@ EOL;
         };
     }
 
-    private function getOutputDiff(PyStringNode $expectedText): string
+    private function getOutputDiff(PyStringNode $expectedText, bool $removeaAbsolutePaths = true): string
     {
         $differ = new Differ(new DiffOnlyOutputBuilder());
 
-        return $differ->diff($this->getExpectedOutput($expectedText), $this->getOutput());
+        return $differ->diff($this->getExpectedOutput($expectedText), $this->getOutput($removeaAbsolutePaths));
     }
 
     private static function clearDirectory($path)
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -457,24 +457,6 @@ EOL;
         );
     }
 
-    /**
-     * Checks whether last command output contains provided string including absolute paths
-     *
-     * @Then the output with absolute paths should contain:
-     *
-     * @param PyStringNode $text PyString text instance
-     */
-    public function theOutputWithAbsolutePathsShouldContain(PyStringNode $text)
-    {
-        if (str_contains($this->getOutput(false), $this->getExpectedOutput($text))) {
-            return;
-        }
-
-        throw new UnexpectedValueException(
-            $this->getOutputDiff($text, false)
-        );
-    }
-
     private function getExpectedOutput(PyStringNode $expectedText)
     {
         $text = strtr($expectedText, [
@@ -511,6 +493,15 @@ EOL;
             // error stacktrace
             $text = preg_replace_callback(
                 '/#\d+ [^:]+:/',
+                function ($matches) {
+                    return str_replace('/', DIRECTORY_SEPARATOR, $matches[0]);
+                },
+                $text
+            );
+
+            // texts with absolute paths
+            $text = preg_replace_callback(
+                '/\{BASE_PATH\}[^\n "]+/',
                 function ($matches) {
                     return str_replace('/', DIRECTORY_SEPARATOR, $matches[0]);
                 },
@@ -577,7 +568,7 @@ EOL;
         return $this->process->getExitCode();
     }
 
-    private function getOutput(bool $removeAbsolutePaths = true): string
+    private function getOutput(): string
     {
         $output = $this->process->getErrorOutput() . $this->process->getOutput();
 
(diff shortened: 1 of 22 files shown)


features/bootstrap/FeatureContext.php
```diff
@@ -457,6 +457,24 @@ public function theOutputShouldContain(PyStringNode $text)
...
     }
 
+    /**
+     * Checks whether last command output contains provided string including absolute paths
+     *
+     * @Then the output with absolute paths should contain:
+     *
+     * @param PyStringNode $text PyString text instance
+     */
+    public function theOutputWithAbsolutePathsShouldContain(PyStringNode $text)
```
carlos-granados, line 467: We add a new step that allows us to check the output without filtering out the absolute paths and instead replacing them with a placeholder

features/convert_config.feature
```diff
@@ -400,6 +400,24 @@ Feature: Convert config
       """
     And the temp "formatters.yaml" file should have been removed
 
+  Scenario: path options
```
carlos-granados, line 403: We test that the yaml config is correctly converted to the PHP config for this new option

phpstan.dist.neon
```diff
@@ -13,6 +13,8 @@ parameters:
                 - src/Behat/Behat/Definition/Printer/ConsoleDefinitionPrinter.php
                 - src/Behat/Behat/EventDispatcher/Cli/StopOnFailureController.php
                 - src/Behat/Behat/Output/Node/Printer/ListPrinter.php
+                - src/Behat/Behat/Output/Node/Printer/Pretty/PrettyPathPrinter.php
```
carlos-granados, line 16: Now we are not using some constructor arguments in some classes but we can't remove them so we tell PhpStan to ignore them

src/Behat/Behat/ApplicationFactory.php
```diff
@@ -83,6 +84,7 @@ protected function getDefaultExtensions()
             new AutoloaderExtension(['' => '%paths.base%/features/bootstrap']),
             new SuiteExtension($processor),
             new OutputExtension('pretty', $this->getDefaultFormatterFactories($processor), $processor),
+            new PathOptionsExtension(),
```
carlos-granados, line 87: Add a new extension that adds the new path options

src/Behat/Behat/Output/Node/Printer/ListPrinter.php
```diff
@@ -45,24 +46,26 @@ final class ListPrinter
...
      * @param ExceptionPresenter $exceptionPresenter deprecated , will be removed in the next major version
-     * @param string $basePath
+     * @param string $basePath deprecated , will be removed in the next major version
      */
     public function __construct(
         ResultToStringConverter $resultConverter,
         ExceptionPresenter $exceptionPresenter,
         TranslatorInterface $translator,
-        $basePath,
+        string $basePath,
```
carlos-granados, line 60: This parameter is not used any more because the configurable path printer service takes care of handling path printing

src/Behat/Behat/Output/Node/Printer/ListPrinter.php
```diff
@@ -45,24 +46,26 @@ final class ListPrinter
...
     ) {
         $this->resultConverter = $resultConverter;
         $this->translator = $translator;
-        $this->basePath = $basePath;
+    }
+
+    public function setConfigurablePathPrinter(ConfigurablePathPrinter $configurablePathPrinter)
```
carlos-granados, line 66: We add a function to be able to set the new configurable path printer service as we cannot inject it through the constructor to keep BC

src/Behat/Behat/Output/Node/Printer/ListPrinter.php
```diff
@@ -84,7 +87,7 @@ public function printScenariosList(OutputPrinter $printer, $intro, $resultCode,
 
         $printer->writeln(sprintf('--- {+%s}%s{-%s}' . PHP_EOL, $style, $intro, $style));
         foreach ($scenarioStats as $stat) {
-            $path = $this->relativizePaths((string) $stat);
+            $path = $this->configurablePathPrinter->processPathsInText((string) $stat);
```
carlos-granados, line 90: All paths are now printed using the new service

src/Behat/Behat/Output/Node/Printer/ListPrinter.php
```diff
@@ -281,35 +284,19 @@ private function printStepStat(
         $printer->writeln();
     }
 
-    /**
```
carlos-granados, line 284: And now this function is not needed

src/Behat/Behat/Output/ServiceContainer/Formatter/PrettyFormatterFactory.php
```diff
@@ -220,6 +220,7 @@ protected function loadCorePrinters(ContainerBuilder $container)
             new Reference('output.node.printer.pretty.width_calculator'),
             '%paths.base%',
         ]);
+        $definition->addMethodCall(('setConfigurablePathPrinter'), [new Reference('configurable.path.printer')]);
```
carlos-granados, line 223: We inject the new service into the classes that need it

src/Behat/Config/Profile.php
```diff
@@ -102,6 +106,13 @@ public function withPrintUnusedDefinitions(bool $printUnusedDefinitions = true):
         return $this;
     }
 
+    public function withPathOptions(bool $printAbsolutePaths): self
```
carlos-granados, line 109: New config option to set this option using PHP

src/Behat/Config/Profile.php
```diff
@@ -215,6 +227,22 @@ private function addUnusedDefinitionsToExpr(Expr &$expr): void
         }
     }
 
+    private function addPathOptionsToExpr(Expr &$expr): void
```
carlos-granados, line 230: And new yaml to php converter for this option

src/Behat/Testwork/PathOptions/Printer/ConfigurablePathPrinter.php
```diff
@@ -0,0 +1,46 @@
...
+ * This file is part of the Behat Testwork.
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Behat\Testwork\PathOptions\Printer;
+
+final class ConfigurablePathPrinter
```
carlos-granados, line 13: This is the new service that prints paths, adding an option to decide if they should be printed as relative (default) or absolute

acoulton (commented):
This will be great @carlos-granados, thanks. Just a couple of thoughts.

acoulton:
I think this will also help with #1601

features/bootstrap/FeatureContext.php
```diff
@@ -569,7 +587,11 @@ private function getOutput()
         }
 
         // Remove location of the project
-        $output = str_replace(realpath(dirname(dirname(__DIR__))) . DIRECTORY_SEPARATOR, '', $output);
+        $output = str_replace(
+            realpath(dirname(dirname(__DIR__))) . DIRECTORY_SEPARATOR,
+            $removeAbsolutePaths ? '' : 'BASE_PATH/',
+            $output
+        );
```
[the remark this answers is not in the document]
carlos-granados: Yes, we can always replace with BASE_PATH, there was one place were I've had to be explicit about this. I also updated the placeholder to something that works on Windows

features/print_absolute_paths.feature
```diff
@@ -0,0 +1,58 @@
...
+      """
+        Scenario:                                    # BASE_PATH/tests/Fixtures/PrintAbsolutePaths/features/test.feature:3
+          Given I have a passing step                # FeatureContext::iHaveAPassingStep()
+          And I have a step that throws an exception # FeatureContext::iHaveAFailingStep()
+            Warning: Undefined variable $b in BASE_PATH/tests/Fixtures/PrintAbsolutePaths/features/bootstrap/FeatureContext.php line 16
+
+      --- Failed scenarios:
+
+          BASE_PATH/tests/Fixtures/PrintAbsolutePaths/features/test.feature:3
+      """
```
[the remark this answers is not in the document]
carlos-granados: I started working on this feature before I finished the config converter and this is why this was here. Now that we have the config converter I agree that we can remove it.

carlos-granados:
@acoulton regarding your comment about the constructors, agree that your suggestion might be a better idea, updated. I also added a constant for the id of the configurable path printer service

acoulton (commented):
Great, thanks @carlos-granados - just one tiny thing I spotted in the diff.

acoulton (approved):

PR #1622, opened 2025-04-15
bug: Keep options for JUnit formatter when converting config

acoulton:
Although the JUnit formatter has no options of its own, it does support the standard (output_path, etc) options. 

These were being lost when converting config from YAML to PHP.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "behat/transliterator": "^1.5",
         "composer-runtime-api": "^2.2",
         "composer/xdebug-handler": "^3.0",
-        "nikic/php-parser": "^5.0",
+        "nikic/php-parser": "^5.2",
         "psr/container": "^1.0 || ^2.0",
         "symfony/config": "^5.4 || ^6.4 || ^7.0",
         "symfony/console": "^5.4 || ^6.4 || ^7.0",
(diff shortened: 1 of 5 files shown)


tests/Fixtures/ConvertConfig/formatters.yaml
```diff
@@ -10,3 +10,14 @@ default:
             other_property: value
     extensions:
         custom_extension.php: ~
+
+with_options:
+    formatters:
+        junit:
+            output_path: build/logs/junit
+        progress:
```
carlos-granados, line 18: Maybe also add one of the base options (`output_path` for example) to the progress and pretty formatters to confirm that they are preserved?

acoulton: Good idea - a couple were covered already in the other profile but worth having more coverage.

composer.json
```diff
@@ -20,7 +20,7 @@
         "behat/transliterator": "^1.5",
         "composer-runtime-api": "^2.2",
         "composer/xdebug-handler": "^3.0",
-        "nikic/php-parser": "^5.0",
+        "nikic/php-parser": "^5.2",
```
carlos-granados, line 23: Why do we need to update the required PHP parser version? 🤔

carlos-granados: Perfect, thanks!

carlos-granados (approved):
Thanks @acoulton

PR #1623, opened 2025-04-15
feature: Convert `output_verbosity` values to constants in PHP config

acoulton:
The `->withOutputVerbosity` formatter config method in PHP is expected to take an `OutputVerbosity::VERBOSITY_*` constant value (as per the phpdoc for that method).

If the user's YAML config contains one of the valid values as an integer we can convert it to a constant reference. This will make the PHP file clearer without e.g. needing a separate comment as might have been present in the YAML.

If we don't recognise the value (it may be invalid, or e.g. an env var) then we just convert it as the original scalar value.


stof (approved):

carlos-granados (approved):
Great addition @acoulton

Issue #1624, opened 2025-04-15
Support PHP config for "tester" options


acoulton:
We haven't yet exposed an explicit PHP interface for configuring the `strict` or `stop-on-failure` runtime behaviour. I started looking at this and realised there's a couple of others:

```yaml
default:
  testers:
    # "Sets the strict mode for result interpretation" (which basically means fail on pending or undefined)
    strict: true
    stop_on_failure: true
    # "Tells tester to skip all tests" - tbh I'm not certain why you'd want this, and it doesn't appear to have feature coverage. 
    # Maybe for a profile that always runs equivalent to `--dry-run`?
    skip: true
  calls:
     # This is in a separate section of the configuration, but feels like it might be logically related to configuring
     # how the testers behave e.g. whether they pass or fail depending on particular conditions.
     # Note this is the only supported config key under `calls` at the moment, we might in future want to
     # add separate settings for collecting / handling deprecations which need to be configured alongside
     # error_reporting but don't really feel like they belong under something called `calls`.
    error_reporting: 32757
```

Arguably these are all behaviours of Behat for a specific `Profile` so they could be direct setters on the `Profile` config object. However that class is already getting quite heavy, and we probably don't want to add a lot more methods to that.

I think we should probably group these together into a new config object, either using `withXXX(true|false)` for each option:

```php
<?php
use Behat\Config\Config;
use Behat\Config\Profile;
use Behat\Config\TesterOptions;

return (new Config())
    ->withProfile((new Profile('default'))
        ->withTesterOptions((new TesterOptions)
            ->withStrictTesters(true)
            ->withStopOnFailure(false)
            ->withSkipAllTests(true)
            ->withErrorReporting(32757)
        )
    );
```

Or possibly with named methods for the on/off boolean states - I find these a little clearer to read, but they are obviously then harder if you wanted to set them based on some variable:

```php
<?php
use Behat\Config\Config;
use Behat\Config\Profile;
use Behat\Config\TesterOptions;

return (new Config())
    ->withProfile((new Profile('default'))
        ->withTesterOptions((new TesterOptions)
            ->failOnPendingOrUndefined() # or passOnPendingOrUndefined()
            ->stopOnFailure() # or continueOnFailure()
            ->skipAllTests() # I'm not convinced this one needs a negation
            ->withErrorReporting(32757)
        )
    );
```

I'm not sure how best to name this group of settings. 

[...]

PR #1625, opened 2025-04-16
test: Refactor features for configurable tester / error_reporting options

acoulton:
Ahead of implementing #1624 migrate the feature coverage to the new `fixture-based` approach so that it's easier to migrate / use the new PHP config for these features.

Also adds coverage for our current handling of `E_DEPRECATED` during steps, and fixes the fact that our error handler did not have a text description for the `Deprecated` error level.


features/config_strict_testers.feature
```diff
@@ -0,0 +1,60 @@
+Feature: Strict result interpretation defined in configuration file
```
acoulton, line 1: Renamed the feature file as `config_strict` was slightly ambiguous about *what* is strict.

features/error_reporting.feature
```diff
@@ -1,146 +1,99 @@
...
   Scenario: With very verbose error reporting
-    When I run "behat -f progress --no-colors -vv features/exception_in_scenario.feature"
-    Then it should fail
-    And the output should contain:
+    When I run behat with the following additional options:
+      | option                                 | value   |
+      | --profile                              | default |
+      | -vv                                    |         |
+      | features/exception_in_scenario.feature |         |
+    Then it should fail with:
```
acoulton, line 71: This scenario isn't actually covering "error_reporting" in the PHP sense at all - it's to do with how we present Exceptions during the steps at different levels of verbosity. It might be better to move it to a separate feature but I haven't done so yet.

features/error_reporting.feature
```diff
@@ -1,146 +1,99 @@
...
   Scenario: With debug verbose error reporting
-    When I run "behat -f progress --no-colors -vvv features/exception_in_scenario.feature"
-    Then it should fail
-    And the output should contain:
+    When I run behat with the following additional options:
+      | option                                 | value   |
+      | --profile                              | default |
+      | -vvv                                   |         |
+      | features/exception_in_scenario.feature |         |
+    Then it should fail with:
```
acoulton, line 90: Likewise this scenario relates to exception handling, not error_reporting.

features/stop_on_failure.feature
```diff
@@ -1,212 +1,124 @@
-Feature: Stop on failure
+Feature: Stop on failure via config
   In order to stop further execution of steps when first step fails
   As a feature developer
-  I need to have a --stop-on-failure option
+  I need to have a stop-on-failure option on the CLI or in config
```
acoulton, line 4: I have combined the previous `stop_on_failure` and `stop_on_failure_config` features. There was a lot of overlap between them, and some things that weren't actually covered that well.

It feels like "stop on failure" is a single feature, and whether it's controlled by config or a CLI flag (and what happens if both are present) are details within that, so I think better we group the coverage together.

tests/Fixtures/ErrorReporting/behat.yaml
```diff
@@ -0,0 +1,9 @@
+default: ~
+
+ignore-all-but-error:
+    calls:
+        error_reporting: 24565
```
acoulton, line 5: We can make these magic numbers much clearer once we have PHP config

carlos-granados (approved):
Thanks @acoulton

PR #1626, opened 2025-04-17
Support (and convert) PHP config for tester options

### LINKED ISSUE #1624: Support PHP config for "tester" options

acoulton:
This branch is based on #1625 so that should be reviewed and merged first.

Adds support for configuring the tester and error_reporting options through PHP, and for converting those configs from YAML.

This includes automatically removing default arguments when generating PHP config methods from YAML so that you get e.g. `->withStopOnFailure()` rather than `->withStopOnFailure(true)`. That change also applies to a couple of existing methods - as shown in the updated convert_config feature file - and as a byproduct means that the config conversion process is now stricter internally about making sure that called methods exist and have been provided with the correct argument list.

Fixes #1624

### DIFF
--- a/features/convert_config.feature
+++ b/features/convert_config.feature
@@ -451,7 +451,7 @@ Feature: Convert config
           ->withProfile(new Profile('default'))
           ->withProfile((new Profile('ignore-errors'))
               ->withTesterOptions((new TesterOptions())
-                  ->withErrorReporting(22527)))
+                  ->withErrorReporting(E_ALL & ~E_DEPRECATED)))
           ->withProfile((new Profile('not-strict'))
               ->withTesterOptions((new TesterOptions())
                   ->withStrictResultInterpretation(false)))
@@ -460,7 +460,7 @@ Feature: Convert config
                   ->withStrictResultInterpretation()
                   ->withStopOnFailure(false)
                   ->withSkipAllTests()
-                  ->withErrorReporting(24565)));
+                  ->withErrorReporting(E_ALL & ~(E_WARNING | E_NOTICE | E_DEPRECATED))));
       """
     And the temp "tester_options.yaml" file should have been removed
 
@@ -518,7 +518,7 @@ Feature: Convert config
           ->withProfile((new Profile('other'))
               ->disableFormatter('pretty')
               ->withTesterOptions((new TesterOptions())
-                  ->withErrorReporting(1)))
+                  ->withErrorReporting(E_ERROR)))
           ->withPreferredProfile('other');
       """
     And the temp "imported.php" file should be like:
--- a/features/convert_config.feature
+++ b/features/convert_config.feature
@@ -366,7 +366,7 @@ Feature: Convert config
 
       return (new Config())
           ->withProfile((new Profile('default'))
-              ->withPrintUnusedDefinitions(true));
+              ->withPrintUnusedDefinitions());
       """
     And the temp "unused_definitions.yaml" file should have been removed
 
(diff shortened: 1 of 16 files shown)


carlos-granados (approved):
@acoulton outstanding job, love what you have done in the `removeDefaultArguments()` and the `errorReportingToConstants` functions. 👍

acoulton:
@carlos-granados thanks - they were interesting problems to get stuck into :-)

It was really satisfying migrating the old features to your fixtures style, that works very well.

Do you mind approving #1625 as well and then I can get that merged first? Thanks!

Issue #1632, opened 2025-04-25
Remove the dependency on behat/transliterator


stof:
the licensing situation of `behat/transliterator` is a huge mess (see https://github.com/Behat/Transliterator/issues/3).

Behat does not rely on the full feature set of `behat/transliterator`. It only uses it as part of its logic generating a method name for a step definition based on the text of the step being used but undefined (for the generated snippet).

I suggest refactoring Behat to remove that dependency:

- for specs written in English, the simple implementation removing the transliteration will work fine. We only need the logic generating a camelCase name, which is mostly (maybe even entirely) outside the Transliterator usage in the existing logic.
- for projects writing specs in other languages that might require transliterations, we can use an **optional** dependency on ext-intl to use its [`Transliterator`](https://www.php.net/manual/en/class.transliterator.php). Such projects are very likely to already rely on ext-intl anyway.

stof:
@acoulton what do you think about that ?

acoulton:
Tbh my first thought yesterday was to just do this and abandon the behat/transliterator package. 

Arguably, there's no functional requirement to transliterate the method names at all - although it's unusual to have non-ascii characters in a method name it's valid (and some people with non-latin-languages might actually prefer that...).

And if the user wants them transliterated they can of course modify before committing - they're going to have to implement the method body anyway, so it's not a huge job to rename it, given it's usually not called by anything directly.

Then I noticed on packagist that Behat/Transliterator has quite a lot of other uses and thought potentially if we can solve the licensing it might be nice to keep it alive...

stof:
> Arguably, there's no functional requirement to transliterate the method names at all - although it's unusual to have non-ascii characters in a method name it's valid (and some people with non-latin-languages might actually prefer that...).

indeed, we could generate Unicode method names (as long as we replace characters that are not allowed in them). That would be even simpler.

acoulton:
Fixed in #1633

Issue #1343, opened 2021-05-06
Conflict of Licenses for the `behat/transliterator` dependency


ricardofiorani:
Hello there,

First I would like to say thank you for creating and maintaining Behat.

I would like to point out that after due diligence in my company, it brought up a conflict of licenses where Behat itself is under the MIT license but the `behat/transliterator`, a [dependency from behat](https://github.com/Behat/Behat/blob/master/composer.json#L20),  is under the artistic license.
The artistic license was declared a liability for us (my company actually), therefore, a problem.

Would it be possible to make `behat/transliterator` an optional dependency OR change its license to MIT as well?

ciaranmcnulty:
I'm not a lawyer, it probably depends on whether we can find a suitable alternative? If there's a package that does something similar but is maintained elsewhere it might be a good idea

stof:
The issue with this package is that it is quite a mess from a licensing PoV. It aggregates some data coming from a Perl package (which is where the Artistic license comes from) and some other helpers which were copied from doctrine 1 IIRC (or maybe other sources).
I'm not sure how to handle a license change for such a package.

ciaranmcnulty:
A quick look on packages shows a few translit packages; I guess this is mostly used for snippet generation?

I wonder if we could move it to a `suggest` and have some dumber fallback transliteration behaviour

acoulton:
This has been fixed by #1633 - we will no longer depend on/use the behat/transliterator package as of the next release.

PR #1633, opened 2025-04-25
change: Do not transliterate when suggesting method names for step definitions

acoulton:
As discussed in #1632, we should just remove our behat/transliterator dependency. We use very little of it, it is out of date, and it avoids the licensing issues discussed in Behat/Transliterator#3 and #1343.

This PR starts by refactoring the duplicated code that generates method names into a single service that implements the new `StepMethodNameGenerator` interface. This would allow people to define their own implementations if they want something more complex (or with more dependencies) than we provide out of the box.

Our default `SimpleStepMethodNameGenerator` then does the bare minimum to produce a unicode name that is syntactically valid for PHP.

This is all backwards compatible at the code level (including if people have registered a custom `PatternPolicy` class).

At a functional level, it obviously changes the method names we'll suggest - including that there are some cases where we'll now just suggest a generic name where previously we would have attempted to transliterate.


features/legacy_snippets.feature
```diff
@@ -264,9 +264,9 @@ Feature: Legacy Snippets
       Feature: Step Pattern
 
         Scenario:                         # features/coffee.feature:2
-          Then 5 should have value of £10 # FeatureContext::shouldHaveValueOfPs()
+          Then 5 should have value of £10 # FeatureContext::shouldHaveValueOf£()
             TODO: write pending definition
-          And 7 should have value of £7.2 # FeatureContext::shouldHaveValueOfPs()
+          And 7 should have value of £7.2 # FeatureContext::shouldHaveValueOf£()
```
acoulton, line 269: I actually think the new `ValueOf£` makes more sense than `ValueOfPs` anyway!

features/legacy_snippets.feature
```diff
@@ -301,7 +301,7 @@ Feature: Legacy Snippets
       --- FeatureContext has missing steps. Define them with these snippets:
 
           #[Given('I have a package v2.5')]
-          public function iHaveAPackageV(): void
+          public function iHaveAPackageV25(): void
```
acoulton, line 304: It would be trivial to strip trailing numbers as before. However I'm not sure it's warranted (bearing in mind this number is in the hardcoded step text, it is not a placeholder).

stof: this case is fine to me

src/Behat/Behat/Context/Snippet/Generator/ContextSnippetGenerator.php
```diff
@@ -122,7 +123,7 @@ public function generateSnippet(Environment $environment, StepNode $step): Snipp
      */
     private function getMethodName(string $contextClass, string $canonicalText, string $pattern): string
     {
-        $methodName = $this->deduceMethodName($canonicalText);
+        $methodName = SimpleStepMethodNameGenerator::cleanupMethodName($canonicalText);
```
acoulton, line 126: We need to keep this here for now in case anyone has a custom PatternPolicy that is not implementing our logic. We can remove it at the next major.

src/Behat/Behat/Definition/Pattern/Pattern.php
```diff
@@ -17,59 +17,34 @@
...
+        return $this->suggestedMethodName;
     }
 
-    /**
-     * Returns canonical step text.
-     *
-     * @return string
-     */
-    public function getCanonicalText()
+    public function getSuggestedMethodName(): string
```
acoulton, line 37: The unified diff looks at a glance like I've just renamed the method - in fact `getCanonicalText` is still present but deprecated, and `getSuggestedMethodName` is a new method.

src/Behat/Behat/Context/Snippet/Generator/ContextSnippetGenerator.php
```diff
@@ -122,7 +123,7 @@ public function generateSnippet(Environment $environment, StepNode $step): Snipp
      */
     private function getMethodName(string $contextClass, string $canonicalText, string $pattern): string
```
stof, line 124: The argument `canonicalText` should probably be renamed, as this is not what will be passed normally

src/Behat/Behat/Definition/Pattern/Pattern.php
```diff
@@ -17,59 +17,34 @@
...
-    /**
-     * @var int
-     */
-    private $placeholderCount;
+    public function __construct(
+        private string $suggestedMethodName,
+        private string $pattern,
+        private int $placeholderCount = 0,
+    ) {
+    }
```
acoulton, line 25: These previously had PHPDoc typehints, I would also be very surprised if anyone externally is creating a `Pattern` instance and certainly not with arguments that won't be able to cast to these types. I think therefore it's safe to make them strict.

src/Behat/Behat/Context/Snippet/Generator/ContextSnippetGenerator.php
stof, line 126: should we trigger a deprecation here when the cleaned value is different from the provided argument, to ask using the StepMethodNameGenerator in the pattern policy ?

src/Behat/Behat/Definition/ServiceContainer/DefinitionExtension.php
```diff
@@ -170,16 +173,26 @@ private function loadDefaultSearchEngines(ContainerBuilder $container)
         $container->setDefinition(self::SEARCH_ENGINE_TAG . '.repository', $definition);
     }
 
+    private function loadStepMethodNameGenerator(ContainerBuilder $container)
```
stof, line 176:
```suggestion
    private function loadStepMethodNameGenerator(ContainerBuilder $container): void
```

src/Behat/Behat/Context/Snippet/Generator/ContextSnippetGenerator.php
acoulton: The problem is we have no neat way to report that deprecation to the user at the moment, and even if we did if they are consuming a PatternPolicy from an extension that could be difficult for them to resolve in the short term.

Considering all the method actually does is lowercase the first letter and backfill empty strings, I almost considered removing it now - worst case someone will get a suggested method with an uppercase first character (or perhaps no name) and have to fix it in their Context class... It's not going to actually break anything. WDYT?

carlos-granados (commented):
Looking mostly good, just added one comment

src/Behat/Behat/Definition/Pattern/SimpleStepMethodNameGenerator.php
```diff
@@ -0,0 +1,35 @@
...
+
+class SimpleStepMethodNameGenerator implements StepMethodNameGenerator
+{
+    /**
+     * @internal
+     *
+     * @deprecated provided only for BC, to allow the ContextSnippetGenerator to continue doing this cleanup for any
+     *             steps that were generated by a third-party PatternPolicy that does not use our StepMethodNameGenerator
+     */
+    public static function cleanupMethodName(string $name): string
```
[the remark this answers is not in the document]
acoulton: Although they will all be suggested as `stepDefinition1` initially [ContextSnippetGenerator::getUniqueMethodName()](https://github.com/Behat/Behat/blob/master/src/Behat/Behat/Context/Snippet/Generator/ContextSnippetGenerator.php#L218) will then convert these to `stepDefinition2`, `stepDefinition3` etc before suggesting the actual snippets. This then also takes account of methods that are already defined in the context (rather than just from this execution).

It doesn't actually look like we have test / feature coverage of that - I could add that and a quick comment here to make the behaviour clearer / more reliable in future?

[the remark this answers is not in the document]
acoulton: @carlos-granados turns out the "numbering duplicate step definitions" behaviour is actually covered already in snippets.feature: https://github.com/Behat/Behat/blob/2ea762db840889161f4513f3e97da31e4878a251/features/snippets.feature#L104-L114

I have added comments to the new code, and also renamed from `StepMethodNameGenerator::generate()` to `StepMethodNameSuggester::suggest()` to make more obvious that these are just suggestions and may not be the final result (also makes it more consistent with $pattern->getSuggestedMethodName()`

src/Behat/Behat/Context/Snippet/Generator/ContextSnippetGenerator.php
acoulton: @carlos-granados I have reflected more on @stof's comment I think the benefit of lowercasing the first character (in the unlikely situation someone is using a custom PatternPolicy) is limited for the cost of exposing this method and changing it again at some future date. I've reworked to just handle the case where the method name is empty, which avoids having to have this method at all.

carlos-granados (approved):
Looking good, great job @acoulton

PR #1635, opened 2025-05-07
test: Update expected output to reflect new gherkin translations

acoulton:
Behat/Gherkin@4.13.0 introduced a number of new translations. Although these are BC for end-users, the new translation options appear in our `--story-syntax` output so the expectation for this needs to be updated.

### DIFF
--- a/features/syntax_help.feature
+++ b/features/syntax_help.feature
@@ -44,34 +44,34 @@ Feature: Syntax helpers
       """
       <?php class FeatureContext implements Behat\Behat\Context\Context {}
       """
-    When I run "behat --no-colors --story-syntax --lang ru"
+    When I run "behat --no-colors --story-syntax --lang el"
     Then the output should contain:
       """
-      # language: ru
-      [Функциональность|Функционал|Свойство|Функция]: Internal operations
+      # language: el
+      [Δυνατότητα|Λειτουργία]: Internal operations
         In order to stay secret
         As a secret organization
         We need to be able to erase past agents' memory
 
-        [Предыстория|Контекст]:
-          [Допустим|Пусть|Дано|*] there is agent A
-          [К тому же|Также|*|И] there is agent B
+        Υπόβαθρο:
+          [Δεδομένου|*] there is agent A
+          [Και|*] there is agent B
 
-        [Сценарий|Пример]: Erasing agent memory
-          [Допустим|Пусть|Дано|*] there is agent J
-          [К тому же|Также|*|И] there is agent K
-          [Когда|Если|*] I erase agent K's memory
-          [Затем|Тогда|То|*] there should be agent J
-          [Иначе|Но|*|А] there should not be agent K
+        [Παράδειγμα|Σενάριο]: Erasing agent memory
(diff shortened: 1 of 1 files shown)


acoulton:
Urgh, of course the output of the `--story-syntax` in Russian now varies between the lowest (v4.12.0) and latest (v4.13.0), so there's no way to have the current feature file pass on all builds in the matrix :-(

I'm not sure of the best solution for this, particularly because of course now that we're updating gherkin regularly it could easily recur in future (and perhaps on a gherkin release that also contains more significant changes).

* We could bump our minimum behat/gherkin to 4.13.0, but it feels wrong to force users to upgrade just to pass one of our tests 
* We could introduce a behat profile with tags to allow us to skip specific scenarios on the composer --lowest build
* We could remove the first two lines (`# language: ru` and the breaking Feature: line) in the output should contain step, which would solve it for now but give the same problem if Scenario / Given / etc were to change in future, and means the feature is less clear about what we actually expect to see.
* We could somehow have that feature do a loose match on the output with some kind of wildcard syntax that maps back to the expected translations.

I think the wildcard approach is the least likely to unexpectedly break in future, with a feature like:

```gherkin
Then the output should contain the following with russian translations:
      """
      # language: ru
      [%GHERKIN_FEATURES_KEYWORDS%]: Internal operations
        In order to stay secret
        As a secret organization
        We need to be able to erase past agents' memory

        [%GHERKIN_SCENARIO_KEYWORDS%]:
          [%GHERKIN_GIVEN_KEYWORDS%] there is agent A
      ... etc ...
      """
```

However, this means we'd be asking gherkin at runtime for both the implementation and the expected test result. Which means the feature provides less confidence about the actual output than it does now...

@carlos-granados what do you think - any better ideas?

carlos-granados:
@acoulton Mmm, it is a difficult problem.

Of the solutions that you mentioned I feel that "We could introduce a behat profile with tags to allow us to skip specific scenarios on the composer --lowest build" is the one that sounded better as it would allow us to confirm the behaviour with both versions. Can you maybe check about the feasibility of this option?

acoulton:
Yeah I think that might be best, but (because we don't pin the upper gherkin version) would still mean the "not-lowest" build will start failing on all branches if/when a future Gherkin release adds more russian translations until we update the features and merge/rebase them into all Behat branches.

Given that, I'm wondering this evening if we should just swap that example from Russian to a language we know is the same in all current Gherkin versions, and hope that it doesn't change in future... 

One of the latin-based languages that's been in gherkin for a long time would probably be fairly stable, ideally one with at least a couple of utf8 diacritics to prove the character set works as expected...

That would definitely be the quickest solution, WDYT?

acoulton:
@carlos-granados I went for Greek as it has extended characters and according to the `git blame` on the translations in upstream cucumber/gherkin it has not changed in the last 8 years so hopefully will be fairly stable.

carlos-granados (approved):
Looking good, I hope that the Greek is ok and we don't get any Greek users saying "what the heck is that?" 🤣

PR #1636, opened 2025-05-07
Backport the fix for Behat/Gherkin file paths to a 3.16.1 release

acoulton:
A number of users have been affected by the change to file paths in Behat/Gherkin v4.13.0 (released yesterday).

This is because:

* Behat/Gherkin has no dependencies, so can happily bump to 4.13.0 on any system running PHP >= 8.1
* Behat/Behat v3.17.0 introduced a dependency on composer/xdebug-handler:v3
* Behat/Behat v3.20.0 introduced a dependency on nikic/php-parser@v5
* We only released #1604 (which is required to support Behat/Gherkin >= 4.13.0) in v3.20.0

Therefore there are projects on PHP >= 8.1 (so still officially supported by us) who will get Behat/Gherkin v4.13.0 but may not be able to get Behat/Behat v3.20.0 due to having other dependencies that are holding them to earlier versions of composer/xdebug-handler or nikic/php-parser.

I propose we do a 3.16.1 release with the fix backported to solve the immediate issue (anyone running PHP 8.1 should be able to get to 3.16.0, and if they are not on 8.1 they won't get the problematic Gherkin version).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [3.16.1] - 2025-05-07
+
+### Changed
+
+* Remove dependency on file location in Gherkin package. The (internal) service container parameters gherkin.paths.lib
+  and gherkin.paths.i18n are no longer defined or used. Minimum behat/gherkin version is now ^4.12.0.
+  By @carlos-granados in #1604, backported from 3.20.0 to fix errors for users stuck on 3.16.0 due to
+  dependency conflicts.
+
 ## [3.16.0] - 2024-11-08
 
 ### Changed
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
     "require": {
         "php": "8.1.* || 8.2.* || 8.3.* || 8.4.* ",
         "ext-mbstring": "*",
-        "behat/gherkin": "^4.10.0",
+        "behat/gherkin": "^4.12.0",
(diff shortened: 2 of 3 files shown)


carlos-granados (approved):
I think this solution is appropriate

swissspidy:
Would it be possible to backport the same fix to a 3.15.1 release? WP-CLI depends heavily on Behat for testing, but also supports PHP 7.2 still, so we're stuck at v3.15 at the moment.

acoulton:
> Would it be possible to backport the same fix to a 3.15.1 release? WP-CLI depends heavily on Behat for testing, but also supports PHP 7.2 still, so we're stuck at v3.15 at the moment.

@swissspidy in principle we could look at it however I'm not sure this should be affecting you.

The gherkin version that changed this was after gherkin dropped support for <8.1, so should not be being installed on your 7.2 build anyway. 

I'd need to look properly when I'm at a PC, but as I recall you should "out of the box" get a Behat & gherkin version that work together on older php versions, and a pair that work together on anything >= 8.1.

stof:
@swissspidy this is a different issue. Latest releases of behat/behat don't declare support for PHP 8.5 yet (the dev version does), while old versions of Behat were not restricting the upper bound.

acoulton:
@swissspidy sorry, appreciate it is affecting you - I should have said that I think there must be something else affecting the dependencies you're getting and it would be better to resolve that than pin behat back to an ancient version on every language version.

As @stof says the issue is we haven't yet released the version that officially supports 8.5. I will aim to do that tomorrow. In the meantime you can either temporarily run on 8.4 or bump to behat 3.x-dev@dev

swissspidy:
Aha 💡  Appreciate the quick response! I will keep an eye on the releases then.

Issue #1596, opened 2025-01-30
Add clickable links to CLI output


carlos-granados:
Some tools like PHPStan or Rector add clickable links to their output. So when they report something about any file, you are able to click on the CLI and directly open the file in the right line in your IDE. The format of these links is configurable so that they can have the format that works with your particular IDE.

We should have equivalent functionality so that when Behat reports a a test failure in a file or shows any other file path, these paths are clickable and can be opened directly in your IDE. The format of the link should be configurable in the Behat config file.

fmatsos:
Hello @carlos-granados 👋🏻

Are you already working on this? I don't see any PR about.
I started to take a look at the code to implement this feature. I can look deeper into it if no one is already looking.

carlos-granados:
@fmatsos I started implementing another feature that would allow us to choose if we want to output absolute paths or relative paths when printing file paths. I planned to later on expand that capability to allow us to print clickable links. So I think it is best if you leave this with me

uuf6429:
Just want to point out that many (most?) terminals do not support such links or only in some very weird/specific formatting. I, for one, still have issues with it on macOS.

So, I would suggest that if this was implemented it, it would:
- be turned off by default (for BC reasons)*
- be customizable

PS: *some IDEs, such as PhpStorm already automatically turn paths in the output terminal to links (it even supports lines), but that doesn't work with formatted links.

carlos-granados:
@uuf6429 yes, the plan is to have this off by default and to be customizable. It will be implemented in the same way that PHPStan or Rector implement this capability

PR #1638, opened 2025-05-12
feat: implement editor url

carlos-granados:
In the same way that PHPStan or Rector do, allow setting an editorURL template that will be used to add links to paths when printing them on the terminal.

The editorURL can be any URL that any IDE recognises to open that specific file, One example is PhpStorm which accepts URLs like this one:

`phpstorm://open?file={path}&line={line}`

The template can use three different placeholders:

{absPath} : absolute path of the file
{relPath}: relative path of the file
{line}: line number in the file

Fixes https://github.com/Behat/Behat/issues/1596

Once this is merged I will add the corresponding documentation in Behat docs

### DIFF
--- a/src/Behat/Behat/Output/Node/Printer/ListPrinter.php
+++ b/src/Behat/Behat/Output/Node/Printer/ListPrinter.php
@@ -61,7 +61,7 @@ final class ListPrinter
     ) {
         $this->resultConverter = $resultConverter;
         $this->translator = $translator;
-        $this->configurablePathPrinter = $configurablePathPrinter ?? new ConfigurablePathPrinter($basePath, printAbsolutePaths: false);
+        $this->configurablePathPrinter = $configurablePathPrinter ?? new ConfigurablePathPrinter($basePath, printAbsolutePaths: false, editorUrl: null);
     }
 
     /**
@@ -285,12 +285,12 @@ final class ListPrinter
                 ),
                 $scope instanceof BeforeScenarioScope,
                 $scope instanceof AfterScenarioScope => $this->configurablePathPrinter->processPathsInText(
-                    $scope->getFeature()->getFile()
-                ) . ':' . $scope->getScenario()->getLine(),
+                    $scope->getFeature()->getFile() . ':' . $scope->getScenario()->getLine()
+                ),
                 $scope instanceof BeforeStepScope,
                 $scope instanceof AfterStepScope => $this->configurablePathPrinter->processPathsInText(
-                    $scope->getFeature()->getFile()
-                ) . ':' . $scope->getStep()->getLine(),
+                    $scope->getFeature()->getFile() . ':' . $scope->getStep()->getLine()
+                ),
                 default => null,
             };
         }
--- a/src/Behat/Config/Profile.php
+++ b/src/Behat/Config/Profile.php
@@ -30,6 +30,7 @@ final class Profile implements ConfigConverterInterface
     private const PRINT_UNUSED_DEFINITIONS_SETTING = 'print_unused_definitions';
     private const PATH_OPTIONS_SETTING = 'path_options';
     private const PRINT_ABSOLUTE_PATHS_SETTING = 'print_absolute_paths';
+    private const EDITOR_URL_SETTING = 'editor_url';
 
     private const DISABLE_FORMATTER_FUNCTION = 'disableFormatter';
     private const FORMATTER_FUNCTION = 'withFormatter';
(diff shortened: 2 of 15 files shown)


src/Behat/Behat/Output/Node/Printer/ListPrinter.php
```diff
@@ -285,12 +285,12 @@ private function getLocationFromScope(?HookScope $scope): ?string
                 ),
                 $scope instanceof BeforeScenarioScope,
                 $scope instanceof AfterScenarioScope => $this->configurablePathPrinter->processPathsInText(
-                    $scope->getFeature()->getFile()
-                ) . ':' . $scope->getScenario()->getLine(),
+                    $scope->getFeature()->getFile() . ':' . $scope->getScenario()->getLine()
```
carlos-granados, line 288: The path printer now expects to find the line number as part of the text so in several places we add it to the text before sending it to printer instead of adding it to the returned text

src/Behat/Config/Profile.php
```diff
@@ -244,18 +250,34 @@ private function addUnusedDefinitionsToExpr(Expr &$expr): void
 
     private function addPathOptionsToExpr(Expr &$expr): void
     {
-        if (!isset($this->settings[self::PATH_OPTIONS_SETTING][self::PRINT_ABSOLUTE_PATHS_SETTING])) {
-            return;
+        $args = [
```
carlos-granados, line 253: When calling the `addMethodCall` function later we need to have values for all parameters as they are needed in the function that checks the default values, so we start with the default values and modify them if any path option setting is found

src/Behat/Testwork/PathOptions/Printer/ConfigurablePathPrinter.php
```diff
@@ -32,15 +35,55 @@ public function setPrintAbsolutePaths(bool $printAbsolutePaths): void
...
+            if ($this->printAbsolutePaths === true) {
+                return $text;
+            }
+
+            return str_replace($this->basePath . DIRECTORY_SEPARATOR, '', $text);
         }
 
-        return str_replace($this->basePath . DIRECTORY_SEPARATOR, '', $text);
+        // Search for paths in the text
+        $basePathPattern = preg_quote($this->basePath . DIRECTORY_SEPARATOR, '/');
```
carlos-granados, line 58: Our text is not a single text with a path, it is a text that can contain multiple paths (for example in an error stack trace) immersed in the text, so we need to loop over any path found and apply our settings

carlos-granados:
Argh, bloody Windows dir separators, let me try to fix the tests

carlos-granados:
Fixed 🎉

src/Behat/Behat/Output/Node/Printer/ListPrinter.php
acoulton: It looks like this is just a code formatting change, unless I'm missing something?

src/Behat/Config/Profile.php
acoulton: This feels a little verbose / hard to follow to me (especially if we add more path options settings in future). Is there a way we could simplify it? Maybe with something like a map of setting name => parameter name and then a foreach over them?

src/Behat/Testwork/PathOptions/Printer/ConfigurablePathPrinter.php
```diff
@@ -10,13 +10,16 @@
...
+use Symfony\Component\Console\Formatter\OutputFormatter;
+
 final class ConfigurablePathPrinter
 {
     private string $basePath;
 
     public function __construct(
         string $basePath,
         private bool $printAbsolutePaths,
+        private ?string $editorUrl,
```
acoulton, line 22: Doesn't this need a default null for BC?

acoulton (commented):
@carlos-granados this looks great and will be super helpful - just a couple of small observations

src/Behat/Testwork/PathOptions/Printer/ConfigurablePathPrinter.php
carlos-granados: updated

src/Behat/Behat/Output/Node/Printer/ListPrinter.php
carlos-granados: if you look carefully you will see that in the previous code the concatenation of the line number happens after the call to `processPathsInText` and now it happens before.

src/Behat/Config/Profile.php
carlos-granados: Let me know if this looks better now

acoulton: Thanks, yes I find that much clearer

src/Behat/Behat/Output/Node/Printer/ListPrinter.php
acoulton: Of course, sorry :facepalm:

acoulton (approved):
Great, thank you :)

Issue #1506, opened 2024-10-31
Add Rector to our code base


carlos-granados:
Once we have a coding standard tool in place, I think we could add [Rector](https://github.com/rectorphp/rector) to our code base. Having this tool available would allow us to easily:
- Update our code base to use the latest PHP capabilities once we have tied down the PHP versions that we support
- Increase our type coverage
- And in general keep a tidier and nicer code base

ttomdewit:
I've had some bad luck with Rector many, many attempts ago. I'm all for adding it, that goes without saying, but I'd be very careful as to what's enabled and how it's ran in CI. I am not a fan of "big bang" PRs that change a lot of different things at the same time. My ideal would be to introduce the tool and start enabling rules/fixers one at a time. Once enabled it'll keep running in CI. I believe this is a more manageable and sensible approach.

ttomdewit:
> I've had some bad luck with Rector many, many attempts ago. I'm all for adding it, that goes without saying, but I'd be very careful as to what's enabled and how it's ran in CI. I am not a fan of "big bang" PRs that change a lot of different things at the same time. My ideal would be to introduce the tool and start enabling rules/fixers one at a time. Once enabled it'll keep running in CI. I believe this is a more manageable and sensible approach.

Sorry, I missed https://github.com/Behat/Behat/issues/1306#issuecomment-2449323357 but it echos my ideas here as well.

carlos-granados:
@ttomdewit yes, with Rector it makes sense to go slowly and not introduce a lot of changes at the same time

acoulton:
@ttomdewit couldn't agree more. Rector (and ECS) are great tools but we've had problems on legacy projects where rules that should have been just code-style / internal have produced breaking changes because our original code was not in an expected state, so being able to review the changes sensibly is vital.

stof:
Upgrading automatically to newer PHP features with Rector can also be risky as Rector does not consider the impact of the change on backward compatibility. So I would not add it as part of a CI step enforcing the code to match the Rector output. In my experience, the output of Rector requires code review on what it does (especially in the context of a package subject to semver).

Regarding increasing type coverage, I think this is more about adding static analysis in the project than about adding Rector (btw, Rector is generally risky to run if you don't have phpstan passing at a high level)

rpkamp:
Agree with @stof. I wouldn't add Rector before PHPStan is also running at at least level 7. Ideally higher.

acoulton:
[...]

PR #1640, opened 2025-05-16
Add Rector

carlos-granados:
Now that we have a full coding standard configuration we are ready to add the Rector tool and start using it.

We need to be careful when using it as some of the changes introduced by Rector can create BC issues so we need to evaluate all of its rules and only apply those which are safe.

This PR introduces the initial setup to use this tool and adds a couple of things:
- The initial level of the Code Quality rules which applies a couple of rules
- Import handling, which makes sure that we are using import statements instead of FQN and removes unused imports

I have separated this in two commits (tool setup and code changes) so that the code changes can be added to the blame ignore file once this PR is approved

Closes https://github.com/Behat/Behat/issues/1506

### DIFF
--- a/src/Behat/Testwork/Exception/Stringer/PHPUnitExceptionStringer.php
+++ b/src/Behat/Testwork/Exception/Stringer/PHPUnitExceptionStringer.php
@@ -11,6 +11,8 @@
 namespace Behat\Testwork\Exception\Stringer;
 
 use Exception;
+use PHPUnit\Framework\TestFailure;
+use PHPUnit\Util\ThrowableToStringMapper;
 use PHPUnit_Framework_Exception;
 use PHPUnit_Framework_TestFailure;
 use Throwable;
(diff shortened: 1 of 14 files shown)


src/Behat/Testwork/Exception/Stringer/PHPUnitExceptionStringer.php
```diff
@@ -62,14 +64,14 @@ public function stringException(Exception $exception, $verbosity)
             // Class names are intentionally fully qualified here to maximise clarity - particularly because a future
             // PHPUnit version may reuse a class name in a different namespace.
 
-            if (class_exists(\PHPUnit\Util\ThrowableToStringMapper::class)) {
+            if (class_exists(ThrowableToStringMapper::class)) {
```
acoulton, line 67: It's probably OK to lose the fully-qualified names here - particularly now we know that people won't be able to use just the assertions from future PHPUnit versions anyway?

If so we should just remove the comment above.

carlos-granados: I had not realised about the comment, agree with you, removed

carlos-granados:
@acoulton do you think this is now OK? I would like to merge it so that we can continue adding more Rector rules

acoulton (approved):

PR #1644, opened 2025-05-24
dx: add option that allows you to remove prefixes in printed paths

carlos-granados:
If your feature or context files are located in some subfolders, for example `tests/behat/features` and `src/behat`, it is not very informative if these folders are always printed when printing paths and you may prefer to remove them, so that for example instead of

```
Scenario: # tests/behat/features/users/login.feature:3
```

you see
```
Scenario: # users/login.feature:3
```

This PR adds a new `removePrefix` option to the configurable path printer that allows you to define a list of prefixes that need to be removed from paths when printing them. This affects only the visible paths, not the ones used in the editorUrl. This option is available in the command line (where the prefixes need to be comma separated), the yaml config and the php config

These prefixes are removed in the order in which they are defined, so you need to list more specific prefixes first

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -468,7 +468,7 @@ EOL;
         // windows path fix
         if ('/' !== DIRECTORY_SEPARATOR) {
             $text = preg_replace_callback(
-                '/[ "]features\/[^\n "]+/',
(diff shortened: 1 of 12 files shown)


acoulton (approved):
@carlos-granados this looks great, thanks - no comments from me. It will be super helpful!

Very sorry that I didn't notice this originally, you should have poked me sooner :)

PR #1648, opened 2025-06-20
Rector code quality

carlos-granados:
Updates the Rector settings to apply the whole set of Rector code quality rules. Most of these rules do not produce changes in our code. For those rules that produce changes I have created one commit per rule to make this easier to review. Once this is fully reviewed, I will modify the last commit to include the git blame ignores for all these commits, please do not merge until this is done

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -10,7 +10,7 @@ return RectorConfig::configure()
         __DIR__ . '/src',
     ])
     ->withRootFiles()
-    ->withCodeQualityLevel(0)
+    ->withPreparedSets(codeQuality: true)
     ->withImportNames(
         removeUnusedImports: true,
     )
--- a/src/Behat/Behat/Context/Snippet/Appender/ContextSnippetAppender.php
+++ b/src/Behat/Behat/Context/Snippet/Appender/ContextSnippetAppender.php
@@ -99,7 +99,7 @@ final class ContextSnippetAppender implements SnippetAppender
      */
     private function logSnippetAddition(AggregateSnippet $snippet, $path)
     {
-        if (!$this->logger) {
+        if ($this->logger === null) {
             return;
         }
 
--- a/src/Behat/Behat/Context/Snippet/Appender/ContextSnippetAppender.php
+++ b/src/Behat/Behat/Context/Snippet/Appender/ContextSnippetAppender.php
@@ -99,7 +99,7 @@ final class ContextSnippetAppender implements SnippetAppender
      */
     private function logSnippetAddition(AggregateSnippet $snippet, $path)
     {
-        if ($this->logger === null) {
+        if (!$this->logger instanceof FilesystemLogger) {
             return;
         }
 
--- a/src/Behat/Behat/Transformation/Transformer/RepositoryArgumentTransformer.php
+++ b/src/Behat/Behat/Transformation/Transformer/RepositoryArgumentTransformer.php
@@ -160,7 +160,7 @@ final class RepositoryArgumentTransformer implements ArgumentTransformer, RegexG
         return array_reduce($transformations, function ($acc, $t) {
             return [
                 $t instanceof SimpleArgumentTransformation ? array_merge($acc[0], [$t]) : $acc[0],
-                !$t instanceof SimpleArgumentTransformation ? array_merge($acc[1], [$t]) : $acc[1],
+                $t instanceof SimpleArgumentTransformation ? $acc[1] : array_merge($acc[1], [$t]),
             ];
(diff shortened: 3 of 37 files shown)


src/Behat/Behat/Transformation/Transformer/RepositoryArgumentTransformer.php
```diff
@@ -160,7 +160,7 @@ private function splitSimpleAndNormalTransformations(array $transformations)
         return array_reduce($transformations, function ($acc, $t) {
             return [
                 $t instanceof SimpleArgumentTransformation ? array_merge($acc[0], [$t]) : $acc[0],
-                !$t instanceof SimpleArgumentTransformation ? array_merge($acc[1], [$t]) : $acc[1],
+                $t instanceof SimpleArgumentTransformation ? $acc[1] : array_merge($acc[1], [$t]),
             ];
         }, [[], []]);
```
acoulton, line 165: This existing logic took a bit of figuring out! It would probably be easier to follow as a pair of simple array_filters (and if the array_reduce was originally a performance concern I'm not sure the repeated array_merge makes it any better).

But I agree this change is correct for the existing implementation.

src/Behat/Behat/Context/Snippet/Appender/ContextSnippetAppender.php
```diff
@@ -99,7 +99,7 @@ private function importClass($class, $contextFileContent)
      */
     private function logSnippetAddition(AggregateSnippet $snippet, $path)
     {
-        if ($this->logger === null) {
+        if (!$this->logger instanceof FilesystemLogger) {
```
acoulton: Only thought about the changes from this `FlipTypeControlToUseExclusiveTypeRector` rule is that in future we might change the property type. For example here `$this->logger` could change to an interface / parent type / something else that is not a `FilesystemLogger`.

The old `===null` wouldn't be affected by that, but the new ` ! instanceof FilesystemLogger` would always be true, so changing the property type could cause an unexpected behaviour change.

Would our tooling detect that (e.g. report that the if condition is always true / always false)?

rector.php
```diff
@@ -10,7 +10,7 @@
         __DIR__ . '/src',
     ])
     ->withRootFiles()
-    ->withCodeQualityLevel(0)
+    ->withPreparedSets(codeQuality: true)
```
acoulton, line 13: Are we at risk of CI breaking unexpectedly if Rector add new rules to the pre-defined "codeQuality" set that we have not yet adopted? We are only pinned to `^2.0`, so every CI build will run with whatever happens to be the latest 2.x release of Rector at the time. I'm not sure their versioning strategy for changes to the config / implementation of rules?

We should perhaps be explicit about the rules - or alternatively require an explicit Rector version in composer.json and have a job (dependabot or a custom one like we have in behat/gherkin for the gherkin dependency) to update to each new Rector version as a standalone PR so that we can deal with any issues that arise without affecting other branches?

carlos-granados: I don't want to be explicit about the rules, because there are a lot to be listed, so instead what I have done is to pin the explicit version of Rector that we are requiring. Since this tool is only used internally for Behat development, we can keep updating the required version when we add more Rector rules

src/Behat/Behat/Context/Snippet/Appender/ContextSnippetAppender.php
carlos-granados: Yes, I think that our tooling would flag this case and the appropriate thing would be to update the check to check the new class

carlos-granados:
@acoulton sorry I took so long to get back to this. Summer is complicated here, hoping to get back to speed in September 😄

src/Behat/Behat/Context/Snippet/Appender/ContextSnippetAppender.php
stof: phpstan can detect such case (but maybe not with our existing config which runs only at level 5 and without strict rules)

carlos-granados: @stof it does detect this case with our current level and setup, just tested it 😄

rector.php
acoulton: That's a good solution.

Probably still worth having dependabot bump it so it doesn't fall too far behind (which could theoretically start holding back other dependencies in CI) but we can tackle that separately.

acoulton (approved):
Great, thanks @carlos-granados - and no worries, summer is busy here too 😁

Issue #1637, opened 2025-05-07
Dependency conflicts may prevent people taking more recent Behat releases


acoulton:
We have added two new composer dependencies in the last few releases:

* v3.17.0 added `"composer/xdebug-handler": "^3.0"` - to allow us to improve performance by disabling xdebug unless requested
* v3.20.0 added `"nikic/php-parser": "^5.0"` (subsequently bumped to `^5.2` in v3.21.0) - to power the YAML-to-PHP `--convert-config` command

Both of these packages are also required by a reasonable number of other packages (particularly for dev tools, though also some runtime). 

This means that there are users out there with potentially complex dependency chains that prevent them from updating to the newer Behat releases. 
That's a shame for users, but also means we have to be more cautious about assuming everyone will bump up Behat/Behat and Behat/Gherkin releases broadly in parallel (as discovered in https://github.com/Behat/Gherkin/issues/317).

I think therefore that we should try to relax these requirements if we can:

* composer/xdebug-handler v2 and v3 appear to be compatible other than PHP version support. I don't know how feasible it would be to also support v1, which is listed as a breaking release.
* we do need the features of nikic/php-parser@^5.2 but only if/when the user runs the one-time `--convert-config` command. I don't imagine we will use it for anything else, so we should make this a suggested dependency, and ask the user to install it at runtime only if they try to use it.

agentrickard:
`"sebastian/diff": "^4.0",` may have the same issue now. Ref: https://www.drupal.org/project/drupal/releases/11.2.2

acoulton:
@agentrickard that shouldn't be the case: `sebastian/diff` is only a dev dependency, so will only be required if you're installing Behat as a root package. If you're requiring Behat from another project, then our dev dependencies will be ignored.

PR #1650, opened 2025-06-21
chore: Support 4.19.x versions of nikic/php-parser

acoulton:
Requiring ^5.2 was causing dependency conflicts that were preventing end-users from upgrading to newer Behat versions.

We only actually *require* nikic/php-parser for running the convert-config command to generate PHP config from the old YAML files. I therefore explored making it a suggested dependency that was only used to run that command.

The problem is that it is also required indirectly by many of our dev dependencies (including PHPUnit, which we use to run our assertions). This means we can't configure a CI run where it is **not** installed, so would not have confidence that Behat does indeed work as expected without it.

The best compromise is to relax the constraint to ^4.19.2 (this version introduced enum support, which we need). That should resolve the conflict for the majority of users.

Refs #1637

### DIFF
--- a/features/convert_config.feature
+++ b/features/convert_config.feature
@@ -177,7 +177,7 @@ Feature: Convert config
               ->withSuite((new Suite('my_suite'))
                   ->withContexts(
                       'MyContext',
-                      'App\AnotherContext'
+                      'AnotherContext'
                   )));
       """
     And the temp "suite_contexts.yaml" file should have been removed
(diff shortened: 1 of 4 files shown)


features/convert_config.feature
```diff
@@ -177,7 +177,7 @@ Feature: Convert config
               ->withSuite((new Suite('my_suite'))
                   ->withContexts(
                       'MyContext',
-                      'App\AnotherContext'
+                      'AnotherContext'
```
acoulton, line 180: The only slight issue is that the 4.x series includes a double backslash in strings e.g. it would produce `App\\AnotherContext` rather than the `App\AnotherContext` that we get from 5.x.

The ["Class references for known extensions and contexts"](https://github.com/Behat/Behat/pull/1650/files#diff-a2faa89adc453668aebf0b150921cf323b46affe2065b08dd197d6c0774a4311R291) scenario already covers behaviour with namespaced classes that are defined and can therefore be converted to imports. This proves we're reading the namespaced class names correctly from YAML and passing them around internally without issues.

Other aspects of that scenario and of these ones cover rendering the class name as a string if it is not a defined class.

IMO therefore we don't need to specifically cover the case where it's a string that contains a namespace separator, so the simplest fix is to just change these examples to ones where both versions of php-parser produce the same output.

stof (approved):

PR #1656, opened 2025-08-26
fix: Ensure that suggested step definitions actually match the step text

### LINKED ISSUE #1653: Behat suggests a snippet that already exists
Given I have a feature file like this:

```feature
Feature: Plan workers for project segments
  Scenario: Extend project segment to Saturday
    When the planner extends the project segment for "Excavation" to include "2025-08-02"
    Then "2025-08-02" (Saturday) is available for planning on project "Excavation"
```

When I run behat command, it suggests I create a step:
```
--- FeatureContext has missing steps. Define them with these snippets:

    #[When('the planner extends the project segment for :arg1 to include :arg2')]
    public function thePlannerExtendsTheProjectSegmentForToInclude($arg1, $arg2): void
    {
        throw new PendingException();
    }

--- Don't forget these 2 use statements:

    use Behat\Behat\Tester\Exception\PendingException;
    use Behat\Step\When;
```

If that step already exists, and I run the test again, then behat suggests that I add this snippet again - the same snippet.

When I do that, then it says:

```
Feature: Plan workers for project segments

  Scenario: Extend project segment to Saturday                                            # features\Acceptance\Planning\PlanWorkersForProjectSegments.feature:2
    When the planner extends the project segment for "Excavation" to include "2025-08-02"
      Step "the planner extends the project segment for :arg1 to include :arg2" is already defined in FeatureContext::thePlannerExtendsTheProjectSegmentForToInclude()

      FeatureContext::thePlannerExtendsTheProjectSegmentForToInclude()
      FeatureContext::thePlannerExtendsTheProjectSegmentForToInclude2()
    Then "2025-08-02" (Saturday) is available for planning on project "Excavation"
      Step "the planner extends the project segment for :arg1 to include :arg2" is already defined in FeatureContext::thePlannerExtendsTheProjectSegmentForToInclude()

      FeatureContext::thePlannerExtendsTheProjectSegmentForToInclude()
      FeatureContext::thePlannerExtendsTheProjectSegmentForToInclude2()

--- Failed scenarios:

    features\Acceptance\Planning\PlanWorkersForProjectSegments.feature:2
```

If I remove parenthesis from the steps in feature file, the problem goes away.

acoulton:
Previously, when generating snippets for undefined step definitions,
there was no guarantee that the pattern would actually match the step
text. In some cases (in particular with turnip) this proved to be an
issue when feature files contained complex or unexpected syntax:

* The first run would apparently provide a valid snippet.
* The second run would still not match this snippet, so would propose a
  new step with the same pattern.
* The third run would fail due to the presence of duplicate patterns in
  the steps from the first and second run.

Instead, check that the generated snippet actually matches the step text
at the time of generating it. If not, show the user a warning with
information about either generating these as regex or manually defining
a step to match their features.

Fixes #1653

### DIFF
--- a/i18n.php
+++ b/i18n.php
@@ -91,7 +91,7 @@ return [
         'snippet_proposal_use' => 'No olvides estas %count% declaraciones de use:',
         'snippet_missing_title' => '<snippet_undefined>Las plantillas para los siguientes pasos en <snippet_keyword>%count%</snippet_keyword> no fueron generadas (revisa tu configuración):</snippet_undefined>',
         'snippet_generation_failure_title' => 'No se pudieron generar automáticamente fragmentos que coincidan con los siguientes pasos:',
-        'snippet_generation_failure_hint' => '(intente usar --snippets-type=regex o defina el paso manualmente)',
+        'snippet_generation_failure_hint' => '(intenta usar --snippets-type=regex o define el paso manualmente)',
         'failed_scenarios_title' => 'Escenarios fallidos:',
         'failed_hooks_title' => 'Hooks fallidos:',
         'failed_steps_title' => 'Pasos fallidos:',


features/snippet_generation_failures.feature
```diff
@@ -0,0 +1,82 @@
...
+        use Behat\Step\Given;
+
+    --- Could not automatically generate snippets matching the following steps:
+        (try using --snippets-type=regex, or manually define the step)
+
+        - a step with (Parentheses) in the actual step text
+    """
+
+  Scenario: See warnings when appending snippets
+    # @todo: Need to reset the state of the FeatureContext in the fixture before each execution
```
acoulton: @carlos-granados there's a bit of an issue here with the new fixture directory implementation. Because we set the working directory to the actual path to the fixtures, runs that need to modify those files (e.g. to append snippets) then affect subsequent runs and/or other scenarios that use the same fixture.

I think potentially we should **copy** the whole fixture directory to a temporary path, and use that as the working directory, so that we have a clean state for every execution?

I can look at this as a separate PR in case it throws up any issues we need to resolve with other features / fixtures.

i18n.php
```diff
@@ -6,6 +6,9 @@
         'snippet_proposal_title' => '<snippet_undefined><snippet_keyword>%count%</snippet_keyword> has missing steps. Define them with these snippets:</snippet_undefined>',
         'snippet_proposal_use' => 'Don\'t forget these %count% use statements:',
         'snippet_missing_title' => '<snippet_undefined>Use <snippet_keyword>--snippets-for</snippet_keyword> CLI option to generate snippets for following <snippet_keyword>%count%</snippet_keyword> suite steps:</snippet_undefined>',
+        // @todo: Add translations for other languages
+        'snippet_generation_failure_title' => 'Could not automatically generate snippets matching the following steps:',
+        'snippet_generation_failure_hint' => '(try using --snippets-type=regex, or manually define the step)',
```
acoulton, line 11: If we're happy with the message in English I can generate some translations. I've made it two separate strings because these lines are rendered indented to match the style of the output above, and it feels like this indent should be the responsibility of the printer rather than the translation texts.

src/Behat/Behat/Definition/Pattern/PatternTransformer.php
```diff
@@ -79,6 +79,24 @@ public function transformPatternToRegex($pattern)
...
 
+    /**
+     * Tests if a pattern matches a given step text.
+     *
+     * If the pattern does not match, returns false. Otherwise, returns an array of the portions of the
+     * step text that match any placeholders within the pattern.
+     *
+     * @return false|array<int|string,string>
+     */
+    public function matchPattern(string $pattern, string $stepText): false|array
```
acoulton, line 90: It doesn't 100% feel like this method belongs in a class called `PatternTransformer` but I felt it was an OK trade-off for being able to be certain that the ContextSnippetGenerator is using the exact same logic as the RepositorySearchEngine to check if the pattern matches the stepText.

PatternTransformer is a final class and used directly by RepositorySearchEngine and ContextSnippetGenerator so there's no BC issue with the new method.

src/Behat/Behat/Snippet/Cli/SnippetsController.php
```diff
@@ -139,6 +143,14 @@ public function printAllSnippets()
         $this->writer->printSnippets($this->printer, $snippets);
     }
 
+    private function printSnippetGenerationFailures(): void
+    {
+        $failures = $this->registry->getGenerationFailures();
+        count($failures) && $this->output->writeln('');
+
+        $this->printer->printSnippetGenerationFailures($failures);
```
acoulton, line 157: Although there is a `SnippetPrinter` interface, which does not contain this new method, the `SnippetsController` depends directly on `ConsoleSnippetPrinter` so this is BC without any additional checks required.

The same is true for the new `getGenerationFailures` method called above - it's not in the `SnippetRepository` interface, but this class depends directly on ``SnippetRegistry` rather than the interface.

features/snippet_generation_failures.feature
stof: Is this modifying the files tracked by git or a copy ? Modifying the tracked files (for something that should not be committed) is also a recipe for mistakes.

acoulton: @stof, yes, currently it modifies the files tracked by git - agreed, that's not a good plan.

I think this is the first example using the new fixtures setup where the files are modified during the execution, hence it hasn't come up till now.

carlos-granados: @acoulton you could do something similar to what we do in the `convert_config.feature`. We copy the files that we need to modify to the temp folder and use them from there.

acoulton: @carlos-granados I looked at that, but I think it's potentially simpler and safer to always run the fixtures in an isolated directory - avoids any issues if one of our scenarios unexpectedly causes Behat to overwrite any of the files, for example. 

I've done a PR for that in #1657

carlos-granados:
@acoulton I guess you would need to rebase and modify this PR once the PR to modify how we run the fixture files is merged. I would prefer to wait until this is done before I review this other PR, please let me know

acoulton:
Absolutely, I'll rebase and update & let you know when it's ready.

acoulton:
@carlos-granados this is ready to review (the only outstanding thing for me is to generate translations of the messages if we're happy with the English versions)

src/Behat/Behat/Snippet/Cli/SnippetsController.php
```diff
@@ -139,6 +143,14 @@ public function printAllSnippets()
         $this->writer->printSnippets($this->printer, $snippets);
     }
 
+    private function printSnippetGenerationFailures(): void
+    {
+        $failures = $this->registry->getGenerationFailures();
+        count($failures) && $this->output->writeln('');
```
carlos-granados: I am not a big fan of using logic expression short circuits to replace if statements, I think they are less legible, so I'd prefer to see a proper `if` here, feel free to ignore

acoulton: Nor am I actually, I must have been slightly blindly following the pattern of the existing methods above.

Thanks for spotting that, will definitely change it.

stof: I agree with @carlos-granados here.

Btw, if we enable phpstan-strict-rules in the future (we are far from it as we run only at level 5, and enabling strict rules before achieving at least level 6 would be weird for sure), it will also complain about such usage.

acoulton: I've converted the pre-existing code in this class to also use explicit `if` as well

i18n.php
```diff
@@ -81,6 +90,8 @@
         'snippet_proposal_title' => '<snippet_undefined>A <snippet_keyword>%count%</snippet_keyword> le faltan pasos. Defínelos con estos pasos:</snippet_undefined>',
         'snippet_proposal_use' => 'No olvides estas %count% declaraciones de use:',
         'snippet_missing_title' => '<snippet_undefined>Las plantillas para los siguientes pasos en <snippet_keyword>%count%</snippet_keyword> no fueron generadas (revisa tu configuración):</snippet_undefined>',
+        'snippet_generation_failure_title' => 'No se pudieron generar automáticamente fragmentos que coincidan con los siguientes pasos:',
+        'snippet_generation_failure_hint' => '(intente usar --snippets-type=regex o defina el paso manualmente)',
```
carlos-granados: In Spanish there are two "voices" for the second-person singular: "tú" which is more informal and "usted" which is more formal. In all other Spanish texts we are using the informal "tú" version but this text uses the formal "usted" version, so it should be changed to
"intenta usar --snippets-type=regex o define el paso manualmente"

acoulton: Thanks - fixed now. Impressed it got that close! 

I guess that's like the French tu and vous? Interestingly we appear to be using vous (-ez) forms for French - someone must have decided to be more polite / formal to francophones 😂

carlos-granados: In French "vous" is much more used than "usted" in Spanish, in French "tu" is reserved to very close people while in Spanish "tú" is much more widely used. This was different a century ago, where in Spanish "usted" was still the most common usage but has changed a lot since the 60s-70s

carlos-granados (approved):
Thanks @acoulton everything looking great now

i18n.php
acoulton: Thanks, that's great to add to my extremely basic knowledge of Spanish! :)

PR #1657, opened 2025-08-26
test: Run fixture-based scenarios in isolated directories

acoulton:
Previously, we were running the scenario in the fixture directory itself (or explicitly copying individual files to a temporary directory). This risks Behat modifying fixture files during the scenario (potentially unexpectedly) and therefore affecting subsequent scenarios.

Instead, run every scenario in an isolated temporary working directory, and simply initialise this from the fixture at the beginning. This guarantees that every scenario runs in an isolated environment.

It also simplifies our step definitions, as we no longer need to provide for copying individual files to a temp dir or having separate steps to assert the content / presence of files in the temp dir.

Placeholders to some paths in features have changed because the paths are now relative to the %%WORKING_DIR%% placeholder (and are not nested under the fixture directory path) rather than to the project root.

This PR also adds some small refactors and enhancements to our features & FeatureContext - including fixing a bug that we were not cleaning up temporary directories created while running our features.

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -370,6 +370,21 @@ EOL;
         Assert::assertEquals($this->getExpectedOutput($text), $fileContent);
     }
 
+    #[Then(':path file should contain exactly:')]
+    public function fileShouldContainExactly(string $path, PyStringNode $text): void
+    {
+        $path = $this->workingDir.'/'.$path;
+        Assert::assertFileExists($path);
+
+        $fileContent = trim(file_get_contents($path));
+        // Normalize the line endings in the output
+        if ("\n" !== PHP_EOL) {
+            $fileContent = str_replace(PHP_EOL, "\n", $fileContent);
+        }
+
+        Assert::assertEquals($text, $fileContent);
+    }
+
     /**
      * Checks whether specified content and structure of the xml is correct without worrying about layout.
      *
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -66,9 +66,7 @@ class FeatureContext implements Context
      */
     public static function cleanTestFolders()
     {
-        if (is_dir($dir = sys_get_temp_dir() . DIRECTORY_SEPARATOR . 'behat')) {
-            self::clearDirectory($dir);
-        }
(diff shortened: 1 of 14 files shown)


features/editor_url.feature
```diff
@@ -47,14 +47,14 @@ Feature: Editor URL
       | --editor-url | 'phpstorm://open?file={absPath}&line={line}' |
     Then the output should contain:
       """
-        Scenario:                                    # <href=phpstorm://open?file={BASE_PATH}tests/Fixtures/EditorUrl/features/test.feature&line=3>features/test.feature:3</>
+        Scenario:                                    # <href=phpstorm://open?file=%%WORKING_DIR%%features/test.feature&line=3>features/test.feature:3</>
```
acoulton, line 50: As a separate PR, I might look at standardising / simplifying the way our FeatureContext handles the various placeholders for directories and filesystem paths - they are not quite consistent with each other, and I think some might be redundant.

But for the purpose of this PR, all paths in features will now be relative to the (previously defined) %%WORKING_DIR%% placeholder.

acoulton:
I'll take a look at the directory separator and path issues on mac/windows tomorrow.

features/bootstrap/FeatureContext.php
```diff
@@ -67,9 +63,12 @@ class FeatureContext implements Context
      */
     public static function cleanTestFolders()
     {
-        if (is_dir($dir = sys_get_temp_dir() . DIRECTORY_SEPARATOR . 'behat')) {
-            self::clearDirectory($dir);
-        }
+        (new Filesystem())->remove(sys_get_temp_dir() . DIRECTORY_SEPARATOR . 'behat');
+    }
+
+    public function __construct(
```
carlos-granados: My preference is to have the constructor as the first function in a class, but that may be personal preference, feel free to ignore this comment

acoulton: Thanks - I'd generally have static functions first, then constructor before any instance methods.

But I wouldn't object to changing it. 

Thinking about it, now that we have promoted properties I can see the logic for the constructor being right at the top with any other property definitions.

acoulton: Done, thanks for the suggestion @carlos-granados

PR #1661, opened 2025-09-14
Fix editorUrl functionality

carlos-granados:
Yesterday I tried to use the new "editor url" functionality in my main project, only to find that it does not really work as expected 😞 . Since our project uses the Symfony console component, the same that Rector and PHPStan use, I had assumed that `<href=` links would be transformed by the console component in the same way as in those tools, so that this would be converted to the proper escape codes that actually show a link in the console. But our project uses a custom console formatter which is much simpler than the standard Symfony formatter and this conversion is not done. So our href links end up being printed verbatim to the console instead of being converted to proper links.

I feel a bit stupid about this because I would have sworn that I had actually tested this output in the console, but obviously I didn't. And the tests themselves should have told me that this was not OK, as the test output that we checked included these links being printed verbatim without being converted 😬 

This PR adds this conversion from href links to escape characters to our console formatter. And this time I did properly check that this works as expected in the console

### DIFF
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -9,6 +9,7 @@
  */
 
 use Behat\Behat\Context\Context;
+use Behat\Behat\Output\Printer\Formatter\ConsoleFormatter;
 use Behat\Gherkin\Node\PyStringNode;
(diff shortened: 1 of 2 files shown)


acoulton:
Good plan - #1662

PR #1664, opened 2025-09-19
dx: add short_summary formatter option

carlos-granados:
When printing the summary of the run with any failed steps/scenarios, the pretty printer prints a short summary which just lists the failed scenarios while the progress printer prints a longer summary which lists all failed steps in detail. If you are running a long list of tests with the pretty printer, you may want to see the longer summary so that you could see the details of the failure without having to scroll up a lot. Conversely, if you are running the progress formatter, in some environments (for example in CI) you may want to just see a smaller summary.

This PR adds a new `short_summary` formatter option which can be used to decide if you want to show the short summary (with just a list of scenarios) or the long summary (with a list of steps). It defaults to true for the pretty printer and to false for the progress printer so that by default they keep the current functionality


src/Behat/Behat/Output/Node/Printer/Pretty/PrettyStatisticsPrinter.php
```diff
@@ -49,12 +50,22 @@ public function printStatistics(Formatter $formatter, Statistics $statistics)
         $hookStats = $statistics->getFailedHookStats();
         $this->listPrinter->printFailedHooksList($printer, 'failed_hooks_title', $hookStats, true);
 
-        $scenarioStats = $statistics->getSkippedScenarios();
-        $this->listPrinter->printScenariosList($printer, 'skipped_scenarios_title', TestResult::SKIPPED, $scenarioStats);
+        $shortSummary = $formatter->getParameter('short_summary');
+        if ($shortSummary) {
```
carlos-granados, line 54: In both printers we print the list of scenarios or the list of steps depending on the new option

src/Behat/Config/Formatter/Formatter.php
```diff
@@ -115,6 +118,9 @@ protected function toPhpExprForNamedFormatter(): Expr
                 $value = ShowOutputOption::from($value);
                 $name = ShowOutputOption::PARAMETER_NAME;
             }
+            if ($name === self::SHORT_SUMMARY_SETTING) {
```
carlos-granados, line 121: The name of the option is `short_summary` but the name of the parameter is `shortSummary` so we need to convert it

tests/Behat/Tests/Config/ProfileTest.php
```diff
@@ -188,18 +189,10 @@ public function testSettingShowOutputOption(): void
                 'progress' => [
                     'timer' => true,
                     'show_output' => 'yes',
+                    'short_summary' => false,
                 ],
                 'junit' => false,
             ],
         ], $profile->toArray());
     }
-
```
carlos-granados, line 196: This test is not needed any longer because now we accept the `InSummary` option for the pretty formatter

src/Behat/Config/Formatter/PrettyFormatter.php
```diff
@@ -24,26 +24,25 @@ final class PrettyFormatter extends Formatter
...
     public function __construct(
         bool $timer = true,
         bool $expand = false,
         bool $paths = true,
         bool $multiline = true,
         ShowOutputOption $showOutput = ShowOutputOption::Yes,
+        bool $shortSummary = true,
         ...$baseOptions,
     ) {
-        if ($showOutput === ShowOutputOption::InSummary) {
```
carlos-granados, line 36: The pretty formatter can now show step output in its summary if the short summary option is set to false, so now this option is valid for this formatter and we can remove this check

features/short_summary.feature
```diff
@@ -0,0 +1,86 @@
...
+          }
+      """
+
+  Scenario: Print short summary in progress printer
+    When I run behat with the following additional options:
+      | option    | value                       |
+      | --profile | progress_with_short_summary |
+    Then the output should contain:
+      """
+      --- Failed hooks:
```
acoulton, line 58: Is it worth including a line of context above this, to make it explicit that the `╳  This suite has a failed teardown (Exception)` etc in the example above is not printed (I assume?) rather than just not included in this example?

carlos-granados: That part of the output is not actually generated by the stats printer and it is not really relevant for this functionality, so what I have done is remove it from the first case

acoulton: Perfect, I think that's clearer, thanks.

carlos-granados:
Thanks @acoulton I will soon add documentation for this and also for the remove prefixes option that was merged recently

PR #1666, opened 2025-09-25
Rector php sets

carlos-granados:
Applies all Rector rules related to our PHP "language level" to upgrade our code to use all language features available until PHP 8.1.

The PR has been divided in a number of commits which apply one or a few rules to make it easier to review. There are a couple of very large commits but in those commits the changes are all very similar so I believe it should be easy to review


acoulton (approved):
Thanks @carlos-granados, I've had a good look over these and all LGTM. Thanks for breaking it into separate commits, that made it much easier :)

PR #1668, opened 2025-10-02
fix: Provide more context when failing with unexpected multiline args

acoulton:
Because of the stage where it is thrown, the exception is treated as an overall fatal error. This meant that Behat does not output any detail on the feature file / step it was parsing when it encountered the unexpected argument. As a result, users had no information on the cause of the error and would have to check their features / steps manually.

Instead, include the name of the function and (where possible) the feature and step within the exception message. This will still mean that users have to fix these one at a time (Behat will exit on the first such exception) but they should at least now be able to locate the issue.

As reported by @AlexSkrypnyk in https://github.com/Behat/Behat/pull/1614#issuecomment-3354127345

### DIFF
--- a/tests/Fixtures/Arguments/features/unexpected-pystring-exception.feature
+++ b/tests/Fixtures/Arguments/features/unexpected-pystring-exception.feature
@@ -1,0 +1,11 @@
+Feature: PystringNodes
+  Scenario:
+    Given a step with no argument
+      """
+      <word1>
+        w
+         o
+    r
+     <word2>
+         d
+      """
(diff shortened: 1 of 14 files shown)


tests/Fixtures/Arguments/features/unexpected-pystring-exception.feature
```diff
@@ -0,0 +1,11 @@
...
+  Scenario:
+    Given a step with no argument
+      """
+      <word1>
+        w
+         o
+    r
+     <word2>
+         d
+      """
```
stof: I suggest adding the fine LF in all those feature files.

acoulton: Good catch, thanks - I'd just captured them from the previous inline strings and hadn't noticed they were missing.

carlos-granados (approved):
@acoulton great stuff, thanks for fixing this (and converting the tests at the same time 😄 )

PR #1674, opened 2025-10-19
Implement JSON formatter

carlos-granados:
Lately JSON has become the lingua franca of tools, which use this format to exchange information everywhere. This PR introduces a JSON formatter which is able to save Behat test results as a JSON file. One possible application for this format is to be able to create an MCP server for Behat, using the generated JSON to return information to the client

This formatter is heavily inspired in the existing JUnit formatter and the exported output is similar but, since it is not constrained by the existing JUnit specification, it follows more closely the structure of our project. The generated JSON basically will contain a list of Suite nodes, which will contain Feature nodes, which will contain Scenario nodes. Each node will have properties indicating some info like the status, the number of tests for each status, file and line and execution time. Also, the JUnit formatter produced one file per suite, this formatter produces just a single JSON file for all suites so the `output_path` needs to be a file name, not a folder name

The PR also defines a JSON schema which can be used by the tools that will use this information

### DIFF
--- a/resources/schema.json
+++ b/resources/schema.json
@@ -11,7 +11,7 @@
         "failed": { "type": "integer", "minimum": 0 },
         "pending": { "type": "integer", "minimum": 0 },
         "undefined": { "type": "integer", "minimum": 0 },
-        "time": { "$ref": "#/$defs/TimeString" },
+        "time": { "$ref": "#/$defs/DurationSeconds" },
         "suites": {
             "type": "array",
             "items": { "$ref": "#/$defs/Suite" }
@@ -19,7 +19,7 @@
     },
 
     "$defs": {
-        "TimeString": {
+        "DurationSeconds": {
             "type": "number",
             "description": "Seconds rounded to 3 decimals (e.g., 0.041)."
         },
@@ -49,7 +49,7 @@
                 "failed": { "type": "integer", "minimum": 0 },
                 "pending": { "type": "integer", "minimum": 0 },
                 "undefined": { "type": "integer", "minimum": 0 },
-                "time": { "$ref": "#/$defs/TimeString" },
+                "time": { "$ref": "#/$defs/DurationSeconds" },
                 "features": {
                     "type": "array",
                     "items": { "$ref": "#/$defs/Feature" }
@@ -74,7 +74,7 @@
                 "failed": { "type": "integer", "minimum": 0 },
                 "pending": { "type": "integer", "minimum": 0 },
                 "undefined": { "type": "integer", "minimum": 0 },
-                "time": { "$ref": "#/$defs/TimeString" },
+                "time": { "$ref": "#/$defs/DurationSeconds" },
                 "scenarios": {
                     "type": "array",
                     "items": { "$ref": "#/$defs/Scenario" }
@@ -94,7 +94,7 @@
             "required": ["name", "time", "status", "file", "line"],
             "properties": {
                 "name": { "type": "string" },
-                "time": { "$ref": "#/$defs/TimeString" },
+                "time": { "$ref": "#/$defs/DurationSeconds" },
                 "status": {
                     "type": "string",
                     "enum": ["passed", "failed", "pending", "undefined", "skipped"]


composer.json
```diff
@@ -31,6 +31,7 @@
 
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^3.68",
+        "opis/json-schema": "^2.5",
```
carlos-granados, line 34: This is used to validate the generated JSON in tests

features/bootstrap/FeatureContext.php
```diff
@@ -421,6 +436,27 @@ private function checkXmlFileContents($path, PyStringNode $text)
...
 
+    private function checkJSONFileContents($path, PyStringNode $text)
+    {
+        Assert::assertFileExists($path);
+
+        $fileContent = trim(file_get_contents($path));
+
+        $fileContent = preg_replace('/"time": "\d\.\d{3}"/U', '"time": "-IGNORE-VALUE-"', $fileContent);
+
+        $directorySeparator = DIRECTORY_SEPARATOR;
```
carlos-granados: When generating the JSON that is saved into the file, in Windows the directory separators (`\`) will be escaped and the generated file will have `\\` in those places. We need to do the same when replacing the directory separator placeholder in our expected results

resources/schema.json
```diff
@@ -0,0 +1,125 @@
+{
```
carlos-granados, line 1: This is the JSON schema that describes the structure of the JSON that our formatter generates

features/convert_config.feature
```diff
@@ -407,6 +408,8 @@ Feature: Convert config
           ->withProfile((new Profile('with_options'))
               ->withFormatter((new JUnitFormatter())
                   ->withOutputPath('build/logs/junit'))
+              ->withFormatter((new JSONFormatter())
```
carlos-granados, line 411: Add this formatter to the tests that check our config converter

features/hooks_failures_json.feature
```diff
@@ -0,0 +1,854 @@
+Feature: Display hook failures location in json printer
```
carlos-granados, line 1: For all the other formatters we have tests that check the hook failures both for annotations and attributes. I don't think we really need to test these two variants for all formatters, being able to use annotations or attributes is not related to the formatter output. So we probably want to remove some of those tests. For now, I have only added the tests with attributes for this new formatter

src/Behat/Behat/Output/Node/EventListener/JSON/JSONDurationListener.php
```diff
@@ -0,0 +1,99 @@
...
+use Behat\Testwork\Counter\Timer;
+use Behat\Testwork\Event\Event;
+use Behat\Testwork\EventDispatcher\Event\AfterExerciseCompleted;
+use Behat\Testwork\EventDispatcher\Event\AfterExerciseSetup;
+use Behat\Testwork\EventDispatcher\Event\AfterSuiteTested;
+use Behat\Testwork\EventDispatcher\Event\BeforeSuiteTested;
+use Behat\Testwork\Output\Formatter;
+use Behat\Testwork\Suite\Suite;
+
+final class JSONDurationListener extends DurationListener
```
carlos-granados, line 23: We enhance the existing duration timer to save new timers for each suite and for the whole exercise

src/Behat/Behat/Output/Node/EventListener/JSON/JSONElementListener.php
```diff
@@ -0,0 +1,153 @@
...
+use Behat\Gherkin\Node\ScenarioLikeInterface;
+use Behat\Testwork\Event\Event;
+use Behat\Testwork\EventDispatcher\Event\AfterExerciseCompleted;
+use Behat\Testwork\EventDispatcher\Event\AfterExerciseSetup;
+use Behat\Testwork\EventDispatcher\Event\AfterSuiteSetup;
+use Behat\Testwork\EventDispatcher\Event\AfterSuiteTested;
+use Behat\Testwork\Output\Formatter;
+use Behat\Testwork\Output\Node\EventListener\EventListener;
+
+final class JSONElementListener implements EventListener
```
carlos-granados, line 34: This is the file that listen to all the events and builds the JSON file by calling the different printers at the appropriate times

src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitDurationListener.php
```diff
@@ -1,114 +1,17 @@
...
-use Behat\Gherkin\Node\KeywordNodeInterface;
-use Behat\Gherkin\Node\ScenarioLikeInterface;
-use Behat\Testwork\Counter\Timer;
-use Behat\Testwork\Event\Event;
-use Behat\Testwork\Output\Formatter;
-use Behat\Testwork\Output\Node\EventListener\EventListener;
+use Behat\Behat\Output\Node\EventListener\Statistics\DurationListener;
 
-final class JUnitDurationListener implements EventListener
+final class JUnitDurationListener extends DurationListener
```
carlos-granados, line 15: I have moved all the functionality in this class to a base DurationListener, so that it can be used and extended by the JSONDurationListener

src/Behat/Behat/Output/Node/EventListener/Statistics/DurationListener.php
```diff
@@ -0,0 +1,132 @@
...
+use Behat\Behat\EventDispatcher\Event\BeforeFeatureTested;
+use Behat\Behat\EventDispatcher\Event\BeforeScenarioTested;
+use Behat\Gherkin\Node\FeatureNode;
+use Behat\Gherkin\Node\ScenarioLikeInterface;
+use Behat\Testwork\Counter\Timer;
+use Behat\Testwork\Event\Event;
+use Behat\Testwork\Output\Formatter;
+use Behat\Testwork\Output\Node\EventListener\EventListener;
+
+class DurationListener implements EventListener
```
carlos-granados, line 24: This class just contains the existing functionality from the JUnitDurationListener so that it can be used and extended by the JSONDurationListener

src/Behat/Behat/Output/Node/Printer/ExercisePrinter.php
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Behat\Behat\Output\Node\Printer;
+
+use Behat\Testwork\Output\Formatter;
+
+interface ExercisePrinter
```
carlos-granados, line 15: Define a new interface similar to the existing FeaturePrinter, ScenarioPrinter, etc...

src/Behat/Behat/Output/Node/Printer/JSON/JSONExercisePrinter.php
```diff
@@ -0,0 +1,57 @@
...
+namespace Behat\Behat\Output\Node\Printer\JSON;
+
+use Behat\Behat\Output\Node\EventListener\JSON\JSONDurationListener;
+use Behat\Behat\Output\Node\Printer\ExercisePrinter;
+use Behat\Behat\Output\Statistics\PhaseStatistics;
+use Behat\Testwork\Output\Formatter;
+use Behat\Testwork\Output\Printer\JSONOutputPrinter;
+use Behat\Testwork\Tester\Result\TestResult;
+
+final class JSONExercisePrinter implements ExercisePrinter
```
carlos-granados, line 20: Define a series of printers which are called by the event listener and that will create the JSON file by calling the JSONOutputPrinter

src/Behat/Behat/Output/ServiceContainer/Formatter/JSONFormatterFactory.php
```diff
@@ -0,0 +1,190 @@
...
+use Behat\Testwork\Output\NodeEventListeningFormatter;
+use Behat\Testwork\Output\Printer\Factory\FileOutputFactory;
+use Behat\Testwork\Output\Printer\JSONOutputPrinter;
+use Behat\Testwork\Output\ServiceContainer\Formatter\FormatterFactory;
+use Behat\Testwork\Output\ServiceContainer\OutputExtension;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Reference;
+
+final class JSONFormatterFactory implements FormatterFactory
```
carlos-granados, line 38: This class creates all the services needed by the new functionality

src/Behat/Behat/Output/ServiceContainer/Formatter/JSONFormatterFactory.php
```diff
@@ -0,0 +1,190 @@
...
+        $container->setDefinition('output.json.statistics.feature', $definition);
+
+        $definition = new Definition(NodeEventListeningFormatter::class, [
+            JSONFormatter::NAME,
+            'Outputs the failures in JSON files.',
+            JSONFormatter::defaults(),
+            $this->createOutputPrinterDefinition(),
+            new Definition(ChainEventListener::class, [
+                [
+                    new Reference(self::ROOT_LISTENER_ID),
```
carlos-granados, line 143: We need to call the ScenarioStatsListener, StepStatsListener and HookStatsListener three times because we have three sets of statistics: per feature, per suite and per exercise

src/Behat/Config/Formatter/JSONFormatter.php
```diff
@@ -0,0 +1,33 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Behat\Config\Formatter;
+
+use PhpParser\Node\Expr;
+
+final class JSONFormatter extends Formatter
```
carlos-granados, line 17: Config file for the JSON formatter

src/Behat/Testwork/Output/Printer/Factory/FileOutputFactory.php
```diff
@@ -0,0 +1,36 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Behat\Testwork\Output\Printer\Factory;
+
+use Behat\Testwork\Output\Exception\MissingOutputPathException;
+use Symfony\Component\Console\Output\StreamOutput;
+
+final class FileOutputFactory extends OutputFactory
```
carlos-granados, line 17: This is an output factory which creates an output stream directed to a single file

src/Behat/Testwork/Output/Printer/JSONOutputPrinter.php
```diff
@@ -0,0 +1,206 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Behat\Testwork\Output\Printer;
+
+use Behat\Testwork\Output\Exception\MissingOutputPathException;
+use Behat\Testwork\Output\Printer\Factory\FileOutputFactory;
+use Symfony\Component\Console\Output\OutputInterface;
+
+final class JSONOutputPrinter extends StreamOutputPrinter
```
carlos-granados, line 17: This is the class that generates the actual JSON, being called at the appropriate times by the different printers. It does this by generating an array of all the nodes and then converting this array to json before saving it to a file

src/Behat/Testwork/Output/Printer/JSONOutputPrinter.php
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        parent::flush();
+    }
+
+    /**
+     * @param array<string, string|int|array> $node
+     * @param array<string, string|int> $attributes
+     */
+    private function addAttributesToNode(array &$node, array $attributes, bool $atStart = false): void
```
carlos-granados, line 202: This function has a flag which can be used to add the new attributes at the beginning of the node or at the end, this is useful to construct a json representation which is readable, with single properties at the beginning of the node and arrays at the end

resources/schema.json
```diff
@@ -0,0 +1,125 @@
...
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
```
stof, line 2: This schema should probably not be inside the `features` folder. It makes sense to ship it in Behat releases for others to use it. The `features` folder is excluded from dist archives as it is used when contributing to Behat.

tests/Fixtures/TestReportFormat/behat.php
carlos-granados, line 1: The fixtures for the JSON formatter and the JUnit formatter were exactly the same, so I moved them to a "TestReport" folder so that they could be shared by the tests of these two formatters

features/bootstrap/schema/schema.json
```diff
@@ -0,0 +1,125 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "https://example.com/schemas/test-report.schema.json",
```
stof: should this use `behat.org` instead of `example.com` ?

features/bootstrap/schema/schema.json
```diff
@@ -0,0 +1,125 @@
...
+        "suites": {
+            "type": "array",
+            "items": { "$ref": "#/$defs/Suite" }
+        }
+    },
+
+    "$defs": {
+        "TimeString": {
+            "type": "string",
+            "description": "Seconds as a string with optional decimals (e.g., '0.041').",
```
stof: Why making a string rather than a float ?

features/bootstrap/schema/schema.json
```diff
@@ -0,0 +1,125 @@
...
+            "pattern": "^[0-9]+(\\.[0-9]+)?$"
+        },
+
+        "FailureEntry": {
+            "type": "object",
+            "additionalProperties": false,
+            "required": ["message"],
```
stof: shouldn't the `type` property be required as well ?

features/bootstrap/schema/schema.json
```diff
@@ -0,0 +1,125 @@
...
+                }
+            ]
+        }
+    }
+}
+
+
+
+
+
```
stof: those extra lines should be removed

resources/schema.json
carlos-granados: @stof you are absolutely right. Where do you think we should store this file? 🤔

stof: I would suggest a `resources` folder at the root

features/bootstrap/FeatureContext.php
acoulton: It might be simpler/more obvious to just do `$text = str_replace('-DIRECTORY-SEPARATOR-', json_encode(DIRECTORY_SEPARATOR), $text)` rather than concatenating manually (which would have taken me a moment to understand without your comment)?

features/bootstrap/FeatureContext.php
```diff
@@ -421,6 +436,27 @@ private function checkXmlFileContents($path, PyStringNode $text)
...
+        if (DIRECTORY_SEPARATOR === '\\') {
+            $directorySeparator .= DIRECTORY_SEPARATOR;
+        }
+        $text = str_replace('-DIRECTORY-SEPARATOR-', $directorySeparator, $text);
+
+        $data = json_decode((string) $fileContent, true);
```
acoulton: Possibly add the JSON_THROW_ON_ERROR here for clearer failure output if the json itself is broken somehow?

features/bootstrap/schema/schema.json
acoulton: I'd argue that time is a value with a fixed precision depending on the instrument used to measure it. It's not an arbitrary floating point value. IMO it's inefficient / unclear for the JSON to encode some durations as very long floats when in fact the extended precision is irrelevant/meaningless.

However in that case we should make the decimals required and consistent at either milliseconds (.000) or micros (.000000).

features/hooks_failures_json.feature
```diff
@@ -0,0 +1,854 @@
...
+  In order to be able to locate the code that generated a failure
+  As a feature developer using the json printer
+  When a hook throws an error I want to see the related item where the code failed
+
+  Background:
+    Given a file named "features/one.feature" with:
```
acoulton, line 7: Out of curiosity, why inline these feature & context files rather than creating a fixture directory?

features/json_format.feature
```diff
@@ -0,0 +1,504 @@
...
+                          "name": "Use multiline titles",
+                          "tests": 2,
+                          "skipped": 0,
+                          "failed": 0,
+                          "pending": 0,
+                          "undefined": 0,
+                          "time": "-IGNORE-VALUE-",
+                          "scenarios": [
+                              {
+                                  "name": "Adding some interesting value",
```
acoulton, line 263: Question: The multiline titles in the fixture are somewhat arbitrary with random linebreaks mid-sentence, so it looks natural to merge them to a single line.

However in practice if a feature author has deliberately split their scenario title into multiple lines, shouldn't we preserve the newlines in the output?

features/json_format.feature
```diff
@@ -0,0 +1,504 @@
...
+                                  "line": 25,
+                                  "failures": [
+                                      {
+                                          "message": "Then I must have 13: Failed asserting that 14 matches expected '13'.",
+                                          "type": "failed"
+                                      }
+                                  ]
+                              },
+                              {
+                                  "name": "Passed & Failed #1",
```
acoulton, line 105: Probably better to use the newer title handling for ExampleNode scenarios from the outset to avoid recreating #1448 for this formatter (we've done the work in Behat/Gherkin already, it's just not yet been applied to the JUnit formatter here). Will mark the implementation below.

src/Behat/Behat/Output/Node/EventListener/JSON/JSONElementListener.php
```diff
@@ -0,0 +1,153 @@
...
+            $this->setupPrinter->printTeardown($formatter, $event->getTeardown());
+            $this->featurePrinter->printFooter($formatter, $event->getTestResult());
+        }
+    }
+
+    private function onOutlineStart(Event $event): void
+    {
+        if ($event instanceof BeforeOutlineTested) {
+            $outline = $event->getOutline();
+            assert($this->scenarioPrinter instanceof JSONScenarioPrinter);
```
acoulton: If this class only expects to be used with a `JSONScenarioPrinter` wouldn't it be clearer to just make that strict in the constructor rather than a runtime assertion?

src/Behat/Behat/Output/Node/Printer/JSON/JSONScenarioPrinter.php
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    public function printFooter(Formatter $formatter, TestResult $result): void
+    {
+        $scenario = $this->currentScenario;
+        $name = implode(' ', array_map(fn ($l) => trim($l), explode("\n", $scenario->getTitle() ?? '')));
+
+        if ($scenario instanceof ExampleNode) {
+            $name = $this->buildExampleName($scenario);
+        }
```
acoulton: Here's where we can use the new logic from Behat/Gherkin#271 that was added to solve #1448 for JUnit (but not yet adopted by our formatter). 

At runtime, the `$scenario` passed to this printer should always be an instance of `NamedScenarioInterface` so you can just call `$scenario->getName()` and remove the special handling of `ExampleNode`.

src/Behat/Behat/Output/Node/Printer/JSON/JSONScenarioPrinter.php
```diff
@@ -0,0 +1,105 @@
...
+            'status' => $this->resultConverter->convertResultToString($result),
+        ];
+
+        $file = $this->currentFeature->getFile();
+        if ($file) {
+            $cwd = realpath(getcwd());
+            $scenarioAttributes['file'] =
+                str_starts_with($file, $cwd) ?
+                    ltrim(substr($file, strlen($cwd)), DIRECTORY_SEPARATOR) : $file;
+        }
```
acoulton, line 65: Should we perhaps support relative / absolute paths here the same as other formatters, rather than internally guessing based on the cwd (which may not be consistent with `%paths.base%` that everything else uses)? There's an open PR / issue around that for the JUnit printer in #1601.

src/Behat/Testwork/Output/Printer/Factory/FileOutputFactory.php
```diff
@@ -0,0 +1,36 @@
...
+        if ($this->getOutputPath() === null) {
+            throw new MissingOutputPathException(
+                'The `output_path` option must be specified.',
+            );
+        }
+
+        $fileName = $this->getOutputPath();
+
+        $stream = new StreamOutput(
+            fopen($fileName, 'w'),
```
acoulton: Should we handle potential failures to open the file (e.g. already exists and is a directory / permissions failures) - it looks like otherwise this will just fail with a 'The StreamOutput class needs a stream as its first argument.' from passing `false` into the `StreamOutput` constructor?

Also - potentially we should create any parent directories that may be required?

acoulton (commented):
This is brilliant, thanks for all the work on this @carlos-granados 

A few small comments / questions that I spotted when going through it, but overall it's in great shape.

features/json_format.feature
stof: In standard-compliant Gherkin parsing (see the compatibility mode we are adding in `behat/gherkin`, the scenario title is never multi-line. A line break ends the scenario title and starts the description.

acoulton: Good point @stof I had momentarily forgotten that. In that case let's leave this as-is (and equivalent to the JSON formatter).

I need to get back to the gherkin compatibility stuff, once people can opt in to parsing titles & descriptions separately we can look at exposing descriptions in formatters as appropriate.

stof: Maybe the new json formatter should already support description, by splitting the title on the first line break (and then adapting that logic once Gherkin is parsed the new compatibility mode)

features/hooks_failures_json.feature
carlos-granados: I believe that this test can probably share the fixtures with the other tests for the hook failures for the other formatters. But that would require changing those tests to use fixtures, so I would prefer to do that as a separate PR

src/Behat/Behat/Output/Node/Printer/JSON/JSONScenarioPrinter.php
carlos-granados: Much better, this allows us to remove all our custom handling for scenario outlines

src/Behat/Behat/Output/Node/Printer/JSON/JSONScenarioPrinter.php
carlos-granados: I wanted to use the ConfigurablePathPrinter that we introduced for the pretty and progress formatters but I would like to do it properly, adding tests, etc... So again I would prefer to do this separately in another PR and change both the JUnit formatter and this one at the same time

carlos-granados:
@acoulton updated, thanks for all your suggestions. I also changed a few properties which I saw had been defined as unnecessarily nullable. Please let me know if it looks ok now

src/Behat/Behat/Output/Node/Printer/JSON/JSONScenarioPrinter.php
acoulton: That makes sense - my only thought is as this is a new machine-readable format, is there potentially an issue if it ships first with the paths relative to `cwd()` and then we subsequently change that behaviour?

The cwd may not be the same as the base path, and we don't support that as an option in ConfigurablePathPrinter (nor should we) so there wouldn't even be a way to configure it to keep the initial behaviour if users have built something that consumes the JSON and relies on finding their files from it...?

IMO the options for the ConfigurablePathPrinter are separate from the formatter options, so there may be an argument the features for the individual formatters don't need to cover every possible variation. Even if we want to add them, those tests could be added as part of fixing the JUnit formatter but we could still use the ConfigurablePathPrinter here from the start?

resources/schema.json
```diff
@@ -0,0 +1,118 @@
...
+        "undefined": { "type": "integer", "minimum": 0 },
+        "time": { "$ref": "#/$defs/TimeString" },
+        "suites": {
+            "type": "array",
+            "items": { "$ref": "#/$defs/Suite" }
+        }
+    },
+
+    "$defs": {
+        "TimeString": {
```
acoulton: I guess `TimeString` is not a great name now that this is a float? Perhaps `ElapsedTime` or `DurationSeconds`?

src/Behat/Behat/Output/Node/Printer/JSON/JSONScenarioPrinter.php
carlos-granados: I plan to work on this separate PR pretty soon, should be my next task, so as long as we don't do a release before the other PR is ready, we should be safe

acoulton (commented):
Great @carlos-granados just one small naming thing in the updated schema and (~the discussion about paths.~ now resolved)

src/Behat/Behat/Output/Node/Printer/JSON/JSONScenarioPrinter.php
acoulton: Fair enough, that sounds good.

resources/schema.json
carlos-granados: Ah, yes, of course

acoulton (approved):
Brilliant, great work @carlos-granados thank you :)

Issue #1663, opened 2025-09-18
Set error_reporting level when running tests


carlos-granados:
There are some tests in our test suite that expect that notices or deprecation warnings are shown. But if you locally have set `error_reporting` to something other than E_ALL, these notices or warnings may not be shown and the tests would fail. We should set `error_reporting` to E_ALL before attempting to run any test in our test suite

carlos-granados:
> Interesting - in the unit tests or the behat features? I thought behat always set error_reporting to E_ALL when running steps unless you customise it in the testerOptions config?

I saw it in some Behat tests. Would need to check

acoulton:
Maybe it doesn't work quite how I thought... I can maybe take a look next week.

PR #1675, opened 2025-10-21
Fix setting error_reporting

carlos-granados:
When the error reporting level was set (either by using `withErrorReporting()` or by using the default value of E_ALL), we were calling `set_error_handler()` with this level but were not calling `error_reporting()`. This meant that event though we had set a handler to be called on certain error levels, this handler might not be called if the error_reporting level set in php.ini was more restrictive. I saw this in a system where error_reporting was set to `E_ALL & ~E_DEPRECATED & ~E_STRICT` and the test `With default error reporting reports all PHP errors and deprecations` failed because the deprecation was not raised

This PR fixes this problem by actually setting the error_reporting level, not just using it in the set_error_handler call

Fixes https://github.com/Behat/Behat/issues/1663

### DIFF
--- a/src/Behat/Testwork/Call/Handler/RuntimeCallHandler.php
+++ b/src/Behat/Testwork/Call/Handler/RuntimeCallHandler.php
@@ -40,7 +40,7 @@ final class RuntimeCallHandler implements CallHandler
      * @param int $errorReportingLevel
      */
     public function __construct(
-        private $errorReportingLevel,
+        private $errorReportingLevel = E_ALL,
     ) {
         $this->validator = new Validator();
     }


features/error_reporting.feature
```diff
@@ -95,6 +95,6 @@ Feature: Error Reporting
           When an exception is thrown # features/exception_in_scenario.feature:7
             Exception: Exception is thrown in features/bootstrap/FeatureContext.php:47
             Stack trace:
-            #0 {BASE_PATH}src/Behat/Testwork/Call/Handler/RuntimeCallHandler.php(98): FeatureContext->anExceptionIsThrown()
-            #1 {BASE_PATH}src/Behat/Testwork/Call/Handler/RuntimeCallHandler.php(54): Behat\Testwork\Call\Handler\RuntimeCallHandler->executeCall(
+            #0 {BASE_PATH}src/Behat/Testwork/Call/Handler/RuntimeCallHandler.php(100): FeatureContext->anExceptionIsThrown()
```
carlos-granados, line 123: Need to change this test because the line numbers in that file have changed

src/Behat/Testwork/Call/Handler/RuntimeCallHandler.php
```diff
@@ -121,6 +123,8 @@ private function getBufferedStdOut()
     private function startErrorAndOutputBuffering(Call $call)
     {
         $errorReporting = $call->getErrorReportingLevel() ?: $this->errorReportingLevel;
+        $this->previousErrorReporting = error_reporting();
+        error_reporting($errorReporting);
```
acoulton: Couldn't this just be a single call to error_reporting?

carlos-granados: ah, yes, of course

carlos-granados:
@acoulton updated with your suggestion and tests modified to check all needed cases

src/Behat/Testwork/Call/Handler/RuntimeCallHandler.php
```diff
@@ -32,13 +32,15 @@ final class RuntimeCallHandler implements CallHandler
...
+    private ?int $previousErrorReporting = null;
+
     /**
      * Initializes executor.
      *
      * @param int $errorReportingLevel
      */
     public function __construct(
-        private $errorReportingLevel = E_ALL,
+        private $errorReportingLevel,
```
acoulton: It could technically be a BC break to make this parameter non-optional?

I *think* we could safely leave it with the default E_ALL, if we know that in practice in our usage we're always providing a value?

carlos-granados: Yeah, you are right, we don't really need to remove this default value to get this working. Restored

PR #1677, opened 2025-10-23
Use configurable path printer in JSON and JUnit formatters

carlos-granados:
The JSON and JUnit formatters were using a very simple management of the printed file paths. This PR replaces this with the ConfigurablePathPrinter that the other formatter use. This brings a couple of advantages:
- It uses the `%paths.base%` to calculate relative paths, this is more correct than using cwd as those formatters were doing
- It allows us to use the options offered by that printer

The editorUrl option of that printer is not appropriate for these formatters, which cannot use the link added to the paths, so I added a flag to mark if this option needs to be applied or not

Adds tests that check the correct working of the printer options in these formatters

### DIFF
--- a/src/Behat/Behat/Output/Node/Printer/JSON/JSONFeaturePrinter.php
+++ b/src/Behat/Behat/Output/Node/Printer/JSON/JSONFeaturePrinter.php
@@ -47,7 +47,7 @@ final class JSONFeaturePrinter implements FeaturePrinter
         assert($outputPrinter instanceof JSONOutputPrinter);
 
         $outputPrinter->extendFeatureAttributes([
-            'name' => $this->currentFeature->getTitle(),
+            'name' => $this->currentFeature->getTitle() ?? '',
             'tests' => $totalCount,
             'skipped' => $stats[TestResult::SKIPPED],
             'failed' => $stats[TestResult::FAILED],
--- a/src/Behat/Behat/Output/Node/Printer/JUnit/JUnitScenarioPrinter.php
+++ b/src/Behat/Behat/Output/Node/Printer/JUnit/JUnitScenarioPrinter.php
@@ -19,6 +19,7 @@ use Behat\Gherkin\Node\OutlineNode;
 use Behat\Gherkin\Node\ScenarioLikeInterface;
 use Behat\Testwork\Output\Formatter;
 use Behat\Testwork\Output\Printer\JUnitOutputPrinter;
+use Behat\Testwork\PathOptions\Printer\ConfigurablePathPrinter;
 use Behat\Testwork\Tester\Result\TestResult;
 
 /**
@@ -42,6 +43,7 @@ final class JUnitScenarioPrinter
         private readonly ResultToStringConverter $resultConverter,
         private readonly JUnitOutlineStoreListener $outlineStoreListener,
         private readonly ?JUnitDurationListener $durationListener = null,
+        private readonly ?ConfigurablePathPrinter $configurablePathPrinter = null,
     ) {
     }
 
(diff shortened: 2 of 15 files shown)


src/Behat/Behat/Output/Node/Printer/JSON/JSONFeaturePrinter.php
```diff
@@ -47,7 +47,7 @@ public function printFooter(Formatter $formatter, TestResult $result)
         assert($outputPrinter instanceof JSONOutputPrinter);
 
         $outputPrinter->extendFeatureAttributes([
-            'name' => $this->currentFeature->getTitle(),
+            'name' => $this->currentFeature->getTitle() ?? '',
```
carlos-granados, line 50: I noticed that this can be null and if that is the case we need to print an empty string

src/Behat/Behat/Output/Node/Printer/JUnit/JUnitScenarioPrinter.php
```diff
@@ -42,6 +43,7 @@ public function __construct(
         private readonly ResultToStringConverter $resultConverter,
         private readonly JUnitOutlineStoreListener $outlineStoreListener,
         private readonly ?JUnitDurationListener $durationListener = null,
+        private readonly ?ConfigurablePathPrinter $configurablePathPrinter = null,
```
carlos-granados, line 46: This needs to be made nullable to keep BC

src/Behat/Behat/Output/Node/Printer/JSON/JSONFeaturePrinter.php
acoulton: Or I suppose we could make the `name` element nullable in the JSON schema? 

But I think you're probably right that it's simpler for end-users if we coalesce to an empty string - I doubt there's a usecase for being able to explicitly tell whether a feature name was provided..

src/Behat/Behat/Output/Node/Printer/JUnit/JUnitScenarioPrinter.php
```diff
@@ -63,11 +65,11 @@ public function printOpenTag(Formatter $formatter, FeatureNode $feature, Scenari
             'time' => $this->durationListener instanceof JUnitDurationListener ? $this->durationListener->getDuration($scenario) : '',
         ];
 
-        if ($file) {
-            $cwd = realpath(getcwd());
-            $testCaseAttributes['file'] =
-                str_starts_with($file, $cwd) ?
-                    ltrim(substr($file, strlen($cwd)), DIRECTORY_SEPARATOR) : $file;
+        if ($file && $this->configurablePathPrinter instanceof ConfigurablePathPrinter) {
```
acoulton, line 68: So if someone has somehow created a JUnitScenarioPrinter outside our DI container (without the path printer) we will just stop including a `file` attribute in the junit. 

I think that's unlikely to happen, and the `file` attribute was already optional so the report is still schema-valid, so I think this is OK.

acoulton (approved):
Great work, thanks @carlos-granados

acoulton:
We'll need to note in the changelog that the junit `file` paths by default will now be relative to `%paths.base%` rather than `cwd` - in most cases I'd expect that to be the same but it could produce a diff for people who are using tooling to aggregate/compare junit reports across multiple runs.

src/Behat/Behat/Output/Node/Printer/JUnit/JUnitScenarioPrinter.php
carlos-granados: Yeah, that was my thinking

PR #1678, opened 2025-10-23
Fix php version constraint in composer.json

carlos-granados:
The latest version of php-cs-fixer showed a warning saying
"Unable to determine minimum supported PHP version from composer.json: Invalid version string "8.1.*""

It seems that, though composer was able to understand our constraint, php-cs-fixer wasn't

I updated it to use a new constraint that should be understood by everyone

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     ],
 
     "require": {
-        "php": "8.1.* || 8.2.* || 8.3.* || 8.4.* ",
+        "php": ">=8.1 <8.5",
         "ext-mbstring": "*",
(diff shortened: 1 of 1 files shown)


stof (approved):

acoulton:
Thanks @carlos-granados 

> I suggest also opening an issue on php-cs-fixer so they might be aware of this.

I agree with @stof, it does feel like a php-cs-fixer bug if they can't parse a valid composer constraint. That said I think the way you've expressed the constraint is much clearer to read anyway so it's a good change to have.

PR #1682, opened 2025-10-27
fix: Include consistent index & placeholders for Example names in junit

### LINKED ISSUE #1448: [Bug?] Reruns of Scenario outlines lead to potentially confusing scenario names

acoulton:
Previously, when a feature file contained scenarios with `Example` nodes the junit report would append an index to each testcase name to differentiate them (e.g. `<testcase name="Some scenario outline #1"`, `<testcase name="Some scenario outline #2`, etc).

These indices were dynamically generated based on the examples that actually ran. As a result, on a `--rerun` the names would start from `#1` in sequence and be inconsistent with the numbering of the first execution. This makes it difficult to compare reports between each test run, for example to link a passing rerun to the previous failure.

With this change, examples are numbered consistently based on the order they appear in the feature file. This fixes #1448.

Additionally, we also now support placeholders in the Scenario Outline title. This allows users to include some or all of the values from the Examples: table in each `<testcase` name for additional clarity. This matches the behaviour of the official cucumber runners.

This PR builds on work done by @uuf6429 in #1459 and in particular the implementation of the numbering & placeholder replacement that he contributed in Behat/Gherkin#270 and Behat/Gherkin#271.

### DIFF
--- a/features/json_format.feature
+++ b/features/json_format.feature
@@ -103,7 +103,7 @@ Feature: JSON Formatter
                                   ]
                               },
                               {
-                                  "name": "Passed & Failed #1",
+                                  "name": "Passed & Failed with value=5 #1",
                                   "time": -IGNORE-VALUE-,
                                   "status": "failed",
                                   "file": "features-DIRECTORY-SEPARATOR-single_feature.feature",
@@ -116,14 +116,14 @@ Feature: JSON Formatter
                                   ]
                               },
                               {
-                                  "name": "Passed & Failed #2",
+                                  "name": "Passed & Failed with value=10 #2",
                                   "time": -IGNORE-VALUE-,
                                   "status": "passed",
                                   "file": "features-DIRECTORY-SEPARATOR-single_feature.feature",
                                   "line": 29
                               },
                               {
-                                  "name": "Passed & Failed #3",
+                                  "name": "Passed & Failed with value=23 #3",
                                   "time": -IGNORE-VALUE-,
                                   "status": "failed",
                                   "file": "features-DIRECTORY-SEPARATOR-single_feature.feature",
@@ -136,14 +136,14 @@ Feature: JSON Formatter
                                   ]
                               },
                               {
-                                  "name": "Another Outline #1",
+                                  "name": "Another Outline (5 = 15) #1",
                                   "time": -IGNORE-VALUE-,
                                   "status": "passed",
                                   "file": "features-DIRECTORY-SEPARATOR-single_feature.feature",
                                   "line": 39
                               },
                               {
-                                  "name": "Another Outline #2",
+                                  "name": "Another Outline (10 = 20) #2",
                                   "time": -IGNORE-VALUE-,
                                   "status": "passed",
                                   "file": "features-DIRECTORY-SEPARATOR-single_feature.feature",
--- a/features/junit_format.feature
+++ b/features/junit_format.feature
@@ -42,15 +42,15 @@ Feature: JUnit Formatter
           <testcase name="Failed" classname="Adding numbers" status="failed" time="-IGNORE-VALUE-" file="features-DIRECTORY-SEPARATOR-single_feature.feature" line="25">
             <failure message="Then I must have 13: Failed asserting that 14 matches expected '13'."/>
           </testcase>
-          <testcase name="Passed &amp; Failed #1" classname="Adding numbers" status="failed" time="-IGNORE-VALUE-" file="features-DIRECTORY-SEPARATOR-single_feature.feature" line="29">
+          <testcase name="Passed &amp; Failed with value=5 #1" classname="Adding numbers" status="failed" time="-IGNORE-VALUE-" file="features-DIRECTORY-SEPARATOR-single_feature.feature" line="29">
             <failure message="Then I must have 16: Failed asserting that 15 matches expected '16'."/>
           </testcase>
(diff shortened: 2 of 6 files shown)


features/json_format.feature
```diff
@@ -103,7 +103,7 @@ Feature: JSON Formatter
                                   ]
                               },
                               {
-                                  "name": "Passed & Failed #1",
+                                  "name": "Passed & Failed with value=5 #1",
```
acoulton, line 107: There are no changes to the JSON implementation, but I have added placeholders to the fixture which is shared between the JSON and junit formatters.

features/junit_format.feature
```diff
@@ -42,15 +42,15 @@ Feature: JUnit Formatter
           <testcase name="Failed" classname="Adding numbers" status="failed" time="-IGNORE-VALUE-" file="features-DIRECTORY-SEPARATOR-single_feature.feature" line="25">
             <failure message="Then I must have 13: Failed asserting that 14 matches expected '13'."/>
           </testcase>
-          <testcase name="Passed &amp; Failed #1" classname="Adding numbers" status="failed" time="-IGNORE-VALUE-" file="features-DIRECTORY-SEPARATOR-single_feature.feature" line="29">
+          <testcase name="Passed &amp; Failed with value=5 #1" classname="Adding numbers" status="failed" time="-IGNORE-VALUE-" file="features-DIRECTORY-SEPARATOR-single_feature.feature" line="29">
```
acoulton, line 46: I have not explicitly added a test to prove that the `#1`, `#2` etc are in fact consistent between runs. 

However, we're proving the placeholders appear in the output, and we know that comes from the implementation in Behat/Gherkin that also appends the index as it parses the gherkin file - and that is tested in that project. I think that provides sufficient confidence that the rerun will now be numbered correclty.

phpstan.dist.neon
```diff
@@ -12,6 +12,7 @@ parameters:
             paths:
                 - src/Behat/Behat/Definition/Printer/ConsoleDefinitionPrinter.php
                 - src/Behat/Behat/EventDispatcher/Cli/StopOnFailureController.php
+                - src/Behat/Behat/Output/Node/Printer/JUnit/JUnitScenarioPrinter.php
```
acoulton, line 15: The constructor now has an unused param that we need to keep for BC

src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitOutlineStoreListener.php
```diff
@@ -29,6 +29,8 @@ final class JUnitOutlineStoreListener implements EventListener
 {
     /**
      * @var array
+     *
+     * @deprecated This will be removed in the next major as the JUnit formatter no longer uses this information
```
carlos-granados: Do we really need to deprecate private properties/methods? I would say we just need to deprecate public/protected ones

carlos-granados (commented):
Looks great, good job @acoulton just a question

src/Behat/Behat/Output/Node/EventListener/JUnit/JUnitOutlineStoreListener.php
stof: private methods can be deleted directly, as they cannot be used from the outside.

acoulton: Good point @carlos-granados I got trigger happy marking things to remind us to delete them. It was partly to record that I had reviewed for any other uses or side-effects of that behaviour and confirmed it will be safe to delete once the public `->getCurrentOutline()` is gone. I felt it also marked that we're expecting to remove these, so discourages anyone from adding any new (internal) feature that uses them.

@stof I deprecated rather than deleted the private method because we need to keep it as-is for now - it collects the state that is later used by the (deprecated and now unused by us) `->getCurrentOutline()` public method.

If the comments make the diff / codebase too noisy then I can happily remove them and we can just work out what to delete if/when we get to the point of deleting the `->getCurrentOutline()` method.

What do you think?

carlos-granados: I still feel we should only deprecate the public interface, I believe that should be enough to see what needs to be removed

acoulton: No problem, done now.

carlos-granados (approved):
Great stuff, thanks @acoulton

timhunt:
I just wanted to add a thank-you, for making this improvment that fixed the issue #1448 that I raised. :-)

PR #1685, opened 2025-10-29
chore: Update .gitattributes file

carlos-granados:
- Adds four files which should only be used for Behat development and don't need to be exported: `.php-cs-fixer.dist.php`, `AGENTS.md`, `phpstan.dist.neon` and `rector.php`
- Removes two files which don't seem to exist any longer `.scrutinizer.yml` and `appveyor.yml`

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,10 +1,15 @@
 /features export-ignore
 /tests export-ignore
+.editorconfig export-ignore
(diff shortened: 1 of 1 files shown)


carlos-granados:
@acoulton I added the first three. `GitHub_Lockup_Dark.svg` is referenced from `README.md`, so I think it is best if we keep it

PR #1686, opened 2025-11-04
fix: Exit with correct exit code on sigint

andrewnicols:
Exit codes for fatal signals is typically 128 + the signal number.

For SIGINT, whose number is 2, it should be 130.

https://www.gnu.org/software/bash/manual/html_node/Exit-Status.html

### DIFF
--- a/src/Behat/Testwork/EventDispatcher/Cli/SigintController.php
+++ b/src/Behat/Testwork/EventDispatcher/Cli/SigintController.php
@@ -51,7 +51,7 @@ final class SigintController implements Controller
     /**
      * Dispatches AFTER exercise event and exits program.
      */
-    public function abortExercise(int $signal): never
+    public function abortExercise(int $signal = SIGINT): never
     {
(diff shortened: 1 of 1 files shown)


andrewnicols:
Hitting Ctrl+C shoudl give an exit code of 130.

src/Behat/Testwork/EventDispatcher/Cli/SigintController.php
```diff
@@ -51,10 +51,10 @@ public function execute(InputInterface $input, OutputInterface $output)
     /**
      * Dispatches AFTER exercise event and exits program.
      */
-    public function abortExercise(): never
+    public function abortExercise(int $signal): never
```
acoulton: Obviously this function is only intended to be called by our signal handler (and is only public because it predates private/protected callables).

However technically it's a public method so I think we should have a default argument to avoid breaking BC if for some reason someone is calling it externally. Since we only actually register it for SIGINT, that seems like a reasonable default.

```suggestion
    public function abortExercise(int $signal = SIGINT): never
```

acoulton (approved):

Issue #1316, opened 2020-07-04
Favor composition over inheritance


rpkamp:
Instead of extending several Symfony classes, we should favour composition instead. For example, for the dispatcher it could be something like:

```
class TestworkEventDispatcher
{
    private $eventDispatcher;

    public function __construct()
    {
        $this->eventDispatcher = new Symfony\Component\Event\EventDispatcher();
    }

    public function addListener($eventName, $event)
    {
        $this->eventDispatcher->addListener($eventName, $event);
    }

    public function dispatch($event)
    {
        $this->dispatcher->dispatch($event);
    }
}
```

This has the advantage that the Symfony Event dispatcher truly becomes an implementation detail, and can be swapped at any moment if it doesn't serve its purpose anymore. Also, it would prevent situations like we just had where a deprecation in the Symfony Dispatcher caused all classes that used the EventDispatcher to change as well.
Had there been encapsulation instead of inheritance the `TestworkEventDispatcher` could have done the BC layer and the rest of the code wouldn't need to be touched.

This is a BC break, since some extensions may depend on the concrete implementation of Symfony. So if we do this there should be a nice upgrade path.

ciaranmcnulty:
We now have a `4.x` branch where we can propose BC breaks

stof:
For the event dispatcher, maybe we should use the PSR-14 interface as type instead of creating our own interface in Behat.

acoulton:
tl;dr - This is too big to be in scope for the 4.0 release in particular because of the impact on extension authors (which will further delay getting real-world upgrades to 4.0). We should definitely consider it for a future major.

That said, there is one dependency we can fairly easily solve - our `Extension` interface should not depend on the symfony `CompilerPassInterface`. This is the first place where we & extension authors are forced into potential BC breaks to support new symfony majors (even if the extension itself never interacts with the DI container). #1796 fixes that so that the `Extension` interface is entirely in our control and we use composition to wire the extension into the DI container.

For the others:

* `symfony/config` - extensions may have a dependency on this if they implement the `Extension::configure()` method to add custom configuration. There's no PSR interface for config-builders, and writing our own generic interface/bridge layer would be fairly complex given the chaining & nesting of calls. It would also be a significant manual job for extension authors to migrate to that.
[...]

PR #1687, opened 2025-11-04
[4.x] feat: Add Symfony 8 support

Kocal:
Related to https://github.com/FriendsOfBehat/SymfonyExtension/pull/218

This PR increase version constraints on `symfony/*` packages to allow Symfony 8, which will be released this month.

~~The CI fails because no version of PHP-CS-Fixer is compatible with Symfony 8, I'm on it.~~
EDIT: switched to `php-cs-fixer/shim` package

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,12 +51,18 @@ jobs:
           - php: 8.2
             os: ubuntu-latest
             composer-mode: update
-            symfony-version: '5.4'
+            symfony-version: '5.4.*'
 
           - php: 8.2
             os: ubuntu-latest
             composer-mode: update
-            symfony-version: '6.4'
+            symfony-version: '6.4.*'
+
+          - php: 8.4
+            os: ubuntu-latest
+            composer-mode: update
+            composer-minimum-stability: RC
+            symfony-version: '8.0.*'
 
     steps:
       - uses: actions/checkout@v4
@@ -69,6 +75,10 @@ jobs:
           ini-values: "phar.readonly=0,zend.exception_ignore_args=Off"
           coverage: none
 
+      - name: Configure Composer minimum stability
+        if: matrix.composer-minimum-stability != ''
+        run: composer config minimum-stability ${{ matrix.composer-minimum-stability }}
+
       - name: Install symfony/flex
         if: matrix.symfony-version != ''
         run: |
@@ -78,7 +88,7 @@ jobs:
       - name: Install latest dependencies
         if: matrix.composer-mode == 'update'
         env:
-          SYMFONY_REQUIRE: ${{ matrix.symfony-version }}.*
+          SYMFONY_REQUIRE: ${{ matrix.symfony-version }}
         run: composer update ${{ matrix.php == '8.5' && '--ignore-platform-req=php+' || '' }}
 
       - name: Install lowest dependencies
--- a/src/Behat/Behat/Tester/ServiceContainer/TesterExtension.php
+++ b/src/Behat/Behat/Tester/ServiceContainer/TesterExtension.php
@@ -96,7 +96,7 @@ class TesterExtension extends BaseExtension
         $this->loadPendingExceptionStringer($container);
     }
 
-    public function process(ContainerBuilder $container)
+    public function process(ContainerBuilder $container): void
     {
         parent::process($container);
 
--- a/src/Behat/Testwork/EventDispatcher/ServiceContainer/EventDispatcherExtension.php
+++ b/src/Behat/Testwork/EventDispatcher/ServiceContainer/EventDispatcherExtension.php
@@ -75,7 +75,7 @@ class EventDispatcherExtension implements Extension
         $this->loadEventDispatchingSuiteTester($container);
     }
 
-    public function process(ContainerBuilder $container)
+    public function process(ContainerBuilder $container): void
     {
         $this->processSubscribers($container);
     }
--- a/src/Behat/Testwork/Hook/ServiceContainer/HookExtension.php
+++ b/src/Behat/Testwork/Hook/ServiceContainer/HookExtension.php
@@ -56,7 +56,7 @@ class HookExtension implements Extension
         $this->loadHookableTesters($container);
     }
 
-    public function process(ContainerBuilder $container)
+    public function process(ContainerBuilder $container): void
     {
     }
 
--- a/src/Behat/Testwork/Tester/ServiceContainer/TesterExtension.php
+++ b/src/Behat/Testwork/Tester/ServiceContainer/TesterExtension.php
@@ -109,7 +109,7 @@ abstract class TesterExtension implements Extension
         $this->loadSpecificationTester($container);
     }
 
-    public function process(ContainerBuilder $container)
+    public function process(ContainerBuilder $container): void
     {
         $this->processExerciseWrappers($container);
         $this->processSuiteTesterWrappers($container);
--- a/composer.json
+++ b/composer.json
@@ -21,12 +21,12 @@
         "composer/xdebug-handler": "^1.4 || ^2.0 || ^3.0",
         "nikic/php-parser": "^4.19.2 || ^5.2",
         "psr/container": "^1.0 || ^2.0",
-        "symfony/config": "^5.4 || ^6.4 || ^7.0",
-        "symfony/console": "^5.4 || ^6.4 || ^7.0",
-        "symfony/dependency-injection": "^5.4 || ^6.4 || ^7.0",
-        "symfony/event-dispatcher": "^5.4 || ^6.4 || ^7.0",
-        "symfony/translation": "^5.4 || ^6.4 || ^7.0",
-        "symfony/yaml": "^5.4 || ^6.4 || ^7.0"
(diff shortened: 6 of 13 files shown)


Kocal:
Oh yeah you're right, I forgot about this package. Thanks! :)

VincentLanglet:
Seems like you forgot one place return type hint @Kocal 
```
Behat\Behat\Tester\ServiceContainer\TesterExtension::process
```

Kocal:
Oops, thanks, it should be fine now! :) 

There is still a small issue here, someone from Symfony must create the tag `3.6.1` on [symfony/translation-contracts](https://github.com/symfony/translation-contracts), and we will be good: 
```
  Problem 1
    - Root composer.json requires symfony/translation ^5.4 || ^6.4 || ^7.0 || ^8.0 -> satisfiable by symfony/translation[v8.0.0-BETA1].
    - symfony/translation v8.0.0-BETA1 requires symfony/translation-contracts ^3.6.1 -> found symfony/translation-contracts[dev-main, 3.6.x-dev (alias of dev-main)] but it does not match your minimum-stability.
```

.github/workflows/build.yml
```diff
@@ -76,8 +86,8 @@ jobs:
       - name: Install latest dependencies
         if: matrix.composer-mode == 'update'
         env:
-          SYMFONY_REQUIRE: ${{ matrix.symfony-version }}.*
-        run: composer update ${{ matrix.php == '8.5' && '--ignore-platform-req=php+' || '' }}
+          SYMFONY_REQUIRE: ${{ matrix.symfony-version }}
+        run: composer update ${{ (matrix.php == '8.4' || matrix.php == '8.5') && '--ignore-platform-req=php+' || '' }}
```
acoulton: Do we actually still need to ignore platform requirements on PHP 8.4? It's passing without this in master currently and I don't think symfony 8 will add any non-php-84 dependencies?

src/Behat/Behat/Tester/ServiceContainer/TesterExtension.php
```diff
@@ -96,7 +96,7 @@ public function load(ContainerBuilder $container, array $config)
         $this->loadPendingExceptionStringer($container);
     }
 
-    public function process(ContainerBuilder $container)
+    public function process(ContainerBuilder $container): void
```
acoulton, line 99: Hmm - this is tricky - TesterExtension is not a final class so this would technically be a BC break... I don't **imagine** people will be extending it, but I've been wrong before...

src/Behat/Testwork/EventDispatcher/ServiceContainer/EventDispatcherExtension.php
```diff
@@ -75,7 +75,7 @@ public function load(ContainerBuilder $container, array $config)
         $this->loadEventDispatchingSuiteTester($container);
     }
 
-    public function process(ContainerBuilder $container)
+    public function process(ContainerBuilder $container): void
```
acoulton, line 78: This one is also a technical BC break

src/Behat/Testwork/Hook/ServiceContainer/HookExtension.php
```diff
@@ -56,7 +56,7 @@ public function load(ContainerBuilder $container, array $config)
         $this->loadHookableTesters($container);
     }
 
-    public function process(ContainerBuilder $container)
+    public function process(ContainerBuilder $container): void
```
acoulton, line 59: Technical BC break

src/Behat/Testwork/Tester/ServiceContainer/TesterExtension.php
```diff
@@ -109,7 +109,7 @@ public function load(ContainerBuilder $container, array $config)
         $this->loadSpecificationTester($container);
     }
 
-    public function process(ContainerBuilder $container)
+    public function process(ContainerBuilder $container): void
```
acoulton, line 112: Also a potential BC break - `TesterExtension` is abstract and theoretically intended as a base class for other testers on the "Testwork" framework though I'm not aware of projects outside Behat using it.

acoulton (commented):
@Kocal thanks for doing this.

There are a few classes where adding the `void` return typehint required by the Symfony 8 `CompilerPassInterface` interface will technically break BC in Behat.

In some of these cases I'm not actually sure **why** our class isn't `final` as most of them are and they don't look like things we'd want users to extend. So the risk of breaking BC on those is fairly small.

However the one that's not so visible in the diff (except in the feature examples) is obviously our `Extension` interface itself, because that extends the symfony one.

So any/every third-party Behat extension a project is using will also need to be updated to add the return typehints before it can be used with symfony 8 (and may not itself have a dependency on the symfony version, since they are apparently only implementing "our" interface).

I'm therefore leaning to thinking this has to be a new major Behat version. We could make that a very small release, with just this and a few other "shouldn't affect anyone but could do" internal things that we've been holding off on, to make it easy for end-users to update. 

There are bigger plans for "a future Behat major" but IMO we should probably keep them separate from one that's required for dependency version support.

This would mean that people will need to wait for third-party extensions to update their composer requirements to allow the new Behat major - but since they'd also need to have the return typehints to work in a project with symfony 8 that probably isn't a big deal...

I'm still deciding what I think about this, opinions/arguments very welcome.

acoulton:
OK, we'll get a 4.x branch going / updated for us to start working towards a (small) 4.0 - will get this merged as soon as we have something clean to merge to (the current 4.x is very out of date / essentially abandoned and we need to work out the best way to solve that - see #1692).

acoulton:
Oops - github automatically closed this when I deleted & recreated the 4.x branch. This can now be rebased.

Kocal:
Will do tonight, thanks for cherry-pickings :)

Kocal:
PR rebased!

.github/workflows/build.yml
```diff
@@ -51,12 +51,18 @@ jobs:
...
             os: ubuntu-latest
             composer-mode: update
-            symfony-version: '6.4'
+            symfony-version: '6.4.*'
+
+          - php: 8.4
+            os: ubuntu-latest
+            composer-mode: update
+            composer-minimum-stability: beta
+            symfony-version: '^8.0.0-BETA-1'
```
acoulton: I think with the minimum-stability set to beta, we don't need to explicitly specify it in the version constraint as well - if I install locally on this branch with SYMFONY_REQUIRE=8.0.* I'm getting the same packages as the actions run installed here.

```suggestion
            symfony-version: '8.0.*'
```

.github/workflows/build.yml
acoulton: I'm still not clear on the rationale for this change. I've just tried installing locally on 8.4 with the SYMFONY_REQUIRE set to 8.0.* and I'm getting the same packages with & without the `--ignore-platform-req` flag.

AFAICS we only need that on 8.5 to allow us to install with dependencies that don't officially support 8.5 yet.

acoulton (changes requested):
Thanks very much @Kocal just a couple of comments on the build.yml

.github/workflows/build.yml
Kocal: I don't really remember why TBH, I will remove it

acoulton (approved):
Great work, thanks for doing this @Kocal

mpdude:
Is there a list of items that need to be worked on for this minimal-changes 4.0 release?

acoulton:
@mpdude sort of - I have started a [4.0 milestone](https://github.com/Behat/Behat/milestone/5), with some of the existing issues that I think are relevant - but there are some things (e.g. adding strict types throughout) that still exist in our heads / in comments on other issues. I'll try to create a proper 4.0 release planning issue later today, or tomorrow.

mpdude:
Thank you!

I was asking because I'd like to help extension maintainers to take care of the necessary changes and make new releases compatible to a future 4.x themselves – something that I expect to take some time and that blocks some of my projects from doing "clean" Symfony 8 upgrades (without installing pending PR branches of things through Composer).

Selling changes to extension maintainers, especially when it comes to declaring compat with Behat 4 would be much easier if 4.0 were already a real thing 😉 .

Keep up the good work! It's amazing to see how much things have been moving here over the past weeks.

stof:
For extension maintainers, the main necessary change will be to use native types when they override/implement methods coming from Behat (and requiring at least PHP 7.2+ so that they can benefit from PHP variance rules to be compatible with both 3.x and 4.x of Behat)

mpdude:
Would you recommend that extension authors declare `symfony/event-dispatcher` and/or `symfony/depencency-injection` in their composer.json files with appropriate versions, since they depend on interfaces from those packages - even if that's hidden behind the `Extension` Behat interface that extends `CompilerPassInterface`?

(Yes, I have seen #1316 and I see that there might be better solutions that make the problem go away.)

acoulton:
I think that's probably one of the first questions we need to research/answer.

My strong preference would be for us to define our own clean interfaces (which would coincidentally be the same as the current symfony ones, but might diverge in future).

And then extension developers would not be tied to any particular symfony dependency unless they were using symfony features directly.

However it needs someone to dig in to how complex that would be to implement / support in the Behat codebase in the first instance.

PR #1688, opened 2025-11-04
refactor: use fixtures for hook failure tests

carlos-granados:
Updates the hook failures tests to use fixtures instead of files created on the fly. Uses a single fixture folder for all hook failures tests. Removes hook failure tests for annotations


features/hooks_failures_json.feature
```diff
@@ -4,49 +4,17 @@ Feature: Display hook failures location in json printer
...
-          When I have a simple step
-      """
-    And a file named "features/two.feature" with:
-      """
-      Feature: Second feature
-
-        Scenario: First scenario
-          When I have a simple step
-      """
+    Given I initialise the working directory from the "HookFailures" fixtures folder
```
carlos-granados, line 7: Use the fixtures from the "HookFailures" folder

features/hooks_failures_json.feature
```diff
@@ -4,49 +4,17 @@ Feature: Display hook failures location in json printer
...
+    And I provide the following options for all behat invocations:
+      | option      | value |
+      | --no-colors |       |
+      | --format    | json  |
```
carlos-granados, line 11: Define the output format for all tests

features/hooks_failures_json.feature
```diff
@@ -4,49 +4,17 @@ Feature: Display hook failures location in json printer
...
-          #[When('I have a simple step')]
-          public function iHaveASimpleStep()
-          {
-          }
-      }
-      """
-    When I run "behat --no-colors --format=json --out=beforesuite.json"
+    When I run behat with the following additional options:
+      | option    | value            |
+      | --profile | beforeSuite      |
```
carlos-granados, line 16: We use the profile that matches what we want to test

features/hooks_failures_junit_annotations.feature
carlos-granados, line 1: Testing this with annotations does not provide any real value. We already test that hook annotations are processed in other tests and the hook failures tests are about the output when one of these failures happen

features/hooks_failures_junit_attributes.feature
```diff
@@ -4,49 +4,17 @@ Feature: Display hook failures location in junit printer using attributes
...
-          When I have a simple step
-      """
-    And a file named "features/two.feature" with:
-      """
-      Feature: Second feature
-
-        Scenario: First scenario
-          When I have a simple step
-      """
+    Given I initialise the working directory from the "HookFailures" fixtures folder
```
carlos-granados, line 7: The tests use the same fixture folder for all formatters

tests/Fixtures/HookFailures/features/bootstrap/BaseContext.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+use Behat\Behat\Context\Context;
+
+abstract class BaseContext implements Context
+{
+    private static bool $thrown = false;
+
+    protected static function throwFailure(string $message): void
```
carlos-granados, line 9: We implement a function that only throws an exception the first time it is called, this allows us to use this with several suites, features, scenarios or steps

PR #1689, opened 2025-11-05
feat: Support PHP 8.5

acoulton:
Adds PHP 8.5 to our build matrix and supported versions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: [8.1, 8.2, 8.3, 8.4]
+        php: [8.1, 8.2, 8.3, 8.4, 8.5]
         os: [ubuntu-latest]
         composer-mode: [update]
         symfony-version: ['']
@@ -77,7 +77,7 @@ jobs:
         if: matrix.composer-mode == 'update'
         env:
           SYMFONY_REQUIRE: ${{ matrix.symfony-version }}.*
-        run: composer update ${{ matrix.php == '8.4' && '--ignore-platform-req=php+' || '' }}
+        run: composer update ${{ matrix.php == '8.5' && '--ignore-platform-req=php+' || '' }}
 
       - name: Install lowest dependencies
         if: matrix.composer-mode == 'lowest'
@@ -95,7 +95,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: [8.1, 8.4]
+        php: [8.1, 8.5]
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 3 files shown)


src/Behat/Behat/Context/Snippet/Generator/FixedContextIdentifier.php
```diff
@@ -26,7 +26,7 @@ public function __construct(
 
     public function guessTargetContextClass(ContextEnvironment $environment): ?string
     {
-        if ($environment->hasContextClass($this->contextClass)) {
+        if (is_string($this->contextClass) && $environment->hasContextClass($this->contextClass)) {
```
acoulton, line 29: See the commit message for details of this fix.

.github/workflows/build.yml
acoulton, line 37: I have not yet bumped the MacOs or Windows jobs to 8.5 - I think we should wait until there's a stable 8.5 release for that to avoid any packaging issues / RC changes affecting the only builds we run on these platforms.

src/Behat/Behat/Context/Snippet/Generator/FixedContextIdentifier.php
stof: I would probably write it as `$this->contextClass !== null` instead.

.github/workflows/build.yml
stof: I agree with that

src/Behat/Behat/Context/Snippet/Generator/FixedContextIdentifier.php
acoulton: I would have too, historically. 

However if `String` was a class rather than a scalar in PHP, then Rector's [FlipTypeControlToExclusiveTypeRector](https://getrector.com/rule-detail/flip-type-control-to-use-exclusive-type-rector) (which is part of the code quality set that we have enabled) would flip that to `$this->contextClass instanceof String`.

Obviously it's not a class, and therefore Rector has no opinion on this condition - but it felt to me like `is_string(..)` was more in the style of how (object-property) conditions are expressed in the rest of our codebase.

The property is strict typed as `?string` so there's no functional difference either way. Happy to change it if you want?

stof: That rector rule is quite opinionated though (and I would probably never use it in my own projects)

stof (approved):
Approved anyway.

src/Behat/Behat/Context/Snippet/Generator/FixedContextIdentifier.php
acoulton: @carlos-granados do you have an opinion? I have no particularly strong feelings either way but conscious you've been most involved in picking Rector rules to apply.

carlos-granados: No strong opinion one way or the other, let's just keep it as it is

PR #1397, opened 2022-07-02
Accept multiple path arguments at CLI

adrienbrault:
This is https://github.com/Behat/Behat/pull/996 rebased. I've removed the change to `CHANGELOG.md`.

I've been using this patch for years now, it's great!

PR #1692, opened 2025-11-07
Merge current 3.x series up to 4.x

acoulton:
Brings the 4.x branch up to date with current master.

**I AM ACTUALLY NOT SURE ABOUT MERGING THIS**.

The merge essentially reverts everything that is in the current 4.x (https://github.com/Behat/Behat/compare/master...4.x). Most of these are workflow / dependency changes that have already been superseded (other than enabling CI on the 4.x branch).

The only functional change in the 4.x branch is #1397, which was targeted to 4.x due to a concern about BC. The 4.x branch stalled shortly afterwards (in 2023) and that PR was never mentioned in the issue for relating to running with multiple paths (#834)

Therefore, we actually merged a similar-but-different contribution from #1611 to the 3.x series and released it in 3.23.0. It looks like we didn't pick up on the potential BC concern about changing that CLI argument to an array, but we have not had reports of any issues. I've reviewed #1397 and there's nothing there that isn't already in 3.x.

Given all that, and that the 4.x branch has never been released and has been stale for several years, I think we might be better off just deleting and recreating it from the current master.

That would give us significantly clearer git history and ensure we can more easily see what has **actually** gone into the 4.x series as we work towards releases on that branch.

Opinions welcome @carlos-granados @stof and anyone else!


acoulton:
Closed in favour of #1693 ahead of starting a new 4.x branch.

PR #1693, opened 2025-11-12
Merge original 4.x branch back to 3.x series

acoulton:
Brings the history of the old abandoned 4.x branch into the current master, ahead of creating a new 4.x branch. This will avoid us losing the record of these contributions when we restart development for 4.0.

There are only very limited changes - there was only one functional PR #1397 that was merged into 4.x in 2023. The same feature was subsequently implemented by a different contributor in #1611, merged into master and released in 3.23.0.

The solutions are essentially identical except for a change to WrongPathsException which I have kept both because it is more correct and in case anyone is already using the 4.x branch.

Once this PR is merged, I will delete the current 4.x branch and recreate it from master.

Replaces #1692

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,7 +2,9 @@ name: Build
 
 on:
   push:
-    branches: [master]
+    branches:
+      - master
+      - 4.x
     paths-ignore:
       - '**.md'
   pull_request:
--- a/src/Behat/Testwork/Tester/Exception/WrongPathsException.php
+++ b/src/Behat/Testwork/Tester/Exception/WrongPathsException.php
@@ -20,25 +20,41 @@ use Symfony\Component\DependencyInjection\Exception\RuntimeException;
 final class WrongPathsException extends RuntimeException implements TesterException
 {
     /**
+     * @var list<string>
+     */
+    private readonly array $paths;
+
+    /**
      * Initializes exception.
      *
      * @param string $message
-     * @param string $path
+     * @param string|list<string> $paths
      */
     public function __construct(
         $message,
-        private $path,
+        string|array $paths,
     ) {
         parent::__construct($message);
+        $this->paths = (array) $paths;
     }
 
     /**
      * Returns path that caused exception.
      *
-     * @return string
+     * @return list<string>
      */
-    public function getPath()
+    public function getPaths(): array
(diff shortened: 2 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -2,7 +2,9 @@ name: Build
 
 on:
   push:
-    branches: [master]
+    branches:
+      - master
+      - 4.x
```
acoulton, line 7: Adding 4.x into the list in current master will minimise the need for differences in our build.yml across branches and therefore future merge conflicts.

src/Behat/Testwork/Tester/Exception/WrongPathsException.php
```diff
@@ -19,26 +19,41 @@
...
     /**
      * Initializes exception.
      *
      * @param string $message
-     * @param string $path
      */
     public function __construct(
         $message,
-        private $path,
+        string|array $paths,
```
acoulton, line 35: Kept as `string|array` in 3.x for BC, but we can make this just `array` in 4.x

src/Behat/Testwork/Tester/Exception/WrongPathsException.php
```diff
@@ -19,26 +19,41 @@
...
      *
-     * @return string
+     * @deprecated
      */
-    public function getPath()
+    public function getPath(): string
     {
-        return $this->path;
+        return implode(', ', $this->paths);
     }
```
acoulton, line 59: We don't actually use this getter, but it is part of the public interface and kept for BC. Should be very limited impact to anyone if we remove it in 4.x.

src/Behat/Testwork/Tester/Exception/WrongPathsException.php
```diff
@@ -19,26 +19,41 @@
...
+    /**
+     * @var list<string>
+     */
+    private readonly array $paths;
+
     /**
      * Initializes exception.
      *
      * @param string $message
-     * @param string $path
```
stof, line 26: should actually be documented as `@param string|list<string> $paths` to have the precise type.

acoulton: Good catch, thanks - fixed.

PR #1694, opened 2025-11-19
[4.x] refactor: Resolve small refactors & deprecations

### LINKED ISSUE #1519: Small refactors and removals for next major version

acoulton:
Resolves a number of deprecations and small refactors - generally around constructor signatures - that were identified in previous PRs and logged for the next major.

Partially addresses #1519

### DIFF
--- a/src/Behat/Behat/Output/Node/Printer/JUnit/JUnitScenarioPrinter.php
+++ b/src/Behat/Behat/Output/Node/Printer/JUnit/JUnitScenarioPrinter.php
@@ -11,7 +11,6 @@
 namespace Behat\Behat\Output\Node\Printer\JUnit;
 
 use Behat\Behat\Output\Node\EventListener\JUnit\JUnitDurationListener;
-use Behat\Behat\Output\Node\EventListener\JUnit\JUnitOutlineStoreListener;
 use Behat\Behat\Output\Node\Printer\Helper\ResultToStringConverter;
 use Behat\Gherkin\Node\FeatureNode;
 use Behat\Gherkin\Node\NamedScenarioInterface;
@@ -30,10 +29,6 @@ final class JUnitScenarioPrinter
 {
     public function __construct(
         private readonly ResultToStringConverter $resultConverter,
-        /*
-         * @deprecated the outlineStoreListener is no longer used but kept for BC. It will be removed in the next major
-         */
-        JUnitOutlineStoreListener $outlineStoreListener,
         private readonly ?JUnitDurationListener $durationListener = null,
         private readonly ?ConfigurablePathPrinter $configurablePathPrinter = null,
     ) {
--- a/phpstan.dist.neon
+++ b/phpstan.dist.neon
@@ -12,7 +12,6 @@ parameters:
             paths:
                 - src/Behat/Behat/Definition/Printer/ConsoleDefinitionPrinter.php
                 - src/Behat/Behat/EventDispatcher/Cli/StopOnFailureController.php
-                - src/Behat/Behat/Output/Node/Printer/JUnit/JUnitScenarioPrinter.php
                 - src/Behat/Behat/Output/Node/Printer/ListPrinter.php
                 - src/Behat/Behat/Tester/Runtime/RuntimeFeatureTester.php
                 - src/Behat/Behat/Transformation/Transformation/ReturnTypeTransformation.php
--- a/phpstan.dist.neon
+++ b/phpstan.dist.neon
@@ -10,12 +10,7 @@ parameters:
         -
             identifier: constructor.unusedParameter
             paths:
-                - src/Behat/Behat/Definition/Printer/ConsoleDefinitionPrinter.php
-                - src/Behat/Behat/EventDispatcher/Cli/StopOnFailureController.php
-                - src/Behat/Behat/Output/Node/Printer/ListPrinter.php
-                - src/Behat/Behat/Tester/Runtime/RuntimeFeatureTester.php
                 - src/Behat/Behat/Transformation/Transformation/ReturnTypeTransformation.php
-                - src/Behat/Testwork/Ordering/Cli/OrderController.php
         -
             identifier: function.impossibleType
             paths:
--- a/src/Behat/Behat/ApplicationFactory.php
+++ b/src/Behat/Behat/ApplicationFactory.php
@@ -49,13 +49,6 @@ use Composer\InstalledVersions;
  */
 final class ApplicationFactory extends BaseFactory
 {
-    /**
-     * @deprecated this constant will not be updated for releases after 3.13.0 and will be removed in the next major.
-     * You can use composer's runtime API to get the behat version if you need it - see getVersion() in this class for
-     * an example. Note that composer's versions will not always be simple numeric values.
-     */
-    public const VERSION = '3.13.0';
-
     protected function getName()
     {
         return 'behat';
--- a/src/Behat/Behat/Definition/Printer/ConsoleDefinitionPrinter.php
+++ b/src/Behat/Behat/Definition/Printer/ConsoleDefinitionPrinter.php
@@ -11,7 +11,6 @@
 namespace Behat\Behat\Definition\Printer;
 
 use Behat\Behat\Definition\Definition;
-use Behat\Behat\Definition\Pattern\PatternTransformer;
 use Behat\Behat\Definition\Translator\DefinitionTranslator;
 use Behat\Gherkin\Keywords\KeywordsInterface;
 use Behat\Testwork\Suite\Suite;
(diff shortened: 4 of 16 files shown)


src/Behat/Behat/Output/Node/Printer/JUnit/JUnitScenarioPrinter.php
```diff
@@ -30,10 +30,6 @@ final class JUnitScenarioPrinter
 {
     public function __construct(
         private readonly ResultToStringConverter $resultConverter,
-        /*
-         * @deprecated the outlineStoreListener is no longer used but kept for BC. It will be removed in the next major
-         */
-        JUnitOutlineStoreListener $outlineStoreListener,
         private readonly ?JUnitDurationListener $durationListener = null,
         private readonly ?ConfigurablePathPrinter $configurablePathPrinter = null,
```
acoulton, line 33: These dependencies - and similar in a number of other classes - were added as nullable params to avoid a BC break for users creating our classes without using our service definitions. I think we should probably make them mandatory in 4.0, but I intend to handle that in a separate PR.

src/Behat/Behat/Transformation/SimpleArgumentTransformation.php
```diff
@@ -21,6 +21,12 @@
  */
 interface SimpleArgumentTransformation extends Transformation
 {
+
+    /**
+     * @phpstan-param callable $callable
+     */
+    public function __construct(string $pattern, callable|array $callable, ?string $description = null);
```
acoulton: The interface here matches existing implementations and run-time usage. We may want to consider whether to drop support for legacy `array` callables in 4.0 - that would obviously be a separate PR as it would have quite wide-ranging impact.

src/Behat/Behat/Transformation/Transformation/ColumnBasedTableTransformation.php
```diff
@@ -32,17 +32,10 @@ public static function supportsPatternAndMethod($pattern, ReflectionMethod $meth
...
-     * @param string      $pattern
-     * @param callable    $callable
-     * @param string|null $description
-     */
     public function __construct(
-        private $pattern,
-        $callable,
-        $description = null,
+        private readonly string $pattern,
+        callable|array $callable,
```
stof: why `callable|array` while the phpdoc said `callable` ?

acoulton: It had been a long day and I had a brainfade and imagined that `callable` as a native type hint only supported modern callables and not the old array style 🤣

Will fix.

acoulton: Urgh, turns out I was right the first time and the brainfade was in my previous reply. We do need `callable|array` to allow us to accept old-style callables (at least according to the CI failure). I'll look at this again later on / tomorrow - at a glance we might be able to update our factory code to pass a modern callable instead in 3.x, which would make this easier to tighten in 4.x.

stof: `callable` accepts any callables, not just Closure.

If tests fail, it might indicate that we have places passing an array that is not actually callable.

stof: Looking at it, those classes extend RuntimeCallee, which supports `callable|array{class-string, string}` as type for `$callable`, where the array can be resolved into an actual callable. See `Call::getBoundCallable` and https://github.com/Behat/Behat/blob/f003e1fe5e12288c386b6979d8995c762807640e/src/Behat/Behat/Context/Environment/InitializedContextEnvironment.php#L118-L127

So we should update the type for those transformation in the 3.x branch.

acoulton: Ah, it's exactly that.

[This code](https://github.com/Behat/Behat/blob/f003e1fe5e12288c386b6979d8995c762807640e/src/Behat/Behat/Transformation/Context/Factory/TransformationCalleeFactory.php#L34-L45) is passing `"callables"` that are not actually callable since PHP 8.0 - they are an array of `['class_name', 'method_name']` but the method is an instance method.

This is where I confused myself about "old-style-callables" the other day - they are arrays, they used to be callable, but they are not any more.

These classes extend [RuntimeCallee](https://github.com/Behat/Behat/blob/f003e1fe5e12288c386b6979d8995c762807640e/src/Behat/Testwork/Call/RuntimeCallee.php#L41-L53) which together with the RuntimeCallHandler and InitializedContextEnvironment do a bunch of work to turn this back into an actual callable bound to the correct context class instance. This will obviously be a different instance for every scenario.

RuntimeCallee was previously also typehinted as taking just `callable`, but @carlos-granados [fixed that as part of updating to Psalm 5 last year](https://github.com/Behat/Behat/pull/1548#discussion_r1862359111) -  however it looks like we didn't catch that the phpdoc on the child classes was also incorrect.

I'll take a proper look at the best way to solve this when I can spend a bit of concentrated time on it - I thought this was going to be a quick branch update in a break today, but I suspect it needs a separate PR.

stof: Those arrays have **never** been callable, even on older versions.
They are a custom Behat things, where we have `['class_name', 'method_name']` as callable definition, that will be bound into an actual callable `[$contextObject, 'method_name']` at a later point (so that the method is called on the context object of the current isolated environment).

As long as only phpdoc types were used and no static analysis was applied, the wrong type was not caught by our CI. and the dynamic instantation of transformations in TransformationCalleeFactory means that psalm could not detect this either.

acoulton: I think that's not strictly true.

Prior to 8.0 an array like `['MyClass', 'someInstanceMethod']` would pass `is_callable` and could be used as a callable (emitting a deprecation in later 7.x versions. 

The method would just be called statically - see https://3v4l.org/nPqTK#veol - as I recall it would then fail at runtime if the method attempted to use `$this` since it was in a static context.

The Behat "magic" is that we then convert those legacy static callbacks into instance callbacks on a context class.

My hunch is we'll still need to allow that as I think the context class instance is not known at the time we parse the hook. But we might then be better introducing an explicit type for that e.g `callable|ContextMethodReference` or something to avoid having to support arbitrary arrays. 

As I say, I'll have a proper look - the code chains around all of this are quite complex - and propose a PR in due course.

stof: Oh, I forgot that on PHP 7.x, the `['MyClass', 'someInstanceMethod']` array was still considered callable but deprecated (and would not be a *working* callable)

> But we might then be better introducing an explicit type for that e.g `callable|ContextMethodReference` or something to avoid having to support arbitrary arrays.

We don't support arbitrary arrays in RuntimeCallee

acoulton: > (and would not be a working callable)

It would be a working callable if it did not access any instance methods or state - see that 3v4l where I am successfully using it with array_map.

Hence why I was confusing myself before, the old PHP memories die hard but become jumbled :laughing:

> We don't support arbitrary arrays in RuntimeCallee

Sorry, I was unclear. 

I *think* (to be confirmed) we only actually *need* to support legacy callables where the class is a `Context` class to be provided by `InitializedContextEnvironment`. We might therefore want to deprecate and then remove support for `[string,string]` callables referencing an instance method on an arbitrary class - these will likely fail at call time anyway.

I really need to do some actual work now :laughing:  but I will come back to this and do some proper investigation. I'll likely drop this commit off this branch and we can look at standalone PR(s) to resolve this bit - possibly some BC changes for 3.x and then others for 4.0.

stof: Well, to me, the solution for now (in 3.x also) is to use the same type than in RuntimeCallee (as we really just pass it to the parent constructor anyway)

acoulton: That may be the most pragmatic thing, but if we can tighten `RuntimeCallee` without too much work (and without breaking BC) I think that would be better. I'll take a look.

In the meantime I've removed the `SimpleArgumentTransformation` changes from this branch as I think whatever we do they merit a separate PR.

acoulton:
The phpstan failure is also present in 4.x and will be fixed in a separate PR

carlos-granados (approved):
Good job in the good direction 🎉

PR #1701, opened 2025-11-24
chore: Replace friendsofphp/php-cs-fixer by php-cs-fixer/shim

acoulton:
Using php-cs-fixer/shim avoids us having any restrictions from php-cs-fixer's own dependency constraints.

This was initially included in #1687 by @Kocal as it was required to allow us to install Symfony 8 ahead of php-cs-fixer adding support.

I've extracted to a separate PR for the 3.x series - we might as well switch now to reduce merge conflicts or any other dependency constraints in future.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,8 +16,8 @@
     "require": {
         "php": ">=8.1 <8.6",
         "ext-mbstring": "*",
-        "behat/gherkin": "^4.12.0",
         "composer-runtime-api": "^2.2",
+        "behat/gherkin": "^4.12.0",
         "composer/xdebug-handler": "^1.4 || ^2.0 || ^3.0",
         "nikic/php-parser": "^4.19.2 || ^5.2",
         "psr/container": "^1.0 || ^2.0",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -30,15 +30,15 @@
...
         "opis/json-schema": "^2.5",
         "phpstan/phpstan": "^2.0",
         "phpunit/phpunit": "^9.6",
         "rector/rector": "2.1.7",
         "sebastian/diff": "^4.0",
         "symfony/polyfill-php84": "^1.31",
         "symfony/process": "^5.4 || ^6.4 || ^7.0",
-        "symfony/filesystem": "^5.4 || ^6.4 || ^7.0"
+        "symfony/filesystem": "^5.4 || ^6.4 || ^7.0",
+        "php-cs-fixer/shim": "^3.89"
```
carlos-granados: I think it is good to keep this list sorted (though I can see that this was already broken with symfony/filesystem) Can you please change this so that the list is sorted?

carlos-granados (changes requested):

composer.json
acoulton: Good catch - I've configured composer to sort packages and re-sorted the requires accordingly.

carlos-granados (approved):
nice @acoulton

PR #1702, opened 2025-11-24
refactor: Add backwards-compatible return types ahead of symfony8

acoulton:
Extracted from work by @Kocal in #1687 - these return types will be required in 4.x for symfony8 support, but can be added in 3.x without any BC impact as they are either final classes or internal test code. 

Adding them now will reduce potential for future conflicts merging 3.x up to 4.x.


stof (approved):

carlos-granados (approved):
Thanks @acoulton

Issue #1695, opened 2025-11-19
Encouraging support for Behat development


acoulton:
We now have a solid year of progress under the new maintainer team, and have started working on a 4.0 release (the first major release since 2014).

I would like us to consider more actively encouraging (financial) support for the project maintainers. I have two motivations:

* As a point of principle, I think open source projects - particularly those that are primarily used for work - should be proactive about encouraging their users to support the ecosystem. Normalising the idea of supporting - or at least considering supporting - open source maintainers benefits the whole community.
* From a personal perspective, my circumstances are likely to change a bit in 2026. I am very keen to continue contributing to Behat - I enjoy working on the project and value the collaboration with the other maintainers & the community. But it would undoubtedly be easier for me to justify spending the kind of time I want to spend if I had some additional sponsorship.

IMO, the project and team are not big enough to justify the overhead and governance of a formal (foundation / OpenCollective etc) structure to allow sponsorship of Behat as an overall entity that pays out to individual developers. Although if others felt that was the way forward, I'd be happy to investigate options (which would need to include getting Konstantin's involvement / blessing as the creator of the project).

I'd propose instead to simply ask users to consider individual sponsorship of the current active maintainers (@stof, @carlos-granados and me).

We are all set up on GitHub sponsors, so I suggest that:
* We add a github sponsor link for each maintainer to the "funding" section of the `composer.json` in Behat/Behat and Behat/Gherkin.
* We also add a github `FUNDING.yml` with the same links to sponsor each maintainer, so that users see sponsorship calls-to-action in the github UI.
* If the maintainer team changes in future, we would add / remove people as required.

We could also add a page to the Behat docs with more information about how / why to support the team, and link to that from our composer / FUNDING files.  We could do that **instead** of linking to individual maintainers, but I think adding direct maintainer sponsorship links rather than a click-through the docs site makes it more obvious & easy for users to contribute.

Obviously this really depends on what @carlos-granados and @stof think - but I think this is a discussion we should be transparent about, and I'm very open to thoughts from the wider Behat community on what you think would be useful / appropriate. 

And hopefully it goes without saying that I know there are many other people who've contributed significant time to the project - some very recently - and that is of course hugely valued. I don't in any way intend this discussion to undermine those contributions.

carlos-granados:
Thanks @acoulton I agree with what you describe and propose, I had been thinking about something like this for some time

stof:
I agree about that as well.

PR #1704, opened 2025-11-24
docs: Add funding links and information

acoulton:
@carlos-granados @stof I've added sponsorship links and some readme text in line with #1695 

I think it might be most logical to put the funding links in alphabetical order. But I wasn't totally comfortable doing that as someone with an "a" name :laughing: It'd be nice if Github would randomise the order over time but I don't think they do...

So I ran the names through a random shuffle and have them in different orders in the composer.json and funding.yml. If we go with that, I can shuffle them again for the files in Behat/Gherkin...

Let me know what you think

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,1 @@
+github: [ acoulton, stof, carlos-granados ]
--- a/README.md
+++ b/README.md
@@ -63,8 +63,30 @@ Useful Links
 Contributors
 ------------
 
-- Konstantin Kudryashov [everzet](https://github.com/everzet) [lead developer]
+- Konstantin Kudryashov [everzet](https://github.com/everzet) [original developer]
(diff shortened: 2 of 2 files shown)


carlos-granados (approved):
Thanks @acoulton, looking great

acoulton:
@stof just wanted to check you're also happy with this? Thanks

stof:
You could reduce the number of places where the list of maintainers has to be maintained. The GitHubDriver of composer supports reading the `.github/FUNDING.yml` file (or the `FUNDING.yml file`) in the default branch of the repo where there is no explicit `funding` key in the composer.json.

acoulton:
Ahh, that's useful to know, thanks!

I'll do that later and then we can get this merged.

acoulton:
That's showing now on https://github.com/Behat/Behat and when you run `composer fund`:
image

It appears that composer alpha-sorts the funding links, however GitHub shows them in the order they appear in the file.

PR #1705, opened 2025-11-25
Add option to set if we want to print skipped steps to pretty formatter

### LINKED ISSUE #1658: Provide an option not to display "skipped" steps in the pretty formatter
When a step fails, the pretty formatter prints the complete scenario including all the following skipped steps (and, by default, multiline arguments).

While this is sometimes useful, it can create a lot of noise during development, forcing users to scroll up the output to see the actual failing step and assertion. The impact is often worst early in development (when often the failure will be early in a scenario, due to the order of implementation).

In discussion on #1612, @jdeniau mentioned that it would be good to have an option not to print skipped steps. I'd also find that useful.

carlos-granados:
If we have long scenarios, we may not be interested in seeing the list of skipped steps that is printed after a failed steps. This PR adds an option to the pretty printer to decide if you want to print skipped steps or not

Fixes https://github.com/Behat/Behat/issues/1658

### DIFF
--- a/features/print_skipped_steps.feature
+++ b/features/print_skipped_steps.feature
@@ -1,0 +1,59 @@
+Feature: Pretty formatter print_skipped_steps option
+  In order to control whether skipped steps are printed
+  As a Behat user
+  I need a pretty formatter option to hide skipped steps
+
+  Background:
+    Given I initialise the working directory from the "PrintSkippedSteps" fixtures folder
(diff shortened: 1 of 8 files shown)


acoulton (approved):
Fantastic, thanks @carlos-granados

jdeniau:
Great!

PR #1706, opened 2025-11-25
Add translation cache

### LINKED ISSUE #1452: Skip or improve translation process
Hi,

The translator component is heavily called during the running of my behat tests.

I collect some metrics of my project with blackfire.

```
977 scenarios (977 passed)
3032 steps (3032 passed)
4m40.23s (1.33Gb)
```

image

It take ~10% of the time to translate steps definitions.

Maybe there should be a way to disable the translation process ? (I'm not using it and use the default `en` language).

Or maybe we should cache the result of the translation process ? (4mil call of the method `Behat\Behat\Definition\Translator\Translator::trans` may not be optimized ?)

carlos-granados:
Adds a simple memory cache so that we don't have to call the `trans` function in the translator if we already have translated the current definition, should improve performance in the cases where steps are used many times

Fixes https://github.com/Behat/Behat/issues/1452

### DIFF
--- a/src/Behat/Behat/Definition/Translator/DefinitionTranslator.php
+++ b/src/Behat/Behat/Definition/Translator/DefinitionTranslator.php
@@ -21,6 +21,11 @@ use Behat\Testwork\Suite\Suite;
 final class DefinitionTranslator
 {
     /**
+     * @var array<string, string>
+     */
+    private array $translationCache = [];
+
+    /**
      * Initialises definition translator.
      */
     public function __construct(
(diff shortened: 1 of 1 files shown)


src/Behat/Behat/Definition/Translator/DefinitionTranslator.php
```diff
@@ -39,8 +44,15 @@ public function translateDefinition(Suite $suite, Definition $definition, $langu
     {
         $assetsId = $suite->getName();
         $pattern = $definition->getPattern();
+        $cacheKey = sprintf('%s|%s|%s', $assetsId, $pattern, $language ?? '');
```
stof: this should use `$this->translator->getLocale()` as default locale, not the empty string, so that an explicit locale and the usage of the default locale with the same locale share the same cache key.

carlos-granados: Good shout, updated

acoulton (approved):
LGTM, thanks @carlos-granados

PR #1707, opened 2025-11-25
Refactor append snippets test to use fixtures

carlos-granados:
I want to start refactoring all existing tests to use the fixtures instead of creating the files on the fly. This PR refactors the "append snippets" test to use the fixtures

### DIFF
--- a/features/append_snippets.feature
+++ b/features/append_snippets.feature
@@ -71,6 +71,10 @@ Feature: Append snippets option
           {
           }
 
+          /**
+           * This dummy method is added just so that PHP-CS-Fixer does not
+           * complain about unused import `use` statements.
+           */
           private function useClasses(PyStringNode $node, TableNode $table)
           {
               throw new PendingException();
@@ -173,6 +177,10 @@ Feature: Append snippets option
           {
           }
 
+          /**
+           * This dummy method is added just so that PHP-CS-Fixer does not
+           * complain about unused import `use` statements.
+           */
           private function useClasses(PyStringNode $node, TableNode $table)
           {
           }
--- a/features/append_snippets.feature
+++ b/features/append_snippets.feature
@@ -4,109 +4,7 @@ Feature: Append snippets option
   I need to be able to autoappend snippets to context
 
   Background:
-    Given a file named "features/bootstrap/FeatureContext.php" with:
-      """
-      <?php
-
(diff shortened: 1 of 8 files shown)


features/append_snippets.feature
```diff
@@ -4,120 +4,18 @@ Feature: Append snippets option
...
     When I run "behat -f progress --append-snippets --snippets-for=FeatureContext --snippets-type=regex"
     Then "features/bootstrap/FeatureContext.php" file should contain:
       """
       <?php
 
-      use Behat\Behat\Context\Context,
-          Behat\Behat\Tester\Exception\PendingException;
-      use Behat\Gherkin\Node\PyStringNode,
-          Behat\Gherkin\Node\TableNode;
+      use Behat\Behat\Context\Context;
```
carlos-granados, line 15: The new fixture files must follow the coding standards of the project, so the expected output needs to be changed to also follow these standards. For example here we do not use `use` statements with multiple imports

features/append_snippets.feature
```diff
@@ -127,42 +25,56 @@ Feature: Append snippets option
...
-          #[Given('/^context parameter "([^"]*)" should be array with (\d+) elements$/')]
-          public function contextParameterShouldBeArrayWithElements($key, $count) {
-              \PHPUnit\Framework\Assert::assertIsArray($this->parameters[$key]);
-              \PHPUnit\Framework\Assert::assertEquals(2, count($this->parameters[$key]));
+          private function doSomethingUndefinedWith()
+          {
           }
 
-          private function doSomethingUndefinedWith() {}
+          private function useClasses(PyStringNode $node, TableNode $table)
```
carlos-granados, line 78: If we add some `use` statements we must have some code which actually uses these classes, otherwise the CS fixer will complain, so I added these dummy methods which just use these classes

acoulton: Might be worth having this as a code comment (partly in case a tool / someone feels like removing the unused method params in future)?

acoulton (commented):
Thanks @carlos-granados just a small suggestion.

It'll be great to get these all onto the newer fixture style :)

features/append_snippets.feature
carlos-granados: Good shout, updated

acoulton (approved):

PR #1708, opened 2025-11-25
change: Drop `dev-master` composer branch alias & update actions branch

### LINKED ISSUE #1703: Rename `master` branch to `3.x`
Moved from https://github.com/Behat/docs/pull/209#issuecomment-3570016576 for visibility/clarity

As we start working towards 4.0, I propose that we rename the `master` branch to `3.x`.

IMO this would be clearer for us and for contributors while we're supporting multiple versions. 

In particular, until we drop support for 3.x we will still want most backwards-compatible changes to go into 3.x first and then be merged up to 4.x. I think that's more obvious/easier to explain if the branches are explicitly named.

For that reason I would also keep 3.x as the default branch until that version goes into security-only support.

As @stof has pointed out, keeping `master` as the older 3.x branch feels wrong, but I think it's also problematic to use it as the non-default branch for 4.x - if we have a `master` then people will instinctively base contributions off it causing friction & annoying rebases/merges.

Explicitly naming all branches follows how e.g. symfony and doctrine operate. It'll also be easier to configure branch protection rules, github actions build triggers etc as anything *.x is a main version branch.

I think we can configure the branch alias the other way round in our composer.json so that anyone depending on `dev-master` through composer isn't affected.

The only downside would be that any external links to specific files in the master branch e.g. https://github.com/Behat/Behat/blob/master/CHANGELOG.md would be broken, although equally linking to a file in a branch like that is never guaranteed to be long-term stable.

acoulton:
We have renamed the `master` branch to `3.x` as part of preparations for working towards a 4.0 release. In future, all main branches will be named for their release series in this fashion.

Anyone directly depending on `dev-master` in their own `composer.json` will see a failure on their next composer update. They will need to migrate to a more specific constraint for the actual version they require.

Github actions will build on all pushes to a `*.x` branch - we will no longer need to update the workflow when starting a new version series.

Fixes #1703

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,8 +3,7 @@ name: Build
 on:
   push:
     branches:
-      - master
-      - 4.x
+      - '*.x'
     paths-ignore:
       - '**.md'
   pull_request:
--- a/composer.json
+++ b/composer.json
@@ -62,12 +62,6 @@
         }
     },
 
-    "extra": {
-        "branch-alias": {
-            "dev-master": "3.x-dev"
-        }
-    },
-
     "scripts": {
         "all-tests": [
             "@cs",


acoulton:
@carlos-granados this will need to be merged before your PRs so that they trigger a build when merged to the new 3.x branch.

carlos-granados:
@acoulton I only saw this after merging my PRs 😞

carlos-granados (approved):
Good stuff @acoulton

acoulton:
> @acoulton I only saw this after merging my PRs 😞 

Sorry, after I did it I thought I should have co-ordinated with you first on the timing.

We've only missed a couple of builds, and your PRs were green and up to date, so not the end of the world.

PR #1709, opened 2025-11-26
Refactor attributes test to use fixtures

carlos-granados:
Refactors `attributes.feature` to use fixtures


acoulton (approved):
Thanks @carlos-granados

PR #1710, opened 2025-11-26
Refactor autowire test to use fixtures

carlos-granados:


features/autowire.feature
```diff
@@ -15,258 +15,73 @@ Feature: Helper services autowire
...
-          /** @Given a step */
-          public function aStep() {}
-      }
-      """
-    When I run "behat --no-colors -f progress features/autowire.feature"
+    When I run behat with the following additional options:
+      | option                    | value            |
+      | --profile                 | constructor_args |
+      | features/autowire.feature |                  |
     Then it should pass
```
acoulton: The existing scenarios are very relaxed on what qualifies as success (basically just that Behat exited 0). When there was only one context class available for it to use that was maybe ok-ish (if we trust that Behat did actually run to completion).

But we're now also trusting that the wiring from the profile specified in the feature file through the config file in the fixture is definitely running the ConstructorArgsContext and not e.g. one of the other contexts that might be passing for an unrelated reason to what we're trying to prove.

That feels like it gets a bit risky long-term.

Could we perhaps move to a "Then it should pass with" and use the pretty format output to assert that the correct context class & step was executed?

This would also make it a little easier for a new user to understand exactly what this scenario is meant to cover?

carlos-granados: I modified the tests to show the actual output

acoulton (approved):
Thanks, I think that helps a lot :)

PR #1712, opened 2025-11-26
Refactor config_inheritance test to use fixtures

carlos-granados:


carlos-granados:
Mmm, this is weird, locally the tests are passing. I need to try to find out why this fails here

carlos-granados:
The tests were passing locally because I still had a behat.yml file which I had not committed and it was using that instead of the behat.php file. And this has brought into view one issue: when using withImport and withProfile we are not merging the configuration as with the yaml files, instead we are replacing full profiles. This requires further investigation, so for now I am going to be closing this PR

acoulton:
I could be wrong, but I feel like I had a vague memory we discussed inheritance between files in the php config implementation and decided that since users have various options to import/reuse/share values within their php files it would be better to let them be explicit about that than to replicate the implicit yaml behaviour, particularly since we only merge some bits of the config.

It is entirely possible I've imagined that though 🤣

PR #1715, opened 2025-11-27
Refactor context consistency tests

carlos-granados:
We had separate feature files for annotations and attributes. In this case we are testing the access to contexts so the fact that steps are defined using annotations or attributes is not important, so I have just left the version that uses attributes (and removed the reference to attributes from the feature name to make it obvious that we don't care about this)


acoulton (approved):
Thanks @carlos-granados

PR #1717, opened 2025-11-27
refactor definitions override tests

carlos-granados:
In this case I have preferred to leave the tests with both annotations and attributes as I think it is significant


carlos-granados:
Mmm, not sure why the static analysis for PHP 8.5 is failing, seems unrelated to this PR, need to investigate

acoulton (approved):
Great - I agree, the annotations & attributes variants add value here.

PR #1718, opened 2025-11-27
test

carlos-granados:
PR created to confirm if phpstan fails in php 8.5 due to some new dependency update


carlos-granados:
OK, I confirmed that PHPStan has started to fail, probably due to some dependency being updated. I will try to find out what is going on

PR #1719, opened 2025-11-27
Fix static analysis in PHP 8.5

carlos-granados:
The latest versions of the Symfony Config component improve the docs for some of the functions on the config objects and now use Template tags so that PHPStan is able to correctly guess the type returned. So some asserts are not needed any more. But they are still needed in older versions of Symfony which don't have these improved docs. The only solution I was able to find was to change these asserts into @var tags, this seems to work for both versions

Had to disable a PHP-CS-Fixer rule which was incorrectly trying to change these @var tag docblocks into normal comments

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -29,5 +29,6 @@ return (new Config())
         ],
         'single_line_throw' => false, //override Symfony
         'yoda_style' => false, //override Symfony
+        'phpdoc_to_comment' => false, //incorrectly changes @var tags
     ])
     ->setFinder($finder);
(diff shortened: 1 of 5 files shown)


.php-cs-fixer.dist.php
```diff
@@ -29,5 +29,6 @@
         ],
         'single_line_throw' => false, //override Symfony
         'yoda_style' => false, //override Symfony
+        'phpdoc_to_comment' => false, //incorrectly changes @var tags
```
acoulton: It might be better to just exclude the specific tag we want to keep?

```suggestion
        'phpdoc_to_comment' => [ // Keep as a phpdoc if it contains the `@var` annotation
            'ignored_tags' => ['var'],
        ],
```

acoulton (approved):
Thanks for figuring this out - I can't think of a better solution. Just a small comment on the cs-fixer config.

.php-cs-fixer.dist.php
carlos-granados: Ah, did not known about that option, updated!

PR #1720, opened 2025-11-27
Refactor definition patterns tests

carlos-granados:
In this case I also considered it was useful to keep the tests for annotations and attributes


acoulton (approved):
Agreed, this one is also worth keeping the annotations & attributes versions

PR #1726, opened 2025-11-28
refactor: migrate execution_order.feature to use fixtures

carlos-granados:


acoulton (approved):
I also never knew about the `--order` feature. This exercise is proving useful for finding bits of Behat that are new to me 😆

PR #1727, opened 2025-11-29
refactor: migrate extension_inheritance.feature to use fixtures

carlos-granados:


features/extension_inheritance.feature
```diff
@@ -4,181 +4,32 @@ Feature: Profile extension overrides
...
-          Then the extension "custom extension 1" should not be loaded
-      """
-    When I run "behat -f progress -p custom_profile -vvv"
+    When I run behat with the following additional options:
+      | option                                 | value          |
+      | --config behat-default-extension.yaml  |                |
+      | --profile                              | custom_profile |
+      | -vvv                                   |                |
+      | features/extensions-disabled.feature   |                |
     Then it should pass
```
acoulton: I think this is another where it'd be good to add expected output here so we're confident the config / profile has been triggered as expected (also helps to describe what the feature does now that the files are all moved to the fixture)?

acoulton (changes requested):

features/extension_inheritance.feature
carlos-granados: Updated

acoulton (approved):
Thanks @carlos-granados , really helpful to have that extra detail in the feature.

PR #1729, opened 2025-11-30
refactor: Convert format_options.feature to use fixtures

carlos-granados:


acoulton (approved):
Thanks @carlos-granados

PR #1731, opened 2025-12-01
refactor: convert hooks tests to use fixtures

carlos-granados:


acoulton (approved):
Thanks @carlos-granados, you're on 🔥 !

PR #1733, opened 2025-12-01
Support Symfony 8

nikow13:


stof:
We cannot add support for Symfony 8 in the Behat 3.x branch because Symfony 8 forces us to add native return types in a bunch of classes where this will be a BC break as they are non-final (and so could be extended in behat extensions).
The changes in this PR cannot be accepted in 3.x as they would make us break semver.

This is why we have started the work of preparing a 4.x branch that will have native return types everywhere and will allow us to add such support (work is still in progress on that)

acoulton:
Just to add - the symfony 8 support is already merged and available with a composer requirement of `^4.0@dev`. 

As @stof says, there are more breaking changes to come in that branch so you should expect it to be unstable for now. It will also likely conflict with requirements from any behat extensions you are using, unless you alias it to "look like" a 3.x version. However, if you want to start working with symfony8 now and are using a composer.lock to pin to explicit commits then you *should* be able to use it - you may just need to make changes in your project / extensions when you run a `composer update`.

mpdude:
Thank you guys for working on this!

PR #1734, opened 2025-12-01
refactor: convert i18n tests to use fixtures

carlos-granados:


acoulton (approved):
Thanks @carlos-granados

PR #1740, opened 2025-12-03
refactor: convert name_filters tests to use fixtures

carlos-granados:


acoulton (approved):

Issue #1403, opened 2022-07-07
Automate merge-ups


ciaranmcnulty:
We should have a gh workflow that tries to merge `master` into the `4.x` branch

PR #1749, opened 2025-12-04
chore: Merge up 3.x development branch to 4.x

acoulton:
Brings 4.x up to date with current work on 3.x, including a couple of fixes for coding style & static analysis (due to third-party updates) that are causing build failures on 4.x.

I'm planning to work on #1403 so that we can get this automated in future, but until then we can just periodically do a manual PR.

These PRs **will need to be merge commits, not squash merges** so that the 4.x branch contains a usable change history.


acoulton:
The phpstan failure is already present in 4.x (it is due to new dependency releases). I'll fix that in a separate PR.

carlos-granados (approved):
Thanks @acoulton

Issue #1746, opened 2025-12-04
[4.x] Move support for stringifying phpunit exceptions into an extension


acoulton:
The PHPUnitExceptionStringer is impossible to test properly, and causes static analysis issues, because we do not have a tightly defined runtime dependency on PHPUnit. Therefore we have to guess at how to handle PHPUnit exceptions based on sniffing classes that are known to exist in different phpunit versions. 

We are also still attempting to support PHPUnit versions that have been dead for a long time.

And we know that using PHPUnit assertions within Behat is not officially supported in newer versions of PHPUnit.

We should remove this functionality from core in 4.0, in favour of shipping a separate `behat-phpunit-extension` which can have a hard dependency on the phpunit version(s) it supports. This might also let us ship a workaround for #1618 - and it will let us make it much clearer to end users that they are depending on phpunit functionality that is not guaranteed to work outside phpunit.

acoulton:
@Shine-neko as offered in https://github.com/Behat/Behat/discussions/1858 if you would like to get involved with this that would be great.

Obviously we will ultimately need to create a repo and package in the Behat organisation, so we'll need to co-ordinate on that - but if you wanted you could start off in a repo of your own.

However I think the very first step is to update our own FeatureContext so that it doesn't use phpunit for assertions. 

We could move to another assertion library (some are suggested in https://docs.behat.org/en/latest/useful_resources.html#assertion-tools) but IMO we should *not* depend on any Behat extension to stringify failures in our own test suite as that will create a circular dependency issue in future. 

Given we have quite a small number of steps, and simple assertions, it might be simplest to just implement our own methods to throw an exception with all the information we want directly in the message. @carlos-granados what do you think?

PR #1750, opened 2025-12-04
Refactor phpunit exceptions tests

carlos-granados:


acoulton (approved):
Thanks @carlos-granados 

Being reminded of the hoops I had to jump through here last year makes me feel even more strongly about #1746 😆

PR #1753, opened 2025-12-04
refactor: convert profile_filters tests to use fixtures with PHP config

carlos-granados:


acoulton (approved):
Thanks @carlos-granados

PR #1756, opened 2025-12-05
refactor: convert rerun tests to use fixtures

carlos-granados:


features/rerun.feature
```diff
@@ -130,43 +52,11 @@ Feature: Rerun
...
-            | 3   | 1     | 1      |
-            | 0   | 4     | 7      |
-            | 2   | 2     | 3      |
-      """
-    When I run "behat --no-colors -f progress features/apples.feature"
-    And I run "behat --no-colors -f progress features/apples.feature --rerun"
-    Then it should fail with:
+  Given I run "behat features/apples.feature"
+  And I copy "features/apples-fixed.feature" to "features/apples.feature"
+  When I run "behat features/apples.feature"
```
acoulton: I think perhaps these initial two "I run" steps should check the result & output (even just the `# scenarios (# failed)` line) to prove that the scenarios ran / failed as expected prior to the final assertion? 

Otherwise for example we don't know for sure that the "fixed" apples feature actually fixed the scenario rather than just removing it, or moving it to a different line number so it doesn't match the rerun cache.

carlos-granados: Yeah, makes sense, updated

acoulton: Thanks - this scenario still doesn't quite make sense to me.

* We run Behat once at the very beginning of the scenario, which fails (presumably with 2 failing the same as line 13).
* Then we copy over the file which only changes one example.
* Then we run **without --rerun**, and we see that now it fails on one scenario (but runs all the others).
* And then we run with rerun and see that it only re-runs the failed scenario.

So the only thing we're really proving is that `--rerun` only re-runs the failed example - but we've already proved that in the "Rerun only failed scenarios" above?

I wonder if this scenario is instead meant to prove:

* Run the first time runs all scenarios, 2 fail, generates a rerun file
* Then fix one failure and run **with --rerun** and now it **only runs the 2 that failed the first time**, one passes, and it overwrites the rerun file to remove it
* Then run with `--rerun` again and now it only runs the scenario that is still failing.

Unless I'm missing something?

carlos-granados: Makes sense, updated

acoulton (approved):
Thanks @carlos-granados

PR #1757, opened 2025-12-07
refactor: Resolve symfony/console deprecation in 7.4, fix phpstan

acoulton:
symfony/console 7.4.0 deprecates the `Application::add` method in favour of a new `Application::addCommand` method. `Application::add` is then removed in 8.0.

We had a fix for this in the 4.0 branch, but it started to cause phpstan errors once symfony 8.0 was released & installed for the phpstan job. Additionally, the errors varied depending on the symfony version installed.

This fix bumps us off the deprecated method for Behat 3.x and symfony >= 7.4.0. It also updates the phpstan ignore to cover all supported symfony versions.

### DIFF
--- a/phpstan.dist.neon
+++ b/phpstan.dist.neon
@@ -17,3 +17,8 @@ parameters:
                 - src/Behat/Behat/Tester/Runtime/RuntimeFeatureTester.php
                 - src/Behat/Behat/Transformation/Transformation/ReturnTypeTransformation.php
                 - src/Behat/Testwork/Ordering/Cli/OrderController.php
+        -
+            # BC wrapper for symfony/console < 7.4.0
+            message: '#method_exists.+ ''(add|addCommand)'' will always evaluate to (true|false)#'
+            paths:
+                - src/Behat/Testwork/Cli/Application.php
--- a/src/Behat/Testwork/Cli/Application.php
+++ b/src/Behat/Testwork/Cli/Application.php
@@ -126,9 +126,24 @@ final class Application extends BaseApplication
             $this->configurationLoader->setConfigurationFilePath($path);
         }
 
-        $this->add($this->createCommand($input, $output));
+        $this->addCommand($this->createCommand($input, $output));
 
         return parent::doRun($input, $output);
+    }
+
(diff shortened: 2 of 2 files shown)


phpstan.dist.neon
```diff
@@ -17,3 +17,8 @@ parameters:
                 - src/Behat/Behat/Tester/Runtime/RuntimeFeatureTester.php
                 - src/Behat/Behat/Transformation/Transformation/ReturnTypeTransformation.php
                 - src/Behat/Testwork/Ordering/Cli/OrderController.php
+        -
+            # BC wrapper for symfony/console < 7.4.0
+            message: '#method_exists.+ ''(add|addCommand)'' will always evaluate to (true|false)#'
```
acoulton, line 22: I have intentionally left out the identifier as it varies between symfony version installed (`function.impossibleType` if the method is not present, `function.alreadyNarrowedType` if it is).

src/Behat/Testwork/Cli/Application.php
```diff
@@ -126,11 +126,26 @@ public function doRun(InputInterface $input, OutputInterface $output): int
...
+    {
+        // Provide compatibility with all supported symfony/console versions
+        // Attempt to use the `addCommand` method added in symfony/console 7.4.0
+        // (`add` was deprecated in 7.4.0 and removed in 8.0.0)
+        if (method_exists(parent::class, 'addCommand')) {
+            return parent::addCommand($command);
+        }
+
+        // Otherwise assert we are on an older version with `add` and call that.
+        assert(method_exists($this, 'add'));
```
acoulton, line 144: This is the main functional difference to the patch in the 4.x branch - that was just using a ternary:
https://github.com/Behat/Behat/blob/a4302410cb7badc3ce95710969318e2f0c918749/src/Behat/Testwork/Cli/Application.php#L130

But of course there is no guarantee that `add` exists just because `addCommand` does not, so once 8.0 was actually released (and therefore installed in the phpstan build) it reported that the `add` method does not exist. Asserting here avoids that.

carlos-granados (approved):
Makes sense

PR #1760, opened 2025-12-08
refactor: convert result_types tests to use fixtures

carlos-granados:


acoulton (approved):
thanks @carlos-granados

Issue #1744, opened 2025-12-04
[4.x] Add strict parameter, property & return types throughout


acoulton:
Likely needs to be done in multiple phases:

First add safe types to 3.x to minimise the future diff between versions:

* [x] #1763 to add return types to final / private classes & methods based on static analysis in 3.x (these are BC as the methods cannot be extended)
* [x] #1779 to add further strict return types (as well as some property types)
* [x] add parameter types to private methods based on static analysis in 3.x (these are BC because we know all the possible call sites and the static analyser can detect exactly what types are actually ever passed).
* [x] **MAYBE** add parameter types to public methods based on static analysis in 3.x if they match the phpdoc. Theoretically these are BC if we believe that the phpdoc provides an external contract. However they could cause a BC break if people are calling with `declare(strict_types=1)` and have previously relied on us typecasting internally. We have accepted this in the past, but the risk of doing lots at once might outweigh the benefit of getting this into 3.x

Then add merge up to 4.x and add remaining types:

* [x] Potentially first using [ingenerator/risky-rector-rules](https://github.com/ingenerator/risky-rector-rules) to generate strict types for all interfaces from their phpdoc. Rector produces different results when the interface has strict types, so we need to update them first.  #1820 
* [x] Then use Rector to add types for all classes / methods based on static analysis #1822 and #1824 
* [x] Then use [ingenerator/risky-rector-rules](https://github.com/ingenerator/risky-rector-rules) to convert remaining phpdoc types into strict types. #1824 
* [x] Finally search for and manually resolve any remaining methods / properties with no type declaration #1831

acoulton:
@carlos-granados @stof how do we feel about whether it's a BC break to convert phpdoc `@param` tags to strict types in the 3.x series?

For example:

```php
interface ExceptionHandler
{
    /**
     * Checks if handler supports exception.
     *
     * @param Throwable $exception
     *
     * @return bool
     */
    public function supportsException($exception);
}
```

Could safely become:

```php
interface ExceptionHandler
{
    /**
     * Checks if handler supports exception.
     *
     * @return bool
     */
    public function supportsException(Throwable $exception);
}
```

without requiring any changes to end-user code. 

And although someone could theoretically be calling it at runtime with something that's not a `Throwable`, in practice our current implementations would fail badly if you did, so we can be pretty sure people aren't.

For parameters that expect objects, it seems reasonable to say you shouldn't be passing anything that doesn't match the `@param`, so it's not a BC break if we start enforcing that.

[...]

PR #1763, opened 2025-12-08
[3.x] refactor: Add initial backwards-compatible return types

acoulton:
This is a first pass at adding backwards-compatible return types that we can safely merge into the 3.x series, as we work towards #1744.

The return types applied are based on:

* Rector's type coverage rules (up to level 21) - these are based on the actual type returned where this can be conclusively determined from static analysis.
* Manual modification to the phpdoc `@return` type of any interface the method implements, if this was less specific than the type that Rector detected. This is the type Rector would have used if the interface had a strict return type, but we cannot add those until 4.x.

I've committed this a level at a time, running phpstan at each commit and resolving any errors it produced (there were a few due to invalid / old-style phpdoc `@return` tags for lists / arrays etc that were previously ignored but now don't match the runtime return type.

Return types are added to 

* public, protected and private methods in `final` classes
* private methods in `non-final` classes
* `final public` methods in `non-final` classes

Rector mostly limited itself to these rules, but there were a few cases where by default it added types to non-final classes (I think because they are not extended anywhere by us, and the parent method has no strict type). I've excluded those changes.

~Note that this depends on #1762 and will be rebased after that is merged.~ now rebased


acoulton:
@carlos-granados This is quite a big PR, but hopefully keeping it as separate commits will make it easier to review.

rector.php
```diff
@@ -13,8 +21,86 @@
...
+            // Would be a BC break
+            __DIR__.'/src/Behat/Behat/Transformation/ServiceContainer/TransformationExtension.php',
+        ],
+        StringReturnTypeFromStrictScalarReturnsRector::class => [
+            // Would be a BC break
+            __DIR__.'/src/Behat/Behat/Snippet/ServiceContainer/SnippetExtension.php',
+            __DIR__.'/src/Behat/Behat/Transformation/ServiceContainer/TransformationExtension.php',
+            __DIR__.'/src/Behat/Testwork/EventDispatcher/ServiceContainer/EventDispatcherExtension.php',
+            __DIR__.'/src/Behat/Testwork/Hook/ServiceContainer/HookExtension.php',
+        ],
```
acoulton: I have kept the changes to Rector config in this branch for now so you can see "how the sausage was made".

I don't want to actually merge it into 3.x as I don't want any of these rules to be in our default set - we only want to run them manually under close supervision.

Once we're happy with the PR I will revert these config changes before we squash merge.

stof (approved):

carlos-granados:
@acoulton I'd like to take a good look at this PR, will try to do it this evening, please don't merge before I check it

acoulton:
> @acoulton I'd like to take a good look at this PR, will try to do it this evening, please don't merge before I check it

I'd be delighted if you did, very happy to wait for your review whenever you get a chance :)

carlos-granados (approved):
Finished reviewing these changes, outstanding work, many thanks @acoulton!

acoulton:
This is now updated to revert the temporary Rector config changes.

I will squash and release it after we release 3.28.0.

acoulton:
@carlos-granados I have added the changelog for 3.29.0 directly to this branch for simplicity. There are no changes to code (other than reverting the rector config) since you reviewed.

carlos-granados (approved):
Let's go ahead with this

PR #1764, opened 2025-12-09
refactor: convert rerun-only tests to use fixtures

carlos-granados:
In this case we have re-used the fixtures that we had for the `rerun.feature` tests as they were almost identical, we have just added a new feature because the behaviour when we were fixing the tests was slightly different

### DIFF
--- a/features/rerun-only.feature
+++ b/features/rerun-only.feature
@@ -4,92 +4,14 @@ Feature: Rerun
   I need to have an ability to rerun failed previously scenarios
 
   Background:
-    Given a file named "features/bootstrap/FeatureContext.php" with:
-      """
-      <?php
+    Given I initialise the working directory from the "Rerun" fixtures folder
+    And I provide the following options for all behat invocations:
+      | option   | value    |
+      | --no-colors |       |
+      | --format | progress |
 
-      use Behat\Behat\Context\Context;
-      use Behat\Step\Given;
-      use Behat\Step\Then;
-      use Behat\Step\When;
-
-      class FeatureContext implements Context
-      {
-          private $apples = 0;
-          private $parameters;
-
-          public function __construct(array $parameters = array()) {
-              $this->parameters = $parameters;
-          }
-
-          #[Given('/^I have (\d+) apples?$/')]
-          public function iHaveApples($count) {
-              $this->apples = intval($count);
-          }
-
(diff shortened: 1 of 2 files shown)


features/rerun-only.feature
```diff
@@ -4,92 +4,14 @@ Feature: Rerun
...
+      | option   | value    |
+      | --no-colors |       |
+      | --format | progress |
 
   Scenario: Run one feature with 2 failed and 3 passing scenarios
-    When I run "behat --no-colors -f progress features/apples.feature"
+    When I run "behat features/apples.feature"
     Then it should fail with:
       """
       ..F.............F....
```
acoulton: @carlos-granados sorry to keep picking on these rerun features.

I think the process you're working through is very helpfully showing where the noise of the setup in the past has made it hard to reason about what the features are actually covering.

I've realised this first scenario is IMO not a scenario at all. It is not testing anything about the rerun feature, only that the fixture we have created fails as expected.

This is actually a precondition of the later scenarios (where we start with the same `Given I run "behat features/apples.feature" but don't assert anything about the result).

The later scenarios assume that if we've run the same steps to setup the fixture, it will fail in the same way as this scenario did - but this is not actually guaranteed (there could have been some filesystem issue for example).

I think the content of this scenario should either be inlined as the first steps of the actual scenarios, or a `Background` if it is consistent for all scenarios (it isn't in this case). 

This would also apply to the feature we looked at the other day, and to #1766

features/rerun-only.feature
```diff
@@ -130,43 +52,22 @@ Feature: Rerun
     """
 
   Scenario: Fixing scenario removes it from the rerun log
```
acoulton: I think this scenario isn't named quite right for what it covers. `Run nothing if there were no previous failures` or something would make more sense?

acoulton: Or maybe this is actually two separate scenarios:

* If Behat has never run before, nothing runs (even with the unchanged `apples.feature` - Behat passes because that failing scenario is never executed).
* If Behat has run & failed, then was fixed, then `--rerun-only` passed, then the next run does not run any scenarios.

carlos-granados: Agree that this is better defined as two separate scenarios, I have updated the code

acoulton (approved):
thanks @carlos-granados that's a great improvement

PR #1765, opened 2025-12-09
refactor: convert syntax_help tests to use fixtures

carlos-granados:


src/Behat/Behat/Context/Annotation/DocBlockHelper.php
```diff
@@ -24,6 +24,9 @@ public function extractDescription(string $docBlock): string
         // Remove indentation
         $description = preg_replace('/^[\s\t]*/m', '', $docBlock);
 
+        // normalize line endings
```
carlos-granados, line 27: While running the tests in CI I discovered that the actual output was different in Windows machines. This had been hidden by the manipulation of line endings that we do on the tests but when using fixtures it became noticeable. The issue is that we later split this description into lines by a newline character, so we need to normalize the line endings in Windows in order for this functionality to behave equally in all OSs

acoulton: Good catch, thanks @carlos-granados.

It would be worth mentioning this change in the commit message when it's merged (though it should make no external difference, as you say I can see that we later explode by "\n" and then print each line using the system line ending).

acoulton (approved):
Thanks @carlos-granados

src/Behat/Behat/Context/Annotation/DocBlockHelper.php
carlos-granados: Added a comment in the extended description of the commit

PR #1766, opened 2025-12-09
refactor: convert rerun_strict tests to use fixtures

carlos-granados:
In this case we could not re-use the fixtures as they were very different from the existing ones

### DIFF
--- a/features/rerun_strict.feature
+++ b/features/rerun_strict.feature
@@ -4,44 +4,15 @@ Feature: Rerun with strict
   I need to have an ability to rerun failed previously scenarios, including those which failed due to the strict option
 
   Background:
-    Given a file named "features/bootstrap/FeatureContext.php" with:
-      """
-      <?php
-
-      use Behat\Behat\Context\Context;
-      use Behat\Step\When;
-
-      class FeatureContext implements Context
-      {
-          #[When('I have a failing step')]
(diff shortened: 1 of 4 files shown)


features/rerun_strict.feature
```diff
@@ -4,44 +4,15 @@ Feature: Rerun with strict
...
+    And I provide the following options for all behat invocations:
+      | option      | value    |
+      | --no-colors |          |
+      | --format    | progress |
 
   Scenario: Rerun feature without strict option
-    When I run "behat --no-colors -f progress -n features/rerun_strict.feature"
-    And I run "behat  --rerun --no-colors -f progress -n features/rerun_strict.feature"
+    When I run "behat features/rerun_strict.feature"
+    And I run "behat --rerun features/rerun_strict.feature"
```
acoulton, line 28: See the comments on #1764 - in this feature we don't have a separate scenario that proves the fixture is broken at first (which is correct IMO).

However our initial `When I run "behat features/rerun_strict.feature"` in both scenarios should assert the expected result & output so that we can see how it differs on the second `--rerun` execution.

carlos-granados: Updated

acoulton (approved):
Perfect, thanks @carlos-granados

PR #1768, opened 2025-12-09
refactor: convert snippets tests to use fixtures

carlos-granados:


features/bootstrap/FeatureContext.php
```diff
@@ -314,7 +314,7 @@ public function iRunBehatInteractively($answerString, $argumentsString)
 
         $this->answerString = $answerString;
 
-        $this->options = '--format-settings=\'{"timer": false}\'';
+        $this->options = '--no-colors --format-settings=\'{"timer": false}\'';
```
carlos-granados, line 317: This step overrides any option set for behat previously, so we need to make sure that it uses the no-colors option

acoulton: Makes sense - but would it be clearer if it didn't override the options (so it behaves consistently with the normal "I run Behat" step? I suppose that might impact on other scenarios that are depending on this behaviour, so we could leave it for now,

acoulton (approved):
Thanks @carlos-granados one small comment / query but happy with it as-is if you are.

PR #1769, opened 2025-12-09
refactor: convert suite tests to use fixtures

carlos-granados:


features/suite.feature
```diff
@@ -588,22 +147,8 @@ Feature: Suites
...
-      class FeatureContext implements \Behat\Behat\Context\Context
-      {
-      }
-      """
-    When I run "behat --no-colors -f progress --suite suite-with-hyphens"
+    Given I clear the default behat options
+    When I run "behat --config=behat-hyphens.php --format progress --suite suite-with-hyphens"
     Then it should pass with:
       """
       No scenarios
```
acoulton: This obviously matches the original, but it doesn't necessarily prove that we have actually correctly matched the `suite-with-hyphens` suite name.

Particularly since we don't appear to have any coverage that it fails with an error if you specify a suite name that does not exist. We could theoretically be falling back to the first / default / all suites in this situation (we're not, but that could change in the future).

I **think** it might actually be viable to just rename the `second` suite in the existing `behat-two-features.php` config file to `suite-with-hyphens` and just reuse those examples to prove that we've selected the correct features to run. I don't think any of the other scenarios here depend on the naming of that suite.

Equally, these are quite minor/theoretical concerns so feel free to ignore & move on!

carlos-granados: Yeah, I see what you mean. I modified the configuration for this suite so that it actually runs some tests and then confirmed that they were run

acoulton (approved):
Thanks @carlos-granados

PR #1770, opened 2025-12-09
refactor: simplify rerun.feature by moving setup to Background

carlos-granados:
@acoulton seeing your comments in https://github.com/Behat/Behat/pull/1764 and https://github.com/Behat/Behat/pull/1766 I agree with what you exposed there and I have decided to start by updating the original `rerun` test to remove the scenario that is unnecessary and to make it more explicit the results of the initial run. Once this is merged, I will update the other PRs

### DIFF
--- a/features/rerun.feature
+++ b/features/rerun.feature
@@ -9,8 +9,6 @@ Feature: Rerun
       | option      | value    |
       | --no-colors |          |
       | --format    | progress |
-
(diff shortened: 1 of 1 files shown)


stof (approved):

acoulton (approved):
Great, thanks for doing this @carlos-granados

PR #1771, opened 2025-12-09
refactor: convert rerun multiple suite tests to use fixtures

carlos-granados:
These tests also re-use the fixtures in the `Rerun` folder. The context file has been modified a little so we have needed to modify existing tests a little bit as well

### DIFF
--- a/features/rerun_with_multiple_suite.feature
+++ b/features/rerun_with_multiple_suite.feature
@@ -10,8 +10,8 @@ Feature: Rerun with multiple suite
       | --no-colors |          |
       | --format    | progress |
 
-  Scenario: Run feature with 6 failed and 12 passing scenarios
-    When I run "behat"
+  Scenario: Rerun only failed scenarios, 4 from default suite and 2 from suite2
+    Given I run "behat"
     Then it should fail with:
       """
       ..F.............F......F.............F......F.............F....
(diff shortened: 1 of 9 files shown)


features/rerun_with_multiple_suite.feature
```diff
@@ -211,77 +81,96 @@ Feature: Rerun with multiple suite
     """
 
   Scenario: Fixing scenarios removes it from the rerun log
-    Given I run "behat --no-colors -f progress features/apples.feature"
-    And there is a file named "features/bananas.feature" with:
+    Given I run "behat features/apples.feature"
```
acoulton: I'm not sure I follow why we run only the apples feature, then immediately copy the `bananas-fixed` feature over the top of the `bananas` feature (without ever having run the bananas feature) and then run both features?

carlos-granados: The original test was like this buy yeah, agree that it does not make much sense. I have updated it so that the first step runs both features, I think it makes much more sense like that

acoulton (approved):
Thanks @carlos-granados yeah, I saw that was how the original feature did it. It's much clearer now.

Thanks for driving all this forward - it makes it so much easier to understand what we are actually testing!

PR #1773, opened 2025-12-10
refactor: convert tag_filters tests to use fixtures

carlos-granados:


acoulton (approved):

PR #1774, opened 2025-12-10
refactor: convert traits tests to use fixtures

carlos-granados:
This is the last test that needed to be converted to use fixtures! 🎉

### DIFF
--- a/features/traits.feature
+++ b/features/traits.feature
@@ -4,86 +4,13 @@ Feature: Support traits
   I need to be able to use definition traits in my context
 
   Background:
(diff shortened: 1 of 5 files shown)


acoulton (approved):
Brilliant work @carlos-granados - thanks for all the effort on all of these!

You have definitely earned a beer / cake / reward of your choice 🥳

acoulton:
Once you've merged this I'll prepare a release (probably tomorrow) with all the recent changes. 

Then I'll update and merge #1763 with the strict return types and we can release it as a standalone release so that it's easy for people to roll back / past it if there's any issues with it (not that I expect any).

PR #1777, opened 2025-12-11
Update all test files to use attributes instead of annotations

carlos-granados:
- Updates our main FeatureContext.php file to use Behat attributes instead of annotations everywhere
- Review all our fixtures and use attributes everywhere except in those cases where we specifically want to test annotations


acoulton (approved):
Thanks @carlos-granados

PR #1778, opened 2025-12-17
[3.x] refactor: Improve type-safety of regex string replacement operations

acoulton:
While starting work on the next iteration of adding strict return types, Rector picked up a number of methods that had phpdoc of `@return string` but could theoretically also have been returning `null`.

This is because `preg_replace` and `preg_replace_callback` can return null in some cases (e.g. if the pattern is valid but it hits the runtime backtrack_limit). We were not always handling this - sometimes casting back to empty string, sometimes just assuming the result would be a string.

This can also happen if there is an issue with the regex itself (e.g. the pattern is invalid) and the runtime PHP error is not captured by an error handler that converts it to an exception - generally in Behat we do, but the static analyser cannot detect this.

This commit introduces thin wrapper functions to ensure that the preg_replace / preg_replace_callback was successful. This will reduce the need for assertions elsewhere in the code as we move towards stricter typing.

Note that the native PHP functions can also return an `array` if passed an `array $subject` however we never use this feature so I have limited the utility functions to `string $subject` and ` string` return type. If we needed to support arrays in future I'd be inclined to do that as separate methods (e.g. `replaceArray`) so that we only have to handle array return types in places we expect them.

### DIFF
--- a/src/Behat/Behat/Util/RegexException.php
+++ b/src/Behat/Behat/Util/RegexException.php
@@ -1,0 +1,14 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Behat\Behat\Util;
+
+use Exception;
+
+/**
+ * @internal
+ */
+class RegexException extends Exception
+{
+}
--- a/src/Behat/Behat/Util/StrictRegex.php
+++ b/src/Behat/Behat/Util/StrictRegex.php
@@ -46,7 +46,8 @@ final class StrictRegex
      *
      * @throws RegexException if the regex fails
      */
-    public static function replaceCallback(array|string $pattern,
+    public static function replaceCallback(
+        array|string $pattern,
         callable $callback,
         string $subject,
     ): string {
--- a/src/Behat/Behat/Util/StrictRegex.php
+++ b/src/Behat/Behat/Util/StrictRegex.php
@@ -41,6 +41,24 @@ final class StrictRegex
     }
 
     /**
+     * @param string|list<string> $pattern
+     * @param callable(array<mixed,string>):string $callback
+     *
+     * @throws RegexException if the regex fails
+     */
+    public static function replaceCallback(array|string $pattern,
+        callable $callback,
+        string $subject,
+    ): string {
+        $result = self::callWithErrorCapture(fn (): ?string => preg_replace_callback($pattern, $callback, $subject));
+        if ($result === null) {
+            throw new RegexException('Regex failed: '.preg_last_error_msg(), preg_last_error());
+        }
+
+        return $result;
+    }
+
+    /**
      * @template T of mixed
      *
      * @param callable():T $callable
--- a/src/Behat/Behat/Util/StrictRegex.php
+++ b/src/Behat/Behat/Util/StrictRegex.php
@@ -1,0 +1,59 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Behat\Behat\Util;
+
+use InvalidArgumentException;
+
+/**
+ * @internal
+ */
+final class StrictRegex
+{
+    /**
+     * @param string|list<string> $pattern
+     * @param string|list<string> $replacement
+     *
+     * @throws RegexException if the regex fails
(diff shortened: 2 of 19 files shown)


src/Behat/Behat/Util/RegexException.php
```diff
@@ -0,0 +1,14 @@
...
+declare(strict_types=1);
+
+namespace Behat\Behat\Util;
+
+use Exception;
+
+/**
+ * @internal
+ */
+class RegexException extends Exception
```
acoulton, line 12: If the preg_xxx functions generate a PHP warning / error, it's difficult to tell whether this is a runtime exception (e.g. due to runtime input) or a code issue (e.g. an invalid regex pattern). I have therefore just made our exception extend `Exception`.

src/Behat/Behat/Util/StrictRegex.php
```diff
@@ -0,0 +1,77 @@
...
+     * @throws RegexException if the regex fails
+     * @throws InvalidArgumentException if passing pattern & replacements as arrays and the number of entries does not match
+     */
+    public static function replace(
+        string|array $pattern,
+        string|array $replacement,
+        string $subject,
+        int $limit = -1,
+    ): string {
+        if (is_array($pattern) && is_array($replacement) && count($pattern) !== count($replacement)) {
```
acoulton, line 27: When passing an array of `$pattern`, native PHP allows three cases:
* Pass a string `$replacement` which will be used for all matches (we use this in some places to change multiple patterns to `''`
* Pass an array of `$replacement` with one entry for each `$pattern` (we use this)
* Pass an array of `$replacement` which has fewer entries than there are `$pattern`s. In this case, any extra patterns will be replaced with an empty string. (We do not use this).

I think it's better to be explicit - if you want everything replaced the same then passing a single string is fine, but if you pass an array you should pass exactly one replacement for each pattern.

src/Behat/Behat/Util/StrictRegex.php
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * @param string|list<string> $pattern
+     * @param callable(array<mixed,string>):string $callback
+     *
+     * @throws RegexException if the regex fails
+     */
+    public static function replaceCallback(array|string $pattern,
+        callable $callback,
+        string $subject,
```
stof, line 52: why this weird splitting of the method signature which is not fully multiline ? This does not match the PSR-12 code styles, which expects being either single-line or having one argument per line, doing it fully.

tests/Behat/Tests/Util/StrictRegexTest.php
```diff
@@ -0,0 +1,213 @@
...
+
+use BadMethodCallException;
+use Behat\Behat\Util\RegexException;
+use Behat\Behat\Util\StrictRegex;
+use InvalidArgumentException;
+use PHPUnit\Framework\TestCase;
+
+final class StrictRegexTest extends TestCase
+{
+    private string $original_backtrack_limit;
```
stof: I would use camelCase to match our coding standards.

src/Behat/Behat/Util/StrictRegex.php
acoulton: Accidental - started as a copy & paste from the native php stub, wasn't auto-formatted correctly & I didn't spot it. Thanks for flagging.

tests/Behat/Tests/Util/StrictRegexTest.php
acoulton: You're absolutely right, sorry, had my head in a different project earlier that is using snake_case and hadn't quite woken up!

acoulton: I even managed to fix a pre-existing example of exactly this in b34f7e4 without noticing I'd added a new one :laughing: :facepalm: 

Those style issues both fixed now.

carlos-granados (approved):
Good idea and good implementation (first time I've seen first class callables doing something useful)

acoulton:
Thanks @carlos-granados

PR #1779, opened 2025-12-17
[3.x] refactor: Add more BC types to method returns and private properties

acoulton:
This is a second pass at adding backwards-compatible return types that can be safely merged into the 3.x series, as work towards #1744 

As previously, I used Rector to apply type coverage rules (up to level 32) and manually reviewed / filtered the commits to exclude changes that would have been a BC break. I also manually improved a couple of signatures where we know more about the return type than Rector was able to detect.

As part of this, Rector identified that some private properties are always initialised from the constructor with a strict-typed parameter. It was therefore able to add property typehints (and in some cases `readonly`, unless the property can also be assigned another way). Some of these are in non-final classes, but as the property is private there's still no BC issue here.

It looks like this PR gets us about as far as Rector can go with return types at the moment - level 33 starts to introduce strict typing to method parameters (starting with private methods). I'll look at that in a separate PR.


src/Behat/Behat/Tester/Result/ExecutedStepResult.php
```diff
@@ -47,17 +49,17 @@ public function getCallResult(): CallResult
...
     }
 
     public function hasException(): bool
     {
-        return null !== $this->getException();
+        return $this->getException() instanceof Exception;
     }
 
-    public function getException()
+    public function getException(): ?Exception
```
stof, line 62: should this be Exception or Traversable ? How are we handling `Error` in Behat ?

carlos-granados: Do you mean Throwable?

stof: yeah indeed.

acoulton: https://github.com/Behat/Behat/pull/833 introduced support for `Throwable` by wrapping it in an `Exception` (to avoid a BC break on all the things that are typed as taking / returning `Exception`).

So this is correct for 3.x - and I think probably 4.x, we could consider a change in future but I'm not sure it's worth the BC break for now.

carlos-granados:
@acoulton As with your previous PR for adding types, please let me review this one carefully. I don't think I will be able to check it today, will try to do it tomorrow

acoulton:
> @acoulton As with your previous PR for adding types, please let me review this one carefully. I don't think I will be able to check it today, will try to do it tomorrow

@carlos-granados no problem at all.

carlos-granados (approved):
Finally found the time to look into this (crazy Xmas....). I looked into all the individual commits, thanks for splitting them up. Everything looks fine

acoulton:
Thanks very much @carlos-granados (and feliz año nuevo!) - no worries, the holidays were pretty busy for me too.

I won't have much time the next few days but hopefully I'll get a chance to work on the next iteration at some point next week.

PR #1780, opened 2025-12-18
[3.x] refactor: Use first-class callables for internal callbacks

acoulton:
We still had a few uses of array callables within our internal code, which I found when starting to look at adding parameter types to internal methods.

Rector has a rule to fix these as part of the PHP 8.1 set that we are running, but it had skipped over them because they were not public methods, even though they are defined and used in their own scope. 

I think this may actually be a Rector bug - see https://github.com/rectorphp/rector-src/pull/7760 for the patch which I have temporarily used locally to convert these.

Refactoring them to first-class callables will help Rector (and possibly phpstan) to make better decisions about parameter & return types in future.


carlos-granados (approved):
Nice!

PR #1783, opened 2026-01-17
fix: Normalize PHP file line numbers in test output to avoid fragile tests

carlos-granados:
There are several places in our tests where we check the expected output and it contains line numbers of php files. This makes these tests fragile because whenever we update those php files, for example when doing a refactor, these line numbers may change. And the actual line numbers in php files are unimportant for our tests

This PR adds a function that convert these line numbers in the output to a placeholder and updates our feature files to use this placeholder. 

This PR does not change the expected line numbers when they are related to feature files. These line numbers are more stable and provide information that may be important


acoulton (approved):
Thanks @carlos-granados

Issue #1743, opened 2025-12-04
4.0 release planning / co-ordination issue


acoulton:
We have started work towards a 4.0.0 release.

We'd hoped to start a new major for a while (there has not been a major release since 3.0.0 in 2014). However the primary motivation for doing this now is that symfony 8.0 incudes BC breaks in the symfony interfaces we extend. Therefore, we cannot support symfony >= 8.0 with Behat 3.x, and this is holding back users who want to upgrade to the newer symfony version.

Therefore the goals for 4.0 are:

* To stick to a scope that allows us to deliver a release relatively soon, even if there are other breaking changes we want to make in future.
* To make it as easy as possible for extension authors to update to 4.0 - most packages should be able to update primarily using automated tooling (e.g. Rector) although some manual changes may also be required.
* To make it as easy as possible for end-users to update to 4.0 - the majority of users should only need to bump their dependency constraints, and perhaps run automated tooling that we provide to update their projects.
* As much as we can within the constraints above, to take the opportunity to make other breaking changes (primarily dropping deprecated features and code and support for edge cases and unexpected usage), to make it easier to work towards subsequent 4.x releases and an eventual v5.

Ensuring that 4.0 is a frictionless update for extensions and end-users is important, because it will allow us to set a relatively short EOL for 3.x. That minimises the overhead of maintaining multiple versions in parallel, particularly when we want to start working on v5.

I propose that we split this into (at least) two releases:

## [4.0.0-alpha1](https://github.com/Behat/Behat/milestone/5)

This will be an "essentially stable" alpha containing all the planned breaking changes that are likely to affect extension authors (strict types, formally marking some classes as `@internal`, etc). 

4.0.0-alpha1 should also be stable enough for end-users to run in CI.

This will provide:
* A clear base for extension authors to update their own code. 
* A chance for us to respond to feedback (e.g. unexpected challenges upgrading) before things are locked down.
* Slightly less time pressure to complete everything that we want to get into 4.0 - users who are keen to bump to symfony8 should be able to start using 4.0.0-alpha1 as soon as we release it.

## [4.0.0](https://github.com/Behat/Behat/milestone/4)

Breaking changes that are unlikely to directly affect extension authors or expected usage, or that remove deprecated functionality, do not need to be prioritised for alpha1. 

These can wait and be shipped before the final 4.0.0 release - so long as users know what's coming, they should be able to update their projects against alpha1 before we actually complete all the deletions / internal refactoring.

acoulton:
[...]

PR #1785, opened 2026-02-07
Allow Symfony 8

Jean85:


Jean85:
Tests executed locally, all green

acoulton:
@Jean85 thanks for the contribution but unfortunately we can't merge this to 3.x as it contains breaking changes.

We are [working towards a Behat 4.0 release](https://github.com/Behat/Behat/issues/1743) which already has support for symfony8 in the 4.x dev branch.

The challenge is that if you're using any third-party Behat extensions they will not yet officially support Behat 4.0 (as we have not yet finalised the 4.0.0-alpha1 release, so cannot yet tell extension authors what changes they need to make).

If you want to live dangerously, you *might* be able to have composer require behat at `4.x-dev as 3.99.99` to work around dependency constraints. It's impossible to say whether that will run without problems - it will depend on the other packages you're using and your own code.

Otherwise I'm afraid you'll need to wait for Behat 4.0 to get Symfony 8.0 support.

PR #1786, opened 2026-02-08
fix: apply coding style fixes

carlos-granados:
The latest PHP-CS-Fixer has added some changes to their rules which results in some changes being applied to our code base, these are mainly converting protected members to private in final classes. This PR adds these changes to our code base

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
     "require-dev": {
         "opis/json-schema": "^2.5",
         "php-cs-fixer/shim": "^3.89",
-        "phpstan/phpstan": "^2.0",
+        "phpstan/phpstan": "2.1.18",
         "phpunit/phpunit": "^9.6",
         "rector/rector": "2.1.7",
         "sebastian/diff": "^4.0",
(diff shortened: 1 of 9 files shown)


carlos-granados:
The BC breaks checker is detecting these changes as BC breaks but they are not actually BC breaks, the classes are final so there should not be any other class using these protected methods which have been converted to private, I think we can ignore this error

carlos-granados:
We had pinned our Rector version but not our PHPStan version so now Rector was erroring because it did not work with the latest PHPStan version. These two dependencies need to be updated in lock-step. I will update them to their latest versions in a separate PR, for now I have pinned PHPStan to a version which is compatible with our Rector version

acoulton:
Thanks @carlos-granados the problem with merging with an expected failure on the BC job is that it always checks the whole project against the last release (not just the current PR changes) so it will continue to fail on other branches meaning we might miss new issues.

I've raised a [bug / fix on the bc-checker](https://github.com/Roave/BackwardCompatibilityCheck/pull/986) so we can either wait for that to be released, or temporarily add these failures into the [roave-backward-compatibility-check.xml](https://github.com/Behat/Behat/blob/3.x/.roave-backward-compatibility-check.xml) ignore list with a ref to that PR.

carlos-granados:
@acoulton your fix on the BC-checker was merged and released yesterday (that was fast!). I have run the bc-compatibility-check again and it is now passing, so this should be ready to go

acoulton (approved):
That was fast indeed, great they got it released so quickly. This all looks good to me, thanks @carlos-granados

PR #1787, opened 2026-02-10
Add MCP server

carlos-granados:
Uses the PHP-MCP Server library (https://github.com/php-mcp/server) to add an MCP server to Behat. This can be started with the new `--mcp-server` command line option. When this option is passed we start a new MCP server which provides a tool that can be used to run Behat tests.

This MCP server can use a stdio transport, ideal for local installations or an HTTP transport, ideal for remote or docker installations. This can be chosen with the `--mcp-transport` command line option which defaults to stdio. If HTTP is chosen, you can specify the host and port using the `--mcp-host` and `--mcp-port` command line options

The tool accepts a number of optional parameters which can be used to select a different config file or which can be used to run a specific profile, suite, feature or scenario


features/bootstrap/FeatureContext.php
```diff
@@ -95,6 +118,29 @@ public function prepareTestFolders(): void
         $this->phpBin = $php;
     }
 
+    #[AfterScenario]
+    public function stopMcpServer(): void
```
carlos-granados, line 122: This function stops any running mcp servers once each scenario is finished. This guarantees that no dangling process is left even if any of the MCP tests fail

features/bootstrap/FeatureContext.php
carlos-granados, line 1: The implemetation of these tests was inspired by the tests available in the PHP-MCP library which show how to start, stop and communicate with these servers from testing tools

src/Behat/Testwork/Cli/McpServerCommand.php
```diff
@@ -0,0 +1,56 @@
...
+namespace Behat\Testwork\Cli;
+
+use PhpMcp\Server\Server;
+use PhpMcp\Server\Transports\StdioServerTransport;
+use PhpMcp\Server\Transports\StreamableHttpServerTransport;
+use Symfony\Component\Console\Command\Command as BaseCommand;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+final class McpServerCommand extends BaseCommand
```
carlos-granados, line 20: This command is run if the `--mcp-server` command line option is added to the behat command

src/Behat/Testwork/Cli/McpServerCommand.php
```diff
@@ -0,0 +1,56 @@
...
+        parent::__construct('mcp-server');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $server = Server::make()
+            ->withServerInfo('Behat MCP Server', '1.0.0')
+            ->build();
+
+        $server->discover(
```
carlos-granados, line 33: This library can auto discover and configure available tools using PHP attributes

src/Behat/Testwork/Cli/McpServerCommand.php
```diff
@@ -0,0 +1,56 @@
...
+            basePath: dirname(__DIR__),
+            scanDirs: ['Mcp']
+        );
+
+        $transportType = $input->getOption('mcp-transport');
+
+        if ($transportType === 'http') {
```
carlos-granados, line 40: We can use two different transports, stdio or http

src/Behat/Testwork/Mcp/McpTestRunner.php
```diff
@@ -0,0 +1,87 @@
...
+namespace Behat\Testwork\Mcp;
+
+use Behat\Behat\ApplicationFactory;
+use PhpMcp\Schema\ToolAnnotations;
+use PhpMcp\Server\Attributes\McpTool;
+use PhpMcp\Server\Attributes\Schema;
+use Symfony\Component\Console\Input\ArrayInput;
+use Symfony\Component\Console\Output\BufferedOutput;
+
+final class McpTestRunner
```
carlos-granados, line 14: This is the class that provides the tool provided by the MCP server. It uses PHP attributes to define the tool and its parameters

src/Behat/Testwork/Mcp/McpTestRunner.php
```diff
@@ -0,0 +1,87 @@
...
+        #[Schema(description: 'Name of a suite to use (optional)')]
+        ?string $suite = null,
+        #[Schema(description: 'List of paths to execute (optional)', items: ['type' => 'string'])]
+        ?array $paths = null,
+        #[Schema(description: 'Additional command-line options as key-value pairs (optional)', type: 'object', additionalProperties: true)]
+        ?array $additionalOptions = null,
+    ): array {
+        $outputFile = tempnam(sys_get_temp_dir(), 'behat_output_');
+
+        $factory = new ApplicationFactory();
```
carlos-granados, line 43: We build a new instance of the Behat application using the passes parameters and run it. This behat instance will use the JSON formatter

src/Behat/Testwork/Mcp/McpTestRunner.php
```diff
@@ -0,0 +1,87 @@
...
+            $inputArgs['--profile'] = $profile;
+        }
+
+        if ($suite !== null) {
+            $inputArgs['--suite'] = $suite;
+        }
+
+        if ($additionalOptions !== null) {
+            foreach ($additionalOptions as $option => $value) {
+                if (!in_array($option, self::PROTECTED_OPTIONS, true)) {
```
carlos-granados, line 67: We don't allow overriding the `format` or `out` options as these need to be set by us to use the JSON format output to a specific file

src/Behat/Testwork/Mcp/McpTestRunner.php
```diff
@@ -0,0 +1,87 @@
...
+
+        $input = new ArrayInput($inputArgs);
+        $output = new BufferedOutput();
+
+        $application->run($input, $output);
+
+        $result = file_get_contents($outputFile);
+        unlink($outputFile);
+
+        return json_decode($result, true);
```
carlos-granados, line 85: The library needs that our tool returns an array with the result. So we load the generated json file, decode it and return this array as the result. The library will then encode it in the way needed to return it as an MCP response

composer.json
```diff
@@ -34,6 +35,9 @@
         "php-cs-fixer/shim": "^3.89",
         "phpstan/phpstan": "2.1.18",
         "phpunit/phpunit": "^9.6",
+        "react/async": "^4.3",
```
carlos-granados, line 38: Used in the tests to interact with the MCP server

carlos-granados:
There seem to be three tests failing:
- Windows
- Lowest dependencies
- Symfony 5.4
Will need to investigate each of them

acoulton:
@carlos-granados thanks very much for working on this - I haven't had a proper look, but I think it could be a useful thing to have.

My immediate concern is that adding `php-mcp/server` as a runtime dependency brings in a significant number of new dependencies (I make it 21 packages) including some that are widely used but possibly with different constraints in people's own projects / other dependencies.

I'd be nervous that making this a core requirement in `composer.json` will increase the risk of dependency conflicts that e.g. prevent end-users getting subsequent point versions of `behat/behat`, even if they don't want to use the MCP server features. I think a test framework should ideally have minimum third-party dependencies, to make it as easy as possible to use it in a wide range of projects.

My gut feel is that this would be better as an opt-in Behat extension, rather than a core feature?

The other advantage of it being an extension is it's much easier to bump the extension's major version (e.g. to support breaking changes in a future php-mcp/server major) than it would be to bump the whole Behat/Behat version and roll that out across the Behat ecosystem.

What do you think?

carlos-granados:
@acoulton yeah, I see your point of view. We also seem to have the issue that this library does not seem to support some of the lowest versions that we support like Symfony 5.4. I'll look into moving this into an independent extension

carlos-granados:
@acoulton in the end I have implemented this as a fully autonomous tool in this repo https://github.com/Behat/mcp-server I think it is what makes more sense.

I have made you and @stof admins of that repo. If any of you would prefer not to be included in that repo, please let me know

Closing this

Issue #1698, opened 2025-11-20
[4.x] Drop support for defining steps, hooks, transforms etc with annotations


acoulton:
I propose that we drop support for using phpdoc `@Given`, `@BeforeStep` etc annotations in feature contexts for version 4.0, in favour of only using PHP attributes to define these. This will allow us to simplify both the runtime codebase and our feature coverage.

This would mean that end-users still using annotations would have to update their `Context` files to use attributes as part of the 4.0 upgrade. 

However, Rector has built-in and robust support for doing this automatically across an entire project. We can provide users with guidance on that as part of the release notes.

acoulton:
@stof fair point. 

Although I also think we should try to keep user-facing changes in 4.0 to the bare minimum, and ideally only things that are global changes they can easily resolve with automated tooling (e.g. this and #1699). 

As discussed, 4.0 might have more impact on extension authors (e.g. some of the strict typing / constructor changes) but once those are resolved it'd be great if we could get as many users as possible onto 4.0 as quickly as possible. The sooner we get 4.0 out, the sooner we can drop 3.x and work entirely in the modernised / strict-typed codebase. That will give us a much better baseline for considering more user-facing changes for 5.0 and beyond.

carlos-granados:
Closed via https://github.com/Behat/Behat/pull/1790

PR #1790, opened 2026-02-12
feat: remove annotation support for steps, hooks, and transformations

carlos-granados:
I believe we need to try to release version 4 as soon as possible as currently any user who wants to update to Symfony 8 is unable to do so if they are using Behat. So I'm going to be working on some of the things that were planned for version 4

In this version annotations will no longer be supported for defining steps, hooks, and transformations. Users will need to use PHP 8 attributes instead. This PR removes all handling of annotations

Removed:
- AnnotationReader interface and AnnotatedContextReader
- DefinitionAnnotationReader, HookAnnotationReader, TransformationAnnotationReader
- All annotation-related test fixtures and feature files
- ANNOTATION_READER_TAG constant from ContextExtension

Updated:
- Service container extensions to remove annotation reader loading
- Test fixtures to remove annotation profiles

Fixes https://github.com/Behat/Behat/issues/1698


carlos-granados:
I did a rebase and must have done something wrong, please let me review what I broke

carlos-granados:
OK, during the rebase somehow I lost a commit. I have recovered it and this is now ready. Regarding the failures:
- The CS and Rector failures need that we merge the latest 3.x into 4.x, @acoulton can you please take care of this? Once this is done, I will rebase this branch
- Not really sure what we need to do about the BC breaks. @acoulton can you advice? Thanks!!

stof:
Getting a job saying there are BC break compared to the latest release is quite expected when you work on preparing the new major version (the fact that we have BC breaks is precisely the reason why this branch is now `4.x`).

The tool checking for BC break might need to be improved to handle the case when the branch-alias for the current branch is at a higher major version than the latest release

acoulton:
I'll try and do the merge up at the weekend, we have the funeral tomorrow.

My thinking was we would add known BC breaks to the ignore file until we get to 4.0, that will give us a clear set of changes (and still flag any unexpected ones) when we come to do releases/release notes.

features/definitions_transformations_annotations.feature
```diff
@@ -1,159 +0,0 @@
...
-
-      4 scenarios (4 passed)
-      12 steps (12 passed)
-      """
-
-  Scenario: Row Table Arguments Transformations
-    When I run behat with the following additional options:
-      | option         | value                             |
-      | --suite        | row_table_argument_transformation |
-    Then it should pass with:
```
acoulton, line 42: These scenarios (and a few others in this file) don't seem to be mirrored in `definitions_transformations_attributes` so I think we might be losing coverage of some types of transformations?

acoulton (commented):
Thanks very much @carlos-granados, just one thought (other than the rebasing & ignoring the BC breaks that you've already mentioned).

features/definitions_transformations_annotations.feature
carlos-granados: You're absolutely right, let me work to add them back

carlos-granados:
@acoulton missing tests added back and BC break exceptions added as well. This is now ready for final review

acoulton (approved):
Perfect, thanks very much @carlos-granados

Issue #1699, opened 2025-11-20
[4.x] Drop support for reading and converting YAML configuration


acoulton:
I propose that we drop support for YAML configuration in 4.0, in favour of the newer PHP configuration files.

We have [already documented that YAML configuration will likely be removed in future](https://docs.behat.org/en/latest/user_guide/configuration/yaml_configuration.html), and provided a tool for users to convert existing YAML to the new PHP configs.

Dropping this in 4.0 would mean we could also drop the `--convert-config` command and associated code. This will simplify our config objects, remove the runtime dependency on `nikic/php-parser`, and make it easier to add new config options and features in future.

The DX of the new PHP config is also better for end-users, since the configuration objects are somewhat self-documenting.

This would require users to run the `--convert-config` command **before** they upgrade from 3.x to 4.0. They will need to spend a little time reviewing the automated migration, although it has proved robust in most common cases.

It may also have an impact on extension authors if they are depending on YAML config directly - but in most cases I would expect they are interacting with config after it is parsed by Behat so there should be limited impact. Generally I think they might only want to update any example configurations in their documentation.

I think the benefits significantly outweigh the - likely small - potential inconvenience to end users of being forced to run the conversion tool before upgrading.

acoulton:
If we do drop YAML config then we should also consider what to do with the `--config-reference` command option that I've just discovered thanks to #1713 😆 

It currently prints out all supported config nodes (including any registered by extensions) as YAML... I'm not sure it's worth keeping / rewriting the code to dump a reference config in PHP, and perhaps if the config is expressed as self-documenting PHP objects the command to dump it at runtime is not that crucial?

carlos-granados:
Closed via https://github.com/Behat/Behat/pull/1791

PR #1791, opened 2026-02-14
feat: drop support for YAML config and finding files in `config` dir

carlos-granados:
- Disables support for loading YAML config files
- Disables support for automatically finding config files in a `config` directory - we will now only automatically load `behat.php` or `behat.dist.php` from the root directory.
- Removes the Convert Config command and all the associated functions in the config objects
- Removes the dependency on php-parser
- Removes the Config Reference command as it just printed a reference of the Yaml config
- Modifies all existing tests and fixtures to use PHP config files instead of YAML files

Fixes https://github.com/Behat/Behat/issues/1699


carlos-granados:
Again the CS and Rector tests will be fixed once we rebase on an updated 4.x version. And we'll have to add the BC changes to the BC tool configuration

features/init.feature
```diff
@@ -36,17 +48,32 @@ Feature: Init
...
+      return (new Config())
+          ->withProfile(
+              (new Profile('default', [
+                  'autoload' => '%paths.base%/contexts',
+              ]))
+                  ->withSuite(
+                      (new Suite('suite1', [
+                          'paths' => ['%paths.base%/scenarios1'],
+                          'contexts' => ['Custom1Context'],
+                      ]))
```
acoulton: Should we make these (and the others in the file) use the `withPaths` and `withContexts` helper methods to show our recommended usage, rather than passing in the raw config arrays?

features/config.feature
```diff
@@ -259,14 +170,3 @@ Feature: Config
       """
       behat.php
       """
-
-  Scenario: Prioritize config file from root
-    Given a file named "behat.yaml.dist"
-    Given a file named "config/behat.yaml"
```
acoulton, line 265: Does anything actually prove that we load a file from e.g. `config/behat.php`? I don't think I see a feature covering that, or that the root takes precedence if both are present.

I almost wonder whether for 4.0 we should simplify the list of "automagic" config locations down to just the root, and drop support for finding a file in `config/`?

composer.json
acoulton, line 28: I might be wrong, but I think we no longer have a direct dependency on symfony/yaml with this PR?

acoulton (commented):
Thanks very much @carlos-granados

composer.json
carlos-granados: We do need this dependency because we still allow loading translations from yaml files. Without this dependency we get an error from the translation component: "Loading translations from the YAML format requires the Symfony Yaml component"

features/config.feature
carlos-granados: Yeah, I think we should drop support for the `config` folder. This seems like an idea inherited from Symfony but no other tool of the PHP ecosystem supports this, so I think it is a sound idea to remove this

features/init.feature
carlos-granados: Yeah, good shout

carlos-granados:
@acoulton code updated and BC exceptions added to the base line, ready for final review

acoulton (approved):
Great, thank you @carlos-granados - I'll updated the PR title & description to reflect us removing support for finding files in `config` folder as well.

Issue #1754, opened 2025-12-04
[4.x] Remove support for "automagically" guessing extension class names from short names


acoulton:
As discussed in https://github.com/Behat/docs/pull/146#discussion_r1820595587 there are historically multiple ways to identify an extension to load, which are not particularly well documented (see also Behat/docs#147).

One of the options is the "short name" which works by `ExtensionManager` attempting to guess the extension class name from the short name e.g. if the user configured `Vendor\Something` we would try to find `Vendor\Something\ServiceContainer\SomethingExtension`.

This was useful in the past to keep verbosity out of the YAML files. However, it came at the cost of some confusion for end users and extension authors. It also forces people to put their main extension class one level down in their package namespace, when it might be more useful / obvious to users if it was in the package root.

Now that we have PHP config files, we've solved the verbosity issue. Users can just use `SomethingExtension::class` and add a `use` statement like for any other class reference. Using a fully qualified class reference also allows you to autocomplete the class name when adding it, and click through from your config file to the extension entrypoint.

Therefore I propose that 4.0 should remove support for guessing extension class names from a short name. The only supported options will be a fully qualified class name, or a path to a file that is absolute, relative to the working directory, or relative to the extensions directory.

stof:
to me, this depends of the target you have for 4.0.

If the goal is to have a small 4.0 release soon to allow supporting Symfony 8, this should not be included IMO, keeping it for 5.0.

Any removal of feature from Behat would require being able to deprecate them first in the existing version to allow a smooth upgrade path. And behat currently does not have a system to report deprecations (be it from itself or from code being tested).

acoulton:
I'm in two minds on this.

I agree that we don't want to create unnecessary friction for upgrading.

Equally, I think we possibly could do this without being able to trigger a runtime deprecation for it. 

It will already have been automatically fixed for anyone who has migrated to PHP config automatically (the `--convert-config` command already converts shorthand extensions to their FQCN). And it will show up as soon as you try to run behat - it's not something that will lurk and cause an unexpected problem later...

So IMO if the release notes are clear and the exception message if an extension cannot be found is clear (we can quite easily check for a value that looks like a short name and tell the user how to fix it) then this is quite a minor change that users could handle fairly easily.

acoulton:
[...]

PR #1793, opened 2026-02-17
fix: remove short name extension loading from ExtensionManager

carlos-granados:
Remove the guessFullExtensionClassName() method that attempted to expand short extension names to full class names. This feature is no longer useful now that YAML config file support has been removed from 4.x.

The error message now suggests using the fully qualified class name if users are attempting to use a short name.

Fixes https://github.com/Behat/Behat/issues/1754


acoulton (approved):
Thanks @carlos-granados - since you were able to delete this entire behaviour without changing any tests I think that reinforces this is definitely worth doing!

Issue #1574, opened 2024-12-19
[RFC] Use deprecation contracts to notice developers about future deprecations


fmatsos:
As we will drop support for annotations to PHP attributes in the near future, I thought it might be useful to start displaying deprecation notices to prepare users for the migration.

For this I suggest using the  package `symfony/deprecation-contracts`, which provides a single global function `trigger_deprecation` that triggers silenced deprecation notices.

This could also be used for any other planned deprecations in the code and/or features.

stof:
The actual need is for Behat is to build a _reporter_ for deprecations.
Making it compatible with the deprecations triggered in `symfony/deprecation-contracts` is indeed a good idea, given that many projects are using this way of reporting deprecations (Symfony of course, but also Drupal for instance), either through `symfony/deprecation-contracts` or using `trigger_error(..., E_USER_DEPRECATED)` directly. `doctrine/deprecations` can also be configured to use the same reporting mechanism.

However, `symfony/deprecation-contracts` won't help us implement the reporter (if Behat itself does not need to _trigger_ deprecations, it won't even need to have `symfony/deprecation-contracts` as a dependency).

fmatsos:
Is there any specification about the reporter needs?
I'm not sure to understand what you mean exactly.

acoulton:
@fmatsos I believe that @stof means we need a new feature for Behat to collect details of deprecations during a run and report them separately from failures & exceptions. We'd probably want a way for end-users to configure whether deprecations should cause the suite to fail or not. I'd expect this to be similar to the way PHPUnit handles this.

We'd also need to work out the best way to make this BC with the existing defaults / options for customising the errorlevel in the behat process.

I was thinking of having a go at this when I get a chance, but if you would like to / have time to look at it that would be great and I'd be happy to talk more about how it could work :)

jdeniau:
@stof Do you want to report all deprecations or only behat's one ? 

FWIW, I use https://github.com/caciobanu/behat-deprecation-extension that report all deprecations after the run. 

It's a good idea, but it's quite annoying as there might be a lot of deprecations where the user can't do anything:

App → package A → package B

If package B trigger a deprecation, then iti's logged, but the app can't do anything about it, as it needs to be taken care of in package A.

carlos-granados:
Closed via https://github.com/Behat/Behat/pull/1794

PR #1794, opened 2026-02-18
feat: add deprecation collector to capture and print Behat deprecations

### LINKED ISSUE #1747: [3.x] Add a deprecation tracking / reporting mechanism
This needs to be able to track deprecations triggered:

* Within our own code (potentially while Behat is still bootstrapping e.g. when reading config files and initialising extensions)
* Within extension code
* Within end-user code (step definitions etc) - bearing in mind that users might have expressed a preference for handling deprecations by configuring Behat's error_reporting setting and registering their own error handler.

We need to provide sane backwards-compatible defaults, including whether the deprecation report is displayed by default (which would be safe with any of our formatters, but perhaps not with a third-party formatter that is outputting structured data to stdout).

Relates to #1574.

We need to add this to 3.x before the final 4.0.0 release, so that we can inform users if they are using deprecated features before they upgrade,

carlos-granados:
Adds a deprecation collection system that:
- Registers an error handler early in Application::doRun() to capture all deprecations from Behat code
- Uses a singleton pattern to ensure the same collector instance is used throughout the application lifecycle
- Prints collected deprecations at the end of the test run

This feature is off by default as I believe it is better not to have any extra output which might break existing pipelines. This can be turned on using a command line option or a configuration option

This deprecation collector does not collect deprecations thrown in the tested code. Behat already provides a mechanism to handle errors (including deprecations) in tested code. If this is not flexible enough, the users can use this extension: https://github.com/caciobanu/behat-deprecation-extension

Fixes https://github.com/Behat/Behat/issues/1747 and https://github.com/Behat/Behat/issues/1574

### DIFF
--- a/features/convert_config.feature
+++ b/features/convert_config.feature
@@ -376,6 +376,24 @@ Feature: Convert config
       """
     And the "unused_definitions.yaml" file should have been removed from the working directory
 
+  Scenario: Print deprecations
+    When I run behat with the following additional options:
+      | option   | value                   |
+      | --config | print_deprecations.yaml |
+    Then it should pass
+    And "print_deprecations.php" file should contain:
+      """
+      <?php
+
+      use Behat\Config\Config;
+      use Behat\Config\Profile;
+
+      return (new Config())
+          ->withProfile((new Profile('default'))
+              ->withPrintDeprecations());
+      """
+    And the "print_deprecations.yaml" file should have been removed from the working directory
+
   Scenario: Formatters
     When I run behat with the following additional options:
       | option   | value           |
@@ -522,6 +540,7 @@ Feature: Convert config
                   ->withFilter(new NameFilter('john'))
                   ->withFilter(new RoleFilter('admin')))
               ->withPrintUnusedDefinitions()
+              ->withPrintDeprecations()
               ->withPathOptions(
                   printAbsolutePaths: true,
                   editorUrl: 'phpstorm://open?file={relPath}&line={line}',
--- a/features/convert_config.feature
+++ b/features/convert_config.feature
@@ -387,10 +387,12 @@ Feature: Convert config
 
       use Behat\Config\Config;
       use Behat\Config\Profile;
+      use Behat\Config\TesterOptions;
 
       return (new Config())
           ->withProfile((new Profile('default'))
-              ->withPrintBehatDeprecations());
+              ->withTesterOptions((new TesterOptions())
+                  ->withPrintBehatDeprecations()));
       """
     And the "print_behat_deprecations.yaml" file should have been removed from the working directory
 
@@ -540,7 +542,6 @@ Feature: Convert config
                   ->withFilter(new NameFilter('john'))
                   ->withFilter(new RoleFilter('admin')))
               ->withPrintUnusedDefinitions()
-              ->withPrintBehatDeprecations()
               ->withPathOptions(
                   printAbsolutePaths: true,
                   editorUrl: 'phpstorm://open?file={relPath}&line={line}',
@@ -550,7 +551,8 @@ Feature: Convert config
                   ]
               )
               ->withTesterOptions((new TesterOptions())
-                  ->withStrictResultInterpretation())
+                  ->withStrictResultInterpretation()
+                  ->withPrintBehatDeprecations())
               ->withExtension(new Extension('custom_extension.php'))
               ->withSuite((new Suite('my_suite'))
                   ->addContext(
--- a/features/convert_config.feature
+++ b/features/convert_config.feature
@@ -376,26 +376,6 @@ Feature: Convert config
       """
     And the "unused_definitions.yaml" file should have been removed from the working directory
 
-  Scenario: Print deprecations
-    When I run behat with the following additional options:
-      | option   | value                   |
-      | --config | print_behat_deprecations.yaml |
-    Then it should pass
-    And "print_behat_deprecations.php" file should contain:
-      """
-      <?php
-
-      use Behat\Config\Config;
-      use Behat\Config\Profile;
-      use Behat\Config\TesterOptions;
-
-      return (new Config())
-          ->withProfile((new Profile('default'))
-              ->withTesterOptions((new TesterOptions())
-                  ->withPrintBehatDeprecations()));
-      """
-    And the "print_behat_deprecations.yaml" file should have been removed from the working directory
-
   Scenario: Formatters
     When I run behat with the following additional options:
       | option   | value           |
@@ -491,12 +471,20 @@ Feature: Convert config
           ->withProfile((new Profile('not-strict'))
               ->withTesterOptions((new TesterOptions())
                   ->withStrictResultInterpretation(false)))
+          ->withProfile((new Profile('print-deprecations'))
+              ->withTesterOptions((new TesterOptions())
+                  ->withPrintBehatDeprecations()))
+          ->withProfile((new Profile('fail-on-deprecations'))
+              ->withTesterOptions((new TesterOptions())
+                  ->withFailOnBehatDeprecations()))
           ->withProfile((new Profile('complete'))
               ->withTesterOptions((new TesterOptions())
                   ->withStrictResultInterpretation()
                   ->withStopOnFailure(false)
                   ->withSkipAllTests()
-                  ->withErrorReporting(E_ALL & ~(E_WARNING | E_NOTICE | E_DEPRECATED))));
+                  ->withErrorReporting(E_ALL & ~(E_WARNING | E_NOTICE | E_DEPRECATED))
+                  ->withPrintBehatDeprecations()
+                  ->withFailOnBehatDeprecations()));
       """
     And the "tester_options.yaml" file should have been removed from the working directory
 
@@ -552,7 +540,8 @@ Feature: Convert config
               )
               ->withTesterOptions((new TesterOptions())
                   ->withStrictResultInterpretation()
-                  ->withPrintBehatDeprecations())
+                  ->withPrintBehatDeprecations()
+                  ->withFailOnBehatDeprecations())
               ->withExtension(new Extension('custom_extension.php'))
               ->withSuite((new Suite('my_suite'))
                   ->addContext(
--- a/features/convert_config.feature
+++ b/features/convert_config.feature
@@ -379,9 +379,9 @@ Feature: Convert config
   Scenario: Print deprecations
     When I run behat with the following additional options:
       | option   | value                   |
-      | --config | print_deprecations.yaml |
+      | --config | print_behat_deprecations.yaml |
     Then it should pass
-    And "print_deprecations.php" file should contain:
+    And "print_behat_deprecations.php" file should contain:
       """
       <?php
 
@@ -390,9 +390,9 @@ Feature: Convert config
 
       return (new Config())
           ->withProfile((new Profile('default'))
-              ->withPrintDeprecations());
+              ->withPrintBehatDeprecations());
       """
-    And the "print_deprecations.yaml" file should have been removed from the working directory
+    And the "print_behat_deprecations.yaml" file should have been removed from the working directory
 
   Scenario: Formatters
     When I run behat with the following additional options:
@@ -540,7 +540,7 @@ Feature: Convert config
                   ->withFilter(new NameFilter('john'))
                   ->withFilter(new RoleFilter('admin')))
               ->withPrintUnusedDefinitions()
-              ->withPrintDeprecations()
+              ->withPrintBehatDeprecations()
               ->withPathOptions(
                   printAbsolutePaths: true,
                   editorUrl: 'phpstorm://open?file={relPath}&line={line}',
--- a/features/print_deprecations.feature
+++ b/features/print_deprecations.feature
@@ -1,0 +1,57 @@
+Feature: Print deprecations
+  In order to identify deprecated code usage in Behat
+  As a developer
+  I need to be able to see deprecation warnings triggered by Behat code
+
+  Background:
+    Given I initialise the working directory from the "Deprecations" fixtures folder
+    And I provide the following options for all behat invocations:
+      | option      | value    |
+      | --no-colors |          |
+      | --format    | progress |
+
+  Scenario: Do not print deprecations by default
+    When I run "behat"
+    Then it should pass with:
+      """
+      ..
+
+      1 scenario (1 passed)
+      2 steps (2 passed)
+      """
+
+  Scenario: Print deprecations using config option
+    When I run behat with the following additional options:
+      | option    | value              |
+      | --profile | print_deprecations |
+    Then it should pass with:
+      """
+      2 deprecations triggered (2 unique):
+
+        ⚠ This extension triggers a deprecation during initialization
+        ⚠ This extension triggers a deprecation during load
+      """
+
+  Scenario: Print deprecations using CLI option
+    When I run behat with the following additional options:
+      | option               | value |
+      | --print-deprecations |       |
+    Then it should pass with:
+      """
+      2 deprecations triggered (2 unique):
+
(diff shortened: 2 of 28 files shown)


src/Behat/Testwork/Cli/Application.php
```diff
@@ -125,12 +129,12 @@ public function doRun(InputInterface $input, OutputInterface $output): int
...
         }
 
-        $this->addCommand($this->createCommand($input, $output));
+        $this->doAddCommand($this->createCommand($input, $output));
 
         return parent::doRun($input, $output);
     }
 
-    public function addCommand(callable|SymfonyCommand $command): ?SymfonyCommand
+    public function doAddCommand(callable|SymfonyCommand $command): ?SymfonyCommand
```
carlos-granados: We had tried to fix a deprecation that was thrown in Symfony console 7.4 where the `add()` function was deprecated and a replacement `addCommand()` was added. However, the fix, though it satisfied PHPStan, did not actually work because the `addCommand()` function in the Symfony code called the `init()` function that checks if the declaring class of the `add()` and the `addCommand()` function are the same and, if they are not, calls the `add()` function. Since we had defined this `addCommand()` function here, this check failed and we ended up calling the `add()` function and the deprecation was thrown. This was only noticed when I added this deprecation collector which printed it. The only way I have found to fix this is to change the name of this function. This is in theory a BC break as this is a public function, but I doubt that anyone is calling a function that has been introduced such a short time ago. If someone has a better idea, please let me know.

Hope all of this makes sense :-)

stof: Do we need this `doAddCommand` to be public ?

carlos-granados: @stof ah, of course, no, it does not need to be public. Updated

src/Behat/Testwork/Cli/Application.php
```diff
@@ -125,12 +129,23 @@ public function doRun(InputInterface $input, OutputInterface $output): int
...
+        $commands[] = new DumpReferenceCommand($this->extensionManager);
+        $commands[] = new DebugCommand($this, $this->configurationLoader, $this->extensionManager);
+        $commands[] = new ConvertConfigCommand($this->configurationLoader);
+
+        return $commands;
+    }
+
+    private function doAddCommand(callable|SymfonyCommand $command): ?SymfonyCommand
     {
         // Provide compatibility with all supported symfony/console versions
```
stof: there is even a simpler way to write code support all versions of Symfony: using `$this->addCommands([$command])` instead (and letting the `addCommands` method of `symfony/console` use non-deprecated method).

carlos-granados: @stof yeah, good shout, updated!

features/convert_config.feature
```diff
@@ -374,6 +374,24 @@ Feature: Convert config
...
+    And "print_deprecations.php" file should contain:
+      """
+      <?php
+
+      use Behat\Config\Config;
+      use Behat\Config\Profile;
+
+      return (new Config())
+          ->withProfile((new Profile('default'))
+              ->withPrintDeprecations());
```
acoulton: I think if this is only concerned with deprecations in Behat itself (and extensions) then we need to be more explicit with the naming so that it's clear this is separate to anything in steps and hooks (like e.g. the difference between `deprecations` and `phpunitDeprecations` in phpunit config)?

I also wonder if this should logically sit on TesterOptions alongside the errorlevel configuration. Partly because I think I would also want an option to fail on these deprecations in order to make this more useful in CI. For example I suspect we (& extension authors) would definitely want a run in our own CI that fails on deprecations to allow us to spot & fix any that are introduced by php versions/dependencies.

That doesn't have to be part of this PR but worth thinking about where to put the config now?

features/print_deprecations.feature
```diff
@@ -0,0 +1,57 @@
...
+      | --print-deprecations |       |
+    Then it should pass with:
+      """
+      2 deprecations triggered (2 unique):
+
+        ⚠ This extension triggers a deprecation during initialization
+        ⚠ This extension triggers a deprecation during load
+      """
+
+  Scenario: Deprecations in step definitions and hooks are not printed
```
acoulton, line 66: We should maybe have a feature proving that this config does not affect the step failing if you've set the errorlevel to include deprecations, just for clarity?

features/print_deprecations.feature
```diff
@@ -0,0 +1,57 @@
...
+  Scenario: Print deprecations using CLI option
+    When I run behat with the following additional options:
+      | option               | value |
+      | --print-deprecations |       |
+    Then it should pass with:
+      """
+      2 deprecations triggered (2 unique):
+
+        ⚠ This extension triggers a deprecation during initialization
+        ⚠ This extension triggers a deprecation during load
```
acoulton, line 47: This output is ok if the deprecation is a nicely worded custom message. However if it's something more like `trim(): passing null to parameter of type string is deprecated ` (or similar more generic deprecations) that could be a nightmare to find & fix.

I think we need at least the first line of the trace as standard, but ideally the option to show the full trace if required?

tests/Fixtures/Deprecations/deprecation_extension.php
```diff
@@ -0,0 +1,36 @@
...
+        return 'deprecation_extension';
+    }
+
+    public function configure(ArrayNodeDefinition $builder): void
+    {
+    }
+
+    public function initialize(ExtensionManager $extensionManager): void
+    {
+        @trigger_error('This extension triggers a deprecation during initialization', E_USER_DEPRECATED);
```
acoulton: We should definitely catch & handle anything that comes to us through `trigger_error` e.g. from PHP or dependencies.

However, we can't guarantee those will get to our error handler. If the user has e.g. bootstrapped a framework then there's a strong chance that's registered an error handler of its own which will now get these calls. It's impossible to predict whether that will display output / ignore them / fail on them but either way it'll be confusing & possibly breaking if Behat's behaviour doesn't follow our own config flags.

Therefore I think instead of calling `trigger_error` in our own code, we should have a standalone way to report a deprecation directly to our collector singleton. This will ensure it's definitely handled & reported by us regardless of the user's runtime environment. This should also be part of our public API for extension authors to use.

acoulton (commented):
Thanks very much for this @carlos-granados I'm travelling so haven't had a chance to look in detail yet but I have a few high-level thoughts/suggestions for discussion.

carlos-granados:
> Thanks very much for this @carlos-granados I'm travelling so haven't had a chance to look in detail yet but I have a few high-level thoughts/suggestions for discussion.

I'm travelling myself, coming back on Monday, so won't be able to look into this until Tuesday, but I'll reply to some of your comments

features/convert_config.feature
carlos-granados: Yeah, I think you are probably right and this better belongs with the tester options. And adding an option to fail if there are any deprecations should be pretty simple, so I'll look into that as well. And your suggestion about naming makes sense as well

features/print_deprecations.feature
carlos-granados: Yep, I'll add it

features/print_deprecations.feature
carlos-granados: I'll print the line and file where the deprecation was generated. I don't want to print a full trace because that would break the capability to merge all instances of the same deprecation and we really don't want to be listing each deprecation individually

tests/Fixtures/Deprecations/deprecation_extension.php
carlos-granados: I have to disagree here. If some framework sets up some kind of error handler, that would be within the code that is executed when running the tests and we are handling this separately from these deprecations. Behat is not a library which is called by an outside project, so the top level error_handler should always be ours. And using trigger_error is much easier to implement than having to call directly the singleton

features/print_deprecations.feature
acoulton: Seems a fair compromise for now, I agree grouping is usually what's needed.

tests/Fixtures/Deprecations/deprecation_extension.php
acoulton: I'm not sure we can assume this.

I know the two legacy frameworks we bootstrap in some of our projects (from an extension) register their error handlers at the point of bootstrapping and it's hard to unregister that.

I don't think we're the only people where that happens - ISTR the symfony2extension also bootstraps the outer layer of the symfony framework during extension load, at least for some symfony versions, and that includes an error handler (couldn't get enough WiFi on the train to pull down a symfony project to test this).

In that case, the framework's error will be registered after ours but outside the user code running in scenarios, meaning it will get anything from extensions that load later or anything of ours that runs after loading extensions.

I would generally avoid global singletons/state but ultimately using `trigger_error` with an assumption that will arrive at a collector we've registered is global state, just hidden. 

Therefore I think in this case it would be justifiable to do something like `BehatDeprecationCollector::report('some message')`direct  and avoid the indirection of hoping `trigger_error` will do the same thing.

carlos-granados: @acoulton I have tested this in a project which uses Symfony and the Behat Symfony extension and in this case our deprecations handler works as expected and registers and reports all needed deprecations. The Symfony error handler is not registered until some actual request is made, at which point we do not need to handle any raised deprecations. Not sure if you could test the behaviour with the legacy frameworks that you say you are loading through an extension.
If possible, I would like to keep the mechanism of just calling `trigger_error()` to raise a deprecation. It is simpler, understood by all developers and does not need to import any class to use it

carlos-granados:
@acoulton I have updated this PR with your suggestions:
- The management of the option has been moved to the TesterOptions config object
- I added a new configuration option that can be set so that the Behat command will fail if there are any Behat deprecations
- Changed the name of the options so that it is clearer that this is used to print deprecations related to Behat itself
- Print the file and line where the deprecation was raised
- Add a new scenario that verifies that when error reporting is set to include deprecations, deprecations in the tested code will still cause test failures, independent of the Behat deprecation printing feature.

features/print_deprecations.feature
```diff
@@ -0,0 +1,113 @@
...
+      """
+      ..
+
+      1 scenario (1 passed)
+      2 steps (2 passed)
+      """
+
+  Scenario: Print deprecations using config option
+    When I run behat with the following additional options:
+      | option    | value              |
```
acoulton, line 25: very minor but it'd be good to fix the table alignment in this feature file?

features/print_deprecations.feature
```diff
@@ -0,0 +1,113 @@
...
+    When I run behat with the following additional options:
+      | option   | value                            |
+      | --config | behat-fail-with-deprecations.php |
+    Then it should fail with:
+      """
+      ..
+
+      1 scenario (1 passed)
+      2 steps (2 passed)
+      """
```
acoulton, line 107: Should `fail on deprecations` imply / enforce `print deprecations`? Or at least print some sort of final output to explain why it failed, even if it doesn't print the deprecations themsevles?

It could be confusing to get output saying everything passed, but the command exited nonzero (especially in CI)?

acoulton (commented):
@carlos-granados my apologies for the slow review - I missed this somehow. Overall this is looking great now. A couple of tiny things I noticed, and I'll reply separately to your comment about `trigger_error` vs a custom method.

features/print_deprecations.feature
stof: yes it should imply it. That was actually [my first feedback to PHPUnit](https://github.com/sebastianbergmann/phpunit/issues/5856) once trying their new reporting system. So let's learn from it and provide the good experience from the start.

tests/Fixtures/Deprecations/deprecation_extension.php
acoulton: @carlos-granados thanks for testing with symfony, I looked at older supported symfony versions too and I see that although the extension bootstraps symfony, that doesn't itself register the error handler.

I do know for certain that two of the legacy frameworks we use do register an error handler as part of bootstrapping, and there's no easy way to avoid that. So for us, any `trigger_error` called within Behat or extensions (after our extension that loads the framework) will definitely come to our handler and not this one.

It won't break our build (generally, we toggle fail-on-deprecation on for specific builds as we work towards a major PHP or dependency update, and then turn it back off once green). But it will mean we don't see the deprecation output regardless of our behat config. And our build will not fail if we set the `failOnBehatDeprecation` in config.

I'm very happy to accept that we're outliers, but I don't think we're the only people using Behat with a legacy framework, and it wouldn't surprise me to find at least a few users whose builds *do* fail on deprecation. I'm also not certain how e.g. Drupal's bootstrapping works, and when they register an error handler.

I take the point on `trigger_error` being well understood.

I also think it's fairly common for projects to prefer a different mechanism for triggering deprecations in their ecosystem - e.g. [symfony/deprecation-contracts](https://github.com/symfony/deprecation-contracts) and [doctrine/deprecations](https://github.com/doctrine/deprecations). 

Doctrine for example have specifically said they prefer their library because it `no side-effects by default, making it a perfect fit for libraries that don't know how the error handler works they operate under`.

So my opinion is:

* Using `trigger_error` will probably work as we hope for most users in most circumstances - but is not guaranteed to work for everyone. There is a small-but-nonzero risk that it breaks builds for existing users.
* We don't actually **want** end-users to be able to replace our deprecation handling by calling `set_error_handler` as that will cause confusion.
* The easiest/only way to guarantee that a deprecation specific to Behat (or an extension) makes it to Behat is to call a dedicated method for that.

Just to be clear - we should still register an error handler in the hope it catches anything from `trigger_error` and any native deprecations.

But in our own code, and our recommendations for extension developers, we should prefer calling a method that is guaranteed to report to our logger, in every case and on every environment. A method on the existing singleton achieves that in a way that `trigger_error` (by design) cannot.

All that said, having something that works for the majority is much better than not having anything at all. So if you feel strongly about it then I am happy to back down :laughing:

carlos-granados: @acoulton So, how would you actually implement this? Should we implement a static function in the DeprecationCollector class, something like DeprecationCollector::trigger('deprecated bla bla')?

acoulton: Yes, I think so - `DeprecationCollector::trigger(string $msg)` would feel familiar to people used to `trigger_error()`.

Or we could just go `DeprecationCollector::add()` for simplicity.

I think the advantage of `trigger` or `add` over e.g. `log` or `report` is it might do more than just log e.g. it might cause the build to fail, depending on the user's configuration.

carlos-granados:
@acoulton @stof I've updated the branch so that:
- Failing on deprecations implies printing them
- We add a `trigger` static function to the DeprecationsCollector which can be used to record a deprecation and use this in our codebase

tests/Fixtures/Deprecations/deprecation_extension.php
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    public function initialize(ExtensionManager $extensionManager): void
+    {
+        DeprecationCollector::trigger('This extension triggers a deprecation during initialization');
+    }
+
+    public function load(ContainerBuilder $container, array $config): void
+    {
+        DeprecationCollector::trigger('This extension triggers a deprecation during load');
```
acoulton: I guess the only thing now is our feature coverage doesn't (I think?) prove we're also catching native deprecations e.g. from native php or vendor code that calls `trigger_error`. 

Maybe one of these calls should use `trigger_error` to cover that?

acoulton (commented):
This is brilliant @carlos-granados, one small query/comment on test coverage but otherwise it looks perfect to me now, thanks.

tests/Fixtures/Deprecations/deprecation_extension.php
carlos-granados: @acoulton you are right, updated

acoulton (approved):
Superb, thank you! 🎉

PR #1796, opened 2026-02-18
[4.0] refactor: Extension interface should not extend third-party interface

acoulton:
For 4.0, the Extension interface no longer extends the symfony CompilerPassInterface. This reduces the risk of having to make breaking changes and/or force extension authors to update their extensions for future symfony majors.

Instead, we satisfy the CompilerPassInterface by wrapping the Extension internally as an implementation detail.

Follows #1795 and refs #1316 - I will rebase this once #1795 is merged to 3.x and 3.x has been merged up.

### DIFF
--- a/.roave-backward-compatibility-check.xml
+++ b/.roave-backward-compatibility-check.xml
@@ -65,5 +65,12 @@
         <ignored-regex>#REMOVED: These ancestors of Behat\\Config\\.+? have been removed: \["Behat\\\\Config\\\\ConfigConverterInterface"\]#</ignored-regex>
         <ignored-regex>#REMOVED: Method Behat\\Config\\.+?\#toPhpExpr\(\) was removed#</ignored-regex>
         <ignored-regex>#REMOVED: Constant Behat\\Config\\Formatter\\PrettyFormatter::PRINT_SKIPPED_STEPS_PARAMETER_NAME was removed#</ignored-regex>
+
+
+        <!--
+        `Behat\Testwork\ServiceContainer\Extension` is now a standalone interface and no longer extends the third-party
+        `CompilerPassInterface`
+        -->
+        <ignored-regex>#REMOVED: These ancestors of .+?Extension have been removed: \["Symfony\\\\Component\\\\DependencyInjection\\\\Compiler\\\\CompilerPassInterface"\]#</ignored-regex>
     </baseline>
 </roave-bc-check>
--- a/src/Behat/Testwork/ServiceContainer/Extension.php
+++ b/src/Behat/Testwork/ServiceContainer/Extension.php
@@ -43,13 +43,29 @@ interface Extension extends CompilerPassInterface
 
     /**
      * Setups configuration for the extension.
+     *
+     * NOTE: If your extension implements this method, your composer.json should declare
+     * a direct dependency on the version(s) of symfony/config that you support.
      */
     public function configure(ArrayNodeDefinition $builder);
 
     /**
      * Loads extension services into temporary container.
      *
+     * NOTE: If your extension implements this method, your composer.json should declare
+     * a direct dependency on the version(s) of symfony/dependency-injection that you support.
+     *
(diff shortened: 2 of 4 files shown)


src/Behat/Testwork/ServiceContainer/Extension.php
```diff
@@ -43,13 +42,29 @@ public function initialize(ExtensionManager $extensionManager);
...
+
+    /**
+     * You can modify the container here before it is dumped to PHP code.
+     *
+     *  NOTE: If your extension implements this method, your composer.json should declare
+     *  a direct dependency on the version(s) of symfony/dependency-injection that you support.
+     *
+     * @return void
+     */
+    public function process(ContainerBuilder $container);
```
acoulton, line 69: Note: adding strict return types to the methods in this interface is intentionally not part of this PR - that will be covered by a separate PR to add strict return types throughout for 4.x as part of #1744

.roave-backward-compatibility-check.xml
```diff
@@ -65,5 +65,12 @@
         <ignored-regex>#REMOVED: These ancestors of Behat\\Config\\.+? have been removed: \["Behat\\\\Config\\\\ConfigConverterInterface"\]#</ignored-regex>
         <ignored-regex>#REMOVED: Method Behat\\Config\\.+?\#toPhpExpr\(\) was removed#</ignored-regex>
         <ignored-regex>#REMOVED: Constant Behat\\Config\\Formatter\\PrettyFormatter::PRINT_SKIPPED_STEPS_PARAMETER_NAME was removed#</ignored-regex>
+
+
+        <!--
+        `Behat\Testwork\ServiceContainer\Extension` is now a standalone interface and no longer extends the third-party
+        `CompilerPassInterface`
+        -->
+        <ignored-regex>#Error: These ancestors of .+?Extension have been removed: \["Symfony\\\\Component\\\\DependencyInjection\\\\Compiler\\\\CompilerPassInterface"\]#</ignored-regex>
```
acoulton: I'll look into why this isn't matching the output...

stof: I think the `Error: ` prefix is not part of the message. It is part of the github formatter (so that they are reported as annotations)

acoulton: Thanks, yes - I was doing this on a train last week on a laptop that couldn't run the tool, so was  looking at the github actions output.

You're right the `Error:` prefix is part of the gha format instead of the initial `REMOVED:` prefix that is used for ignoring. I did think it was odd that the line didn't match the format of the other ignore patterns but it makes sense now. I'll know for next time!

Fixed now.

src/Behat/Testwork/ServiceContainer/Extension.php
```diff
@@ -43,13 +42,29 @@ public function initialize(ExtensionManager $extensionManager);
...
+     *
+     * NOTE: If your extension implements this method, your composer.json should declare
+     * a direct dependency on the version(s) of symfony/config that you support.
      */
     public function configure(ArrayNodeDefinition $builder);
 
     /**
      * Loads extension services into temporary container.
      *
+     * NOTE: If your extension implements this method, your composer.json should declare
```
stof, line 54: maybe this should say `if your extension implements logic in this method`. All extensions are required to implement the method when implementing the interface, even if it is empty.

acoulton: That's a valid point. This comment is in 3.x already, so I would probably change it there after this is merged, rather than redoing all the merges now. It can then get updated in 4.x on the next normal merge-up.

carlos-granados (approved):
Looking good to me now

src/Behat/Testwork/ServiceContainer/Extension.php
acoulton: Fixed by https://github.com/Behat/Behat/pull/1804

PR #1798, opened 2026-02-21
feat: Move `gherkin` configuration to dedicated object

acoulton:
Currently, the only way to set Gherkin configuration through the new config objects is via `Profile::withFilter()`, which internally adds the filter to the `$settings['gherkin']['filters']` array. This is the only config setter we expose for any settings in the `gherkin` group.

However, the `GherkinExtension` already defines a `cache` option (which we do not have a setter for) and we need to add an option in #1799 to configure the gherkin parser mode. It's possible that we will need other gherkin settings in future (and that users will want to share these settings between profiles).

Therefore, I have extracted the Gherkin `withFilter()` setting to a standalone DTO. The existing `Profile::withFilter()` method is retained but deprecated and will be removed in 4.0.

### DIFF
--- a/features/profile_filters.feature
+++ b/features/profile_filters.feature
@@ -9,8 +9,8 @@ Feature: Filters
       | option      | value |
       | --no-colors |       |
 
-  Scenario: Tag filters
-    When I run "behat --config=behat-tag-filter.php"
+  Scenario Outline: Tag filters
+    When I run "behat --config=<config_file>"
     Then it should pass with:
       """
       @tag2
@@ -48,6 +48,11 @@ Feature: Filters
       4 scenarios (4 passed)
       10 steps (10 passed)
       """
+
+    Examples:
+      | config_file                     | comment                                   |
+      | behat-tag-filter.php            | Using new method on GherkinOptions object |
+      | behat-tag-filter-on-profile.php | Using deprecated method on Profile object |
 
   Scenario: Role filters
     When I run "behat --config=behat-role-filter.php"
--- a/src/Behat/Config/GherkinOptions.php
+++ b/src/Behat/Config/GherkinOptions.php
@@ -1,0 +1,85 @@
+<?php
+
+namespace Behat\Config;
+
+use Behat\Config\Converter\ConfigConverterTools;
+use Behat\Config\Filter\FilterInterface;
+use Behat\Config\Filter\NameFilter;
+use Behat\Config\Filter\NarrativeFilter;
+use Behat\Config\Filter\RoleFilter;
+use Behat\Config\Filter\TagFilter;
+use Behat\Testwork\ServiceContainer\Exception\ConfigurationLoadingException;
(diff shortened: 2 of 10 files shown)


features/profile_filters.feature
```diff
@@ -49,6 +49,11 @@ Feature: Filters
       10 steps (10 passed)
       """
 
+    Examples:
+      | config_file                     | comment                                   |
```
acoulton, line 53: The `comment` column is not required in this table, but unfortunately we do not support syntax like:

```
| behat-tag-filter.php | # Using new method on GherkinOptions object
```

until later versions of Behat/Gherkin.

src/Behat/Config/GherkinOptions.php
```diff
@@ -0,0 +1,111 @@
...
+
+    public function toArray(): array
+    {
+        return $this->settings;
+    }
+
+    /**
+     * Sets the parser cache directory (defaults to the system tmp dir, if writable).
+     */
+    public function withCacheDir(string $dir): self
```
acoulton, line 35: Note that we do not actually have any feature coverage that proves this option is properly passed to/used by the Gherkin parser.

I thought of writing a feature to prove either that at least some files are written to the custom dir, or that Behat fails if the cache directory is not writable / cannot be created, but both of these felt a little contrived / coupled. I've left as-is for now.

`'cache'` is defined as an explicit option by the config builder in `GherkinExtension` and therefore the argument name is validated at runtime - we just don't have confirmation that the `GherkinExtension` then passes that in correctly to the Gherkin layer.

features/profile_filters.feature
```diff
@@ -9,8 +9,8 @@ Feature: Filters
       | option      | value |
       | --no-colors |       |
 
-  Scenario: Tag filters
-    When I run "behat --config=behat-tag-filter.php"
+  Scenario Outline: Tag filters
```
acoulton, line 12: I've switched this to a scenario outline to prove that the result is identical whether calling `->withFilters` on the `Profile` or the `GherkinOptions` object.

I think it's enough to prove that with one type of filter, we don't need to duplicate all the scenarios here.

src/Behat/Config/Profile.php
```diff
@@ -80,15 +75,15 @@ public function withExtension(ExtensionConfigInterface $extension): self
         return $this;
     }
 
+    /**
+     * @deprecated see withGherkinOptions()->withFilter(). This method will be removed in 4.0
```
carlos-granados: I would probably say

@deprecated use withGherkinOptions()->withFilter() instead

carlos-granados (approved):
@acoulton looking good, just a tiny comment

src/Behat/Config/Profile.php
acoulton: Thanks @carlos-granados, fixed

carlos-granados (approved):

Issue #1645, opened 2025-05-26
[3.x] Support configuring the gherkin parser compatibility mode


acoulton:
The next gherkin release will add an optional `GherkinCompatibilityMode` feature flag to the Parser. See https://github.com/Behat/Gherkin/pull/349

This is because the legacy / default behaviour parses feature files in a way that has some edge cases / small details that differ from the official cucumber/gherkin parsers. The new compatibility mode option allows consumers to opt-in to parser behaviour that matches the official parsers.

Once that is released, we will want to expose it as a new configuration option in Behat.

We will also need to update Behat functionality (particularly the Pretty formatter) that depended on the legacy parsing.

acoulton:
I'd ideally like to get this into 3.x before we release 4.0.0, so that 4.0.0 can switch the default to the gherkin-compliant mode (while still allowing users to configure it back).

acoulton:
This was implemented in #1799 and is now documented at https://docs.behat.org/en/latest/user_guide/gherkin/parser_mode.html

We still need to update the Pretty formatter to cope with it (which already has an issue #1646), and check for any other potential issues in GHERKIN_32 mode. 

However once we merge #1812 to change the default in 4.0 we have done enough for `4.0.0-alpha1`.

PR #1799, opened 2026-02-21
feat: Support configuring `GherkinCompatibilityMode`

acoulton:
This PR is based on #1798 (the first two commits can be reviewed in that PR & I will rebase after it is merged) and is the first part of the work required for #1645.

I have limited the feature coverage here to the simplest case that proves which parser mode is active (pass or fail on presence of an invalid language tag, using the pretty formatter).

There is more work to do to make the output of the pretty formatter the same regardless of the parsing mode (padding is handled slightly differently, and there are more significant changes to how escapes and newlines are processed within tables). I will tackle these in separate PRs once we have the config option.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         "php": ">=8.2 <8.6",
         "ext-mbstring": "*",
         "composer-runtime-api": "^2.2",
-        "behat/gherkin": "^4.12.0",
+        "behat/gherkin": "^4.15.0",
         "composer/xdebug-handler": "^1.4 || ^2.0 || ^3.0",
         "nikic/php-parser": "^4.19.2 || ^5.2",
         "psr/container": "^1.0 || ^2.0",
--- a/features/gherkin_compatibility_mode.feature
+++ b/features/gherkin_compatibility_mode.feature
@@ -1,0 +1,34 @@
+Feature: Gherkin compatibility mode
+  In order to have my feature files parsed as they would be by other cucumber-based tools
+  As a tester
+  I need to be able to configure the Gherkin parser compatibility mode
+
+  Background:
+    Given I initialise the working directory from the GherkinCompatibility fixtures folder
+    And   I provide the following options for all behat invocations:
+      | option      | value    |
+      | --no-colors |          |
+      | --format    | progress |
+
+  Scenario: Legacy parsing by default (ignores invalid language)
+    When I run behat with the following additional options:
+      | option    | value   |
+      | --profile | default |
+    Then it should pass with:
+      """
+      UUU
+
+      1 scenario (1 undefined)
+      3 steps (3 undefined)
+      """
+
+  Scenario: Opt in to cucumber-compatible parsing
+    When I run behat with the following additional options:
+      | option    | value      |
+      | --profile | gherkin-32 |
+    Then it should fail with:
+      """
+      In KeywordsDialectProvider.php line XX:
+
+        Language not supported: martian
+      """
--- a/src/Behat/Behat/Gherkin/ServiceContainer/GherkinExtension.php
+++ b/src/Behat/Behat/Gherkin/ServiceContainer/GherkinExtension.php
@@ -100,13 +100,10 @@ final class GherkinExtension implements Extension
             ->enumNode('compatibility')
             ->info('Controls the extent to which gherkin is parsed equivalent to other cucumber tools')
                 // enumFqcn is not available until symfony 7.3
-                ->values([
-                    ...GherkinCompatibilityMode::cases(),
-                    ...array_map(
-                        fn (GherkinCompatibilityMode $m) => $m->value,
-                        GherkinCompatibilityMode::cases(),
-                    ),
-                ])
+                ->values(array_map(
+                    fn (GherkinCompatibilityMode $m) => $m->value,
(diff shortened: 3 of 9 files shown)


features/gherkin_compatibility_mode.feature
```diff
@@ -0,0 +1,34 @@
...
+
+  Scenario: Opt in to cucumber-compatible parsing
+    When I run behat with the following additional options:
+      | option    | value      |
+      | --profile | gherkin-32 |
+    Then it should fail with:
+      """
+      In KeywordsDialectProvider.php line XX:
+
+        Language not supported: martian
```
acoulton, line 33: I've created #1800 as a separate issue to improve the handling / output of the error in this case (e.g. to show the name of the feature file that was being parsed at the time).

composer.json
```diff
@@ -17,7 +17,7 @@
         "php": ">=8.2 <8.6",
         "ext-mbstring": "*",
         "composer-runtime-api": "^2.2",
-        "behat/gherkin": "^4.12.0",
+        "behat/gherkin": "^4.15.0",
```
acoulton, line 20: It would be possible to build this feature to support older behat/gherkin versions (e.g. by checking at runtime for the existence of the `GherkinCompatibilityMode` class). 

However, I don't think there are any dependency constraints that should prevent people bumping to the newer gherkin version and I think it's worth pushing users in that direction as we work towards 4.0, to simplify the number of cases we have to handle.

acoulton:
@carlos-granados I've now rebased this on to the updated 3.x with #1798 merged

src/Behat/Behat/Gherkin/ServiceContainer/GherkinExtension.php
```diff
@@ -95,6 +96,16 @@ public function configure(ArrayNodeDefinition $builder): void
                         : null
                 )
         ;
+        $childrenBuilder
+            ->enumNode('compatibility')
+            ->info('Controls the extent to which gherkin is parsed equivalent to other cucumber tools')
+                // enumFqcn is not available until symfony 7.3
+                ->values(array_map(
+                    fn (GherkinCompatibilityMode $m) => $m->value,
+                    GherkinCompatibilityMode::cases(),
```
stof: I think we should allow *both* the backing values and the enum cases themselves here (as done when using `enumFqcn()`, making it easier to switch to it in a future version)

stof (commented):
Is the `Rule` keyword the only remaining difference between `GherkinCompatibilityMode::GHERKIN_32` and cucumber?

carlos-granados (approved):
All looking OK but I second the comment made by @stof about using the backing values and the enum cases

src/Behat/Behat/Gherkin/ServiceContainer/GherkinExtension.php
acoulton: @stof @carlos-granados good idea - I've done that now.

src/Behat/Behat/Gherkin/ServiceContainer/GherkinExtension.php
```diff
@@ -100,10 +100,13 @@ public function configure(ArrayNodeDefinition $builder): void
...
-                    GherkinCompatibilityMode::cases(),
-                ))
+                ->values([
+                    ...GherkinCompatibilityMode::cases(),
+                    ...array_map(
+                        fn (GherkinCompatibilityMode $m) => $m->value,
+                        GherkinCompatibilityMode::cases(),
+                    ),
+                ])
                 ->defaultValue(GherkinCompatibilityMode::LEGACY->value)
```
stof, line 107: I suggest using `GherkinCompatibilityMode::LEGACY` here

acoulton: @stof - yeah, I missed that, thanks.

Unfortunately it looks like we can't support the enum values for as long as we support symfony 5.4.

symfony/config [before 6.3 calls `array_unique` on the values when building an EnumNodeDefinition](https://github.com/symfony/config/blob/5.4/Definition/Builder/EnumNodeDefinition.php#L30) and this then fails because array_unique compares values as strings and it's not possible to cast an enum to string.

I suppose we could try to detect symfony 5.4 and only add the case values if we're on a newer version, but then we would still have to set our actual config and default values as `$mode->value` so it would be fairly limited use.

From an end-user perspective, they will always call `->withGherkinOptions()->withCompatibilityMode(GherkinCompatibilityMode::WHATEVER)` so IMO it doesn't make much difference whether we are internally turning that into `->value` and back again, or passing it as the raw enum.

I've therefore reverted the change I made to go back to just using the strings.

stof:
It looks like `symfony/config` 5.4 does not handle enums properly in EnumNode (see the failing job using Symfony 5.4).

acoulton:
> Is the `Rule` keyword the only remaining difference between `GherkinCompatibilityMode::GHERKIN_32` and cucumber?

Yes, as far as I know. I plan to look at `Rule` support after we release 4.0.0-alpha1 and ideally before 4.0.0. I have an idea for doing that while maintaining BC at the interface / object level. As that adds support for syntax that currently causes a parse error, I think it's safe enough to add that later even if people have started using the new mode.

carlos-granados (approved):
Looks good after reverting that change

PR #1804, opened 2026-03-06
docs: Clarify advice for on dependencies for extension authors

acoulton:
As per https://github.com/Behat/Behat/pull/1796#discussion_r2848376629 the previous wording was not totally clear. All extensions have to implement the methods in the interface, and in theory you could implement logic that does not reference the third-party objects that are passed in.

Make the wording explicit that the dependency issue only arises if your implementation references the third-party objects.

### DIFF
--- a/src/Behat/Testwork/ServiceContainer/Extension.php
+++ b/src/Behat/Testwork/ServiceContainer/Extension.php
@@ -44,7 +44,7 @@ interface Extension extends CompilerPassInterface
     /**
      * Setups configuration for the extension.
      *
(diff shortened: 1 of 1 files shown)


stof (approved):

stof:
> n theory you could implement logic that does not reference the third-party objects that are passed in.

other than the no-op logic, I don't think that's actually possible though

acoulton:
> > n theory you could implement logic that does not reference the third-party objects that are passed in.
> 
> other than the no-op logic, I don't think that's actually possible though

Not in any useful/sane way I agree

PR #1805, opened 2026-03-14
Add deprecation messages

carlos-granados:
- Trigger deprecations in all code marked as @deprecated
- Trigger deprecation when loading YAML config files
- Trigger deprecations when using annotations

### DIFF
--- a/src/Behat/Behat/Context/Argument/CompositeFactory.php
+++ b/src/Behat/Behat/Context/Argument/CompositeFactory.php
@@ -10,6 +10,7 @@
 
 namespace Behat\Behat\Context\Argument;
 
+use Behat\Testwork\Deprecation\DeprecationCollector;
 use Behat\Testwork\Suite\Suite;
 
 /**
@@ -27,6 +28,11 @@ final class CompositeFactory implements SuiteScopedResolverFactory
      * @var SuiteScopedResolverFactory[]
      */
     private $factories = [];
+
+    public function __construct()
+    {
+        DeprecationCollector::trigger('CompositeFactory is deprecated and will be removed in 4.0. Use CompositeArgumentResolverFactory instead.');
+    }
 
     /**
      * Registers factory.
--- a/src/Behat/Behat/Context/Snippet/Generator/ContextInterfaceBasedContextIdentifier.php
+++ b/src/Behat/Behat/Context/Snippet/Generator/ContextInterfaceBasedContextIdentifier.php
@@ -12,6 +12,7 @@ namespace Behat\Behat\Context\Snippet\Generator;
 
 use Behat\Behat\Context\Environment\ContextEnvironment;
 use Behat\Behat\Context\SnippetAcceptingContext;
+use Behat\Testwork\Deprecation\DeprecationCollector;
 
 /**
  * Identifier that uses context interfaces to guess which one is target.
@@ -26,6 +27,8 @@ final class ContextInterfaceBasedContextIdentifier implements TargetContextIdent
     {
         foreach ($environment->getContextClasses() as $class) {
             if (in_array(SnippetAcceptingContext::class, class_implements($class))) {
+                DeprecationCollector::trigger('ContextInterfaceBasedContextIdentifier is deprecated in favour of --snippets-for and will be removed in 4.0.');
+
                 return $class;
             }
         }
--- a/src/Behat/Behat/Definition/Context/Annotation/DefinitionAnnotationReader.php
+++ b/src/Behat/Behat/Definition/Context/Annotation/DefinitionAnnotationReader.php
@@ -15,6 +15,7 @@ use Behat\Behat\Definition\Call\Given;
 use Behat\Behat\Definition\Call\RuntimeDefinition;
 use Behat\Behat\Definition\Call\Then;
 use Behat\Behat\Definition\Call\When;
+use Behat\Testwork\Deprecation\DeprecationCollector;
 use ReflectionMethod;
 
 /**
@@ -42,6 +43,8 @@ final class DefinitionAnnotationReader implements AnnotationReader
         if (!preg_match(self::$regex, $docLine, $match)) {
             return null;
         }
+
+        DeprecationCollector::trigger('Using annotations to define steps is deprecated and will be removed in Behat 4.0. Use PHP attributes instead.');
 
         $type = strtolower($match[1]);
         $class = self::$classes[$type];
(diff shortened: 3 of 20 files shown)


src/Behat/Behat/Context/Argument/CompositeFactory.php
```diff
@@ -28,6 +29,11 @@ final class CompositeFactory implements SuiteScopedResolverFactory
      */
     private $factories = [];
 
+    public function __construct()
+    {
+        DeprecationCollector::trigger('CompositeFactory is deprecated and will be removed in 4.0. Use CompositeArgumentResolverFactory instead.');
```
carlos-granados, line 34: In classes which have been deprecated we trigger a deprecation in the constructor

src/Behat/Behat/Context/Snippet/Generator/ContextInterfaceBasedContextIdentifier.php
```diff
@@ -26,6 +27,8 @@ public function guessTargetContextClass(ContextEnvironment $environment)
     {
         foreach ($environment->getContextClasses() as $class) {
             if (in_array(SnippetAcceptingContext::class, class_implements($class))) {
+                DeprecationCollector::trigger('ContextInterfaceBasedContextIdentifier is deprecated in favour of --snippets-for and will be removed in 4.0.');
```
carlos-granados, line 30: In classes which don't have a constructor, we trigger the deprecation in methods which use the functionality of the class

src/Behat/Behat/Definition/Context/Annotation/DefinitionAnnotationReader.php
```diff
@@ -43,6 +44,8 @@ public function readCallee($contextClass, ReflectionMethod $method, $docLine, $d
             return null;
         }
 
+        DeprecationCollector::trigger('Using annotations to define steps is deprecated and will be removed in Behat 4.0. Use PHP attributes instead.');
```
carlos-granados, line 47: We trigger deprecations when annotations are used

src/Behat/Behat/Definition/Pattern/Pattern.php
```diff
@@ -27,10 +29,12 @@ public function __construct(
     /**
      * Returns canonical step text.
      *
-     * @deprecated see getSuggestedMethodName
+     * @deprecated Use getSuggestedMethodName() instead. Will be removed in 4.0.
      */
     public function getCanonicalText(): string
     {
+        DeprecationCollector::trigger('Pattern::getCanonicalText() is deprecated. Use getSuggestedMethodName() instead. It will be removed in 4.0.');
```
carlos-granados, line 36: In deprecated methods we trigger the deprecation if the method is used

src/Behat/Behat/EventDispatcher/Event/AfterBackgroundSetup.php
```diff
@@ -47,8 +47,6 @@ public function getFeature(): FeatureNode
     /**
      * Returns scenario node.
      *
-     * @deprecated Use getBackground() instead
```
carlos-granados, line 50: We cannot really deprecate this code. This class implements the ScenarioLikeTested interface which requires that the getScenario method is implemented in the class

src/Behat/Behat/Output/Statistics/StepStatV2.php
```diff
@@ -32,7 +32,6 @@ public function __construct(
         private readonly ?string $error = null,
         private readonly ?string $stdOut = null,
     ) {
-        parent::__construct($stepText, $stepPath, $resultCode, $error, $stdOut);
```
carlos-granados, line 35: Even though for now StepStatV2 needs to continue extending StepStat, we don't call the constructor so that the deprecation in that class is not triggered. Instead we reimplement all the methods defined in the base class so that we don't actually use it

src/Behat/Testwork/Output/Printer/StreamOutputPrinter.php
```diff
@@ -40,8 +41,13 @@ public function setOutputPath($path)
         $this->flush();
     }
 
+    /**
```
carlos-granados, line 44: These methods had not yet been marked as deprecated but they should be because they implement an interface where these methods have been deprecated

src/Behat/Testwork/ServiceContainer/Configuration/ConfigurationLoader.php
```diff
@@ -197,6 +198,8 @@ private function loadFileConfiguration($configPath, $profile): array
 
             $config = $phpConfig->toArray();
         } else {
+            DeprecationCollector::trigger(printf('Loading YAML configuration files is deprecated and will be removed in Behat 4.0. Please convert "%s" to a PHP configuration file. You can use the --convert-config option to do this automatically.', $configPath));
```
carlos-granados: We trigger a deprecation if any configuration is loaded from a YAML file

src/Behat/Behat/EventDispatcher/Event/AfterBackgroundSetup.php
acoulton: I think potentially we should deprecate the `ScenarioLikeTested` interface? 

It looks like we only use it in a couple of places internally, and I think possibly only because originally we couldn't use union types there. Ultimately all the `BackgroundTested` and `ScenarioTested` events are different types of things and generally need to be handled differently...

I could take a look at that.

I think potentially we could keep the `@deprecated` tag on this method so that e.g. phpstan will report a problem if you are calling it directly on an instance of this class. But you're right we can't emit a runtime deprecation as long as it's part of the interface.

src/Behat/Testwork/Output/Printer/StreamOutputPrinter.php
acoulton: I think generally analysis tools will cascade the `@deprecated` from parent interfaces, but I agree it's better to be explicit.

acoulton (changes requested):
@carlos-granados sorry for the slow reply, this looks great. 

Just the comment about keeping the `@deprecated` phpdoc on inappropriate methods from `ScenarioLikeTested` that I think we should solve differently.

src/Behat/Behat/EventDispatcher/Event/AfterBackgroundSetup.php
carlos-granados: @acoulton I've restored these `@deprecated` annotations and will let you handle the deprecation of the interface

acoulton (approved):
Brilliant thanks @carlos-granados

Issue #1745, opened 2025-12-04
[4.x] Review which parts of the codebase should be marked as `@internal`


acoulton:
There are several areas of the codebase that we don't really expect people to use directly, but that are not marked up as `@internal`.

We should review and mark up as required for 4.0 so that the supported public API (for end users & extension authors) is more clearly defined. This will make it easier to add features & fixes in future without BC breaks.

acoulton:
@carlos-granados @stof I'd like to propose a different solution for this. 

Instead of marking things as `@internal`, I think we should follow the same strategy of allow-listing that [phpstan uses for BC for extensions](https://phpstan.org/developing-extensions/backward-compatibility-promise):

* Everything is assumed to be internal by default
* Non-final classes with `@api` in their PHPDoc can be extended.
* Objects of classes with `@api` in their PHPDoc cannot be created via new unless there’s also @api above the public constructor (they should instead be obtained from the DI container)
* All public methods from classes with `@api` in their PHPDoc can be called.
* Interfaces with `@api` in their PHPDoc can be implemented and extended.
* All methods from interfaces with `@api` in their PHPDoc can be called.
* Public methods with `@api` in their PHPDoc (or in the PHPDoc of the declaring class/interface) can be called.

I chatted to Ondrej about their approach at PHPUK today. My main concern was about DX considering e.g. IDEs will not automatically highlight unsupported usage as they would when using `@internal` tags. PHPStan have not had complaints / problems about that, and it's possible to configure phpstan with custom rules based on the [Restricted Usage Extensions](https://phpstan.org/developing-extensions/restricted-usage-extensions) to detect & enforce incorrect usage of internal API.

My logic is:

* Like PHPStan, Behat's primary purpose is to run / call / interact with user-supplied code. This is the opposite to e.g. a library / framework where the primary usage is to be called *by* the user. Therefore, the majority of our codebase (certainly implementations but also some interfaces) never need to be called from outside - in fact many end-users will use Behat successfully without ever calling our code other than the attributes & perhaps event objects.
* Making "internal" code part of the public API is BC, but "internalising" things that were previously public is a BC break. Therefore our starting point for 4.0.0-alpha1 should be to make basically everything "internal" apart from a well-defined public API. That allows us to make future development much more efficient, but we can still add things back in later 4.x releases if they cause people problems.
* Allow-listing `@api` methods will be a much smaller initial diff (therefore easier to review and reason about) than marking essentially the whole codebase as `@internal` with a few exceptions.
[...]

PR #1807, opened 2026-03-22
docs: Explicitly mark up interfaces, classes & methods in the public API

acoulton:
Introduces explicit `@api` tags on interfaces, classes and methods that form the Behat public API for end-users and extension authors.

From 4.0, our backwards compatibility promise will only apply to code that is tagged as `@api` - everything else will be treated as `@internal`.

I have limited this initial PR to things I am fairly certain need to be public, including:

* DTO classes for things like events, hooks, results, configuration.
* Interfaces that I think people should / might need to implement to usefully customise Behat's behaviour.

It might be easier to review the individual commits, which explain why groups of classes/interfaces are included.

For the moment, I have not included any of our runtime implementations of the interfaces. I think generally we'd want people to either re-implement the interface, or use the decorator pattern to wrap our implementation - though we might make some exceptions.

I've also not included the `ScenarioLikeTested` interface for now, per discussion on #1805. We definitely shouldn't make that part of our future public API.

I think we can merge this to 3.x now - we can always add more `@api` tags in future PRs.

I've also begun work on the phpstan extension to check that Behat extensions follow our recommended practices. This includes checking that they only use our public API. 

Before we release 4.0, I will complete that tooling and run it against some of the common Behat extensions, to see if it finds any more API that we need to make public.

Works towards #1745

### DIFF
--- a/src/Behat/Behat/Tester/Result/FailedStepSearchResult.php
+++ b/src/Behat/Behat/Tester/Result/FailedStepSearchResult.php
@@ -17,6 +17,8 @@ use Behat\Testwork\Tester\Result\ExceptionResult;
  * Represents a step test result with a failed definition search.
  *
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * @api
  */
 final class FailedStepSearchResult implements StepResult, ExceptionResult
 {
--- a/src/Behat/Hook/AfterFeature.php
+++ b/src/Behat/Hook/AfterFeature.php
@@ -14,11 +14,15 @@ use Attribute;
 
 /**
  * Represents an Attribute for AfterFeature hook.
+ *
+ * @api
  */
 #[Attribute(Attribute::TARGET_METHOD | Attribute::IS_REPEATABLE)]
 final class AfterFeature implements Hook
 {
     /**
+     * @api
+     *
      * @param string|null $filterString
      */
     public function __construct(
--- a/src/Behat/Behat/Context/Argument/ArgumentResolver.php
+++ b/src/Behat/Behat/Context/Argument/ArgumentResolver.php
@@ -19,6 +19,8 @@ use ReflectionClass;
  * @see ContextEnvironmentHandler
  *
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * @api
  */
 interface ArgumentResolver
 {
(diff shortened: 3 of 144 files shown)


src/Behat/Hook/AfterFeature.php
```diff
@@ -14,11 +14,15 @@
...
 /**
  * Represents an Attribute for AfterFeature hook.
+ *
+ * @api
  */
 #[Attribute(Attribute::TARGET_METHOD | Attribute::IS_REPEATABLE)]
 final class AfterFeature implements Hook
 {
     /**
+     * @api
```
carlos-granados, line 24: I'm not really sure if we really need to mark the constructor as @api if we have already marked the class as @api. Not saying that it should not be done, just wondering. If you have an explanation one way or the other it would be good to hear it

src/Behat/Behat/Tester/Result/FailedStepSearchResult.php
carlos-granados, line 25: For example in this class we don't mark the constructor as @api. Not sure why we need to mark it in some classes and not in others

carlos-granados (commented):
@acoulton thanks for your efforts on this, great job. I just have a question as there is something I am unsure of

src/Behat/Hook/AfterFeature.php
acoulton: @carlos-granados so this follows the phpstan pattern:

* If the class has `@api` then you can use it (e.g. to typehint a dependency/method call parameter, etc) and you can call all the public methods it declares - except...
* You cannot do `new Whatever(...)` (= calling __construct) **unless** the constructor has an explicit `@api`).

This is to allow for cases where e.g. we want extensions to be able to depend on services we provide through the DI container, or event objects that we create and pass around. 

Keeping the constructor internal by default allows us to change the class dependencies over time, knowing that end users are only expecting to use/wrap the instance we provide.

On the attribute (and config) classes therefore the constructor gets an `@api` tag because we expect end users to create these in their code so the constructor has to keep BC.

I'll obviously be documenting the meaning of this.

carlos-granados: Understood! Makes sense 😄

carlos-granados (approved):
Great job @acoulton

PR #1808, opened 2026-03-22
[3.x] refactor: Add remaining BC strict types to method params, returns & properties

acoulton:
This is the final pass at adding backwards-compatible strict types in the 3.x series, continuing from #1779 and working towards #1744. 

This set applies all remaining Rector rules, adding exceptions at each level to exclude any changes that would have caused a BC break. Each individual commit has been checked with phpstan and the bc-checker tooling.

I also manually fixed a couple of incorrect phpdoc types along the way. And I have improved type-safety / error-handling within the `ContextSnippetAppender` class, to avoid cases where theoretically we could have passed `false` into methods that were only valid to call with a string.

Once this is merged to 3.x, we can merge up to 4.0 and then look at adding strict types to public methods and non-final classes where these could break BC.


src/Behat/Behat/Definition/Call/RuntimeDefinition.php
```diff
@@ -29,13 +29,12 @@ abstract class RuntimeDefinition extends RuntimeCallee implements Stringable, De
...
      * @param callable    $callable
-     * @param string|null $description
      */
     public function __construct(
         private $type,
         private $pattern,
-        $callable,
-        $description = null,
+        callable|array $callable,
+        ?string $description = null,
```
acoulton, line 40: It is safe to add these strict types here, because the values are passed straight into the parent constructor which already has strict types defined - so any calling code that was passing incorrect types would already be failing.

rector.php
```diff
@@ -15,11 +29,119 @@
     ->withRootFiles()
     ->withPreparedSets(codeQuality: true)
     ->withPhpSets(php82: true)
+    ->withTypeCoverageLevel(51)
```
acoulton: As previously, I've left the temporary changes to Rector config committed so the diff shows the process but I will revert these before merging the PR.

src/Behat/Testwork/ServiceContainer/ExtensionManager.php
```diff
@@ -25,10 +25,10 @@ final class ExtensionManager
      */
     private $extensions = [];
     /**
-     * @var Extension[string]
+     * @phpstan-var array<string,Extension>
```
carlos-granados: Why do we need to use here @phpstan-var instead of @var?

carlos-granados (approved):
@acoulton I've gone through all the changes and they all seem fine. I just added a question about one change where I was not sure why something was done

src/Behat/Testwork/ServiceContainer/ExtensionManager.php
acoulton: Because everyone else was having a lie in so I hadn't had a coffee yet 🤣

Will fix

acoulton: @carlos-granados fixed now

src/Behat/Behat/Definition/Call/RuntimeDefinition.php
```diff
@@ -29,13 +29,12 @@ abstract class RuntimeDefinition extends RuntimeCallee implements Stringable, De
      * @param string      $type
      * @param string      $pattern
      * @param callable    $callable
```
stof: shouldn't we remove those phpdoc with non-matching type ?

acoulton:
@carlos-granados In fixing that `@var` tag, I realised we had quite a few places where our phpdoc was specifying either `@var SomeClass[]` or `@return SomeClass[]` that could safely be made more specific e.g. `@var list<SomeClass>`.

I've therefore added a couple more commits to improve all the ones that I think are safe / BC to change now. I've only done that where the code guarantees the more-specific type (so for example I haven't changed `@return SomeClass[]` on interface methods because people could have validly implemented those to return a map / filtered list with non-sequential keys etc).

I deliberately changed a couple that weren't safe to change, and PHPStan reported errors, so I'm pretty confident these are all fine.

src/Behat/Testwork/Ordering/OrderedExercise.php
```diff
@@ -27,15 +27,12 @@
  */
 final class OrderedExercise implements Exercise
 {
-    /**
-     * @var Orderer
-     */
-    private $orderer;
+    private NoopOrderer|Orderer $orderer;
```
stof: should just be `Orderer` here, as `NoopOrderer` is a subtype.

src/Behat/Behat/Definition/Call/RuntimeDefinition.php
acoulton: Good catch.

I ran an inspection for mismatched phpdoc and in most cases these relate to our custom `callable` support (where the native type is `callable|array` to allow for the "callables" that we map to a `Context` instance at runtime. 

I have taken the opportunity to define that as a `phpstan-type` and reference that throughout the project to make clear that these aren't expected to be any kind of array.

src/Behat/Testwork/Ordering/OrderedExercise.php
acoulton: Also a good catch. I've checked for unions that were introduced by Rector and found a couple of others that could be simplified / fixed as well.

carlos-granados (approved):
Latest changes looking good, let's go ahead with this

carlos-granados:
@acoulton once you merge this, maybe it would be a good time to do a new release so that people can start looking into deprecations.

Also perhaps good to merge all this into the 4.x branch

acoulton:
Thanks @carlos-granados and absolutely, was just holding off to get this in before the next release.

I'll merge this and do a release and merge up to 4.x tomorrow.

Issue #1646, opened 2025-05-26
Pretty formatter should support `gherkin-32` compatibility mode


acoulton:
The new `gherkin-32` compatibility mode introduces some differences in how Behat/Gherkin parses feature files, including:

* changing how left-padding and indentation of lines is trimmed / normalised e.g. https://github.com/Behat/Gherkin/issues/209
* parsing escape sequences and newlines within table cells
* parsing multiline scenario etc titles as a one-line title + separate description

In most cases these will not directly affect Behat itself (end-users will obviously be affected if e.g. they are expecting a literal `\n` within a table cell and receive a multiline string instead).

However, they will affect our pretty formatter because it attempts to render the parsed content back in the original gherkin format and therefore needs to reverse some of the work the parser has done.

PR #1812, opened 2026-03-27
[4.0] feat: Use GherkinCompatibilityMode::GHERKIN_32 by default

acoulton:
In the 4.0 series, by default we will switch to parsing feature files in a way that is compatible with the official cucumber/gherkin parsers. Users can switch back to legacy mode if they have any issues.

Although the GHERKIN_32 mode is still technically experimental, we plan to complete it and mark it stable before Behat v4.0.0.

See https://docs.behat.org/en/latest/user_guide/gherkin/parser_mode.html

We therefore no longer need behat.yml.dist (which was being ignored in 4.0 anyway once we dropped support for YAML config) as this was only there to run our own tests in gherkin-32 mode, which is now the default.

### DIFF
--- a/features/gherkin_compatibility_mode.feature
+++ b/features/gherkin_compatibility_mode.feature
@@ -11,27 +11,28 @@ Feature: Gherkin compatibility mode
       | --format    | progress |
 
   @gherkin-mode:has-explicit
-  Scenario: Legacy parsing by default (ignores invalid language)
+  Scenario: Cucumber-compatible parsing by default (fails on invalid language)
     Given the test runner does not configure a default gherkin compatibility mode
     When I run behat with the following additional options:
       | option    | value   |
       | --profile | default |
+    Then it should fail with:
+      """
+      In KeywordsDialectProvider.php line XX:
+
+        Language not supported: martian
+      """
+  
+    
+  @gherkin-mode:has-explicit
+  Scenario: Opt in to legacy parsing (ignores invalid language)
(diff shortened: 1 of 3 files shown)


acoulton:
Interesting - our Behat is all failing on comparing whitespace output from the pretty formatter. 

This is ultimately due to #1646 - we knew the pretty formatter would need changes. 

But it suggests that #1801 (which was supposed to have switched our suite to run in `gerkin-32` mode in the 3.x branch) didn't actually do anything. Which is probably why it went green so easily :facepalm: 

I will investigate and fix. This also means that #1646 is needed for `4.0.0-alpha1`.

acoulton:
Of course - #1801 did correctly set the new gherkin mode for *our* behat process that runs our features.

But it didn't set the mode for the behat processes that we run *within* that process to verify their behaviour. So they were still running in `legacy` mode, and therefore the pretty formatter output that we're asserting against didn't change until I changed the default parsing mode for all processes in this PR.

acoulton:
This branch should now be fine, we just need to:

* Merge #1847 and #1848 to 3.x
* Merge 3.x up to 4.x
* Rebase this branch onto 4.x
* Update FeatureContext::$gherkinCompatibilityMode to `null` so that we're not overriding the gherkin mode by default (I think it'd be worth keeping the mechanism that #1848 introduces in case we want to test in a different default gherkin mode in future).

carlos-granados (approved):
Thanks @acoulton

PR #1814, opened 2026-03-31
Remove all deprecated code

carlos-granados:
Removes all the code that had been marked as @deprecated in the 3.x branch

The only @deprecated tags left are the ones related to the `getScenario()` functions. @acoulton said he would be looking into them separately


src/Behat/Behat/Context/Environment/Handler/ContextEnvironmentHandler.php
```diff
@@ -43,16 +41,12 @@ final class ContextEnvironmentHandler implements EnvironmentHandler
     /**
      * Initializes handler.
      *
-     * @param ArgumentResolverFactory|SuiteScopedResolverFactory $resolverFactory
+     * @param ArgumentResolverFactory $resolverFactory
```
stof: this should become a native parameter type (btw, the type is wrong in 3.x as `null` is accepted)

src/Behat/Behat/Output/Statistics/TotalStatistics.php
```diff
@@ -44,11 +44,11 @@ final class TotalStatistics implements Statistics
      */
     private array $skippedScenarioStats = [];
     /**
-     * @var list<StepStat>
+     * @var list<StepStatV2>
```
stof, line 47: shouldn't we keep `StepStat` as the name for the type being returned?

src/Behat/Behat/Context/Environment/Handler/ContextEnvironmentHandler.php
carlos-granados: Good shout, updated

src/Behat/Behat/Output/Statistics/TotalStatistics.php
carlos-granados: So, I had many doubts about that. In 3.0 we have two classes, StepStat, which is the one which is being deprecated and removed and StepStatV2 which is the one which remains. I wasn't sure if I should rename StepStatV2 to StepStat. On one hand, this would remove the need of the suffix but on the other hand it may make the life of people upgrading more difficult and perhaps create confusion. @stof @acoulton what do you both think?

stof: Well, as TotalStatistics documents `StepStat` in its type, code consuming it might rely on `StepStat` in some parameter type, which would also be broken during the upgrade with your change.

And keep `StepStatV2` as name looks weird to me.

Thus, you also removed some methods of `StepStatV2` that were not marked as deprecated (all the inherited methods), while output formatters might use them.

carlos-granados: TotalStatistics no longer documents `StepStat`, I changed everything to `StepStatV2`. And I copied all the inherited methods from the old `StepStat` to `StepStatV2`, everything is working fine with all tests and all PHPStan checks passing.
So I think that the only question is if we should keep this `StepStatV2` name or change it to `StepStat`

acoulton: I'm in two minds about this. The naming of `StepStatV2` does indeed look weird.

[StepStat was deprecated in 3.6.0, ten years ago](https://github.com/Behat/Behat/commit/088df8062fea9547a8a30d3423b747fbc7437d27#diff-0e4d216fb7b477a04b4e56f9694aeee28e2895ffe92db3cde959a8c4c356de8eR17-R20) and ever since then we've been telling people it would be deleted in 4.0 and to use StepStatV2. 

That commit shows that `StepStatV2` has always inherited and re-declared all the methods that were in `StepStat` - so the concern about removing inherited methods doesn't apply. 

It looks like the only reason for adding `StepStatV2` was to allow adding more required parameters to the constructor (a good example of why we're being strict about whether the constructor itself is considered part of the public API from 4.0).

### Merging into to `StepStat` and deleting `StepStatV2`

I think the history means it's **much more likely** that people will see a breaking change in 4.0 if we remove `StepStatV2`.

`StepStat` has been deprecated since before a lot of Behat extensions / end-user implementations were written, so I doubt many are actually referencing it.

People could fix that by just searching/replacing the class name (or getting Rector to do it) but that will make it hard for them to support Behat v3.x and 4.0 at the same time, because the `StepStat` in 3.x doesn't define some methods that they might need.

It would also be the exact opposite of what our deprecation and messaging has said we were going to do, for a very long time.

### Keeping `StepStatV2` and deleting `StepStat`

If this breaks for a few people, it's an easy search-and-replace for the class name to fix - nothing else changes. And they can then support 3.x and 4.0 without any issues, because `StepStatV2` is present in both.

So I think we should probably keep `StepStatV2` and accept this is a bit of a historical awkwardness that isn't easy to fix neatly in the next major.

carlos-granados: I agree with you so I'm keeping `StepStatV2`

carlos-granados:
All the BC ignores have now been added, this should be fully ready for final review

tests/Fixtures/ProfileFilters/behat-tag-filter-on-profile.php
```diff
@@ -4,11 +4,14 @@
...
+use Behat\Config\GherkinOptions;
 use Behat\Config\Profile;
 
 return (new Config())
     ->withProfile(
         (new Profile('default'))
-            ->withFilter(new TagFilter('tag2'))
+            ->withGherkinOptions((new GherkinOptions())
+                ->withFilter(new TagFilter('tag2'))
+            )
```
acoulton: I think this example from the fixture can actually be removed, and the `Scenario Outline` in `profile_filters.feature:12` converted to a straight scenario - it was only there to prove that 3.x would behave the same regardless of whether the filter was on the profile or the GherkinOptions.

acoulton (commented):
Thanks @carlos-granados just one small comment on a feature/fixture that I think is now redundant but other than that this looks great to me.

tests/Fixtures/ProfileFilters/behat-tag-filter-on-profile.php
carlos-granados: Good shout, updated

acoulton (approved):
Superb, thanks @carlos-granados

Issue #1748, opened 2025-12-04
[4.x] Review which parts of the codebase should be marked as `final`


acoulton:
There are a number of classes that are not `final` but that we do not actually expect users to extend.

We should make these `final` for 4.0.0 to reduce the public API and make it easier to change implementations in future without breaking BC.

stof:
We should also add a `@final since 3.28.0` phpdoc tag on them in a 3.28.0 release (exact version to be adjusted based on what the next 3.x minor version will be at the time this is done of course)

acoulton:
> We should also add a `@final` since 3.28.0 phpdoc tag

I'm not sure about this, for two reasons:

* They're not actually `final` since then - we'll still be guaranteeing BC on them for the rest of the 3.x series. The only change in 3.x is deprecating the idea of extending them, but I don't think that's quite the same as marking them `@final`.
* Even at level 1, phpstan by default counts extending an `@final` class as an error. So this would likely break CI in for end-users that are using 3.x with phpstan, even though they are not required to change anything until they upgrade to 4.x.

If we think it's important to give advance warning in 3.x then it feels like it'd be more appropriate to trigger a deprecation in the constructor `if ($this::class !== self::class)` (assuming we add a way to report that).

stof:
@acoulton The Symfony BC policy makes a distinction between tags marked as `@final since ...` and other `@final` tags when it comes to what *contributors* can do on those classes (basically, contributors must treat the class as non-final until the next semver-breaking version).

> * Even at level 1, phpstan by default counts extending an `@final` class as an error. So this would likely break CI in for end-users that are using 3.x with phpstan, even though they are not required to change anything until they upgrade to 4.x.

Does this depend whether you turn on the reporting of deprecations in phpstan ?

acoulton:
> > * Even at level 1, phpstan by default counts extending an `@final` class as an error. So this would likely break CI in for end-users that are using 3.x with phpstan, even though they are not required to change anything until they upgrade to 4.x.
> 
> Does this depend whether you turn on the reporting of deprecations in phpstan ?

As far as I can tell, no - it is not affected by phpstan config. If I:
* do a `composer require phpstan/phpstan` in a clean directory, without installing the `phpstan-deprecation-rules` extension
* then run it at level 1 against this [playground example](https://phpstan.org/r/ea814d26-c3e2-441c-9a5d-5fda2ff4778c) 
[...]

PR #1815, opened 2026-04-02
[4.0] refactor: Finalise all non-API classes that are not extended

acoulton:
This PR starts work towards #1748 with the goal that we use `final` classes wherever possible in 4.0.

As a first step, I have finalised all classes that:

* Are not tagged `@api` (and therefore should not be used or extended in third-party code from 4.0 onwards).
* Are not extended by Behat itself.

Making these classes final also allows us to make a number of methods `private` rather than `protected`.

### DIFF
--- a/.roave-backward-compatibility-check.xml
+++ b/.roave-backward-compatibility-check.xml
@@ -107,5 +107,157 @@
         <ignored-regex>#CHANGED: Default parameter value for parameter \$defaultVerbosity of Behat\\Testwork\\Exception\\ExceptionPresenter\#__construct\(\) changed#</ignored-regex>
         <ignored-regex>#CHANGED: The number of required arguments for Behat\\Testwork\\Exception\\ExceptionPresenter\#__construct\(\) increased#</ignored-regex>
         <ignored-regex>#CHANGED: The parameter \$.+? of Behat\\Testwork\\Exception\\ExceptionPresenter\#__construct\(\) changed#</ignored-regex>
+
+        <!--
+        Make all non-API classes final unless they are extended by Behat (also "removes" protected methods)
+        NB that methods that changed from protected to private are listed twice - once as "removed" and once with the
+        visibility changes.
+        -->
+        <ignored-regex>#CHANGED: Class Behat\\Testwork\\PathOptions\\Cli\\PathOptionsController became final#</ignored-regex>
+        <ignored-regex>#CHANGED: Class Behat\\Testwork\\Call\\Exception\\FatalThrowableError became final#</ignored-regex>
+        <ignored-regex>#CHANGED: Class Behat\\Testwork\\Output\\Printer\\Factory\\FilesystemOutputFactory became final#</ignored-regex>
+        <ignored-regex>#REMOVED: Method Behat\\Testwork\\Output\\Printer\\Factory\\FilesystemOutputFactory\#configureOutputStream\(\) was removed#</ignored-regex>
+        <ignored-regex>#CHANGED: Method configureOutputStream\(\) of class Behat\\Testwork\\Output\\Printer\\Factory\\FilesystemOutputFactory visibility reduced from protected to private#</ignored-regex>
+        <ignored-regex>#CHANGED: Class Behat\\Testwork\\Output\\Cli\\OutputController became final#</ignored-regex>
+        <ignored-regex>#REMOVED: Method Behat\\Testwork\\Output\\Cli\\OutputController\#configureFormatters\(\) was removed#</ignored-regex>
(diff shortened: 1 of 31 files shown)


acoulton:
I originally included Rector changes in this branch to add more strict types to methods that are now private. 

However I realised that diff was too big - it was also picking up type changes that we'd deliberately left out of 3.x because they're not BC, and the changes that I've now backported into #1815.

I'll look at Rector strict type rules on 4.0 as a separate PR after this is merged.

src/Behat/Behat/EventDispatcher/ServiceContainer/EventDispatcherExtension.php
acoulton: This line is removed in #1814 so the phpstan warning will be fixed when I rebase after that is merged.

src/Behat/Behat/Output/ServiceContainer/Formatter/PrettyFormatterFactory.php
```diff
@@ -65,7 +65,7 @@
  *
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
  */
-class PrettyFormatterFactory implements FormatterFactory
+final class PrettyFormatterFactory implements FormatterFactory
```
acoulton, line 68: It is possible that people are extending our formatter factory classes - but I don't think we want to encourage that in future.

As with the PRs that first tagged up our expected public API, I'd prefer to start by making all our now-not-public-api `final` in 4.0.0-alpha1 and then we can always consider relaxing that for specific cases if we find a good reason to extend them.

src/Behat/Behat/Transformation/ServiceContainer/TransformationExtension.php
```diff
@@ -152,7 +152,7 @@ protected function processArgumentsTransformers(ContainerBuilder $container)
      *
      * @todo Remove method in next major version
      */
-    protected function getDefinitionArgumentTransformerId()
+    private function getDefinitionArgumentTransformerId()
```
acoulton: This is removed in #1814 so the phpstan warning will be fixed when I rebase after that is merged.

carlos-granados:
Thanks @acoulton I would prefer to review this once it has been rebased and all tests are green. Please let me know

acoulton:
> Thanks @acoulton I would prefer to review this once it has been rebased and all tests are green. Please let me know

Of course, it was more to sense-check the thinking behind the PR but no problem.

It's rebased and green now. The list of BC breaks in the ignore file is quite big, but that is partly because it lists every method/property that was previously `protected` twice. It's also slightly misleading on the wording of some of the removals but I've explained that in the file itself.

.roave-backward-compatibility-check.xml
```diff
@@ -107,5 +107,157 @@
...
+        NB that methods that changed from protected to private are listed twice - once as "removed" and once with the
+        visibility changes.
+        -->
+        <ignored-regex>#CHANGED: Class Behat\\Testwork\\PathOptions\\Cli\\PathOptionsController became final#</ignored-regex>
+        <ignored-regex>#CHANGED: Class Behat\\Testwork\\Call\\Exception\\FatalThrowableError became final#</ignored-regex>
+        <ignored-regex>#CHANGED: Class Behat\\Testwork\\Output\\Printer\\Factory\\FilesystemOutputFactory became final#</ignored-regex>
+        <ignored-regex>#REMOVED: Method Behat\\Testwork\\Output\\Printer\\Factory\\FilesystemOutputFactory\#configureOutputStream\(\) was removed#</ignored-regex>
+        <ignored-regex>#CHANGED: Method configureOutputStream\(\) of class Behat\\Testwork\\Output\\Printer\\Factory\\FilesystemOutputFactory visibility reduced from protected to private#</ignored-regex>
+        <ignored-regex>#CHANGED: Class Behat\\Testwork\\Output\\Cli\\OutputController became final#</ignored-regex>
+        <ignored-regex>#REMOVED: Method Behat\\Testwork\\Output\\Cli\\OutputController\#configureFormatters\(\) was removed#</ignored-regex>
```
carlos-granados, line 122: Maybe for these cases where there were several functions moved from protected to private in a class we can use a regular expression pattern like:
`#REMOVED Method Behat\\Testwork\\Output\\Cli\\OutputController\#.?\(\) was removed#` so that we don't have to list so many lines. This should not be problematic because there are no other protected functions that may be changed in those classes

carlos-granados (commented):
Looks good overall, just a comment on the BC break file

.roave-backward-compatibility-check.xml
acoulton: Oh yeah, good shout.

acoulton: Updated

carlos-granados (approved):
All looking good now

PR #1816, opened 2026-04-02
fix: Return type on ScenarioLikeTested, and add strict returns to `final` events

acoulton:
The `ScenarioLikeTested::getScenario()` method had a phpdoc claiming that it would always return a `ScenarioInterface`.

In fact, that has never been true:

* The `BackgroundTested` family of implementations in fact return a `BackgroundNode`. This does *not* implement `ScenarioInterface` - it implements the parent `ScenarioLikeInterface`.
* The `ScenarioTested` family of implementations return a `ScenarioLikeInterface` - this might be a `ScenarioInterface` but there is no guarantee that it is.

The typehint on the interface is (and always has been) incorrect.

Additionally, the event objects here have always been `final` but the previous PRs to add backwards compatible return types missed them. I'm not sure if I accidentally excluded them as non-BC, or if Rector has got better at spotting them.

Either way, this is safe to go into 3.x.

Continues #1744

### DIFF
--- a/src/Behat/Behat/EventDispatcher/Event/ScenarioLikeTested.php
+++ b/src/Behat/Behat/EventDispatcher/Event/ScenarioLikeTested.php
@@ -11,7 +11,7 @@
 namespace Behat\Behat\EventDispatcher\Event;
 
 use Behat\Gherkin\Node\FeatureNode;
-use Behat\Gherkin\Node\ScenarioInterface;
+use Behat\Gherkin\Node\ScenarioLikeInterface;
 
 /**
  * Represents an event of scenario-like structure (Scenario, Background, Example).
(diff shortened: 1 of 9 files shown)


src/Behat/Behat/EventDispatcher/Event/ScenarioLikeTested.php
```diff
@@ -30,7 +31,7 @@ public function getFeature();
     /**
      * Returns scenario node.
      *
-     * @return ScenarioInterface
```
acoulton, line 33: As described in the commit message, this is (and always has been) incorrect. Classes that implement this will return a `ScenarioLikeInterface`. This could be a `ScenarioInterface`, but it could also be something else.

Since this is both wrong, and only a phpdoc type declaration, I don't consider it to be a BC break to change it in 3.x

carlos-granados (approved):
A step in the good direction, thanks @acoulton

PR #1817, opened 2026-04-03
refactor: Bump phpstan & rector and apply new refactorings

acoulton:
We have pinned rector to a specific version, to avoid unexpected changes in CI. This means that we also had to pin phpstan as (in #1786) we found that our Rector version was not fully compatible with newer phpstan.

This PR updates both dependencies to their latest version, and applies a small number of new internal refactorings that Rector proposed.


carlos-granados (approved):
Thanks @acoulton

PR #1818, opened 2026-04-09
refactor: Apply further strict types to 3.x

acoulton:
In #1808 I took Rector up to the maximum set of type coverage rules that we could apply in 3.x without BC breaks. However, this was before #1817 where I bumped Rector to the latest version.

Rector is now able to find a few more places where we can safely add strict types in 3.x - and this also highlighted some phpdoc types that I was able to tighten manually.

Continues #1744


carlos-granados (approved):
Great stuff @acoulton

PR #1820, opened 2026-04-10
[4.0] refactor: Add strict types to interface method params & returns

acoulton:
Adds strict parameter / return types to all methods that are defined in an interface. The types are applied both to the interface and to all implementations of the method.

I have used my [ingenerator/risky-rector-rules](https://github.com/ingenerator/risky-rector-rules) package to:

* generate the interface method signatures from their phpdoc, removing redundant phpdoc at the same time
* explicitly mark interface methods as `void` if they were not documented as returning anything
* add parameter types in all methods that extend / implement a parent method (if the parameter was previously untyped). This is theoretically risky if a child class was intentionally accepting a wider type than the parent, but I have not yet found anywhere that it actually causes a problem.

I have also used standard Rector rules to add method return types based on the new interface method returns, and to refactor a few `==` that are now safe to be `===` because they are comparing known types.

Along the way, I've made some manual fixes / changes to the auto-generated types where our phpdoc was wrong/incomplete.

I've left in the commits that added the extra Rector package & rules for transparency, but reverted them out at the end. The risky-rector-rules are only intended to be used for a one-time migration.

This continues #1744

After this is merged, the next step will be to run Rector's own type-declaration rules again to add types based on analysing the new method parameter / return type information.

### DIFF
--- a/src/Behat/Behat/EventDispatcher/Event/StepTested.php
+++ b/src/Behat/Behat/EventDispatcher/Event/StepTested.php
@@ -11,6 +11,7 @@
 namespace Behat\Behat\EventDispatcher\Event;
 
 use Behat\Gherkin\Node\FeatureNode;
+use Behat\Gherkin\Node\NodeInterface;
 use Behat\Gherkin\Node\StepNode;
 use Behat\Testwork\EventDispatcher\Event\LifecycleEvent;
 
@@ -42,7 +43,7 @@ abstract class StepTested extends LifecycleEvent implements GherkinNodeTested
      */
     abstract public function getStep();
 
-    final public function getNode()
+    final public function getNode(): NodeInterface
     {
         return $this->getStep();
     }
--- a/src/Behat/Behat/EventDispatcher/Tester/EventDispatchingFeatureTester.php
+++ b/src/Behat/Behat/EventDispatcher/Tester/EventDispatchingFeatureTester.php
@@ -42,7 +42,7 @@ final class EventDispatchingFeatureTester implements SpecificationTester
     ) {
     }
 
-    public function setUp(Environment $env, $spec, $skip): Setup
+    public function setUp(Environment $env, mixed $spec, bool $skip): Setup
     {
         $event = new BeforeFeatureTested($env, $spec);
 
@@ -57,12 +57,12 @@ final class EventDispatchingFeatureTester implements SpecificationTester
         return $setup;
     }
 
-    public function test(Environment $env, $spec, bool $skip): TestResult
+    public function test(Environment $env, mixed $spec, bool $skip): TestResult
     {
         return $this->baseTester->test($env, $spec, $skip);
     }
 
-    public function tearDown(Environment $env, $spec, bool $skip, TestResult $result): Teardown
+    public function tearDown(Environment $env, mixed $spec, bool $skip, TestResult $result): Teardown
     {
         $event = new BeforeFeatureTeardown($env, $spec, $result);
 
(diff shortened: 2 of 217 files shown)


src/Behat/Behat/EventDispatcher/Tester/EventDispatchingFeatureTester.php
```diff
@@ -40,7 +42,7 @@ public function __construct(
     ) {
     }
 
-    public function setUp(Environment $env, $spec, $skip)
+    public function setUp(Environment $env, mixed $spec, bool $skip): Setup
```
acoulton, line 45: `$spec` has to be `mixed` here because we've defined the `Behat\Testwork\Tester\SpecificationTester` interface as a generic that can take any type of `$spec`.  Our implementations are then all defined as ` @implements SpecificationTester<FeatureNode>`.

The original idea was that the `Testwork` package could be extracted to create any type of test runner - but I don't think anyone has ever done that. However we can't easily tighten the `SpecificationTester` interface without creating a dependency from the `Testwork` namepace to the `Behat` classes and that would break our current architectural model. I think probably we leave this and look at reducing / removing some of the abstraction / layers in a future major release if we think it's worth it.

src/Behat/Behat/EventDispatcher/Event/StepTested.php
stof, line 42: shouldn't those abstract methods also get a return type ?

acoulton: To keep this at least vaguely reviewable, I've limited this PR to methods that come from interfaces. 

I will pick them up in a later PR to deal with any remaining class methods.

src/Behat/Behat/EventDispatcher/Tester/EventDispatchingFeatureTester.php
stof: When working on building the runner on top of the concept of pickles (see `cucumber/gherkin`), we will probably have to break the current abstraction anyway (unless the layer of abstraction we have in Testwork could actually be used to create `SpecificationTester<Pickle>` or something like that).
So I agree it is not worth breaking this now.

acoulton:
NB that we've already configured the bc-checker ignore file in the 4.x branch to ignore all warnings about parameters and returns changing from `no type` to anything, so I didn't have to add any more expected BC breaks in this PR.

stof:
For the `Callee::getDescription` issue, I suggest fixing the bad phpdoc in the 3.x branch.

src/Behat/Behat/EventDispatcher/Event/StepTested.php
stof: well, abstract methods in abstract classes are quite similar to interface methods. But I'm fine with handling them as a separate PR (as long as they are not forgotten)

acoulton: Agreed, it's just my Rector rule only has a filter for applying types to `interface` or `everything`. They definitely won't be forgotten, I've multiple passes of type declarations to do ending with a check that absolutely every method param / return / property has a strict type before we consider #1744 complete for 4.0.0-alpha1.

acoulton:
> For the `Callee::getDescription` issue, I suggest fixing the bad phpdoc in the 3.x branch.

Good shout - https://github.com/Behat/Behat/pull/1821

carlos-granados (approved):
I was finally able to take a good look at this, did not see anything that seemed wrong

acoulton:
Thanks @carlos-granados

PR #1822, opened 2026-04-17
[4.0] refactor: Add strict types based on Rector type-declaration rules

acoulton:
Continues #1744 to add more strict types based on Rector's type declaration rules.

This includes:

* Types that Rector could detect in the 3.x branch (including `void` methods) but that we skipped because they would have caused a BC break.
* Types that Rector can detect now that we have strict types on our interfaces in 4.x (e.g. a method that returns the result of calling an interface that now has a strict type).

As usual, I have committed each type coverage level separately to aid review.

### DIFF
--- a/.roave-backward-compatibility-check.xml
+++ b/.roave-backward-compatibility-check.xml
@@ -25,6 +25,9 @@
         <!-- Addition of strict return types is expected throughout the 4.0 release -->
         <ignored-regex>#CHANGED: The return type of .+? changed from no type to#</ignored-regex>
 
+        <!-- Addition of strict property types is expected throughout the 4.0 release -->
+        <ignored-regex>#CHANGED: Type of property .+? changed from having no type to#</ignored-regex>
+
         <!--
         When running in the 4.x branch, the BC checker installs symfony 8, therefore reports BC breaks in symfony
         classes that we extend (since the 3.x release it compares to installed symfony 7)
--- a/rector.php
+++ b/rector.php
@@ -14,7 +14,7 @@ return RectorConfig::configure()
     ->withRootFiles()
     ->withPreparedSets(codeQuality: true)
     ->withPhpSets(php82: true)
-    ->withTypeCoverageLevel(13)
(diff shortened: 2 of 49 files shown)


carlos-granados (commented):
Looks good but I wonder if now that we seem to be reaching the end of applying all Rector rules, if it would be good to *keep*  the rector configuration with the type coverage level, so that the code continues following all these rules for future code additions

acoulton:
> Looks good but I wonder if now that we seem to be reaching the end of applying all Rector rules, if it would be good to *keep*  the rector configuration with the type coverage level, so that the code continues following all these rules for future code additions

I agree, I took them back out in this branch as they're not fully safe until we have all the phpdoc types converted.

But I have left them in in the next PR https://github.com/Behat/Behat/pull/1824#discussion_r3103762910

carlos-granados:
> > Looks good but I wonder if now that we seem to be reaching the end of applying all Rector rules, if it would be good to _keep_  the rector configuration with the type coverage level, so that the code continues following all these rules for future code additions
> 
> I agree, I took them back out in this branch as they're not fully safe until we have all the phpdoc types converted.
> 
> But I have left them in in the next PR [#1824 (comment)](https://github.com/Behat/Behat/pull/1824#discussion_r3103762910)

Ah, great, I had not looked into the other PR yet

carlos-granados (approved):

acoulton:
> > > Looks good but I wonder if now that we seem to be reaching the end of applying all Rector rules, if it would be good to _keep_  the rector configuration with the type coverage level, so that the code continues following all these rules for future code additions
> > 
> > 
> > I agree, I took them back out in this branch as they're not fully safe until we have all the phpdoc types converted.
> > But I have left them in in the next PR [#1824 (comment)](https://github.com/Behat/Behat/pull/1824#discussion_r3103762910)
> 
> Ah, great, I had not looked into the other PR yet

Makes sense, there were a lot of them :laughing:

PR #1823, opened 2026-04-17
fix: Type hinting / safety of TableColumnTransformation

acoulton:
The phpdoc declared that `$argumentValue` would be either a `TableNode` or an `array`.

However, this is not compatible with the parent interface, where the argumentValue is effectively `mixed`.

In our own / expected usage, the caller should first call `supportsDefinitionAndArgument` to check that the argument is suitable for this transformation.

So in practice within Behat we can only receive a `TableNode|array` but in theory another caller could pass something different.

We cannot convert this into a strict parameter type in 4.0 because of the parent type compatibility issue. Instead, we should narrow the type at runtime to enforce that the caller has checked (or knows) that the argumentValue is supported.

We can safely do this in 3.x without a BC break.


carlos-granados (approved):
Looks good and agree with your explanation

PR #1824, opened 2026-04-17
[4.0] refactor: Add remaining auto-generated strict types

acoulton:
This temporarily reintroduces ingenerator/risky-rector-rules to convert the remaining phpdoc parameter, return and property types into strict types.

First, I generated strict types for all objects -`@param` and `@return` types, then types that Rector could now infer automatically, then finally any `@var` types. Then I ran Rector again which generated some changes, including making the newly-typed properties readonly.

While reviewing that first pass, I realised that #1820 had missed some interface `@param` and `@return` where they had union types. This was because my ingenerator/risky-rector-rules didn't support converting union types to strict types.

I've fixed that, so the second half of this branch follows the same sequence of commits - interface methods, class methods, Rector's type declaration rules, phpdoc `@var` - for the union-types that were missed in the first pass.

I think this PR gets us as far as we can go with generating/converting strict types for #1744. 

The next step will be to review any remaining methods / properties with missing types and add them manually.

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -13,9 +13,11 @@ return RectorConfig::configure()
         __DIR__ . '/src',
     ])
     ->withRootFiles()
-    ->withPreparedSets(codeQuality: true)
+    ->withPreparedSets(
+        codeQuality: true,
+        typeDeclarations: true,
+    )
     ->withPhpSets(php82: true)
-//    ->withTypeCoverageLevel(66)
     ->withSkip([
         StringableForToStringRector::class,
         ReadOnlyClassRector::class,
--- a/rector.php
+++ b/rector.php
@@ -1,6 +1,8 @@
 <?php
 
 declare(strict_types=1);
+
+use Ingenerator\RiskyRectorRules\PhpDocToStrictTypes\AddPropertyTypeFromPhpDocRector;
 use Rector\Config\RectorConfig;
 use Rector\Php80\Rector\Class_\StringableForToStringRector;
 use Rector\Php81\Rector\Array_\ArrayToFirstClassCallableRector;
@@ -12,9 +14,9 @@ return RectorConfig::configure()
         __DIR__ . '/src',
     ])
     ->withRootFiles()
-    ->withPreparedSets(codeQuality: true)
-    ->withPhpSets(php82: true)
-    ->withTypeCoverageLevel(49)
+//    ->withPreparedSets(codeQuality: true)
+//    ->withPhpSets(php82: true)
+//    ->withTypeCoverageLevel(66)
     ->withSkip([
         StringableForToStringRector::class,
         ReadOnlyClassRector::class,
@@ -24,4 +26,7 @@ return RectorConfig::configure()
     ->withImportNames(
         removeUnusedImports: true,
     )
+    ->withRules([
+        AddPropertyTypeFromPhpDocRector::class,
+    ])
 ;
--- a/rector.php
+++ b/rector.php
@@ -14,6 +14,7 @@ return RectorConfig::configure()
     ->withRootFiles()
     ->withPreparedSets(codeQuality: true)
     ->withPhpSets(php82: true)
+    ->withTypeCoverageLevel(0)
     ->withSkip([
         StringableForToStringRector::class,
         ReadOnlyClassRector::class,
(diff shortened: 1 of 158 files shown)


src/Behat/Behat/Context/Exception/ContextNotFoundException.php
```diff
@@ -21,23 +21,18 @@ final class ContextNotFoundException extends InvalidArgumentException implements
...
      * Initializes exception.
-     *
-     * @param string $message
-     * @param string $class
      */
     public function __construct(
-        $message,
-        private $class,
+        string $message,
+        private readonly string $class,
```
acoulton, line 27: Now that properties have strict types, Rector is able to mark them as readonly if they are private / the class is final and nothing writes to them.

src/Behat/Behat/Gherkin/ServiceContainer/GherkinExtension.php
```diff
@@ -194,16 +194,14 @@ private function loadParser(ContainerBuilder $container, string $compatibilityMo
...
      */
-    private function loadDefaultLoaders(ContainerBuilder $container, $cachePath): void
+    private function loadDefaultLoaders(ContainerBuilder $container, string $cachePath): void
     {
         $definition = new Definition(GherkinFileLoader::class, [
             new Reference('gherkin.parser'),
         ]);
 
-        if ($cachePath) {
+        if ($cachePath !== '') {
```
acoulton, line 204: Now that the type is known, Rector highlighted a few places where we should be explicit about what "empty" values we're checking for.

src/Behat/Behat/Gherkin/Specification/Locator/FilesystemFeatureLocator.php
```diff
@@ -142,10 +140,6 @@ private function findAbsolutePath(string $path): string|false
             return realpath($path);
         }
 
-        if (null === $this->basePath) {
-            return false;
-        }
```
acoulton, line 147: `basePath` has always been (phpdoc) typed as `string` as far as I can see, and it always gets a value when we construct this class. 

It looks like this null check (which was caught by phpstan) has always been redundant.

src/Behat/Behat/Hook/Call/RuntimeFeatureHook.php
```diff
@@ -54,18 +52,13 @@ public function filterMatches(HookScope $scope): bool
...
-     */
-    private function isMatch(FeatureNode $feature, $filterString)
+    private function isMatch(FeatureNode $feature, string $filterString): bool
     {
         if (str_contains($filterString, '@')) {
             return $this->isMatchTagFilter($feature, $filterString);
         }
 
-        if (!empty($filterString)) {
+        if ($filterString !== '') {
```
acoulton, line 61: Previously this would also have ignored the value `'0'` (due to `empty()` behaviour) but I think that's not actually expected and we should only be ignoring the empty string. There were a couple of other places like this.

rector.php
```diff
@@ -12,7 +12,10 @@
         __DIR__ . '/src',
     ])
     ->withRootFiles()
-    ->withPreparedSets(codeQuality: true)
+    ->withPreparedSets(
+        codeQuality: true,
+        typeDeclarations: true,
```
acoulton, line 17: I have left the `typeDeclarations` Rector rules enabled as part of this PR.

Now that we have strict types (almost) everywhere, there should be very little risk of Rector proposing invalid type changes in future on the 4.x branch.

Leaving these rules active will help us to get through the final manual steps to prepare for strict types in 4.0, and in future will highlight any potential issues with new / contributed code. 

We can also be fairly confident that the backwards-compatibility-checker action will highlight any proposed type changes that would be a BC break.

acoulton:
@carlos-granados this is ready for review, although no worries if you'd prefer to wait until #1822 is merged and I've rebased this.

It's thrown up a couple of places where either our phpdoc was incorrect and/or we should have been handling `null` values. I've got fixes for those in this branch at the moment but I've also backported them to 3.x in #1823, #1825 and #1826.

acoulton:
This is rebased now, it was green last night but is now failing due to the phpunit security vuln released this morning.

We won't be able to run CI on it until #1827 is merged, released, merged up to 4.x, and then this branch rebased again.

All for a vulnerability that doesn't affect us! :facepalm: but better to be secure by default :laughing:

acoulton:
@carlos-granados this is now rebased and green for when you have a chance to take a look.

carlos-granados (approved):
I took a good look and could not find anything that did not look right, let's go ahead with this

PR #1825, opened 2026-04-17
fix: Incorrect not-null string phpdoc typehints

acoulton:
In both of these methods the phpdoc specifies that `$type` will/must always be a `string`. 

In practice it is very often null, and Behat correctly treats that as the 'default' case. The internal methods that are called all accept `string|null` and therefore so should the public method.


carlos-granados (approved):
Makes sense

PR #1826, opened 2026-04-17
fix: Always populate language in `TranslatedDefinition`

acoulton:
If no language was passed to `DefinitionTranslator` (which accepts a null language) then it will use the default locale.

Previously we were then passing that null through to the resultant `TranslatedDefinition`. 

But the phpdoc for that class expects that the `$language` will always be `string` - both on the property itself and on the getter.

It seems more logical to pass in the language that was actually used (e.g. the default locale) to the DTO that carries the translation, rather than allowing the object to accept / carry a `null` language.


carlos-granados (approved):
Seems reasonable

PR #1831, opened 2026-04-19
[4.0] refactor: Add remaining strict types and improve runtime type checking

acoulton:
This PR adds strict types for all remaining properties, method parameters and returns. 

I have ensured that is true by bumping our PHPStan to level 6 (ignoring errors about array shapes and generics, which are not relevant to our PHP native strict typehints).

As a result, I have also included a few internal refactorings:

* To be stricter about asserting at runtime that our code is being used as expected / with the expected types
* Where there is a more efficient / neater implementation that works better with strict types.

I've explained these changes in the individual commit messages.

None of these should affect any users or extension authors, as they do not touch our public API. The runtime type checks I've added only enforce that parameters match those that the phpdoc previously specified.

This completes #1744


carlos-granados (approved):
Great explanations in all commits, nice job

PR #1832, opened 2026-04-20
refactor: Add runtime deprecations for ScenarioLikeTested interface

acoulton:
There are three places this could affect users:

* If they are calling `instanceof ScenarioLikeTested`. There's nothing we can do to detect / warn about this other than marking the interface `@deprecated` for their IDE / CI tooling.

* If they are creating and dispatching custom events that implement `ScenarioLikeTested` but *do not* extend either `ScenarioTested` or `BackgroundTested`. We can detect these at the point we handle the event.

* If they are calling `getScenario` instead of `getBackground` on any of the `BackgroundTested` event family.

I have refactored so that our own code only calls `getBackground` on a background node. Any remaining calls to `getScenario` will trigger a deprecation.


carlos-granados (approved):
I think this covers all possible use cases, good job

PR #1833, opened 2026-04-20
ci: Fail on Behat deprecations in CI

acoulton:
We should not be using any deprecated code within Behat itself.

Obviously we might do inside our own features (e.g. to prove that deprecations are reported) but these are controlled by their own Behat config within the test fixture.

Therefore, the outer Behat process that runs the features can safely fail on any Behat deprecation.

In theory this could be configured in behat.yml, but I think it's clearer (and future proof) to take the opportunity to convert to using PHP config for our own Behat run.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -7,7 +7,7 @@
 .github export-ignore
 .php-cs-fixer.dist.php export-ignore
 AGENTS.md export-ignore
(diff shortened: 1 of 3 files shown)


carlos-granados (approved):
Nice moving into PHP configuration

PR #1834, opened 2026-04-20
[4.0] refactor: Remove ScenarioLikeTested in favour of separate events

acoulton:
Separate out the BackgroundTested and ScenarioTested event families - they represent different things. A background is a group of steps, but it is not a "scenario" in its own right.

This allows us to remove the deprecated `getScenario` getter on `BackgroundTested` events, which has always returned the Background node rather than the Scenario that is actually being tested at that moment.


carlos-granados (approved):
Looks good, thanks @acoulton

PR #1836, opened 2026-04-26
Use Traversable as iterator aggregate return type

stof:
Rector inferred a return type based on the implementation of the method. However, I don't think we should guarantee the kind of iterator being returned, to reduce the BC surface of those classes.

Refs #1744

### DIFF
--- a/src/Behat/Testwork/Call/CallResults.php
+++ b/src/Behat/Testwork/Call/CallResults.php
@@ -13,6 +13,7 @@ namespace Behat\Testwork\Call;
 use ArrayIterator;
 use Countable;
 use IteratorAggregate;
+use Traversable;
 
 /**
  * Aggregates multiple call results into a collection and provides an informational API on top of that.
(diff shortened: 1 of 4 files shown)


src/Behat/Testwork/Call/CallResults.php
```diff
@@ -80,7 +81,7 @@ public function count(): int
     /**
      * Returns collection iterator.
      */
-    public function getIterator(): ArrayIterator
```
stof, line 92: It looks like this was treated as `ArrayIterator<mixed>` here (while reporting a `missingTypes.generics` error that is ignored for now), which was hiding other static analysis issues elsewhere.

stof: See https://github.com/Behat/Behat/pull/1837 implementing generic types to document this properly. Once merged, I will rebase this PR (I confirmed locally that it indeed fixes it)

stof:
For reference, I caught this while working on resolving all `missingTypes.generics` errors in phpstan to remove the ignore rule (PR coming next for that)

acoulton (commented):
Thanks @stof, this makes sense and LGTM once rebased on #1837 to fix the generic type error.

acoulton (approved):
Thanks @stof

PR #1837, opened 2026-04-26
Add generic types for CallResult

stof:
This allows documenting that the HookedSetup and HookedTeardown classes are dealing with call results for HookCall, not for arbitrary calls.


acoulton (approved):
Thanks @stof - really helpful

PR #1839, opened 2026-04-26
Add more types for iterable values

stof:
This does not resolve all of them. The remaining ones cover 3 big categories (and maybe other things I missed):

- suite settings
- container raw extension config in the various `Config` value objects
- container processed extension config in extension classes

Those will be handled separately.

### DIFF
--- a/src/Behat/Behat/Context/Snippet/Generator/CachedContextIdentifier.php
+++ b/src/Behat/Behat/Context/Snippet/Generator/CachedContextIdentifier.php
@@ -19,6 +19,9 @@ use Behat\Behat\Context\Environment\ContextEnvironment;
  */
 final class CachedContextIdentifier implements TargetContextIdentifier
 {
+    /**
+     * @var array<string, string|null>
+     */
     private array $contextClasses = [];
 
     /**
--- a/src/Behat/Behat/HelperContainer/BuiltInServiceContainer.php
+++ b/src/Behat/Behat/HelperContainer/BuiltInServiceContainer.php
@@ -30,6 +30,8 @@ final class BuiltInServiceContainer implements PsrContainerInterface
 
     /**
      * Initialises container using provided service configuration.
+     *
+     * @param array<string, array{class?: string, arguments?: mixed[], factory_method?: string|null}|string|null> $schema
      */
     public function __construct(
         private array $schema,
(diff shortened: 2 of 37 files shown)


src/Behat/Behat/Context/Snippet/Generator/CachedContextIdentifier.php
```diff
@@ -19,6 +19,9 @@
  */
 final class CachedContextIdentifier implements TargetContextIdentifier
 {
+    /**
+     * @var array<string, string|null>
```
acoulton, line 23: I *think* this could be `array<string, class-string<Context>|null>`?

src/Behat/Behat/HelperContainer/BuiltInServiceContainer.php
```diff
@@ -74,6 +76,8 @@ private function createInstance(string $id): object
     /**
      * Gets and validates a service configuration for a service with given ID.
      *
+     * @return array{class: string, arguments: mixed[], factory_method?: string|null}
```
acoulton, line 79: Is there value in extracting & referencing a `TServiceDefinitionSchema` or similar for all the methods here that take / return the same `$schema` structure?

acoulton (commented):
Thanks @stof, just a couple of comments

src/Behat/Behat/Context/Snippet/Generator/CachedContextIdentifier.php
stof: no. `TargetContextIdentifier` returns `string|null`

src/Behat/Behat/HelperContainer/BuiltInServiceContainer.php
stof: They don't all deal with the same type. The public API has many optional properties, while this method receives a normalized version where default behavior for `class` and `arguments` have been resolved by the caller.

src/Behat/Behat/Context/Snippet/Generator/CachedContextIdentifier.php
acoulton: Looks to me like the return type for `TargetContextIdentifier` is too loose, as the expectation is it will be a context class and as far as I can see all the implementations do indeed return a context class.

However, I'm OK if we do that in a separate PR.

acoulton (approved):

Issue #1800, opened 2026-02-21
Improve error output when feature file contains invalid language tag


acoulton:
In the new `gherkin-32` compatibility mode, the parser throws an exception if a feature file contains a `# language:` tag for an unknown language (previously, it silently fell back to English).

As seen in #1799, the error output in this case is not ideal:

```
In KeywordsDialectProvider.php line XX:

    Language not supported: martian
``` 

We would ideally catch this `NoSuchLanguageException` and render an error message that includes the path to the invalid feature file. In an ideal world, this would be treated like a failure of a single feature (rather than an outright fatal error) but I suspect that will be difficult to achieve.

stof:
I tried the behavior of Behat with invalid gherkin files., and it is also a fatal error:

```
In Parser.php line 754:
                                                                                                                                                         
  Table row '1' is expected to have 1 columns, got 0 in file /home/stof/src/Behat/Behat/tests/Fixtures/InvalidGherkin/features/invalid_language.feature  
                                                                                                                                                         

In TableNode.php line 60:
                                                      
  Table row '1' is expected to have 1 columns, got 0  
                                                      
```

I think the fix for this issue would be that the Gherkin Parser should wrap the `NoSuchLanguageException` in a parser exception as well, providing the file name.

Treating it like a single feature error would be hard as we would have to generate a FeatureNode for that invalid gherkin file and make that FeatureNode fail during execution. We don't have the infrastructure for that.

PR #1840, opened 2026-04-27
Add a scenario covering the behavior of invalid gherkin files

stof:
While investigating https://github.com/Behat/Behat/issues/1800, I figured out that this case was not covered by our scenarios.

### DIFF
--- a/tests/Fixtures/InvalidGherkin/features/invalid_gherkin.feature
+++ b/tests/Fixtures/InvalidGherkin/features/invalid_gherkin.feature
@@ -1,0 +1,11 @@
+Feature: Invalid gherkin
+  In order to have a good developer experience
+  As a tester
+  I need to get reports about invalid feature files
+
+  Scenario: Reports
+      Given I have a feature file with invalid content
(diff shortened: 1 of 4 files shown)


tests/Fixtures/InvalidGherkin/features/invalid_gherkin.feature
```diff
@@ -0,0 +1,11 @@
+Feature: Invalid gherkin
+  In order to have a good developer experience
+  As a tester
+  I need to get reports about invalid feature files
+
+  Scenario: Enforce language validation
```
acoulton: Only one small observation that the scenario title here doesn't really match the content - but I don't think it's crucial.

acoulton (approved):
Thanks @stof

tests/Fixtures/InvalidGherkin/features/invalid_gherkin.feature
stof: yeah, left-over of a copy paste. I'll change it.

Issue #1813, opened 2026-03-27
Symfony 8 support on version 3.x?


azambon:
Hello,

we're currently maintaining a project that we'd like to port to PHP 8.5 and Symfony 8. And Behat is the last dependency of that project that still doesn't support Symfony 8.

I know that [Symfony 8 support](https://github.com/Behat/Behat/pull/1687) is planned for [version 4.0.0 alpha 1](https://github.com/Behat/Behat/milestone/6), but the backlog for version 4 doesn't seem to be moving all that quickly.

Looking at the Symfony 8 support PR, it only changed 13 files, so it doesn't seem impossible to backport that to version 3.x.

So... any possibility of anticipating Symfony 8 support without having to wait for version 4?

acoulton:
@azambon unfortunately although the diff for supporting symfony8 is small, the changes are not backwards compatible (because they add return types which will break all end-user / extension code that extends those interfaces and classes). 

Therefore we cannot merge them into the 3.x branch as that will break for all existing users (regardless of whether they're upgrading symfony or not).

We are moving as quickly as we can on the backlog for `4.0.0-alpha1`. 

The Behat project does not have any commercial backing, and the volunteer maintainers don't receive much by way of [sponsorship](https://github.com/Behat/Behat?tab=readme-ov-file#support-the-project), but we're doing our best to make time to work on this.

I actually think we're getting quite close to `alpha1` in developer time - most of the more complex things have been done, including quite a bit of progress towards several of the issues that are still open.

Obviously I can't really predict how long it will be in "wall time" as that depends on how much time we can make around our other commitments.

`4.0.0-alpha1` should be stable enough to upgrade to, with the caveat that there might be subsequent changes that will be breaking if you're doing unusual things or using APIs/features that we've said will not be supported in the final `4.0`.

The other factor of course is that any extensions you're using will also need to be updated to support Behat 4.0. We're hoping this process will be as easy for extension authors as it can be.

Contributions - either to Behat itself, or to start looking at upgrading extensions - are always welcome.

PR #1841, opened 2026-05-04
Feature/symfony8 compatibility

frenchcomp:
Symfony 8 has been released 6 months ago and Behat is only compliant with symfony component from versions 5 to 7.
If Behat is added in dev require of a symfony project, the project is locked in 7.

This commit allow symfony 8 in composer require and fix some Symfony Components' interfaces implementations.

Tests are green localy

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -57,7 +57,12 @@ jobs:
             composer-mode: update
             symfony-version: '6.4'
 
-          - php: 8.3
+          - php: 8.4
+            os: ubuntu-latest
+            composer-mode: update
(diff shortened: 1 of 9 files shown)


acoulton:
@frenchcomp thank you for the contribution but unfortunately we cannot merge this.

As you'll see from the pinned issue #1813 we cannot support symfony 8 with the 3.x series because adding the return types creates BC breaks.

We have been working solidly towards a Behat 4.0 release which will support symfony 8 (and can, experimentally, already be used). See #1743 for more details of our plans. 

We are now getting very close to an alpha-1 release so that extension authors can start to add support for the 4.0 series.

PR #1842, opened 2026-05-17
feat: Support GHERKIN_32 parsing mode in pretty formatter

### LINKED ISSUE #1646: Pretty formatter should support `gherkin-32` compatibility mode

acoulton:
Ensures our pretty formatter can properly / consistently render features that were parsed in the newer `GHERKIN_32` compatibility mode.

Two changes in parser behaviour are particularly relevant.

#### Separate `title` and `description` properties

Anything that can take a description (scenarios, outlines, examples tables etc) is parsed with a one-line `title` and possibly a multi-line `description` as separate properties. 

In `LEGACY` parser mode, the entire text appears in `title` (which may or may not contain linebreaks).

#### Whitespace / indentation not trimmed inside the `description`

Only the start and end of the `description` are trimmed - `LEGACY` mode also attempted to left-trim each line to match the indentation of the keyword.

With a feature file like:

```
    Scenario: Do something
      Because I would like
      to prove that it works.
```

* `GHERKIN_32` mode gives `{"title": "Do something", "description": "Because I would like\n    to prove that it works"}`.
* `LEGACY` mode gives `{"title": "Do something\n  Because I would like\n  to prove that it works"}`.

---

Currently, the pretty formatter **does not** guarantee to render lines with the same indentation as the feature file. Instead it attempts to render with a "normalised" indentation style.  This is because `behat/gherkin` nodes do not contain any information about their original indentation.

In theory the indentation for different types of line can be configured by constructor arguments to e.g. `PrettyFeaturePrinter` and `PrettyScenarioPrinter` - however we don't expose these in our service container configuration.

In practice, the current output of the pretty formatter sometimes matches some feature files - but there are a lot of edge cases and variations if the feature file is indented in unusual ways.

We do not generally expect end-users to be relying on the exact output of the pretty formatter - it should be safe to assume that most build tooling would use one of the machine-readable formats.

However, our own feature suite extensively asserts the output of the pretty formatter (either because we're asserting what the formatter does, or because it is the most readable way to show expected behaviour).

Therefore:
* It seems more important that the output of the pretty formatter for "normally formatted" feature files is consistent regardless of which `GherkinCompatibilityMode` is configured. This avoids us having to modify feature files where the specified behaviour hasn't actually changed.
* Ideally the output of the pretty formatter for "normally formatted" feature files will match previous Behat versions.
* It seems acceptable to slightly change whitespace in pretty output for "unusually formatted" feature files - both between parser modes and/or between Behat versions.

This PR modifies the formatter on that basis, and also normalises the formatting of a couple of our feature files where the indentation isn't relevant to the behaviour that's being specified.

Fixes #1646

### DIFF
--- a/.roave-backward-compatibility-check.xml
+++ b/.roave-backward-compatibility-check.xml
@@ -14,5 +14,11 @@
         <ignored-regex>#CHANGED: The return type of Behat\\Testwork\\Environment\\Call\\EnvironmentCall\#getEnvironment\(\) changed#</ignored-regex>
         <ignored-regex>#CHANGED: The return type of Behat\\Testwork\\Environment\\Call\\EnvironmentCall\#getCallee\(\) changed#</ignored-regex>
         <ignored-regex>#CHANGED: The return type of Behat\\Testwork\\Environment\\Call\\EnvironmentCall\#getArguments\(\) changed#</ignored-regex>
+
+        <!-- The default indentation has changed to produce the same output in most cases as previous versions.
+             We officially don't support configuring this, if end-users are creating an instance with a different
+             indentation then output may change slightly but we don't guarantee human-readable output will be identical
+             between versions anyway. -->
+        <ignored-regex>#CHANGED: Default parameter value for parameter \$subIndentation of Behat\\Behat\\Output\\Node\\Printer\\Pretty\\PrettyScenarioPrinter\#__construct\(\) changed from 2 to 4#</ignored-regex>
     </baseline>
 </roave-bc-check>
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -467,7 +467,12 @@ EOL;
         }
 
         throw new UnexpectedValueException(
-            'Expected previous command to ' . strtoupper($success) . ' but got exit code ' . $this->getExitCode()
+            sprintf(
+                "Expected previous command to %s but got exit code %d with output:\n%s",
+                strtoupper($success),
+                $this->getExitCode(),
+                $this->getOutput()
+            )
         );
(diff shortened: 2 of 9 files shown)


composer.json
```diff
@@ -88,5 +88,24 @@
...
+    "repositories": [{
+        "type": "package",
+        "package": {
+            "name":         "behat/gherkin",
+            "version":      "dev-feat-full-step-text",
+            "dist": {
+                "type": "zip",
+                "url": "https://api.github.com/repos/acoulton/Gherkin/zipball/3eaf20b2e31c2cb8f0a4f13b8be8d04c3314a078",
+                "reference": "3eaf20b2e31c2cb8f0a4f13b8be8d04c3314a078",
+                "shasum": ""
```
acoulton: This will be reverted and rebased out once we have a new Behat/Gherkin release with this feature

stof (approved):

acoulton:
@stof thanks for reviewing before I'd even requested :laughing: - I've updated the PR description but nothing changed in the branch since you saw it.

PR #1843, opened 2026-05-17
test: Prove JUnit scenario description handling in gherkin compat modes

acoulton:
Historically, the `name` attribute in the JUnit file has been the full Scenario "multi-line title" e.g. the name and the description.

In GherkinCompatibilityMode::LEGACY we will keep this behaviour, to avoid any BC break for external systems which may depend on the testcase name being the same over time.

In GherkinCompatibilityMode::GHERKIN_32, the attribute **only** contains the Scenario title e.g. the first line of the scenario. This is what's already happening because the description isn't included in `getTitle` any more.

In most cases, JUnit output is used for summary output on CI systems and for tracking pass/fail trends over time. IMO these will usually benefit from having a short and fairly stable testcase name, so excluding the description here is probably an improvement.

If required we could in future add an option to the JUnit formatter to control whether the description should be included.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         "php": ">=8.2 <8.6",
         "ext-mbstring": "*",
         "composer-runtime-api": "^2.2",
-        "behat/gherkin": "^4.15.0",
+        "behat/gherkin": "^4.16.0",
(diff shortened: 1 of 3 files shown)


carlos-granados (approved):
Ok, good checks. thanks @acoulton

stof (approved):

PR #1845, opened 2026-05-18
fix: Only print ` ` between Step keyword & text if correct for language

acoulton:
Historically, our formatters always printed a space after the keyword of
a step (e.g. `Given I have 20 apples`). This is correct for the majority
of languages, but not all.

For example:
* Languages like Japanese never have spaces after keywords
* Languages like French may not have a space depending on the keyword
  e.g. `Et qu'` when the step begins with a vowel.

The gherkin parser understood this distinction, but this was not
previously exposed to us.

This commit uses the new `$step->getFullText()` everywhere that we print
the keyword and text, rather than concatenating the separate keyword and
text properties.

This will slightly change the formatter output for steps in languages
with no separating space - but to what it should always have been.

This will be consistent across all Gherkin parsing modes.

### DIFF
--- a/src/Behat/Behat/Output/Node/Printer/Pretty/PrettySkippedStepPrinter.php
+++ b/src/Behat/Behat/Output/Node/Printer/Pretty/PrettySkippedStepPrinter.php
@@ -73,9 +73,11 @@ final class PrettySkippedStepPrinter implements StepPrinter
 
         if ($result instanceof DefinedStepResult && $result->getStepDefinition()) {
             $definition = $result->getStepDefinition();
-            // We only paint the step text, but we render it as part of the full text (e.g. with the keyword prefix)
-            // The step text is always at the end, so use preg_replace to guarantee that we never modify the keyword
-            // even the edge case that it contains the step text (e.g. in languages with logograms).
+            // We render the full text, but should only paint the step text - the keyword should not be modified.
+            // It's theoretically possible (e.g. in languages with logograms) that the keyword contains the step text
+            // so we can't just str_replace.
+            // Instead, use a substring to extract the prefix by excluding the last {length of step text} chars.
+            // We can then prepend that to the painted step text to get the final value.
             $stepText = $this->textPainter->paintText(
                 $step->getText(),
                 $definition,
--- a/src/Behat/Behat/Output/Node/Printer/Pretty/PrettySkippedStepPrinter.php
+++ b/src/Behat/Behat/Output/Node/Printer/Pretty/PrettySkippedStepPrinter.php
@@ -25,6 +25,8 @@ use Behat\Testwork\Output\Printer\OutputPrinter;
 use Behat\Testwork\Tester\Result\IntegerTestResult;
 use Behat\Testwork\Tester\Result\TestResult;
 
+use function strlen;
+
 /**
  * Prints steps as skipped.
  *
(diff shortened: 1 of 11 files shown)


src/Behat/Behat/Output/Node/Printer/Pretty/PrettySkippedStepPrinter.php
```diff
@@ -57,31 +57,34 @@ public function __construct(
...
+            // We only paint the step text, but we render it as part of the full text (e.g. with the keyword prefix)
+            // The step text is always at the end, so use preg_replace to guarantee that we never modify the keyword
+            // even the edge case that it contains the step text (e.g. in languages with logograms).
             $stepText = $this->textPainter->paintText(
-                $stepText,
+                $step->getText(),
                 $definition,
                 new IntegerTestResult(TestResult::SKIPPED)
             );
+            $fullStepText = preg_replace('/'.preg_quote($step->getText(), '/').'$/', $stepText, $step->getFullText());
```
stof: why using preg_replace instead of just taking a substring ? `$prefix = substr($step->getFullText(), 0, -\strlen($step->getText())`

acoulton: No particular reason, using the one-line regex to replace the end of the string was just what felt natural as I was working I guess. Have updated to use substring instead.

src/Behat/Behat/Output/Node/Printer/Pretty/PrettySkippedStepPrinter.php
```diff
@@ -57,31 +59,35 @@ public function __construct(
...
      */
-    private function printText(OutputPrinter $printer, $stepType, $stepText, StepResult $result): void
+    private function printText(OutputPrinter $printer, StepNode $step, StepResult $result): void
     {
         $style = $this->resultConverter->convertResultCodeToString(TestResult::SKIPPED);
 
         if ($result instanceof DefinedStepResult && $result->getStepDefinition()) {
             $definition = $result->getStepDefinition();
+            // We only paint the step text, but we render it as part of the full text (e.g. with the keyword prefix)
+            // The step text is always at the end, so use preg_replace to guarantee that we never modify the keyword
```
carlos-granados: Guess you need to change this comment now that you have changed the implementation

src/Behat/Behat/Output/Node/Printer/Pretty/PrettyStepPrinter.php
```diff
@@ -83,19 +85,23 @@ public function printStep(Formatter $formatter, Scenario $scenario, StepNode $st
...
-     * @param string        $stepText
      */
-    private function printText(OutputPrinter $printer, $stepType, $stepText, StepResult $result): void
+    private function printText(OutputPrinter $printer, StepNode $step, StepResult $result): void
     {
         if ($result instanceof DefinedStepResult && $result->getStepDefinition()) {
             $definition = $result->getStepDefinition();
-            $stepText = $this->textPainter->paintText($stepText, $definition, $result);
+            // We only paint the step text, but we render it as part of the full text (e.g. with the keyword prefix)
+            // The step text is always at the end, so use preg_replace to guarantee that we never modify the keyword
```
carlos-granados: Same here

carlos-granados (approved):
Looks good, just a couple of small comments

src/Behat/Behat/Output/Node/Printer/Pretty/PrettySkippedStepPrinter.php
acoulton: Good catch, thanks @carlos-granados

carlos-granados (approved):

PR #1847, opened 2026-05-20
fix: JSON formatter should support `GHERKIN_32` mode

### LINKED ISSUE #1844: JSON formatter tests fail in GHERKIN_32 parsing mode
Our specifications for the JSON format fail on [this scenario](https://github.com/Behat/Behat/blob/3.x/features/json_format.feature#L232) if running in `GHERKIN_32` parser compatibility mode:

This is because the JSON formatter renders the `name` property from `$scenario->getTitle()` and this no longer includes the description (which is in the separate `$scenario->getDescription()` in this parsing mode). It's possible there are similar issues with features & other types of nodes.

Unlike JUnit, there are probably cases where the description is useful for end-users in JSON format.

Best solution is probably to add `description` as a separate property in the JSON output. That way:

* If the file was parsed in `GherkinCompatibilityMode::LEGACY` then users will still get the full string in `name` for BC
* If the file was parsed in `GherkinCompatibilityMode::GHERKIN_32` then users will get separate `name` and `description` values to make it easier to use /  format them to suit their usecase.

acoulton:
As with other formatters, the main difference is in the way scenario names and descriptions are parsed.

Unlike the junit formatter, there may be a broader range of usecases for the JSON output (and we control our own schema). Therefore it makes sense to publish the `description` when we have one, and to keep it in a separate property rather than merging into the name. That gives end-users the most flexibility to decide how to present / use it.

I have made `description` an optional property in the schema:

* In LEGACY parsing mode, the JSON output is identical to previous Behat versions ("multiline title" merged together into a one-line "name").
* In GHERKIN_32 parsing mode, "name" is the title / first line as expected, and ""description" will only be present if it has content.

When rendering a GHERKIN_32 description, I think it will be more consistent / useful if we remove any indentation on the second & subsequent lines. That makes it easy for the user to add consistent indentation to their output if required to suit their usecase.

Fixes #1844

### DIFF
--- a/src/Behat/Behat/Output/Node/Printer/JSON/JSONScenarioPrinter.php
+++ b/src/Behat/Behat/Output/Node/Printer/JSON/JSONScenarioPrinter.php
@@ -13,6 +13,7 @@ namespace Behat\Behat\Output\Node\Printer\JSON;
 use Behat\Behat\Output\Node\EventListener\JSON\JSONDurationListener;
 use Behat\Behat\Output\Node\Printer\Helper\ResultToStringConverter;
 use Behat\Behat\Output\Node\Printer\ScenarioPrinter;
+use Behat\Gherkin\Node\DescribableNodeInterface;
 use Behat\Gherkin\Node\FeatureNode;
 use Behat\Gherkin\Node\NamedScenarioInterface;
 use Behat\Gherkin\Node\ScenarioLikeInterface;
@@ -50,11 +51,14 @@ final class JSONScenarioPrinter implements ScenarioPrinter
     {
         $scenario = $this->currentScenario;
         assert($scenario instanceof NamedScenarioInterface);
-        $name = implode(' ', array_map(fn ($l): string => trim((string) $l), explode("\n", $scenario->getName() ?? '')));
+        ['name' => $name, 'description' => $description] = $this->getNameAndDescription($scenario);
 
         $scenarioAttributes = [
             'name' => $name,
         ];
+        if ($description !== null) {
+            $scenarioAttributes['description'] = $description;
+        }
 
         if ($formatter->getParameter('timer')) {
             $scenarioAttributes['time'] = (float) $this->durationListener->getDuration($scenario);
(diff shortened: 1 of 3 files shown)


src/Behat/Behat/Output/Node/Printer/JSON/JSONScenarioPrinter.php
```diff
@@ -50,11 +51,14 @@ public function printFooter(Formatter $formatter, TestResult $result): void
     {
         $scenario = $this->currentScenario;
         assert($scenario instanceof NamedScenarioInterface);
-        $name = implode(' ', array_map(fn ($l): string => trim((string) $l), explode("\n", $scenario->getName() ?? '')));
+        ['name' => $name, 'description' => $description] = $this->getNameAndDescription($scenario);
```
stof, line 54: We could simplify the code a lot as the handling of the name and description is totally independent in this formatter

acoulton: I'm not sure if that's true.

The value of `name` depends on whether the scenario has a description (because if there's no description, then we run `name` through the code to split it into lines, trim it, and join it into a single string).

We could process name and description separately, but that would mean having two places that were branching based on trying to identify the content of the node and how it has been parsed.

IMO it's much clearer to keep that together, so you can see how both properties vary based on one set of rules about the input...

stof: We run name through the splitting only if it contains `\n`, which will never be the case in gherkin-32 mode anyway.

acoulton: I guess in practice that's true... But IMO there is still a logical relationship between `name` and `description` that follows the gherkin mode.

It's extremely unlikely, but a future cucumber/gherkin could introduce syntax for "a scenario title that is very long and wraps to two lines that may or may not be followed by a description". At which point `name` could indeed contain `\n`.

In my head, the defining thing that changes the behaviour of the formatter isn't whether the name contains newlines. It's that the feature comes from a parser that has the concept of a scenario having a description as a distinct separate property.

So to me it makes much more sense to keep that logic together and commented that way - I don't feel the function I've written is particularly complex?

stof: > It's extremely unlikely, but a future cucumber/gherkin could introduce syntax for "a scenario title that is very long and wraps to two lines that may or may not be followed by a description". At which point `name` could indeed contain `\n`.

and your logic would *also* break in that case, as you don't change behavior based on the parsing mode but based on the presence of a description (so having a multiline title in this potential new mode would be treated differently depending on the presence of a description).

acoulton: Very fair point. Though arguably might be a smaller/easier fix to support the imaginary new syntax if we already have the method coping with the other variants.

To be perfectly honest I don't think my implementation is complex/problematic enough to justify the time either of us have spent thinking about & discussing it. It's proved correct by tests and commented to describe the different states.

If you disagree I'll happily find some time to rewrite it tomorrow (instead of doing something else) but I'm not really seeing a blocker here.

stof (approved):

Issue #1772, opened 2025-12-10
Show the suite for a failing scenario, if there are multiple suites


acoulton:
Users may have multiple suites that share some or all of the same feature files (but for example with different step definitions / a different environment e.g. some sort of pluggable backend driver).

Currently, if we run both suites in the same execution, the output only shows details of the feature / scenario / step. There's nothing to show which suite we were executing at the time. 

See for example the expected output in [suite.feature](https://github.com/Behat/Behat/blob/3.x/features/suite.feature) and particularly the failure output in [rerun_with_multiple_suite.feature](https://github.com/Behat/Behat/blob/3.x/features/rerun_with_multiple_suite.feature).

It would be much easier to understand / debug what has actually failed if the output included details of the suite that the scenario was executed in.

We probably don't want to add this noise in cases where there is only one suite executing. 

It may however be reasonable to add it any time there are multiple suites (rather than trying to work out at runtime whether a specific scenario is included in more than one suite).

stof:
I think that including this info as soon as there are multiple suites is a reasonable behavior.

Amoifr:
I'd like to pick this up — thanks @stof for confirming the direction. Before coding, I'd like to align on the approach and the exact output, since a few things are open.

**Where the ambiguity shows up.** The failure summaries print a scenario/step path with no suite, e.g. in the progress format's `--- Failed steps:` list (from `ListPrinter::printStat`, fed by `StepStatV2`) and in the `--- Failed scenarios:` list (`ListPrinter::printScenariosList`, fed by `ScenarioStat`). When suites share feature files, the same `features/apples.feature:9` can come from two suites, so both lists are affected (e.g. `features/rerun_with_multiple_suite.feature`).

**Proposed approach (self-contained, no `Statistics` interface change):**
- carry the suite name on `StepStatV2` and `ScenarioStat` (the `After*Tested` events expose it via `getSuite()->getName()`, captured in the existing stat listeners);
- in `ListPrinter`, render the suite **only when the displayed failures span more than one suite**, so single-suite runs stay noise-free.

**Two questions:**
1. **Detection** — is "the failures span >1 suite" good enough, or do you specifically want "more than one suite was *executed*" (which would need tracking executed suites in `Statistics`, a small interface addition)?
2. **Format** — preference for the annotation? e.g.
   - `... # features/apples.feature:9 [suite: default]`, or
   - `... # [default] features/apples.feature:9`

[...]

PR #1848, opened 2026-05-20
test: Run features in `gherkin-32` and `legacy` compatibility modes

acoulton:
Prove that our gherkin-32 mode is now compatible with legacy mode by running all scenarios in both `gherkin-32` **and** `legacy` mode unless they specify otherwise.

The most reliable way I could find to do this without interfering with the CLI options or environment variables set by other steps is to merge the default into the environment variables immediately before we run Behat.

If any scenarios are specifying the mode in their config file, that will automatically take precedence over the environment variable.

We now run the features in three suites:
* `explicit-gherkin-mode` runs only those scenarios tagged `@gherkin-mode:has-explicit`, they only run once and we do not force a default gherkin mode.
* `gherkin-legacy` and `gherkin-32` run all other scenarios, forcing the default gherkin mode accordingly. These scenarios therefore run twice and are expected to behave the same in both cases.

### DIFF
--- a/features/gherkin_compatibility_mode.feature
+++ b/features/gherkin_compatibility_mode.feature
@@ -10,6 +10,7 @@ Feature: Gherkin compatibility mode
       | --no-colors |          |
       | --format    | progress |
 
+  @gherkin-mode:has-explicit
   Scenario: Legacy parsing by default (ignores invalid language)
     Given the test runner does not configure a default gherkin compatibility mode
     When I run behat with the following additional options:
@@ -23,6 +24,7 @@ Feature: Gherkin compatibility mode
       3 steps (3 undefined)
       """
 
+  @gherkin-mode:has-explicit
   Scenario: Opt in to cucumber-compatible parsing
     When I run behat with the following additional options:
       | option    | value      |
--- a/features/gherkin_compatibility_mode.feature
+++ b/features/gherkin_compatibility_mode.feature
@@ -11,6 +11,7 @@ Feature: Gherkin compatibility mode
       | --format    | progress |
 
   Scenario: Legacy parsing by default (ignores invalid language)
+    Given the test runner does not configure a default gherkin compatibility mode
     When I run behat with the following additional options:
       | option    | value   |
       | --profile | default |
--- a/behat.dist.php
+++ b/behat.dist.php
@@ -1,17 +1,41 @@
 <?php
 
 use Behat\Config\Config;
+use Behat\Config\Filter\TagFilter;
 use Behat\Config\GherkinOptions;
 use Behat\Config\Profile;
+use Behat\Config\Suite;
 use Behat\Config\TesterOptions;
 use Behat\Gherkin\GherkinCompatibilityMode;
 
+/*
+ * Scenarios that specify a gherkin mode in their own config (or specify that they should use Behat's
+ * own default) should be tagged '@gherkin-mode:has-explicit'. They will only run in this suite.
+ */
+$explicitGherkinSuite = (new Suite('explicit-gherkin-mode'))
+    ->addContext(FeatureContext::class, ['gherkinCompatibilityMode' => null])
+    ->withFilter(new TagFilter('@gherkin-mode:has-explicit'));
+
+/*
+ * Scenarios that do not have the `@gherkin-mode:has-explict` tag will run twice - once with the default forced to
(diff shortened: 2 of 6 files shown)


acoulton:
This build will fail because this branch does not include the fixes for the JSON formatter tests in gherkin-32 mode (which proves it has successfully now forced that scenario to run in gherkin-32 mode).

The branch should be green once we merge #1847 and merge it into this branch.

stof:
Should we instead run all our scenario twice (except when specific to the legacy format) ? We could have 2 suites configured that both include all our features (using a `~@legacy-mode` filter to exclude features targeting only the legacy mode in the suite running in modern mode), with suite settings changing this default mode in the context.

This way, we keep coverage for both modes in Behat 3.x.

Our suite runs fast enough that running 2 suites is not an issue.

acoulton:
> Should we instead run all our scenario twice (except when specific to the legacy format) ? We could have 2 suites configured that both include all our features (using a `~@legacy-mode` filter to exclude features targeting only the legacy mode in the suite running in modern mode), with suite settings changing this default mode in the context.
> 
> This way, we keep coverage for both modes in Behat 3.x.
> 
> Our suite runs fast enough that running 2 suites is not an issue.

That's a reasonable idea. Might need a little thought about the suites / tagging / filtering because in reality we have:

* Some scenarios that explicitly specify a gherkin mode (e.g. in their `behat.php` profile) plus the one that explicitly specifies not to override Behat's own default. Our runtime default mode is irrelevant to them, and there's no benefit to running them more than once.
* Other scenarios that should pass on both `legacy` and `gherkin-32` mode.

For the ones that specify a mode in their own config, I'd be keen to keep that there. That way they're guaranteed to run in the expected mode even if you're running the fixture directly from the CLI (e.g. to check it / iterate on it).

So maybe there are actually three suites:

* One that runs everything tagged `@gherkin-mode:has-explicit`, with no default mode set (e.g. they'll use whatever Behat's own default is).
* One that runs everything NOT tagged `@gherkin-mode:has-explicit` in `gherkin-32` mode.
* One that runs everything NOT tagged `@gherkin-mode:has-explicit` in `legacy` mode.

Worst case scenario, if we forget to add a `@gherkin-mode:has-explicit` to a scenario then unless it depends on the default it will just run a second (redundant) time.

WDYT?

stof:
looks good to me

acoulton:
@stof done - as expected it still fails on the json format tests until #1847 is resolved.

This also demonstrates #1772 - the failure output just shows that `features/json_format.feature:23` failed without saying in which suite. You can only really tell by looking at how far the failure comes in the progress output and guessing based on the order the suites run in. I think it's OK for us for now, we can take a look at that issue later.

features/gherkin_compatibility_mode.feature
```diff
@@ -10,7 +10,9 @@ Feature: Gherkin compatibility mode
       | --no-colors |          |
       | --format    | progress |
 
+  @gherkin-mode:has-explicit
   Scenario: Legacy parsing by default (ignores invalid language)
+    Given the test runner does not configure a default gherkin compatibility mode
```
stof, line 15: is this necessary ? `@gherkin-mode:has-explicit` does not configure a default mode in the runner.

acoulton: Kind-of-not, except that:

* I don't think it's bad to be explicit in the scenario that this is the expected behaviour when we don't force a mode, rather than relying on implicitly understanding what that tag means.
* It adds a level of future-proofing / protection if someone runs that scenario other than with our suite configuration (e.g. if someone is changing the way our suites work in future / has their own `behat.php` overriding our `behat.dist.php`.

It's kind-of a no-op in the same way as the scenarios that specify they run in `legacy` mode when that's the default - but I think it helps if the scenario is self-explaining when the mode is directly relevant to the behaviour we're testing?

stof (approved):

acoulton:
Updated based on #1847 and as expected it's now green, I have just added the gherkin-mode tags to the new json_format scenarios that were added in that PR.

acoulton:
@carlos-granados actually this one ideally also needs to be merged to 3.x before the next merge-up if you have a chance to review it.

carlos-granados (approved):
I think that all standing issues were fixed after @stof' s comments, so let's go ahead and merge this

Issue #1846, opened 2026-05-20
Symfony YAML collection alias limit


phil-davis:
https://github.com/symfony/yaml/releases/tag/v7.4.12 was just released.

It adds a limit to the number of collection aliases allowed when parsing YAML. The default is 128.

I happen to have a https://github.com/owncloud/core/blob/master/tests/acceptance/config/behat.yml that has 2 different collection aliases that each are used 106 times for a total of 212. See https://github.com/symfony/symfony/issues/64308 for details of how it fails.

I think that there will need to be some way for users to override the limit. Maybe providing some parameter or env var when running Behat. And then Behat can call the Parser constructor specifying the desired limit.

I'm happy to provide a PR, if someone confirms that the ability to increase this limit is needed, and gives a preferred way for user to specify it (Behat command line parameter, env var or?)

acoulton:
@phil-davis thanks for raising this and offering a PR.

That's an impressive `behat.yml`! I'm not sure how widely this will be an issue for others, and I wonder if there'd be a way to restructure your yaml to workaround it?

I'd also highlight that YAML config is now deprecated and will not be suppported in Behat 4.0. You might want to look at migrating to PHP config - that's possibly a more obvious way to reuse config values anyway, and it would avoid this altogether.

The[ `--convert-config` command](https://docs.behat.org/en/v3.x/user_guide/configuration/yaml_configuration.html#converting-your-configuration) should do a reasonable job of building a PHP config file from your YAML, though it won't know about aliases etc so I expect you'll have to refactor out some duplication from the version it generates.

If you wanted to contribute a PR to allow customising the YAML options in the 3.x series then I'd be OK with that - I'd say:

* It should definitely be a CLI argument rather than an env var, so that it's visible / discoverable
* The design of that CLI argument needs a little thought - we probably don't want standalone CLI args for every aspect of yaml parsing, so it might need to be a single `--yaml-config-options='{"some":"json"}'` or something?
* It should be feasible to add that option in `Behat\Testwork\Cli\Application` and pass it through to `Behat\Testwork\ServiceContainer\Configuration\ConfigurationLoader` in a backwards compatible way (probably with a new setter on that class?)

phil-davis:
> That's an impressive behat.yml

Yeh, it "just grew" as we made more test scenarios, and put them into test suites that each run for up to about 30 minutes. If you find somewhere that has enough runners, you can run about 40 hours of test scenarios in 30 minutes!

[...]

PR #1849, opened 2026-05-21
fix: Handle empty argument lists when converting Context config to PHP

### LINKED ISSUE #1851: Bug @ vendor/behat/behat/src/Behat/Config/Suite.php
Hi there,
there comes an issue with your config converter "--convert-config". 
As your converter changed our **baht.yml** into **behat.php**, it defines the _$constructorArgs_ in calling method addContext as _null_ . 
when our classes do not have any constructor args. T
But your function Suite::addContext does not support nullable values

`public function addContext(string $context, array $constructorArgs = []): self`

this leads to 

`PHP Fatal error:  Uncaught TypeError: Behat\Config\Suite::addContext(): Argument #2 ($constructorArgs) must be of type array, null given, called in src/test/behat/behat.php on line 137 and defined in behat/behat/src/Behat/Config/Suite.php:51`

Could you please fix it? 

cheers, björn

acoulton:
There are (valid) edge cases in YAML config where a Context definition can be parsed as an array even though it has no arguments. For example, if the Context class name has a trailing `:` e.g.

```yaml
# suite etc config
   contexts:
     - MyContext:
```

The config conversion was incorrectly treating these as having arguments, and assuming the structure would always be `[$contextClass => [$arg1,...]]`.

In this case we actually get `[$contextClass => null]` meaning the converter generates a statement like `->addContext('MyContext', null)` which causes a `TypeError` because that method does not accept `null`.

Fix this by normalising contexts with no arguments back to a simple string before the rest of the processing.

Fixes an issue reported by @phil-davis in the comments of #1846 

Fixes #1851

### DIFF
--- a/src/Behat/Config/Suite.php
+++ b/src/Behat/Config/Suite.php
@@ -114,6 +114,8 @@ final class Suite implements ConfigConverterInterface
         unset($this->settings[self::CONTEXTS_SETTING]);
 
         // Contexts can be configured with or without constructor arguments.
+        // Contexts with no arguments can be represented in different ways, so first normalise them.
+        $contexts = array_map($this->normaliseContextWithNoArgs(...), $contexts);
         $hasAnyWithArguments = (bool) array_filter($contexts, is_array(...));
 
         if (!$hasAnyWithArguments) {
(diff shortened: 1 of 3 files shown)


phil-davis (commented):
LGTM, the test case changes confirm both possible inputs.
Thanks

carlos-granados (approved):
Good find, thanks @acoulton and @phil-davis

PR #1853, opened 2026-06-19
fix: Unexpected uses / creation of `RuntimeCallee` callables

acoulton:
RuntimeCallee::throwIfInstanceMethod is used by our Feature and Suite hook classes to check that the hook is a static method.

It accepts a `callable` argument, but did not actually check the type of that callable. Instead, it checks the type of the `$this->callable` (through `$this->reflection`). The callable passed into the method is only used for rendering the exception. So in theory the method could false-positive and/or fail with an exception message that refers to the wrong callable.

I have fixed this, but also deprecated that usage - I don't think there's any need to support arbitrary callables in this method, and it is unlikely end-users are calling it or extending it.

Also deprecates creating a `RuntimeCallee` with a non-callable array that is not a Behat `Context` class method reference. 
The codebase assumes that a non-callable array will be an array of `[$contextClass, $contextMethod]`, which will later be resolved by `InitializedContextEnvironment`. However, this is never checked.

In practice, creating with other types of array would likely cause a runtime error somewhere in the chain, but we should deprecate it closer to source ahead of tightening the type-safety in 4.0.

Works towards #1519

### DIFF
--- a/src/Behat/Testwork/Call/RuntimeCallee.php
+++ b/src/Behat/Testwork/Call/RuntimeCallee.php
@@ -41,6 +41,9 @@ class RuntimeCallee implements Callee
         private readonly ?string $description = null,
     ) {
         if (is_array($callable)) {
+            if (!$this->isCallableOrBehatContextCallable($callable)) {
+                DeprecationCollector::trigger('Creating '.static::class.' with a non-callable array other than a Behat context method reference is deprecated');
+            }
             $this->reflection = new ReflectionMethod($callable[0], $callable[1]);
             $this->path = $callable[0] . '::' . $callable[1] . '()';
         } else {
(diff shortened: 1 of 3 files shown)


src/Behat/Testwork/Call/RuntimeCallee.php
```diff
@@ -112,26 +128,26 @@ public function isAnInstanceMethod()
     }
 
     /**
-     * @param callable|array{class-string, string} $callable
+     * @param callable|array{class-string, string}|string $callableOrHookType
      */
-    protected function throwIfInstanceMethod(callable|array $callable, string $hookType): void
+    protected function throwIfInstanceMethod(callable|array|string $callableOrHookType, ?string $hookType = null): void
```
acoulton: Urgh, I wasn't quite switched on this morning and forgot that widening these param types is a BC break if anyone is extending the method.

I'll think if I can solve this easily in a BC way, otherwise maybe we just skip the deprecation and change it in 4.x.

carlos-granados (approved):
Looking good, @acoulton

PR #1861, opened 2026-06-27
feat: Add 'inline_failures' option to progress formatter

### LINKED ISSUE #1860: Progress formatter: show failures inline instead of only at the end
We run some big Behat suites in CI - the slowest ones are half an hour plus. The progress formatter is the right tool for those runs, except for one thing: when a step fails you get an `F` in the stream and then nothing else until the whole run finishes and the summary prints. On a long suite that's a frustrating wait just to see what actually broke. And if the job dies before it gets there - a CI timeout, an out-of-memory, someone cancelling it - the summary never prints at all, so you're left staring at an `F` somewhere in a wall of dots with no idea why.

We've been working around this for years with a small extension, [drevops/behat-format-progress-fail](https://github.com/drevops/behat-format-progress-fail). It's the progress formatter with one change: it prints the failing step, its file and line, and the exception inline, the moment the step fails, then carries on with the dots.

```
..
--- FAIL ---
  Then I should have 3 apples # (features/apples.feature):11
    Failed asserting that 2 matches expected 3.
------------
......U.......
--- FAIL ---
  Then I should have 8 apples # (features/apples.feature):25
    Failed asserting that 7 matches expected 8.
------------
.....UU
```

It does the job, but it's a separate package we have to keep in step with every Behat release, and honestly it feels like something that belongs in core. The `show_output` part of it already landed upstream as `ShowOutputOption`, so inline failures are about the only thing left that the built-in progress formatter doesn't cover.

Before getting into how it'd be built, the real question is whether this is something you'd want in core at all - whether inline failures fit where you see the progress formatter going.

If it's a direction you're open to, I'm glad to do the work and put a proper PR together as the next step. My instinct is an option on the existing progress formatter - failures shown inline vs in the summary, defaulting to today's behaviour so it stays backwards compatible, with a `.feature` to cover it - though a separate `progress_fail` formatter would work too if you'd rather keep them apart. Whether the PR's good enough to merge is on me to get right; I just didn't want to build it if the feature itself isn't something you'd take.

And if it's not for you, no worries at all - we'll carry on maintaining the extension the way we do now.

If it helps, the relevant bits are `ProgressStepPrinter` (the per-step `F`) and `ProgressStatisticsPrinter`, which is where all the failure detail currently waits until the end of the run.

AlexSkrypnyk:
Closes #1860

## Summary

Adds an opt-in `inline_failures` option to the existing `progress` formatter that prints each failed, pending, or undefined step's detail (step text, `feature:line`, exception) inline as it happens, instead of only in the end-of-run summary. The option defaults to `false`, so all existing output is unchanged. This implements the format proposed in the issue body, using Behat-native `ExceptionPresenter` for verbosity-aware exception rendering and `ConfigurablePathPrinter` for path resolution (honouring `--print-absolute-paths` and editor URL settings).

## Changes

- **`ProgressFormatter`** - new `bool $inlineFailures = false` constructor parameter and `inline_failures` setting constant; defaults off for full backwards compatibility.
- **`Formatter` (YAML-to-PHP converter)** - added `inline_failures` → `inlineFailures` mapping so the option round-trips through `convert_config`.
- **`CurrentFeatureListener`** - new lightweight event listener that captures the current `FeatureNode` on `BeforeFeatureTested` and exposes it for the duration of that feature; required because undefined steps carry no call result from which to read the feature file.
- **`ProgressStepPrinter`** - new `shouldPrintInline()` / `printInlineProblem()` / `printException()` / `resolveStepPath()` methods; when the option is on and the result is failed/pending/undefined, the inline block is printed instead of the single-character marker (`F`/`P`/`U`); existing newline and 70-step line-wrap coordination is preserved.
- **`ProgressFormatterFactory`** - registers `CurrentFeatureListener` as a service, injects it (plus `ExceptionPresenter` and `ConfigurablePathPrinter`) into `ProgressStepPrinter`, and prepends it to the `ChainEventListener` so it fires before the root listener.
- **`features/progress_inline_failures.feature`** - new integration test covering: default (no inline output), `--format-settings` JSON activation, and PHP-config-profile activation.
- **`tests/Fixtures/PrintInlineFailures/`** - new fixture with a `behat.php` config (default + `inline` profiles), a `FeatureContext` with passing/failing/pending steps, and an `inline.feature` with three scenarios (passing-then-failing, pending, undefined).
- **`tests/Behat/Tests/Config/ProfileTest.php`** - added `inline_failures: false` to the expected defaults in two assertions that enumerate progress-formatter settings.

## Configuration

Enable via PHP config:

```php
new ProgressFormatter(inlineFailures: true)
```

Enable via YAML config:

```yaml
formatters:
  progress:
    inline_failures: true
```

Enable via CLI:

```
behat --format=progress --format-settings='{"inline_failures": true}'
```

The formatter name stays `progress` - this is not a new format name.

## Scope note

This PR implements the format from the **issue body**: an additive `inline_failures` option on the **existing** `progress` formatter. It deliberately does not implement the per-scenario layout floated in [acoulton's comment](https://github.com/Behat/Behat/issues/1860#issuecomment-4804492245) (a `feature:line` prefix before each scenario, one line per scenario), because that is a fundamentally different rendering model. That approach would require a **new formatter**: a scenario-aware renderer that prints a path prefix per scenario, emits one line per scenario, and replaces the 70-step line wrapping - a separate, larger piece of work. The two are complementary and can coexist: this option enriches the classic progress stream, while a future dedicated formatter could offer the per-scenario view as an opt-in choice. The reusable inline failure-block rendering introduced here would support that future formatter.

## Screenshots

N/A (CLI output formatter; example output is below).

## Before / After

**Before** (default, `inline_failures` not set or `false`):

```
.FPU

--- Failed steps:

001 Scenario: Passing then failing   # features/inline.feature:3
      And a failing step             # features/inline.feature:5
        step failed as supposed (RuntimeException)
...
```

**After** (`inline_failures: true`):

```
.
--- FAILED ---
    And a failing step # features/inline.feature:5
      step failed as supposed (RuntimeException)
------------
--- PENDING ---
    Given a pending step # features/inline.feature:8
      TODO: write pending definition
------------
--- UNDEFINED ---
    Given an undefined step # features/inline.feature:11
------------

--- Failed steps:

001 Scenario: Passing then failing   # features/inline.feature:3
      And a failing step             # features/inline.feature:5
        step failed as supposed (RuntimeException)
...
```

The end-of-run summary is still printed in both cases - the inline output is additive.

### DIFF
--- a/src/Behat/Behat/Output/Node/EventListener/AST/CurrentFeatureListener.php
+++ b/src/Behat/Behat/Output/Node/EventListener/AST/CurrentFeatureListener.php
@@ -10,8 +10,8 @@
 
 namespace Behat\Behat\Output\Node\EventListener\AST;
 
+use Behat\Behat\EventDispatcher\Event\AfterFeatureTested;
 use Behat\Behat\EventDispatcher\Event\BeforeFeatureTested;
-use Behat\Behat\EventDispatcher\Event\FeatureTested;
 use Behat\Gherkin\Node\FeatureNode;
 use Behat\Testwork\Event\Event;
 use Behat\Testwork\Output\Formatter;
@@ -34,7 +34,7 @@ final class CurrentFeatureListener implements EventListener
             $this->currentFeature = $event->getFeature();
         }
 
-        if ($eventName === FeatureTested::AFTER) {
+        if ($event instanceof AfterFeatureTested) {
             $this->currentFeature = null;
         }
     }
--- a/src/Behat/Behat/Output/Node/EventListener/AST/CurrentFeatureListener.php
+++ b/src/Behat/Behat/Output/Node/EventListener/AST/CurrentFeatureListener.php
@@ -1,0 +1,46 @@
+<?php
+
+/*
+ * This file is part of the Behat.
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Behat\Behat\Output\Node\EventListener\AST;
+
+use Behat\Behat\EventDispatcher\Event\BeforeFeatureTested;
+use Behat\Behat\EventDispatcher\Event\FeatureTested;
+use Behat\Gherkin\Node\FeatureNode;
+use Behat\Testwork\Event\Event;
+use Behat\Testwork\Output\Formatter;
+use Behat\Testwork\Output\Node\EventListener\EventListener;
+
+/**
+ * Keeps the feature currently being tested available for the whole feature run.
+ *
+ * Step results do not all carry a reference back to their feature - an undefined
+ * step, for example, has no call result to read it from - so the feature node is
+ * captured here once per feature and exposed for the duration of that feature.
+ */
+final class CurrentFeatureListener implements EventListener
+{
+    private ?FeatureNode $currentFeature = null;
+
+    public function listenEvent(Formatter $formatter, Event $event, $eventName): void
+    {
+        if ($event instanceof BeforeFeatureTested) {
+            $this->currentFeature = $event->getFeature();
+        }
+
+        if ($eventName === FeatureTested::AFTER) {
+            $this->currentFeature = null;
+        }
+    }
+
+    public function getCurrentFeature(): ?FeatureNode
+    {
+        return $this->currentFeature;
+    }
+}
--- a/features/progress_inline_failures.feature
+++ b/features/progress_inline_failures.feature
@@ -1,0 +1,55 @@
+Feature: Progress formatter inline failures
+  In order to diagnose problems during long progress-formatter runs
+  As a developer
+  I need the option to print failing, pending and undefined steps inline as they happen
+
+  Background:
+    Given I initialise the working directory from the "PrintInlineFailures" fixtures folder
+    And I provide the following options for all behat invocations:
+      | option      | value |
+      | --no-colors |       |
+
+  Scenario: Step problems are only listed in the summary by default
+    When I run "behat --format=progress features/inline.feature"
+    Then it should fail with:
+      """
+      .FPU
+
+      --- Failed steps:
+      """
+
+  Scenario: Problems are printed inline when enabled via format settings
+    When I run "behat --format=progress --format-settings='{\"inline_failures\": true}' features/inline.feature"
+    Then it should fail with:
+      """
+      .
+      --- FAILED ---
+          And a failing step # features/inline.feature:5
+            step failed as supposed (RuntimeException)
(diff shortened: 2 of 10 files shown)


src/Behat/Behat/Output/Node/EventListener/AST/CurrentFeatureListener.php
```diff
@@ -0,0 +1,46 @@
...
+{
+    private ?FeatureNode $currentFeature = null;
+
+    public function listenEvent(Formatter $formatter, Event $event, $eventName): void
+    {
+        if ($event instanceof BeforeFeatureTested) {
+            $this->currentFeature = $event->getFeature();
+        }
+
+        if ($eventName === FeatureTested::AFTER) {
```
acoulton: Shouldn't this be checking `instanceof AfterFeatureTested` ?

acoulton (changes requested):
@AlexSkrypnyk thanks for the contribution and sorry for the slow review.

One small adjustment I think.

I wasn't entirely keen on the CurrentFeatureListener, as it feels like it's a bit of a workaround that might introduce complexity for the future. I had a look into the calls we make and I see that everywhere we call `->printStep` we're calling it from a context that has an `AfterStepTested` event, and that event *does* carry the FeatureNode. 

I thought about introducing a new `AfterStepTestedPrinter` interface, which ProgressStepPrinter could implement in parallel with the existing. Then when calling we could check if the StepPrinter implements that interface, and call that instead. And then in future (e.g. 5.x) we could retire the StepPrinter interface.

However, that would mean changes in a few places, and would still be a slightly annoying interface as the AfterStepTested doesn't carry the ScenarioNode.

If you want to spend any time looking into that, you could. But my gut feel is we merge this and then once 4.0 is released we could have a proper look at the events we dispatch and the way those go to the formatters as I think there are quite a lot of places that we (and end users) are having to use workarounds to get information that should be available.

src/Behat/Behat/Output/Node/EventListener/AST/CurrentFeatureListener.php
AlexSkrypnyk: Good catch - done in 18322d4f. Switched to `instanceof AfterFeatureTested`.

It's behaviourally equivalent (`AfterFeatureTested` is the only event dispatched under `FeatureTested::AFTER`), but it's symmetric with the `BeforeFeatureTested` check just above it and matches how `FeatureListener` in the same namespace detects the same "feature finished" moment. Thanks!

acoulton (approved):
Great, thanks @AlexSkrypnyk

carlos-granados (approved):
Looking great, thanks @AlexSkrypnyk

PR #1863, opened 2026-08-18
feat: Add support for Cucumber tag expressions

### LINKED ISSUE #1642: Add support for the standard Cucumber tag expressions to filter by tags
Currently, the tag filters in the Suite configuration and the CLI argument are using a filtering syntax inspired by the old Cucumber versions (while not being exactly the same AFAICT, as we support `&&` which was not in it).

Cucumber has since introduced [Tag Expressions](https://github.com/cucumber/tag-expressions/tree/main), which allow more complex boolean expressions.

I suggest adding support for them in Behat.

We will probably have to use a new CLI flag than `--tags` and a new config setting as guessing whether we use the `behat/gherkin` tag filter syntax or a tag expression might be ambiguous.

Amoifr:
Fixes #1642

Adds support for [Cucumber Tag Expressions](https://github.com/cucumber/tag-expressions) through the official `cucumber/tag-expressions` PHP port (published from cucumber/tag-expressions#194).

As suggested in the issue, the existing `--tags` syntax is untouched and the expressions get their own entry points, so there is no ambiguity between the two syntaxes:

- a new `--tag-expression` CLI option (repeatable, like `--tags`);
- a new `tag_expression` filter type for the profile and suite `filters` settings;
- a new `Behat\Config\Filter\TagExpressionFilter` class for the PHP config API.

The filter matches the `TagFilter` semantics: scenarios inherit feature tags, and outline example tables are filtered individually by their own tags. Tags are normalised to their `@`-prefixed form before evaluation, so both parsing modes (legacy and gherkin-32) behave the same, which the new `features/tag_expression_filters.feature` exercises in both modes. An invalid expression fails with the parser's message, e.g. `Tag expression "@a and (" could not be parsed because of syntax error: Unmatched (.`

I can send a follow-up to the docs repository once the shape is validated.

### DIFF
--- a/features/tag_expression_filters.feature
+++ b/features/tag_expression_filters.feature
@@ -48,11 +48,66 @@ Feature: Tag expressions
     Then it should pass
     And the output should contain:
       """
+      Feature: Feature N2
+
+        Background:
+          Given Some normal step N21
+
+        @fast
+        Scenario:
+          Given Some fast step N24
+          And Some fast step N25
+      """
+    And the output should contain:
+      """
+      Feature: Feature N3
+
+        Background:
+          Given Some normal step N21
+
+        @normal
+        Scenario:
+          Given Some normal step N38
+
+        @fast
+        Scenario Outline:
+          Given Some fast step N<num>
+
+          Examples:
+            | num |
+            | 33  |
+            | 34  |
+
+        @normal @fast
+        Scenario Outline:
+          Given Some normal step N<num>
+          And Some fast step N37
+
+          Examples:
+            | num |
+            | 35  |
+            | 36  |
+      """
+    And the output should contain:
+      """
+      Feature: Feature N4
+
+        @normal
+        Scenario:
+          Given Some normal step N41
+          And Some fast step N42
+
+        @fast
+        Scenario:
+          Given Some slow step N43
+      """
+    And the output should contain:
+      """
       8 scenarios (8 passed)
       18 steps (18 passed)
       """
 
-  Scenario: Outline examples are filtered by the expression
+  Scenario: Outlines are filtered by the expression
     When I run "behat --tag-expression '@normal and @fast'"
     Then it should pass
     And the output should contain:
@@ -78,9 +133,77 @@ Feature: Tag expressions
       6 steps (6 passed)
       """
 
+  Scenario: Outline examples are filtered by the expression
+    When I initialise the working directory from the "TagExpressionExamples" fixtures folder
+    And I run "behat features/outline_examples.feature --tag-expression '@quick'"
+    Then it should pass
+    And the output should contain:
+      """
+      Feature: Feature N5
+
+        Scenario Outline:
+          Given Some normal step N<num>
+
+          Examples:
+            | num |
+            | 51  |
+            | 52  |
+      """
+    And the output should contain:
+      """
+      2 scenarios (2 passed)
+      2 steps (2 passed)
+      """
+
   Scenario: Tag expression suite filter in the config file
     When I run "behat --config=behat-with-tag-expression-filter.php"
     Then it should pass
+    And the output should contain:
+      """
+      Feature: Feature N2
+
+        Background:
+          Given Some normal step N21
+
+        @fast
+        Scenario:
+          Given Some fast step N24
+          And Some fast step N25
+      """
+    And the output should contain:
+      """
+      Feature: Feature N3
+
+        Background:
+          Given Some normal step N21
+
+        @fast
+        Scenario Outline:
+          Given Some fast step N<num>
+
+          Examples:
+            | num |
+            | 33  |
+            | 34  |
+
+        @normal @fast
+        Scenario Outline:
+          Given Some normal step N<num>
+          And Some fast step N37
+
+          Examples:
+            | num |
+            | 35  |
+            | 36  |
+      """
+    And the output should contain:
+      """
+      Feature: Feature N4
+
+        @fast
+        Scenario:
+          Given Some slow step N43
+      """
     And the output should contain:
       """
       6 scenarios (6 passed)
--- a/features/tag_expression_filters.feature
+++ b/features/tag_expression_filters.feature
@@ -1,0 +1,96 @@
+Feature: Tag expressions
+  In order to run only needed features
+  As a Behat user
+  I need Behat to support Cucumber tag expressions
+
+  Background:
+    Given I initialise the working directory from the "TagFilters" fixtures folder
+    And I provide the following options for all behat invocations:
+      | option            | value              |
+      | --no-colors       |                    |
+      | --format-settings | '{"paths": false}' |
+
+  Scenario: And expression
+    When I run "behat --tag-expression '@slow and @fast'"
+    Then it should pass
+    And the output should contain:
+      """
+      @slow
+      Feature: Feature N1
+
+        Background:
+          Given Some slow step N11
+
+        @fast
+        Scenario:
+          Given Some fast step N14
+      """
+    And the output should contain:
+      """
+      Feature: Feature N2
+
+        Background:
+          Given Some normal step N21
+
+        @slow @fast
+        Scenario:
+          Given Some slow step N22
+          And Some fast step N23
+      """
+    And the output should contain:
+      """
+      2 scenarios (2 passed)
+      5 steps (5 passed)
+      """
+
+  Scenario: Not expression
+    When I run "behat --tag-expression 'not @slow'"
+    Then it should pass
+    And the output should contain:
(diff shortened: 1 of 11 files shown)


features/tag_expression_filters.feature
```diff
@@ -0,0 +1,96 @@
...
+      """
+
+  Scenario: Not expression
+    When I run "behat --tag-expression 'not @slow'"
+    Then it should pass
+    And the output should contain:
+      """
+      8 scenarios (8 passed)
+      18 steps (18 passed)
+      """
```
acoulton, line 108: I think it would be clearer (and more robust for the future) if this feature was explicit about *which* scenarios ran. Otherwise it relies on knowing that the fixture contains 8 scenarios that don't have the `@slow` tag, and assuming it was those 8 scenarios.

features/tag_expression_filters.feature
```diff
@@ -0,0 +1,96 @@
...
+      14 steps (14 passed)
+      """
+
+  Scenario: Invalid tag expression
+    When I run "behat --tag-expression '@a and ('"
+    Then it should fail
+    And the output should contain:
+      """
+      Tag expression "@a and (" could not be parsed because of syntax error: Unmatched (.
+      """
```
acoulton: Minor thing, please add a newline at the EOF.

```suggestion
      """

```

features/tag_expression_filters.feature
```diff
@@ -0,0 +1,96 @@
...
+  Scenario: Not expression
+    When I run "behat --tag-expression 'not @slow'"
+    Then it should pass
+    And the output should contain:
+      """
+      8 scenarios (8 passed)
+      18 steps (18 passed)
+      """
+
+  Scenario: Outline examples are filtered by the expression
```
acoulton: I would describe this as being `Outlines are filtered by the expression`.

To me, `Outline examples are filtered` means the case where the Scenario has multiple `Examples:` tables, and only the `Examples:` tables with matching tags are included e.g. https://github.com/Behat/Gherkin/blob/master/tests/Filter/TagFilterTest.php#L289

It might be sensible to extend this feature (and the fixture if necessary) to prove that the table tags are also considered? I can see that's covered in the unit tests though, so it's not crucial.

features/tag_expression_filters.feature
```diff
@@ -0,0 +1,96 @@
...
+      2 scenarios (2 passed)
+      6 steps (6 passed)
+      """
+
+  Scenario: Tag expression suite filter in the config file
+    When I run "behat --config=behat-with-tag-expression-filter.php"
+    Then it should pass
+    And the output should contain:
+      """
+      6 scenarios (6 passed)
```
acoulton, line 209: Again here I'd prefer to be explicit about *which* scenarios ran

src/Behat/Behat/Gherkin/Filter/TagExpressionFilter.php
```diff
@@ -0,0 +1,98 @@
...
+            }
+
+            return false;
+        }
+
+        return $this->evaluateTags(array_merge($feature->getTags(), $scenario->getTags()));
+    }
+
+    /**
+     * @param array<array-key, string> $tags
```
acoulton: I *think* we can rely on the getTags() methods from Gherkin always returning list<string> so we could be strict here?

```suggestion
     * @param list<string> $tags
```

acoulton (changes requested):
@Amoifr thanks for a great contribution!

I have a few small comments, mostly just about the feature file - the implementation looks great :)

Amoifr:
Thanks for the careful read @acoulton! All five points addressed in 3e0efe16:

- both the `not @slow` and the config-file scenarios now assert the full pretty output of every feature that runs; combined with the totals, that pins down exactly which scenarios were selected;
- renamed the outline scenario to `Outlines are filtered by the expression`, you're right about the distinction;
- and took your suggestion: a new `Outline examples are filtered by the expression` scenario proves the per-`Examples`-table filtering end to end, using a small dedicated fixture (an outline with a `@quick` and a `@lazy` table) layered on top of the shared one and targeted by path, so none of the shared counts move;
- `list<string>` applied, PHPStan agrees since the call site merges two lists;
- EOF newline added.

The feature file now runs 12/12 in both gherkin compatibility modes.

features/tag_expression_filters.feature
Amoifr: Done in 3e0efe16: the scenario now asserts the full pretty output of Features N2, N3 and N4 before the totals, so the eight are named rather than counted.

features/tag_expression_filters.feature
Amoifr: Same treatment for the config-file scenario, the six selected scenarios are spelled out above the totals.

features/tag_expression_filters.feature
Amoifr: Added.

features/tag_expression_filters.feature
Amoifr: Renamed to `Outlines are filtered by the expression`, you were right about the distinction. I also took the extension you suggested: a separate `Outline examples are filtered by the expression` scenario covers the per-`Examples`-table case, on a small dedicated fixture (an outline with a `@quick` and a `@lazy` table) targeted by path so none of the shared counts move.

src/Behat/Behat/Gherkin/Filter/TagExpressionFilter.php
Amoifr: Applied, PHPStan agrees since the call site merges two lists.

src/Behat/Behat/Gherkin/Cli/FilterController.php
```diff
@@ -56,6 +57,14 @@ public function configure(Command $command): void
                 'Only execute the features or scenarios with tags' . PHP_EOL .
                 'matching tag filter expression.'
             )
+            ->addOption(
+                '--tag-expression',
+                null,
+                InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY,
```
stof: What happens when you supply the option multiple times ? Is it a AND or a OR ? This needs to be explained in the option description IMO.
Given that tag expressions support boolean combinators and braces, maybe we don't need to support passing multiple ones though.

acoulton:
@Amoifr thanks for the updates - I've merged 3.x into this to pick up #1873 which should resolve the CI failures.

src/Behat/Behat/Gherkin/Cli/FilterController.php
acoulton: @stof good catch - this is implemented in Behat/Gherkin and I believe it is an AND ([processing stops on the first filter that doesn't match](https://github.com/Behat/Gherkin/blob/a30421614d271aa677033afe386758aa5e4841fa/src/Gherkin.php#L127-L135)). ~Although I'm not sure it's quite that simple, on a quick glance at the implementation I think potentially if all the filters partially match a feature file, it may then only run the scenarios that match the last filter...~ (edit: verified it filters the same feature multiple times, so it is indeed an AND).

That behaviour seems to be somewhat undefined: there is [only one test in Behat/Gherkin that actually passes multiple filters](https://github.com/Behat/Gherkin/blob/master/tests/GherkinTest.php#L27) and it doesn't really assert anything about the combined effect of the filtering, only that the mock filters are called.

*However* in Behat, when configuring filters through a config file, [we only allow one filter of each type](https://github.com/Behat/Behat/blob/a081eaa9097f9a584ee722e26d8f76f9ccb6a017/src/Behat/Config/Suite.php#L67-L75).

Therefore I think for consistency with config, we should only allow a single `--tag-expression` filter on the CLI (I'm aware the existing `--tags` and `--name` options support arrays - we might consider deprecating that).

Amoifr: Done, `--tag-expression` is `VALUE_REQUIRED` only now. It follows `--role` and `--narrative`, the other single-value filters, and it matches the config side where `Suite` allows one filter per type.

No scenario passed it twice, so nothing moved in the feature file. 12/12 green in both gherkin modes, PHPStan and CS clean.

Thanks for merging 3.x in, @acoulton. I left the `--tags` and `--name` arrays alone, deprecating those feels like its own PR.

acoulton: @Amoifr great, thank you - agreed, if we did decide to deprecate arrays for the other options that's a separate thing. I'll have a think about that.

acoulton (approved):
Thanks very much @Amoifr - this will need a second maintainer to review but it LGTM now.

carlos-granados (approved):
All good, great job @Amoifr

PR #1864, opened 2026-08-19
feat: Add DataTable and DocString step argument types

### LINKED ISSUE #1639: Proposition: add a dedicated Behat represention of multiline step arguments for usage in step definitions
When writing a step definition for a step taking an argument (either a doc string or a data table), Behat currently passes directly the node from `behat/gherkin` to the step definition (note that `behat/gherkin` talks about PyStringNode and TableNode while cucumber now calls those doc strings and data tables).

If we ever want to migrate to `cucumber/gherkin` in a future version, it would be great to provide a way to abstract from those nodes in step definitions (I'm not focusing on events and hooks here, which can give access to other kind of gherkin nodes, as those are a totally different kind of things for which we will probably need a hard BC break in a major version, but are less common outside extensions).

My idea would be to have some `DocString` and `DataTable` classes (names to be bikeshed) in Behat. Arguments with such parameter types in step definitions would be recognized as arguments expecting a special argument (same than when using one of the gherkin classes as parameter type) and Behat would automatically wrap/convert the Gherkin node into the Behat class.

We have 2 options here:
- make the new class expose the same API than the behat/gherkin one so that it is a drop-in replacement during the migration by just changing the parameter type (and changing the parameter type in any function where that argument is forwarded of course)
- take the opportunity to design a more modern interface for those if relevant (no idea whether this is needed). The `DataTable` class could be modeled based on the [`cucumber-jvm`](https://github.com/cucumber/cucumber-jvm/tree/main/datatable#datatable-object) or [`cucumber-js`](https://github.com/cucumber/cucumber-js/blob/main/docs/support_files/data_table_interface.md) classes (which are quite similar). for doc strings, cucumber passes the content as a normal Java/JS string.

To be future-proof, we should make it possible in the future to refactor those arguments to be wrappers around (or converted from) the `PickleDocString` and `PickleTable` objects of `cucumber/messages` while providing the same public API (besides the constructor of course) so that step definitions don't have to change if/when we make a new major version relying on `cucumber/gherkin` and using their concept of pickles and testcases in the runner instead of the gherkin AST directly (opening the door to supporting the `Rule` keyword of Gherkin in Behat). the prerequisite for that is mostly to make sure that the public API does not leak things specific to behat/gherkin (in particular, `DataTable` must wrap the `TableNode` even all its methods are directly forwarded to some existing methods of `TableNode`, it should not try to extend it)

Amoifr:
Closes #1639.

This adds the dedicated Behat representation of multiline step arguments discussed in the issue:

- `Behat\Step\DataTable` wraps `TableNode` (it never extends it) and exposes an API modelled on the cucumber DataTable classes: `asLists()`, `asMaps()`, `asMap()`, `cell()`, `row()`, `column()`, `width()`, `height()`, `isEmpty()`, `transpose()`, plus `Stringable`. Nothing from behat/gherkin leaks into the public API besides the constructor, so it can later wrap a `PickleTable` without breaking step definitions.
- `Behat\Step\DocString` is the almost-empty `Stringable` DTO suggested by @acoulton: `getContent()` and `__toString()`. Chosen over passing a raw `string` so the #1614 checks keep working unchanged.

Wrapping happens during definition matching, and only when no parameter accepts the raw Gherkin node while one accepts the wrapper: every existing step definition keeps receiving exactly what it received before (covered by a test where a `TableNode|DataTable` parameter gets the raw node). The #1614 unused-argument check recognizes the wrappers too.

Bikeshedding welcome on names and namespace; I went with `Behat\Step` since that's where the user-facing step definition API already lives. Follow-ups not included here: snippet generation still proposes `TableNode`/`PyStringNode`, and a docs PR on Behat/docs once the API settles.

### DIFF
--- a/src/Behat/Behat/Definition/Search/RepositorySearchEngine.php
+++ b/src/Behat/Behat/Definition/Search/RepositorySearchEngine.php
@@ -176,9 +176,10 @@ final class RepositorySearchEngine implements SearchEngine
      * Wraps multiline arguments into their dedicated Behat representation
      * when the definition asks for it.
      *
-     * The raw Gherkin node always wins: an argument is only wrapped when no
-     * parameter of the definition accepts the node itself, and some parameter
-     * accepts the wrapper.
+     * A step cannot mix both representations of the same multiline argument, so
+     * asking for the wrapper anywhere in the signature is enough to wrap. When no
+     * parameter accepts it, the raw Gherkin node is passed through unchanged and
+     * existing definitions keep working.
      *
      * @param list<ArgumentInterface> $multiline
      *
@@ -189,14 +190,12 @@ final class RepositorySearchEngine implements SearchEngine
         return array_map(
             function (ArgumentInterface $argument) use ($function) {
                 if ($argument instanceof TableNode
-                    && !$this->someParameterAccepts($function, $argument::class)
                     && $this->someParameterAccepts($function, DataTable::class)
                 ) {
                     return new DataTable($argument);
                 }
 
                 if ($argument instanceof PyStringNode
-                    && !$this->someParameterAccepts($function, $argument::class)
                     && $this->someParameterAccepts($function, DocString::class)
                 ) {
                     return new DocString($argument);
--- a/src/Behat/Behat/Definition/Search/RepositorySearchEngine.php
+++ b/src/Behat/Behat/Definition/Search/RepositorySearchEngine.php
@@ -18,11 +18,19 @@ use Behat\Behat\Definition\SearchResult;
 use Behat\Behat\Definition\Translator\DefinitionTranslator;
 use Behat\Gherkin\Node\ArgumentInterface;
 use Behat\Gherkin\Node\FeatureNode;
+use Behat\Gherkin\Node\PyStringNode;
 use Behat\Gherkin\Node\StepNode;
+use Behat\Gherkin\Node\TableNode;
+use Behat\Step\DataTable;
+use Behat\Step\DocString;
 use Behat\Testwork\Argument\ArgumentOrganiser;
 use Behat\Testwork\Argument\Exception\UnexpectedMultilineArgumentException;
 use Behat\Testwork\Deprecation\DeprecationCollector;
 use Behat\Testwork\Environment\Environment;
+use ReflectionFunctionAbstract;
+use ReflectionIntersectionType;
+use ReflectionNamedType;
+use ReflectionUnionType;
 
 /**
  * Searches for a step definition using definition repository.
@@ -95,7 +103,7 @@ final class RepositorySearchEngine implements SearchEngine
         }
 
         $function = $definition->getReflection();
-        $match = array_merge($match, array_values($multiline));
+        $match = array_merge($match, $this->wrapMultilineArguments($function, array_values($multiline)));
 
         try {
             $arguments = $this->argumentOrganiser->organiseArguments($function, $match);
@@ -162,5 +170,68 @@ final class RepositorySearchEngine implements SearchEngine
         $namedGroups = count(array_filter(array_keys($match), is_string(...)));
 
         return count($match) - 1 - $namedGroups;
+    }
+
+    /**
+     * Wraps multiline arguments into their dedicated Behat representation
+     * when the definition asks for it.
+     *
+     * The raw Gherkin node always wins: an argument is only wrapped when no
+     * parameter of the definition accepts the node itself, and some parameter
+     * accepts the wrapper.
+     *
+     * @param list<ArgumentInterface> $multiline
+     *
+     * @return list<ArgumentInterface|DataTable|DocString>
+     */
+    private function wrapMultilineArguments(ReflectionFunctionAbstract $function, array $multiline): array
+    {
+        return array_map(
+            function (ArgumentInterface $argument) use ($function) {
+                if ($argument instanceof TableNode
+                    && !$this->someParameterAccepts($function, $argument::class)
+                    && $this->someParameterAccepts($function, DataTable::class)
+                ) {
+                    return new DataTable($argument);
+                }
+
+                if ($argument instanceof PyStringNode
+                    && !$this->someParameterAccepts($function, $argument::class)
+                    && $this->someParameterAccepts($function, DocString::class)
+                ) {
+                    return new DocString($argument);
+                }
+
+                return $argument;
+            },
+            $multiline,
+        );
+    }
+
+    /**
+     * @param class-string $class
+     */
+    private function someParameterAccepts(ReflectionFunctionAbstract $function, string $class): bool
+    {
+        foreach ($function->getParameters() as $parameter) {
+            $type = $parameter->getType();
+            $namedTypes = match (true) {
+                $type instanceof ReflectionNamedType => [$type],
+                $type instanceof ReflectionUnionType,
+                $type instanceof ReflectionIntersectionType => array_filter(
+                    $type->getTypes(),
+                    static fn ($member) => $member instanceof ReflectionNamedType,
+                ),
+                default => [],
+            };
+
+            foreach ($namedTypes as $namedType) {
+                if (!$namedType->isBuiltin() && is_a($class, $namedType->getName(), true)) {
+                    return true;
+                }
+            }
+        }
+
+        return false;
     }
 }
--- a/src/Behat/Step/DataTable.php
+++ b/src/Behat/Step/DataTable.php
@@ -10,8 +10,9 @@
 
 namespace Behat\Step;
 
-use Behat\Gherkin\Exception\NodeException;
 use Behat\Gherkin\Node\TableNode;
+use InvalidArgumentException;
+use OutOfBoundsException;
 use Stringable;
 
 /**
@@ -25,9 +26,6 @@ use Stringable;
  */
 final class DataTable implements Stringable
 {
-    /**
-     * @internal instances are created by Behat, from the Gherkin node of the step
-     */
     public function __construct(
         private readonly TableNode $tableNode,
     ) {
@@ -59,27 +57,43 @@ final class DataTable implements Stringable
      * Returns a two-column table as an associative array, mapping the cells
      * of the first column to the cells of the second column.
      *
-     * A row that has more than two cells maps to the list of its remaining
-     * cells, and a single-column row maps to an empty list.
+     * Every row of a single-column table maps to null.
      *
-     * @return array<string, string|list<string>>
+     * @return array<string, string|null>
+     *
+     * @throws InvalidArgumentException when the table has more than two columns
      */
     public function asMap(): array
     {
-        return $this->tableNode->getRowsHash();
+        $width = $this->width();
+
+        if ($width > 2) {
+            throw new InvalidArgumentException(sprintf(
+                'A table with %d columns cannot be converted to a map, it must have at most two columns.',
+                $width,
+            ));
+        }
+
+        $map = [];
+
+        foreach ($this->asLists() as $row) {
+            $map[$row[0]] = $row[1] ?? null;
+        }
+
+        return $map;
     }
 
     /**
      * Returns the value of a single cell by zero-based row and column indexes.
      *
-     * @throws NodeException when the row or the column does not exist
+     * @throws OutOfBoundsException when the row or the column does not exist
      */
     public function cell(int $row, int $column): string
     {
         $cells = $this->row($row);
 
         if (!array_key_exists($column, $cells)) {
-            throw new NodeException(sprintf('Column #%s does not exist in table.', $column));
+            throw new OutOfBoundsException($this->columnIsOutOfBounds($column));
         }
 
         return $cells[$column];
@@ -89,20 +103,38 @@ final class DataTable implements Stringable
      * Returns a row as a list of cell values, by zero-based index.
      *
      * @return list<string>
+     *
+     * @throws OutOfBoundsException when the row does not exist
      */
     public function row(int $index): array
     {
-        return $this->tableNode->getRow($index);
+        $rows = $this->asLists();
+
+        if (!array_key_exists($index, $rows)) {
+            throw new OutOfBoundsException(sprintf(
+                'Row #%d does not exist in this table, which has %d rows.',
+                $index,
+                count($rows),
+            ));
+        }
+
+        return $rows[$index];
     }
 
     /**
      * Returns a column as a list of cell values, by zero-based index.
      *
      * @return list<string>
+     *
+     * @throws OutOfBoundsException when the column does not exist
      */
     public function column(int $index): array
     {
-        return $this->tableNode->getColumn($index);
+        if ($index < 0 || $index >= $this->width()) {
+            throw new OutOfBoundsException($this->columnIsOutOfBounds($index));
+        }
+
+        return array_column($this->asLists(), $index);
     }
 
     /**
@@ -148,5 +180,14 @@ final class DataTable implements Stringable
     public function __toString(): string
     {
         return $this->tableNode->getTableAsString();
+    }
+
+    private function columnIsOutOfBounds(int $index): string
+    {
+        return sprintf(
+            'Column #%d does not exist in this table, which has %d columns.',
+            $index,
+            $this->width(),
+        );
     }
 }
--- a/src/Behat/Step/DataTable.php
+++ b/src/Behat/Step/DataTable.php
@@ -1,0 +1,163 @@
+<?php
+
+/*
+ * This file is part of the Behat.
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Behat\Step;
+
+use Behat\Gherkin\Node\TableNode;
+use Stringable;
+
+/**
+ * Represents a Gherkin data table passed to a step definition.
+ *
+ * Use this type instead of {@see TableNode} in step definition parameters to
+ * keep them decoupled from the Gherkin AST. The API is modelled on the
+ * cucumber DataTable classes.
+ *
+ * @api
+ */
+final class DataTable implements Stringable
+{
+    /**
+     * @api
+     */
+    public function __construct(
+        private readonly TableNode $tableNode,
+    ) {
+    }
+
+    /**
+     * Returns the whole table, including the header row if any, as a list of
+     * rows, each row being a list of cell values.
+     *
+     * @return list<list<string>>
+     *
+     * @api
+     */
+    public function asLists(): array
+    {
+        return array_values(array_map(array_values(...), $this->tableNode->getTable()));
+    }
+
+    /**
+     * Returns the table as a list of associative arrays, using the first row
+     * as keys for the values of every following row.
+     *
+     * @return list<array<string, string>>
+     *
+     * @api
+     */
+    public function asMaps(): array
+    {
+        return $this->tableNode->getHash();
+    }
+
+    /**
+     * Returns a two-column table as an associative array, mapping the cells
+     * of the first column to the cells of the second column.
+     *
+     * @return array<string, string>
+     *
+     * @api
+     */
+    public function asMap(): array
+    {
+        return $this->tableNode->getRowsHash();
+    }
+
+    /**
+     * Returns the value of a single cell by zero-based row and column indexes.
+     *
+     * @api
+     */
+    public function cell(int $row, int $column): string
+    {
+        return $this->row($row)[$column];
+    }
+
+    /**
+     * Returns a row as a list of cell values, by zero-based index.
+     *
+     * @return list<string>
+     *
+     * @api
+     */
+    public function row(int $index): array
+    {
+        return $this->tableNode->getRow($index);
+    }
+
+    /**
+     * Returns a column as a list of cell values, by zero-based index.
+     *
+     * @return list<string>
+     *
+     * @api
+     */
+    public function column(int $index): array
+    {
+        return $this->tableNode->getColumn($index);
+    }
+
+    /**
+     * Returns the number of rows.
+     *
+     * @api
+     */
+    public function height(): int
+    {
+        return count($this->tableNode->getRows());
+    }
+
+    /**
+     * Returns the number of columns.
+     *
+     * @api
+     */
+    public function width(): int
+    {
+        return count($this->tableNode->getRows()[0] ?? []);
+    }
+
+    /**
+     * Returns whether the table has no rows at all.
+     *
+     * @api
+     */
+    public function isEmpty(): bool
+    {
+        return $this->tableNode->getRows() === [];
+    }
+
+    /**
+     * Returns a new table with rows and columns swapped.
+     *
+     * @api
+     */
+    public function transpose(): self
+    {
+        $rows = array_values(array_map(array_values(...), $this->tableNode->getTable()));
+        $transposed = [];
+        foreach ($rows as $rowIndex => $row) {
+            foreach ($row as $columnIndex => $cell) {
+                $transposed[$columnIndex][$rowIndex] = $cell;
+            }
+        }
+
+        return new self(new TableNode($transposed));
+    }
+
+    /**
+     * @api
+     */
+    public function __toString(): string
+    {
+        return $this->tableNode->getTableAsString();
+    }
+}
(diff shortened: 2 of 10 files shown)


tests/Fixtures/Arguments/features/bootstrap/FeatureContext.php
```diff
@@ -68,4 +70,42 @@ public function iHaveNumberAndNumber($number1, $number2)
...
+    public function aDataTable(DataTable $table)
+    {
+        $this->input = $table;
+    }
+
+    #[Then('/^the data table must be equals to table (\d+)$/')]
+    public function theDataTableMustBeEqualsToTable($number)
+    {
+        PHPUnit\Framework\Assert::assertInstanceOf(DataTable::class, $this->input);
+        PHPUnit\Framework\Assert::assertEquals($this->tables[intval($number)], $this->input->asMaps());
```
acoulton: There's now a conflict between these new assertions and the work done in #1872 to move away from using phpunit assertions in our own features & fixtures. Please could you rebase and update to use the new assertion helpers?

tests/Behat/Tests/Step/DataTableTest.php
```diff
@@ -0,0 +1,104 @@
...
+            ['plum', 'purple'],
+        ], $this->table->asLists());
+    }
+
+    public function testAsMapsUsesTheFirstRowAsKeys(): void
+    {
+        $this->assertSame([
+            ['name' => 'apple', 'colour' => 'green'],
+            ['name' => 'plum', 'colour' => 'purple'],
+        ], $this->table->asMaps());
```
acoulton: Please could you add some coverage to define the behaviour of `asMaps` if there is only one row?

tests/Behat/Tests/Step/DataTableTest.php
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    public function testAsMapMapsTheFirstColumnToTheSecondOne(): void
+    {
+        $this->assertSame([
+            'name' => 'colour',
+            'apple' => 'green',
+            'plum' => 'purple',
+        ], $this->table->asMap());
```
acoulton: Please could you add some coverage to define the behaviour of `asMap` if there is either 1 or >2 columns?

tests/Behat/Tests/Step/DataTableTest.php
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    public function testRowIsAccessedByZeroBasedIndex(): void
+    {
+        $this->assertSame(['apple', 'green'], $this->table->row(1));
+    }
+
+    public function testColumnIsAccessedByZeroBasedIndex(): void
+    {
+        $this->assertSame(['colour', 'green', 'purple'], $this->table->column(1));
```
acoulton, line 123: Please could you add coverage to define the behaviour of accessing a cell / row / column that does not exist?

src/Behat/Step/DocString.php
```diff
@@ -0,0 +1,51 @@
...
+     * @api
+     */
+    public function getContent(): string
+    {
+        return $this->pyStringNode->getRaw();
+    }
+
+    /**
+     * @api
+     */
```
acoulton:
```suggestion
```

src/Behat/Step/DocString.php
```diff
@@ -0,0 +1,51 @@
...
+ * Use this type instead of {@see PyStringNode} in step definition parameters
+ * to keep them decoupled from the Gherkin AST.
+ *
+ * @api
+ */
+final class DocString implements Stringable
+{
+    /**
+     * @api
+     */
```
acoulton: What's the usecase for external code needing to create an instance of this? If we can keep this as something that Behat creates for you that would be good.

src/Behat/Step/DataTable.php
```diff
@@ -0,0 +1,163 @@
...
+                $transposed[$columnIndex][$rowIndex] = $cell;
+            }
+        }
+
+        return new self(new TableNode($transposed));
+    }
+
+    /**
+     * @api
+     */
```
acoulton:
```suggestion
```

src/Behat/Step/DataTable.php
```diff
@@ -0,0 +1,163 @@
...
+    public function isEmpty(): bool
+    {
+        return $this->tableNode->getRows() === [];
+    }
+
+    /**
+     * Returns a new table with rows and columns swapped.
+     *
+     * @api
```
acoulton:
```suggestion
```

src/Behat/Step/DataTable.php
```diff
@@ -0,0 +1,163 @@
...
+     */
+    public function width(): int
+    {
+        return count($this->tableNode->getRows()[0] ?? []);
+    }
+
+    /**
+     * Returns whether the table has no rows at all.
+     *
+     * @api
```
acoulton:
```suggestion
```

src/Behat/Step/DataTable.php
```diff
@@ -0,0 +1,163 @@
...
+     */
+    public function height(): int
+    {
+        return count($this->tableNode->getRows());
+    }
+
+    /**
+     * Returns the number of columns.
+     *
+     * @api
```
acoulton:
```suggestion
```

src/Behat/Step/DataTable.php
```diff
@@ -0,0 +1,163 @@
...
+     */
+    public function column(int $index): array
+    {
+        return $this->tableNode->getColumn($index);
+    }
+
+    /**
+     * Returns the number of rows.
+     *
+     * @api
```
acoulton:
```suggestion
```

src/Behat/Step/DataTable.php
```diff
@@ -0,0 +1,163 @@
...
+    {
+        return $this->tableNode->getRow($index);
+    }
+
+    /**
+     * Returns a column as a list of cell values, by zero-based index.
+     *
+     * @return list<string>
+     *
+     * @api
```
acoulton:
```suggestion
```

src/Behat/Step/DataTable.php
```diff
@@ -0,0 +1,163 @@
...
+    {
+        return $this->row($row)[$column];
+    }
+
+    /**
+     * Returns a row as a list of cell values, by zero-based index.
+     *
+     * @return list<string>
+     *
+     * @api
```
acoulton:
```suggestion
```

src/Behat/Step/DataTable.php
```diff
@@ -0,0 +1,163 @@
...
+     */
+    public function asMap(): array
+    {
+        return $this->tableNode->getRowsHash();
+    }
+
+    /**
+     * Returns the value of a single cell by zero-based row and column indexes.
+     *
+     * @api
```
acoulton:
```suggestion
```

src/Behat/Step/DataTable.php
```diff
@@ -0,0 +1,163 @@
...
+        return $this->tableNode->getHash();
+    }
+
+    /**
+     * Returns a two-column table as an associative array, mapping the cells
+     * of the first column to the cells of the second column.
+     *
+     * @return array<string, string>
+     *
+     * @api
```
acoulton:
```suggestion
```

src/Behat/Step/DataTable.php
```diff
@@ -0,0 +1,163 @@
...
+        return array_values(array_map(array_values(...), $this->tableNode->getTable()));
+    }
+
+    /**
+     * Returns the table as a list of associative arrays, using the first row
+     * as keys for the values of every following row.
+     *
+     * @return list<array<string, string>>
+     *
+     * @api
```
acoulton:
```suggestion
```

src/Behat/Step/DataTable.php
```diff
@@ -0,0 +1,163 @@
...
+    ) {
+    }
+
+    /**
+     * Returns the whole table, including the header row if any, as a list of
+     * rows, each row being a list of cell values.
+     *
+     * @return list<list<string>>
+     *
+     * @api
```
acoulton:
```suggestion
```
The `@api` tag is redundant - the class itself is `@api` so all public methods are part of the API by default.

src/Behat/Step/DocString.php
```diff
@@ -0,0 +1,51 @@
...
+    public function __construct(
+        private readonly PyStringNode $pyStringNode,
+    ) {
+    }
+
+    /**
+     * Returns the text of the doc string.
+     *
+     * @api
```
acoulton:
```suggestion
```

src/Behat/Step/DataTable.php
```diff
@@ -0,0 +1,163 @@
...
+     * Returns a two-column table as an associative array, mapping the cells
+     * of the first column to the cells of the second column.
+     *
+     * @return array<string, string>
+     *
+     * @api
+     */
+    public function asMap(): array
+    {
+        return $this->tableNode->getRowsHash();
```
acoulton: Aside from any other consideration, the `asLists` / `asMap` / `asMaps` naming makes much more immediate sense to me. I always end up making a mistake between `getRowsHash` and `getColumnsHash` sooner or later 😆

src/Behat/Step/DataTable.php
```diff
@@ -0,0 +1,163 @@
...
+ * keep them decoupled from the Gherkin AST. The API is modelled on the
+ * cucumber DataTable classes.
+ *
+ * @api
+ */
+final class DataTable implements Stringable
+{
+    /**
+     * @api
+     */
```
acoulton: What's the usecase for external code needing to create an instance of this? If we can keep this as something that Behat creates for you that would be good.

tests/Fixtures/Arguments/features/bootstrap/FeatureContext.php
```diff
@@ -68,4 +70,42 @@ public function iHaveNumberAndNumber($number1, $number2)
...
+    {
+        PHPUnit\Framework\Assert::assertInstanceOf(DocString::class, $this->input);
+        PHPUnit\Framework\Assert::assertEquals($this->strings[intval($number)], $this->input->getContent());
+    }
+
+    #[Given('/^a table that could be wrapped or not:$/')]
+    public function aTableThatCouldBeWrappedOrNot(TableNode|DataTable $table)
+    {
+        $this->input = $table;
+    }
```
acoulton: I think to guarantee BC, we maybe need a feature / step to prove that it is still TableNode  / DocString if there are mixed params and no typehints on the step at all e.g.

```
#[Given('a step that takes :text and these values:')]
public function anOldStep($text, $table) {
   ...
}
```

src/Behat/Behat/Definition/Search/RepositorySearchEngine.php
```diff
@@ -112,4 +120,67 @@ private function match(Definition $definition, string $stepLocation, string $ste
...
+     *
+     * @return list<ArgumentInterface|DataTable|DocString>
+     */
+    private function wrapMultilineArguments(ReflectionFunctionAbstract $function, array $multiline): array
+    {
+        return array_map(
+            function (ArgumentInterface $argument) use ($function) {
+                if ($argument instanceof TableNode
+                    && !$this->someParameterAccepts($function, $argument::class)
+                    && $this->someParameterAccepts($function, DataTable::class)
```
acoulton: I think perhaps we could be more assertive here about expected use - also saving some cycles in the runtime parameter checks.

I'd say it's not valid to have a step that has both a `DataTable $something` *and* a `TableNode $other`. For the moment there can only be one multiline argument per step - that may change in future but even then the author should specify either the new classes or the old ones, not a mix.

So IMO if there is any argument that accepts a `DataTable`, we wrap the table. If that causes a problem for other arguments, the existing argument type checking / matching will handle that.

If no arguments specify `DataTable` then we return the raw argument, for BC.

acoulton (changes requested):
Thanks @Amoifr this is great progress - a few comments and suggestions to take a look at please.

tests/Fixtures/Arguments/features/bootstrap/FeatureContext.php
Amoifr: Rebased on 3.x, so #1872 is in, and the fixture now uses `Behat\Tests\Fixtures\Assert`. The helper has no `assertInstanceOf`, so I went with `Assert::assertSame(DataTable::class, $this->input::class)` rather than adding a method to your new class. It is stricter too, which is what these assertions actually want.

The rebase also picks up #1873, so the red 8.4 and 8.5 jobs should go green.

tests/Behat/Tests/Step/DataTableTest.php
Amoifr: Added `testAsMapsIsEmptyWhenTheTableOnlyHasItsHeaderRow`: a table reduced to its header row gives `[]`, which is what `getHash()` already did.

tests/Behat/Tests/Step/DataTableTest.php
Amoifr: Added both cases, and they turned up something worth deciding rather than just documenting.

`asMap()` was annotated `@return array<string, string>` and that was not true: with one column you get `['name' => [], 'apple' => []]`, and with three you get `['name' => ['colour', 'size']]`. I pinned the current behaviour in two tests and fixed the annotation to `array<string, string|list<string>>`.

The alternative, closer to Cucumber, is to make `asMap()` throw when the table does not have exactly two columns. I did not want to make that call for you, so say the word and I will switch it.

tests/Behat/Tests/Step/DataTableTest.php
Amoifr: Added four tests, and one of them found a bug. `row()` and `column()` already threw a clean `NodeException`, but `cell()` with an out of range column produced `Warning: Undefined array key 9` followed by `TypeError: Return value must be of type string, null returned`. It now throws the same `NodeException` with the message Gherkin uses.

src/Behat/Step/DocString.php
Amoifr: Removed, thanks. Kept the class level `@api` only.

src/Behat/Step/DocString.php
Amoifr: Same as on `DataTable`: constructor marked `@internal`, since only Behat builds these.

src/Behat/Step/DataTable.php
Amoifr: Removed, thanks. Kept the class level `@api` only.

src/Behat/Step/DataTable.php
Amoifr: Removed, thanks. Kept the class level `@api` only.

src/Behat/Step/DataTable.php
Amoifr: Removed, thanks. Kept the class level `@api` only.

src/Behat/Step/DataTable.php
Amoifr: Removed, thanks. Kept the class level `@api` only.

src/Behat/Step/DataTable.php
Amoifr: Removed, thanks. Kept the class level `@api` only.

src/Behat/Step/DataTable.php
Amoifr: Removed, thanks. Kept the class level `@api` only.

src/Behat/Step/DataTable.php
Amoifr: Removed, thanks. Kept the class level `@api` only.

src/Behat/Step/DataTable.php
Amoifr: Removed, thanks. Kept the class level `@api` only.

src/Behat/Step/DataTable.php
Amoifr: Removed, thanks. Kept the class level `@api` only.

src/Behat/Step/DataTable.php
Amoifr: Removed, thanks. Kept the class level `@api` only.

src/Behat/Step/DataTable.php
Amoifr: Removed on every method here, and on `DocString` too. Kept the class level `@api`, which as you say already covers the public methods.

src/Behat/Step/DocString.php
Amoifr: Removed, thanks. Kept the class level `@api` only.

src/Behat/Step/DataTable.php
Amoifr: Glad the naming lands well. `getRowsHash` versus `getColumnsHash` catches me out every single time too. 😄

src/Behat/Step/DataTable.php
Amoifr: No real usecase, agreed. Marked the constructor `@internal instances are created by Behat, from the Gherkin node of the step`. I stopped short of making it private since Behat itself has to build it from `RepositorySearchEngine`, but nothing in the public API hands one out now.

tests/Fixtures/Arguments/features/bootstrap/FeatureContext.php
Amoifr: Good catch, and the scenario I had was actually asserting the old rule, so it had to go anyway. Replaced with exactly what you describe: a step with no typehints at all, `($text, $table)`, which must still receive a raw `TableNode`, plus the same for `PyStringNode`, and an assertion that the other argument still arrives intact.

src/Behat/Behat/Definition/Search/RepositorySearchEngine.php
Amoifr: Agreed, and done. Wrapping now only asks whether some parameter accepts `DataTable` or `DocString`; if none does, the raw node goes through unchanged. The `!someParameterAccepts($function, $argument::class)` half is gone, so a mixed signature is no longer a supported shape and the existing argument matching reports it.

This is what removed the `TableNode|DataTable` scenario in the fixture: under the new rule a union resolves to the wrapper, so that test was encoding the behaviour you are asking me to drop.

src/Behat/Step/DataTable.php
acoulton: @Amoifr sorry, I should have been clearer - this `@internal` tag is redundant.

Just removing the `@api` tag was enough, per our API policy:

> Objects can only be created via new if the PHPDoc of the public constructor is directly tagged with `@api`. In all other cases (even if the class PHPDoc has `@api`), instances of the class should only be created by / retrieved from Behat.

src/Behat/Step/DocString.php
acoulton: As above, the `@internal` tag is redundant

tests/Behat/Tests/Step/DataTableTest.php
acoulton: Very minor nitpick, but if you're in this test file anyway I'd prefer to have the tests for each function together (the new test is further down) - or better yet to use data providers.

tests/Behat/Tests/Step/DataTableTest.php
acoulton: I think if we're using Cucumber's naming convention we should follow their behaviour and:
* throw if there are >2 columns. 
* return array<string, null> if there is 1 column (which looks like [that's cucumber's behaviour](https://github.com/cucumber/cucumber-jvm/blob/main/datatable/src/test/java/io/cucumber/datatable/DataTableTypeRegistryTableConverterTest.java#L628)

Being strict about the columns helps ensure sure the feature file matches the step author's assumptions about the table.

If in future someone *needs* support for the old Behat `'col1' => ['col2', 'col3', ...]` behaviour we could provide that as a separate method.

tests/Behat/Tests/Step/DataTableTest.php
```diff
@@ -0,0 +1,164 @@
...
+            'apple' => ['green', 'small'],
+        ], $table->asMap());
+    }
+
+    public function testAccessingARowThatDoesNotExistThrows(): void
+    {
+        $this->expectException(NodeException::class);
+        $this->expectExceptionMessage('Rows #9 does not exist in table.');
+
+        $this->table->row(9);
```
acoulton: Using an index one more than the highest possible (e.g row 3) would help to prove no zero-index bugs.

tests/Behat/Tests/Step/DataTableTest.php
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    public function testAccessingAColumnThatDoesNotExistThrows(): void
+    {
+        $this->expectException(NodeException::class);
+        $this->expectExceptionMessage('Column #9 does not exist in table.');
+
+        $this->table->column(9);
```
acoulton: Using an index one more than the highest possible (e.g column 2) would help to prove no zero-index bugs.

tests/Behat/Tests/Step/DataTableTest.php
acoulton: As we're trying to break the coupling on Gherkin with these new classes, I think we shouldn't be throwing (or bubbling) `NodeException`. Also because really it's not an issue with the `Node` but with the calling code.

I'd suggest all these row / column / cell index failures should be just `OutOfBoundsException`?

acoulton (changes requested):
Thanks @Amoifr - just a few more comments

src/Behat/Step/DataTable.php
Amoifr: Ah, understood — thanks for pointing me at the policy, that's much cleaner. `@internal` dropped from the constructor: the class keeps `@api`, the constructor stays untagged, and that already says "Behat builds these".

src/Behat/Step/DocString.php
Amoifr: Same here, `@internal` removed from the constructor for the same reason.

tests/Behat/Tests/Step/DataTableTest.php
Amoifr: Fair nitpick, and a habit worth acquiring. I'll move the new test back next to the other `asMaps` one, and fold the variants of each method into data providers while I'm in there.

tests/Behat/Tests/Step/DataTableTest.php
Amoifr: Agreed, and thanks for the cucumber-jvm reference. If I borrow the naming I should borrow the semantics, so `asMap()` will throw when the table has more than two columns and return `array<string, null>` for a single column. That also lets me drop the `string|list<string>` annotation I had introduced, which never felt right.

The old permissive `'col1' => ['col2', ...]` behaviour stays out; happy to add it back as a separate method if anyone ever misses it.

For the >2 columns case I'll throw an `InvalidArgumentException` (the table doesn't have the shape the step author asked for) unless you'd rather have a dedicated exception class in `Behat\Step` — say the word and I'll switch.

tests/Behat/Tests/Step/DataTableTest.php
Amoifr: Good point — bubbling a Gherkin exception out of the very classes meant to hide Gherkin does rather defeat the purpose, and you're right that the fault is in the calling code, not in the node. I'll throw a plain `OutOfBoundsException` from `row()`, `column()` and `cell()` with our own messages, and update the tests accordingly.

tests/Behat/Tests/Step/DataTableTest.php
Amoifr: Right, `row(9)` proves nothing at all. Switching to `row(3)`, the first index past the end of the fixture table.

tests/Behat/Tests/Step/DataTableTest.php
Amoifr: Same here, `column(2)`.

Amoifr:
@acoulton thank you for the thorough review. Twenty-something comments is a lot of your time spent on my code, and most of them are things I should have caught before pushing: the edge cases, the API policy on `@api`/`@internal`, and above all sticking to cucumber's semantics once I had borrowed its names. Sorry for the extra round-trips.

I've turned the whole review into a checklist I'll run through before my next PR, so hopefully the next one costs you rather less red ink. Changes for all the open threads coming shortly.

acoulton (approved):
@Amoifr this is looking great to me now, thanks for the iterations

stof (approved):

acoulton:
Thanks @Amoifr - if you'd like to also work on the followup to add these to the documentation and to the snippet generation that would be brilliant :)

PR #1865, opened 2026-08-20
feat: Show the suite name in failure summaries when several suites ran

### LINKED ISSUE #1772: Show the suite for a failing scenario, if there are multiple suites

Amoifr:
Fixes #1772

When several suites share feature files, the failure summaries printed by the `progress` and `pretty` formatters only showed `features/apples.feature:9`, with no way to tell which suite the failure came from. This PR appends the suite name to those lists, following the format suggested in the issue:

```
--- Failed steps:

001 Scenario: I'm little hungry    # features/bananas.feature:9 (default)
      Then I should have 3 bananas # features/bananas.feature:11
        Failed asserting that 2 matches expected 3.

002 Scenario: I'm little hungry    # features/bananas.feature:9 (suite2)
      Then I should have 3 bananas # features/bananas.feature:11
        Failed asserting that 2 matches expected 3.
```

and, with `short_summary` enabled (or in the pretty formatter):

```
--- Failed scenarios:

    features/bananas.feature:9 (default) (on line 11)
    features/bananas.feature:9 (suite2) (on line 11)
```

### How it works

- `ScenarioStat` and `StepStatV2` get an optional `suiteName` (new trailing constructor argument and getter, so no BC break), captured by the existing stats listeners on `BeforeSuiteTested` / `AfterSuiteTested`.
- `TotalStatistics::getSuiteNames()` returns the suites that executed at least one scenario. `PhaseStatistics` delegates to it. The `Statistics` interface is `@api`, so it is left untouched; the two statistics printers check for `TotalStatistics`, which is what they are always wired with.
- `ListPrinter` prints the suite name when asked to, and now also matches the failing step to its scenario by suite (two suites can produce the same scenario path).
- The annotation is only printed when more than one suite ran scenarios, so single-suite runs are unchanged.

### One note on "more than one suite was executed"

I went with "more than one suite executed at least one scenario" rather than counting `BeforeSuiteTested` events. The reason is that running `behat features/apples.feature` dispatches `BeforeSuiteTested` for every configured suite, including the ones that end up with zero matching scenarios. Counting those would add the annotation for anyone with two suites who runs a single file, which is exactly the noise the issue wanted to avoid. If a suite ran scenarios, it is counted, so the "only one scenario fails, which suite was it in?" case from the discussion is covered. Happy to adjust if you would rather have the stricter interpretation.

### Tests

- `features/rerun_with_multiple_suite.feature` updated (both suites run there, the `--suite suite2` and the final `--rerun` with only one suite left stay unannotated).
- New `features/multiple_suites_summary.feature` covering the failed steps list, the short summary list, the pretty formatter and the single-suite case.

### DIFF
--- a/features/multiple_suites_summary.feature
+++ b/features/multiple_suites_summary.feature
@@ -1,0 +1,106 @@
+Feature: Suite names in the summary lists
+  In order to identify the suite a scenario failed in when several suites share feature files
+  As a feature developer
+  I need the failure summaries to name the suite when more than one suite was executed
+
+  Background:
+    Given I initialise the working directory from the "Rerun" fixtures folder
+    And I provide the following options for all behat invocations:
+      | option      | value |
+      | --no-colors |       |
+
+  Scenario: Failed steps are annotated with the suite name when several suites run
+    When I run "behat --format progress"
+    Then it should fail with:
+      """
+      ..F.............F......F.............F......F.............F....
+
+      --- Failed steps:
+
+      001 Scenario: I'm little hungry   # features/apples.feature:9 (default)
+            Then I should have 3 apples # features/apples.feature:11
+              Failed asserting that 2 matches expected 3.
+
+      002 Example: | 0   | 4     | 8      | # features/apples.feature:29 (default)
+            Then I should have 8 apples     # features/apples.feature:24
+              Failed asserting that 7 matches expected 8.
+
+      003 Scenario: I'm little hungry    # features/bananas.feature:9 (default)
+            Then I should have 3 bananas # features/bananas.feature:11
+              Failed asserting that 2 matches expected 3.
+
+      004 Example: | 0   | 4     | 8      | # features/bananas.feature:29 (default)
+            Then I should have 8 bananas    # features/bananas.feature:24
+              Failed asserting that 7 matches expected 8.
+
+      005 Scenario: I'm little hungry    # features/bananas.feature:9 (suite2)
+            Then I should have 3 bananas # features/bananas.feature:11
+              Failed asserting that 2 matches expected 3.
+
+      006 Example: | 0   | 4     | 8      | # features/bananas.feature:29 (suite2)
+            Then I should have 8 bananas    # features/bananas.feature:24
+              Failed asserting that 7 matches expected 8.
+
+      18 scenarios (12 passed, 6 failed)
+      63 steps (57 passed, 6 failed)
+      """
+
+  Scenario: Failed scenarios are annotated with the suite name when several suites run
+    When I run behat with the following additional options:
+      | option            | value                     |
+      | --format          | progress                  |
+      | --format-settings | '{"short_summary": true}' |
+    Then it should fail with:
+      """
+      ..F.............F......F.............F......F.............F....
+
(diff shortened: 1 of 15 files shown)


features/multiple_suites_summary.feature
```diff
@@ -0,0 +1,106 @@
+Feature: Suite names in the summary lists
+  In order to identify the suite a scenario failed in when several suites share feature files
+  As a feature developer
+  I need the failure summaries to name the suite when more than one suite was executed
+
+  Background:
+    Given I initialise the working directory from the "Rerun" fixtures folder
```
acoulton: I think ideally I wouldn't reuse the `Rerun` fixture for this - it couples this feature to any changes in the rerun feature, when there's no relation between these features. 

I think it should either have a standalone fixture, or perhaps be added into the `TestReportFormat` fixture, since this is related to test report formatting.

acoulton (commented):
Thanks @Amoifr - just a comment on the test fixture for the feature file, but the implementation looks great, good job :)

features/multiple_suites_summary.feature
Amoifr: Good point, reusing the `Rerun` fixture was lazy of me. I went with the standalone option: a new `MultipleSuitesSummary` fixture with its own `behat.php` declaring two suites that share `bananas.feature`, so the annotation is exercised on a feature that genuinely runs twice.

I did look at folding it into `TestReportFormat`, but its `multiple_suites_*` features have no failing scenarios, so I would have had to add my own features and a second config in there anyway, which reintroduces the coupling you are pointing at, this time with the junit and json format tests. Happy to move it if you would still prefer that.

Nice side effect: sizing the fixture for this one need brought the expectations down from 18 scenarios to 6, so the docstrings are much easier to read. Every expected block is captured from a real run rather than hand counted.

Thanks as well for merging 3.x into the branch, that picks up #1873 and turns the red 8.4/8.5 jobs green. I pushed on top of your merge rather than force pushing a rebase over it.

acoulton (approved):
Great, thanks for updating the fixture - a separate fixture makes sense. This all looks good to me now - it will need a second maintainer's review to merge.

Out of interest, some of your replies on this and your other PR read like an AI agent - can I ask how much of the work was AI vs human? It's not a problem, just curious 😄

carlos-granados (approved):
Thanks @Amoifr Great implementation and well tested

Issue #1697, opened 2025-11-20
Pretty formatter does not accurately show scenarios with multiple Examples tables


acoulton:
Related to #1696 and #1082.

If a Scenario has multiple Examples tables, these are executed correctly. However, the pretty printer outputs as though they were in a single table, losing any table titles and rendering inconsistent column padding. For example, a feature like:

```gherkin
        Scenario: Grouped examples
          When I input <name>
          Then I should see <result>

          Examples: valid cases
            | name  | result   |
            | Bob   | Hi Bob   |
            | Jenny | Hi Jenny |

          Examples: invalid cases
            | name   | result                             |
            | 123456 | "123456" doesn't look like a name? |
            | Brian  | Sorry Brian, you're banned         |
```

will output like:

```
        Scenario: Grouped examples     # features/test.feature:6
          When I input <name>          # FeatureContext::input()
          Then I should see "<result>" # FeatureContext::assertSee()
      
          Examples:
            | name  | result   |
            | Bob   | Hi Bob   |
            | Jenny | Hi Jenny |
            | 123456 | '123456' doesn't look like a name? |
            | Brian | Sorry Brian, you're banned |
```

This is because Behat uses the original Behat/Gherkin API to fetch `OutlineNode::getExamples()` and `OutlineNode::getExampleTable()`. When Behat/Gherkin added support for parsing multiple tables, a new API was added for fetching the tables separately and the pre-existing methods were updated to convert the parsed data into a single array of `ExampleNode` and a single `ExampleTableNode` to avoid any BC impact.

This works for our test runner, and means that all the information the formatters display about individual examples (including the line numbers in the file) are correct. 

However, printing separate headings and correctly padding the columns would require updating the pretty printer to use the new Behat/Gherkin API to fetch the original parsed nodes rather than the modified copies from the BC layer.

It may also need refactoring to the pretty printer's event handling - currently the `PrettyOutlinePrinter` renders the start of the outline during the `BeforeOutlineTested` event. This includes the header of the examples table, with the assumption that there is only one. The `PrettyExampleRowPrinter` then prints the individual table rows after each scenario / example is tested. From an initial review of the code, my hunch is that it will be tricky to update this to render separate table headings without a BC break in those classes.

We could potentially consider this for 4.0, or perhaps mark the various individual printer classes that the pretty printer uses as `@internal` to allow us to make breaking changes to them later in the 4.x series.

acoulton:
[...]

PR #1866, opened 2026-08-22
fix: multiple example tables in pretty printer

carlos-granados:
The pretty formatter fetched outline examples through the deprecated `OutlineNode::getExampleTable()`, which merges every examples table into one, so a scenario with multiple tables printed as a single unnamed `Examples:` block with padding computed across all of them — losing each table's tags, name and description. It now uses `getExampleTables()` and prints a heading before the first example of each table, padded independently. 

This fixes two further consequences of the merge: `--expand` had the same broken heading, and any outline whose tables differed in column names or ordering crashed the pretty formatter outright.

Fixes https://github.com/Behat/Behat/issues/1697

### DIFF
--- a/src/Behat/Behat/Output/Node/Printer/ExampleRowPrinter.php
+++ b/src/Behat/Behat/Output/Node/Printer/ExampleRowPrinter.php
@@ -12,6 +12,7 @@ namespace Behat\Behat\Output\Node\Printer;
 
 use Behat\Behat\EventDispatcher\Event\AfterStepTested;
 use Behat\Gherkin\Node\ExampleNode;
+use Behat\Gherkin\Node\ExampleTableNode;
 use Behat\Gherkin\Node\OutlineNode;
 use Behat\Testwork\Output\Formatter;
 
@@ -27,7 +28,8 @@ interface ExampleRowPrinter
     /**
      * Prints example row result using provided printer.
      *
+     * @param ExampleTableNode $exampleTable the examples table that $example was created from
      * @param AfterStepTested[] $events
      */
-    public function printExampleRow(Formatter $formatter, OutlineNode $outline, ExampleNode $example, array $events): void;
+    public function printExampleRow(Formatter $formatter, OutlineNode $outline, ExampleTableNode $exampleTable, ExampleNode $example, array $events): void;
 }
--- a/src/Behat/Behat/Output/Node/Printer/ExampleRowPrinter.php
+++ b/src/Behat/Behat/Output/Node/Printer/ExampleRowPrinter.php
@@ -12,7 +12,6 @@ namespace Behat\Behat\Output\Node\Printer;
 
 use Behat\Behat\EventDispatcher\Event\AfterStepTested;
 use Behat\Gherkin\Node\ExampleNode;
-use Behat\Gherkin\Node\ExampleTableNode;
 use Behat\Gherkin\Node\OutlineNode;
 use Behat\Testwork\Output\Formatter;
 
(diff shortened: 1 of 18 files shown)


src/Behat/Behat/Output/Node/Printer/Helper/ExampleTableResolver.php
```diff
@@ -0,0 +1,67 @@
...
+use OutOfBoundsException;
+
+/**
+ * Maps an example back to the examples table it was created from.
+ *
+ * An outline can have any number of examples tables, but {@see OutlineNode::getExamples()} returns a flat list of
+ * the examples from all of them. Formatters that want to render the tables as they appear in the feature file need
+ * to know which table each example belongs to, and where it sits within that table.
+ *
+ * Examples are matched on their line number, which is unique within a feature file.
```
acoulton, line 25: Good solution :)

src/Behat/Behat/Output/Node/Printer/OutlinePrinter.php
```diff
@@ -29,6 +30,11 @@ interface OutlinePrinter
      */
     public function printHeader(Formatter $formatter, FeatureNode $feature, OutlineNode $outline): void;
 
+    /**
+     * Prints the header of one of the outline's examples tables using the provided printer.
+     */
+    public function printExamplesTableHeader(Formatter $formatter, ExampleTableNode $table): void;
```
acoulton, line 36: Could this be an issue for any third-party formatters that have already updated to 4.0-alpha and are implementing this interface?

src/Behat/Behat/Output/Node/Printer/OutlineTablePrinter.php
```diff
@@ -32,6 +33,11 @@ interface OutlineTablePrinter
      */
     public function printHeader(Formatter $formatter, FeatureNode $feature, OutlineNode $outline, array $results): void;
 
+    /**
+     * Prints the header of one of the outline's examples tables using the provided printer.
+     */
+    public function printExamplesTableHeader(Formatter $formatter, ExampleTableNode $table): void;
```
acoulton, line 39: Could this be an issue for any third-party formatters that have already updated to 4.0-alpha and are implementing this interface?

src/Behat/Behat/Output/Node/Printer/ExampleRowPrinter.php
```diff
@@ -27,7 +28,8 @@ interface ExampleRowPrinter
     /**
      * Prints example row result using provided printer.
      *
+     * @param ExampleTableNode $exampleTable the examples table that $example was created from
      * @param AfterStepTested[] $events
      */
-    public function printExampleRow(Formatter $formatter, OutlineNode $outline, ExampleNode $example, array $events): void;
+    public function printExampleRow(Formatter $formatter, OutlineNode $outline, ExampleTableNode $exampleTable, ExampleNode $example, array $events): void;
```
acoulton, line 34: Could this be an issue for any third-party formatters that have already updated to 4.0-alpha and are implementing this interface?

acoulton (commented):
Thanks @carlos-granados - this looks great. 

Just wondering if we need to be concerned about / can reduce the impact on anyone who has already updated to 4.0-alpha (without causing too much extra complexity in our own codebase)?

carlos-granados:
> Thanks @carlos-granados - this looks great.
> 
> Just wondering if we need to be concerned about / can reduce the impact on anyone who has already updated to 4.0-alpha (without causing too much extra complexity in our own codebase)?

Good suggestion - I modified the implementation so that there's no break against alpha1 now.

printExampleRow() keeps its original signature: the printer resolves the table itself via ExampleTableResolver rather than being passed it.

printExamplesTableHeader() moved out of OutlinePrinter/OutlineTablePrinter into a new ExamplesTableHeaderPrinter interface, which the listeners check with `instanceof`. A formatter that hasn't implemented it just doesn't get the per-table headings, which is its behaviour today.

acoulton (approved):

Issue #1691, opened 2025-11-07
Warn when number of parameters does not match


carlos-granados:
Reported in Twitter by @TomasVotruba

When the number of parameters in the step definition and the function does not match we should show some kind of warning or error: 

<img width="586" height="169" alt="Image" src="https://github.com/user-attachments/assets/2a4f306f-06e7-40c1-949d-b5e4d311b872" />

acoulton:
Thanks @carlos-granados - we thought about this briefly as part of #1609 - 

Throwing an error would certainly match the behaviour of cucumber-ruby. However it would be a BC break for us as there may be people particularly with regex where they have a capturing group that they don't actually require as a parameter eg:

```php
#[When('/(I|somebody) hits the (?P<button>red|green) button'/)] 
public function hitButton($button) { /*... */ }
```

A warning could work, but we'd need to work out where/how to show warnings and how that works with different formatters.

Or it might logically work as an error controlled by an opt-in config (perhaps in testerOptions)? 

The advantage of an optional error is we could potentially then flip that to opt-out in a future major release to match the other cucumber runners "out of the box"?

stof:
Once Behat supports proper reporting of deprecations (see the discussion in #1574), Behat could trigger a deprecation for such usage.
The solution then would be to use a non-capturing group in the regex for the use case of @acoulton.

acoulton:
This was fixed by @carlos-granados in #1867 (deprecation in 3.x) and #1887 (exception in 4.x)

PR #1867, opened 2026-08-23
Throw deprecation if definition and function number of parameters do not match

carlos-granados:
When a step pattern captures more arguments than the definition method has parameters, Behat silently discards the extras, so a stray capturing group goes unnoticed. This reports a deprecation via `DeprecationCollector`, anchored to the definition's own file and line, naming the pattern, the method and both counts, and pointing at non-capturing groups `(?:...)` as the fix for patterns that intentionally capture more than they use. Variadic definitions are skipped so `FriendsOfBehat/VariadicExtension` is unaffected; the missing-value direction already throws via `Argument\Validator`. The `4.x` branch will turn this into an error.

`DeprecationCollector::trigger()` gained optional `$file/$line`. Existing deprecations describe Behat's own APIs, so the backtraced call site suits them; this one describes user code, and without it the printed location was `RepositorySearchEngine.php:139`.

Implements the first part of the fix for https://github.com/Behat/Behat/issues/1691

### DIFF
--- a/features/print_deprecations.feature
+++ b/features/print_deprecations.feature
@@ -75,6 +75,18 @@ Feature: Print deprecations
       2 steps (2 passed)
       """
 
+  Scenario: Patterns providing more arguments than the definition accepts are deprecated
+    When I run behat with the following additional options:
+      | option   | value                        |
+      | --config | behat-unused-arguments.php   |
+    Then it should pass with:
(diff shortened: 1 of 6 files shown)


acoulton (approved):
Brilliant, thanks @carlos-granados

Issue #1237, opened 2019-09-24
Many final classes miss proper interfaces


AlwinGarside:
I'm currently writing an extension for Behat which requires me two extend and/or wrap some specific behat classes. I attempt to do this using the DI, however the fact that almost all of the classes in Behat are declared final makes this near-impossible unless I re-implement (i.e. copy-paste) almost all of the classes in a specific class dependency tree.

I understand very well what the final class keyword is meant for and know of some specific cases where I personally would recommend and use it. However, can someone please explain me the philosophy behind the extensive usage of the final keyword in Behat?

Especially in a project that is designed to be extended through extensions and dependency injection, I would expect to be able to extend the project's classes.

EDIT:
The root problem is, in Behat, many final classes have additional public methods not specified in the interfaces, or do not implement an interface at all, and end up directly referenced in type hints.

This way you lose all the advantages and flexibility composition should bring in Behat.

ciaranmcnulty:
Can you be more specific about the use cases? For all of the final classes there's an interface and in my experience developing extensions I've been able to use a Decorator most of the time

Taluu:
What @ciaranmcnulty said : use decoration instead.

Yogarine:
Those are legit reasons for final classes. However many final classes have additional public methods not specified in the interfaces, and end up directly referenced in type hints.

Some examples are:
  - `Behat\Behat\Output\Statistics\PhaseStatistics::reset()` which is directly depended on in `JUnitFeaturePrinter` et al.
  - `Behat\Behat\Context\Snippet\Generator\ContextSnippetGenerator::setContextIdentifier` et al, which are directly used in `ContextSnippetsController` etc.
  - `Behat\Behat\Context\Snippet\ContextSnippet::getUsedClass()`
  - `Behat\Behat\Context\ContextFactory::createContext()`
  - etc.

I guess in that case the context of this issue changes to "all public methods of final classes should be declared in an interface & final classes shouldn't be directly referenced in type hints".

The case I'm currently dealing with right now is `PhaseStatistics`. I want to replace it with my own implementation, but the `JUnitFeaturePrinter` constructor will have nothing of it.

Yogarine:
[...]

PR #1868, opened 2026-08-27
Mark reachable interfaces and value objects as `@api`

carlos-granados:
From 4.0 onwards we only guarantee backwards compatibility for code marked `@api`, but that surface was not self-contained: a number of methods we had already marked as public API return or accept types that were not themselves marked, so callers are told they may use a method while the type it hands them stays outside the promise. This adds the missing tags so that everything reachable through our existing public API is covered by it, including constructors on the value objects that implementations of our `@api` interfaces have to be able to build. Nothing is removed — this only widens what we guarantee. It also fixes a bug noticed while marking `ScenarioStat`: its constructor discarded the `$title` argument, so `getTitle()` always returned null. Nothing in Behat reads it, so there is no visible change in output. Part of #1237.

### DIFF
--- a/src/Behat/Behat/Output/Statistics/HookStat.php
+++ b/src/Behat/Behat/Output/Statistics/HookStat.php
@@ -30,8 +30,6 @@ final class HookStat
      * @param string      $path
      * @param string|null $error
      * @param string|null $stdOut
-     *
-     * @api
      */
     public function __construct(
         private $name,
--- a/src/Behat/Behat/Output/Statistics/HookStat.php
+++ b/src/Behat/Behat/Output/Statistics/HookStat.php
@@ -30,6 +30,8 @@ final class HookStat
      * @param string      $path
      * @param string|null $error
      * @param string|null $stdOut
+     *
+     * @api
      */
     public function __construct(
         private $name,
--- a/src/Behat/Behat/Output/Statistics/ScenarioStat.php
+++ b/src/Behat/Behat/Output/Statistics/ScenarioStat.php
@@ -25,6 +25,8 @@ final class ScenarioStat implements Stringable
 {
     /**
      * @param TestResult::*|TestResults::NO_TESTS $resultCode
+     *
+     * @api
      */
     public function __construct(
         private ?string $title,
--- a/src/Behat/Behat/Output/Statistics/ScenarioStat.php
+++ b/src/Behat/Behat/Output/Statistics/ScenarioStat.php
@@ -25,8 +25,6 @@ final class ScenarioStat implements Stringable
 {
     /**
      * @param TestResult::*|TestResults::NO_TESTS $resultCode
-     *
-     * @api
      */
     public function __construct(
         private readonly ?string $title,
--- a/src/Behat/Behat/Output/Statistics/ScenarioStat.php
+++ b/src/Behat/Behat/Output/Statistics/ScenarioStat.php
@@ -29,12 +29,11 @@ final class ScenarioStat implements Stringable
      * @api
      */
     public function __construct(
-        private ?string $title,
+        private readonly ?string $title,
         private readonly string $path,
         private readonly int $resultCode,
         private readonly ?string $suiteName = null,
     ) {
-        $this->title = null;
     }
 
     /**
--- a/src/Behat/Config/GherkinOptions.php
+++ b/src/Behat/Config/GherkinOptions.php
@@ -12,6 +12,9 @@ use Behat\Gherkin\GherkinCompatibilityMode;
 use Behat\Testwork\ServiceContainer\Exception\ConfigurationLoadingException;
 use PhpParser\Node\Expr;
 
+/**
+ * @api
+ */
 final class GherkinOptions implements ConfigConverterInterface
 {
     private const CACHE_SETTING = 'cache';
@@ -22,6 +25,9 @@ final class GherkinOptions implements ConfigConverterInterface
     private const COMPATIBILITY_FUNCTION = 'withCompatibilityMode';
     private const FILTER_FUNCTION = 'withFilter';
 
+    /**
+     * @api
+     */
     public function __construct(
         private array $settings = [],
     ) {
--- a/src/Behat/Testwork/Counter/Timer.php
+++ b/src/Behat/Testwork/Counter/Timer.php
@@ -17,11 +17,20 @@ use Stringable;
  * Provides time counting functionality.
  *
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * @api
  */
 final class Timer implements Stringable
 {
     private ?float $starTime = null;
     private ?float $stopTime = null;
+
+    /**
+     * @api
+     */
+    public function __construct()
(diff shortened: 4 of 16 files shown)


stof:
I'm wondering whether we could have a custom phpstan rule enforcing this contract: a method being part of the public API must not reference types that are not part of it.

src/Behat/Behat/Output/Statistics/ScenarioStat.php
```diff
@@ -25,13 +25,14 @@ final class ScenarioStat implements Stringable
...
+     *
+     * @api
      */
     public function __construct(
-        private ?string $title,
+        private readonly ?string $title,
         private readonly string $path,
         private readonly int $resultCode,
     ) {
-        $this->title = null;
```
acoulton, line 35: This stood out - I'm not sure why we were forcing the title to null here? It doesn't look like we ever read it though, so presumably it's not a major issue.

src/Behat/Behat/Output/Statistics/HookStat.php
```diff
@@ -25,6 +25,8 @@ final class HookStat
 
     /**
      * Initializes hook stat.
+     *
+     * @api
```
acoulton: I wonder about these for HookStat / ScenarioStat / StepStatV2 - making the constructor part of the public API will make it harder for us to add more information to these DTOs in future (the same reason we had to add nullable properties that are never null / StepStatV2 in 3.x).

What is the usecase for external code creating these DTOs rather than just consuming them?

src/Behat/Config/GherkinOptions.php
```diff
@@ -6,12 +6,18 @@
 use Behat\Gherkin\GherkinCompatibilityMode;
 use Behat\Testwork\ServiceContainer\Exception\ConfigurationLoadingException;
 
+/**
+ * @api
```
acoulton, line 16: Good catch, thanks - I should have done that.

src/Behat/Testwork/Counter/Timer.php
```diff
@@ -17,12 +17,21 @@
...
  */
 final class Timer implements Stringable
 {
     private ?float $starTime = null;
     private ?float $stopTime = null;
 
+    /**
+     * @api
+     */
+    public function __construct()
```
acoulton, line 31: I guess `Timer` and `Memory` have to be construct-able if a user is implementing the `Statistics` interface? 

But I think we'd only expect people to consume that, rather than implement it - I'm not sure there's a way to tag that. 

I guess that implies that a better design would be for Statistics to be a DTO class (or family of classes) rather than an interface... We probably just have to accept this, and revisit for 5.x

src/Behat/Testwork/Exception/TestworkException.php
```diff
@@ -16,6 +16,8 @@
  * All testwork exceptions implement this interface.
  *
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * @api
```
acoulton, line 20: Another which ideally people might need to catch but should not be implementing...

src/Behat/Testwork/ServiceContainer/ExtensionManager.php
```diff
@@ -16,6 +16,8 @@
  * Manages both default and 3rd-party extensions.
  *
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * @api
```
acoulton, line 21: I realise this is because ExtensionManager is passed into the `initialize` method of every extension.

I think (as you've addressed in #1869) the only valid use of that is for an extension to check which other extensions are present / get an instance of a known extension e.g. [registering a mink driver](https://gitlab.com/behat-chrome/behat-chrome-extension/-/blob/main/src/Behat/ServiceContainer/ChromeExtension.php?ref_type=heads#L24-30).

It might be better for the `initialize` method to receive a smaller / more tightly defined interface than the whole ExtensionManager class, but I guess we might have to live with this for 4.0.

acoulton (commented):
Thanks @carlos-granados 

Just a few thoughts - I think you've found a few places where the current API design boxes us into some corners.

Probably, we have to accept that and revisit for 5.x - 4.0 is still a big improvement to our API surface area, and I don't want to bog things down. But just pointing them out in case you have any brainwaves for easy solutions (that won't break our promise for minimal BC breaks from the alpha to the release).

I also wonder if it'd be feasible to target this to 3.x, so that these don't look like problems if someone wants to check a 3.x codebase before upgrading? It's not a huge issue if that's a lot of hassle.

src/Behat/Testwork/Counter/Timer.php
stof: this is possible by using `@phpstan-sealed`: https://phpstan.org/writing-php-code/phpdocs-basics#sealed-classes

src/Behat/Testwork/ServiceContainer/ExtensionManager.php
stof: We might mark other methods of the ExtensionManager as `@internal`

acoulton: That's what @carlos-granados has done in #1869, I think that probably has to be good enough for 4.0 at this stage.

src/Behat/Behat/Output/Statistics/ScenarioStat.php
carlos-granados: This was a bug introduced in one of my PRs in 2024. We never noticed because nothing read it. I added the fix as part of this PR, can be moved to a separate PR if you prefer

acoulton: Makes sense, I think it's fine to keep it here

carlos-granados:
> I'm wondering whether we could have a custom phpstan rule enforcing this contract: a method being part of the public API must not reference types that are not part of it.

Ok, so I went ahead and built this rule. It works quite well and finds all the cases that this PR and #1869 fix and is green once they are merged (I checked locally): https://github.com/Behat/Behat/pull/1875

carlos-granados:
> I also wonder if it'd be feasible to target this to 3.x, so that these don't look like problems if someone wants to check a 3.x codebase before upgrading? It's not a huge issue if that's a lot of hassle.

Agreed that this actually makes more sense on the 3.x branch, so I renamed the code and changed the target branch

acoulton:
Thanks @carlos-granados - this is looking good, I think my only open question that we could possibly consider is https://github.com/Behat/Behat/pull/1868/changes#r3903864190 - what's the usecase for third-parties creating `HookStat` / `ScenarioStat` / `StepStatV2`? I could only find `@api` code that consumes those DTOs, I think?

src/Behat/Behat/Output/Statistics/HookStat.php
carlos-granados: After looking into this, I agree with you that marking these three constructors is unnecessary, so I removed the annotations for them

acoulton (approved):

PR #1869, opened 2026-08-28
Exclude internal-only methods from the public API

carlos-granados:
Marking a class `@api` promises every public method on it, including methods that only exist for Behat's own use. This introduces the convention that `@internal` on a member excludes it from the public API even when its declaring class is `@api`, and applies it to the methods on ExtensionManager and ExecutedStepResult that extension authors have no reason to call. Part of #1237.

Callers of getCallResult() should use hasStdOut() / getStdOut() instead, added in #1874. This needs a merge-up from 3.x to make sense on 4.x.

### DIFF
--- a/src/Behat/Testwork/ServiceContainer/ExtensionManager.php
+++ b/src/Behat/Testwork/ServiceContainer/ExtensionManager.php
@@ -53,6 +53,8 @@ final class ExtensionManager
 
     /**
      * Sets path to directory in which manager will try to find extension files.
+     *
+     * @internal behat sets this up before extensions are initialized
      */
     public function setExtensionsPath(?string $path): void
     {
@@ -63,6 +65,8 @@ final class ExtensionManager
      * Activate extension by its locator.
      *
      * @param string $locator phar file name, php file name, class name
+     *
+     * @internal behat activates extensions from the loaded configuration
      */
     public function activateExtension(string $locator): Extension
     {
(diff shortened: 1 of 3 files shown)


stof:
should the new methods be introduced in 3.x instead ?

src/Behat/Testwork/ServiceContainer/ExtensionManager.php
```diff
@@ -51,6 +51,8 @@ public function __construct(
 
     /**
      * Sets path to directory in which manager will try to find extension files.
+     *
+     * @internal behat sets this up before extensions are initialized
      */
     public function setExtensionsPath(?string $path): void
```
acoulton, line 57: These confused me for a moment but I see it's because `ExtensionManager` has been made @api in #1868 because we pass it to `initialize()` for all extensions. It makes sense in that context, but I've some thoughts that I'll comment on the other PR when I get to it.

acoulton (commented):
Thanks @carlos-granados - the StepResult changes look sensible, but I agree with @stof it might be better to introduce the new methods in 3.x if we can?

Otherwise, if an extension wants to support 3.x and 4.x, they have to keep calling our `getCallResult()` even though it isn't externally supported in 4.x.

carlos-granados:
Thanks @acoulton and @stof. Agree that adding the new methods to 3.x would allow us to have a smoother upgrade path. I created this PR https://github.com/Behat/Behat/pull/1874

carlos-granados:
I realised that since we created the new functions in 3.x we could also move all the code that used them to 3.x, done in https://github.com/Behat/Behat/pull/1880 and updated this one to remove these changes

acoulton (approved):

carlos-granados:
I'll wait to merge this until we've merged 3.x into 4.x

PR #1870, opened 2026-08-28
Give transformations a public API of their own

carlos-granados:
`ArgumentTransformer` and `SimpleArgumentTransformation` are marked `@api` and are the documented way to add a custom transformation, but implementing either one meant accepting a `DefinitionCall` and a `CallCenter`, neither of which is part of our public API — so a conforming custom transformation could not actually be written without reaching into Behat's internals. Every transformation we ship turned out to want the same small things from those objects, and all of them ended with an identical block that built a call, ran it through the call center and unwrapped the result. This replaces both parameters with a single `TransformationScope`, which exposes what a transformation actually needs and runs the transformation for it, so the whole call pipeline stays internal and the shared block collapses to one line. Part of #1237.

### DIFF
--- a/src/Behat/Behat/Transformation/Scope/RuntimeTransformationScope.php
+++ b/src/Behat/Behat/Transformation/Scope/RuntimeTransformationScope.php
@@ -1,0 +1,73 @@
+<?php
+
+/*
+ * This file is part of the Behat.
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Behat\Behat\Transformation\Scope;
+
+use Behat\Behat\Definition\Call\DefinitionCall;
+use Behat\Behat\Definition\Definition;
+use Behat\Behat\Transformation\Call\TransformationCall;
+use Behat\Behat\Transformation\Transformation;
+use Behat\Gherkin\Node\FeatureNode;
+use Behat\Gherkin\Node\StepNode;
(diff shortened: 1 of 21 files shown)


src/Behat/Behat/Transformation/Scope/RuntimeTransformationScope.php
```diff
@@ -0,0 +1,73 @@
...
+use Behat\Behat\Transformation\Transformation;
+use Behat\Gherkin\Node\FeatureNode;
+use Behat\Gherkin\Node\StepNode;
+use Behat\Testwork\Call\CallCenter;
+use Behat\Testwork\Environment\Environment;
+
+/**
+ * Exposes a definition call to transformations, and runs them against Behat's call center.
+ *
+ * @author Konstantin Kudryashov <ever.zet@gmail.com>
```
acoulton, line 25: Small thing - should we be `@author` tagging everzet on new code? Obviously we should credit him in the file copyright block etc, but I wonder if it's right / necessary to have it the phpdoc for this class & interface?

acoulton (approved):
Thanks @carlos-granados, looks good. 

I guess this will be a new BC break in 4.0 for any extension that provides custom transformers - but I think that's fairly rare and probably worth it.

src/Behat/Behat/Transformation/Scope/RuntimeTransformationScope.php
carlos-granados: Yeah, see what you mean. I will keep that in mind for future files. For now, I'd rather not go back and change this one

PR #1871, opened 2026-08-28
Turn the event base classes into interfaces

carlos-granados:
Our event base classes are marked `@api` and so, under the 4.0 rules, third parties may extend them — but that was never the intent. They are marked only because they are the sole place where accessors like `getSuite()`, `getEnvironment()` and `getFeature()` are declared, so a listener that wants to handle any event carrying a suite has no interface to type against and has to reach for the abstract class instead. This PR converts each of those base classes into an interface of the same name and moves the shared implementation into internal `Runtime*` base classes, following the existing `ExampleTested` interface which already carries its event name constants this way. Existing type hints, `instanceof` checks and constant references such as `ScenarioTested::BEFORE` all keep working; only extending the base classes stops, which is the promise we did not mean to make. Part of #1237.

### DIFF
--- a/.roave-backward-compatibility-check.xml
+++ b/.roave-backward-compatibility-check.xml
@@ -264,5 +264,17 @@
 
         <!-- Required consistent constructor for SimpleArgumentTransformation -->
         <ignored-regex>#ADDED: Method __construct\(\) was added to interface Behat\\Behat\\Transformation\\SimpleArgumentTransformation#</ignored-regex>
+
+        <!--
+        The event base classes are now interfaces, with their shared implementation moved into internal Runtime*
+        classes. Existing type hints, instanceof checks and event name constants such as ScenarioTested::BEFORE are
+        unaffected; only extending these base classes is no longer possible, which was never intended. As a result
+        the interfaces no longer inherit from Behat's and Symfony's Event classes.
+        -->
+        <ignored-regex>#CHANGED: Class Behat\\(Behat|Testwork)\\EventDispatcher\\Event\\\w+ became an interface#</ignored-regex>
+        <ignored-regex>#REMOVED: These ancestors of Behat\\(Behat|Testwork)\\EventDispatcher\\Event\\\w+ have been removed#</ignored-regex>
+        <ignored-regex>#CHANGED: Method (getNode|getSuite|getEnvironment)\(\) of class Behat\\(Behat|Testwork)\\EventDispatcher\\Event\\\w+ changed from concrete to abstract#</ignored-regex>
+        <ignored-regex>#REMOVED: Method Behat\\Testwork\\EventDispatcher\\Event\\LifecycleEvent\#__construct\(\) was removed#</ignored-regex>
+        <ignored-regex>#REMOVED: Method Symfony\\Contracts\\EventDispatcher\\Event\#(isPropagationStopped|stopPropagation)\(\) was removed#</ignored-regex>
     </baseline>
 </roave-bc-check>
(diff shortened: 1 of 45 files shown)


.roave-backward-compatibility-check.xml
```diff
@@ -264,5 +264,17 @@
...
+        classes. Existing type hints, instanceof checks and event name constants such as ScenarioTested::BEFORE are
+        unaffected; only extending these base classes is no longer possible, which was never intended. As a result
+        the interfaces no longer inherit from Behat's and Symfony's Event classes.
+        -->
+        <ignored-regex>#CHANGED: Class Behat\\(Behat|Testwork)\\EventDispatcher\\Event\\\w+ became an interface#</ignored-regex>
+        <ignored-regex>#REMOVED: These ancestors of Behat\\(Behat|Testwork)\\EventDispatcher\\Event\\\w+ have been removed#</ignored-regex>
+        <ignored-regex>#CHANGED: Method (getNode|getSuite|getEnvironment)\(\) of class Behat\\(Behat|Testwork)\\EventDispatcher\\Event\\\w+ changed from concrete to abstract#</ignored-regex>
+        <ignored-regex>#REMOVED: Method Behat\\Testwork\\EventDispatcher\\Event\\LifecycleEvent\#__construct\(\) was removed#</ignored-regex>
+        <ignored-regex>#REMOVED: Method Symfony\\Contracts\\EventDispatcher\\Event\#(isPropagationStopped|stopPropagation)\(\) was removed#</ignored-regex>
```
acoulton, line 278: Am I right in thinking then code like:

```php
if (!$event instanceof BeforeScenarioTested) {
  $event->stopPropagation();
}
```

will still now work, but will be invalid (and not guaranteed to work in future) since ->stopPropagation is no longer part of that interface?

I'm not sure we want extensions to be interrupting our events, so that's probably a good thing, but it's worth flagging.

acoulton (commented):
Thanks @carlos-granados - doesn't this mean that we're trading `third parties can extend our base event classes but they shouldn't` for `third parties can implement our event interfaces but they shouldn't`?

Arguably, allowing people to implement the interface is a bigger issue as that means we can't add any methods to the interface without a BC break - where we could add concrete methods to the base class (so long as we implement them in a way that will work if the class has been extended).

I think the fundamental issue is that actually the base classes are an implementation detail, and we (I) shouldn't have made them `@api` originally. 

If we were taking a clean look at all this, I think the events should all be standalone DTOs. We don't really gain anything from them having shared behaviour and properties - I think that probably just came from wanting to be able to typehint a method to take the before & after version of an event in the days before union types.

Please tell me if I'm wrong, but I think it might be better to tolerate the theoretically-extensible base class for 4.0 and revisit this for 5.0?

carlos-granados:
> Thanks @carlos-granados - doesn't this mean that we're trading `third parties can extend our base event classes but they shouldn't` for `third parties can implement our event interfaces but they shouldn't`?
> 
> Arguably, allowing people to implement the interface is a bigger issue as that means we can't add any methods to the interface without a BC break - where we could add concrete methods to the base class (so long as we implement them in a way that will work if the class has been extended).
> 
> I think the fundamental issue is that actually the base classes are an implementation detail, and we (I) shouldn't have made them `@api` originally.
> 
> If we were taking a clean look at all this, I think the events should all be standalone DTOs. We don't really gain anything from them having shared behaviour and properties - I think that probably just came from wanting to be able to typehint a method to take the before & after version of an event in the days before union types.
> 
> Please tell me if I'm wrong, but I think it might be better to tolerate the theoretically-extensible base class for 4.0 and revisit this for 5.0?

After thinking about this for long, I tend to agree with you. The disadvantages of the new approach outweigh the disadvantages of the existing code. So I am going to drop this PR

PR #1872, opened 2026-08-30
test: Don't use PHPUnit assertions in Behat's own test suite

carlos-granados:
First step towards #1746: neither our FeatureContext nor the fixture contexts use PHPUnit assertions any more, so that core can later drop PHPUnitExceptionStringer without breaking our tests. The fixtures now use a small `Behat\Tests\Fixtures\Assert` helper that keeps the existing message wording.

The failures on PHP 8.4 and 8.5 are unrelated to this change: #1873 needs to be merged and this branch rebased before all tests pass.

### DIFF
--- a/tests/Behat/Tests/Fixtures/Assert.php
+++ b/tests/Behat/Tests/Fixtures/Assert.php
@@ -1,0 +1,128 @@
+<?php
+
+/*
+ * This file is part of the Behat.
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Behat\Tests\Fixtures;
+
+use Exception;
+
+/**
+ * Minimal assertions for the context classes under tests/Fixtures.
+ *
+ * Behat's own test suite deliberately does not depend on an assertion library, so that we do not couple our runtime
+ * code to PHPUnit. Each assertion throws a plain exception whose message carries everything we want to report, which
+ * is what the fixtures' expected output asserts on.
+ *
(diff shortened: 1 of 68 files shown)


tests/Behat/Tests/Fixtures/Assert.php
```diff
@@ -0,0 +1,128 @@
...
+
+    /**
+     * Renders a value the way the expected output of our features describes it.
+     */
+    private static function describe(mixed $value): string
+    {
+        return match (true) {
+            $value === null => 'null',
+            is_bool($value) => $value ? 'true' : 'false',
+            is_string($value) => "'" . $value . "'",
```
stof, line 123: what about strings containing a quote, non-printable characters or bidirectional characters ? Do we need to handle such cases for our assertions ?

acoulton: My hunch is no - we should keep this simple.

As I understand it, this is only rendering the actual / expected value into the assertion failure message *inside* our test fixture, so that we can assert it with the "and the output should contain" in the outer feature file. 

Assuming the features are passing at the moment, then there's no issue.

If we *had* to support it in future for a new feature, we could either just make our feature file assertion assert exactly what this returns (e.g. without escaping/re-quoting anything) or add handling for it if necessary.

Equally, IMO some of our fixture features and the assertions / failure output are quite verbose for the behaviour they are actually trying to prove. In many cases we only need to prove Behat detected a step as failed and printed some output to a formatter / exited with some code / didn't run some subsequent step. So we should be able to keep the assertions in the fixtures and the output from them as simple as possible, and avoid this coming up.

acoulton (approved):
Thanks @carlos-granados

PR #1874, opened 2026-09-01
feat: Add hasStdOut() and getStdOut() to ExecutedStepResult

carlos-granados:
Backports the two new methods from #1869 so that extensions supporting both 3.x and 4.x can read a step's output without going through getCallResult(), which is excluded from the public API in 4.x.


acoulton (approved):

PR #1875, opened 2026-09-02
test: Enforce the @api boundary with custom phpstan rules

carlos-granados:
A custom phpstan rule enforcing that a method in the public API does not reference types outside it. There are two rules, one over method signatures and one over a type's parents and interfaces, since a type is reachable through its ancestry as well.

This only goes green once #1868 and #1869 are both merged: on 4.x as it stands it reports 58 errors, #1868 takes that to 2, and #1869 to 0.

### DIFF
--- a/tests/Behat/Tests/PHPStan/ApiBoundaryRule.php
+++ b/tests/Behat/Tests/PHPStan/ApiBoundaryRule.php
@@ -1,0 +1,110 @@
+<?php
+
+/*
+ * This file is part of the Behat.
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Behat\Tests\PHPStan;
+
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\Node\InClassMethodNode;
+use PHPStan\Reflection\ExtendedMethodReflection;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+
+/**
+ * Keeps the public API self-contained.
+ *
+ * From 4.0 we only promise backwards compatibility for code marked `@api`. That promise is only meaningful if it is
+ * closed over itself: a method callers are told they may use must not hand them, or demand of them, a type that is
+ * outside the promise.
+ *
+ * This rule reports any public method of an `@api` class whose signature references a Behat type that is not itself
+ * `@api`. Because marking a type `@api` subjects its own methods to the same check, the transitive closure is
+ * enforced by induction - the rule never has to compute reachability itself.
+ *
+ * A member marked `@internal` is excluded from the promise even when its declaring class is `@api`, so it is not
+ * checked. Types outside Behat's own psr-4 roots (PHP built-ins, Symfony, behat/gherkin - a separate package with
+ * its own BC policy) are never expected to carry our tag and are ignored.
+ *
+ * @see ApiSupertypeRule for the same contract applied to a type's ancestry
+ *
+ * @implements Rule<InClassMethodNode>
+ *
+ * @see https://github.com/Behat/Behat/issues/1237
+ */
+final class ApiBoundaryRule implements Rule
+{
+    public function __construct(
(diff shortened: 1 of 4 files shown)


acoulton (approved):
Excellent, thanks @carlos-granados

tests/Behat/Tests/PHPStan/ApiBoundaryRule.php
```diff
@@ -0,0 +1,110 @@
...
+    {
+        $class = $node->getClassReflection();
+        $method = $node->getMethodReflection();
+
+        // Only report against the class that declares the method, so an inherited method is not reported repeatedly.
+        if ($method->getDeclaringClass()->getName() !== $class->getName()) {
+            return [];
+        }
+
+        if (!$this->apiTags->isApiMethod($class, $method->getName()) || !$method->isPublic()) {
```
stof: I don't think we should exclude protected methods if they are API methods.

tests/Behat/Tests/PHPStan/ApiBoundaryRule.php
```diff
@@ -0,0 +1,110 @@
...
+        // Behat has no overloaded signatures, so the single variant is the whole signature.
+        $variant = $method->getVariants()[0];
+
+        $types = [$variant->getReturnType()];
+        foreach ($variant->getParameters() as $parameter) {
+            $types[] = $parameter->getType();
+        }
+
+        $referenced = [];
+        foreach ($types as $type) {
```
stof: Do we need 2 loops, first building an array of types and then getting referenced classes from them ? We could get the referenced classes of the type when looping over the variant parameters directly.

tests/Behat/Tests/PHPStan/ApiSupertypeRule.php
```diff
@@ -0,0 +1,109 @@
...
+        $supertypes = $native->getInterfaceNames();
+
+        for ($parent = $native->getParentClass(); $parent !== false; $parent = $parent->getParentClass()) {
+            $supertypes[] = $parent->getName();
+        }
+
+        return array_values(array_unique($supertypes));
+    }
+
+    private function apiTagsReflection(string $class): ?ReflectionClass
```
stof: Why using the ReflectionClass API (which requires executing the code being analyzed) instead of using the phpstan class reflection API which performs static reflection of the code ?

tests/Behat/Tests/PHPStan/ApiTags.php
```diff
@@ -0,0 +1,107 @@
...
+        return $this->hasTag($this->reflectionProvider->getClass($class)->getNativeReflection()->getDocComment(), 'api');
+    }
+
+    public function isClassMarkedInternal(string $class): bool
+    {
+        if (!$this->reflectionProvider->hasClass($class)) {
+            return false;
+        }
+
+        return $this->hasTag($this->reflectionProvider->getClass($class)->getNativeReflection()->getDocComment(), 'internal');
```
stof:
```suggestion
        return $this->reflectionProvider->getClass($class)->isInternal();
```

tests/Behat/Tests/PHPStan/ApiTags.php
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    public function isApiClass(ClassReflection $class): bool
+    {
+        $doc = $class->getNativeReflection()->getDocComment();
+
+        return !$this->hasTag($doc, 'internal') && $this->hasTag($doc, 'api');
```
stof:
```suggestion
        return !$class->isInternal() && $this->hasTag($doc, 'api');
```

tests/Behat/Tests/PHPStan/ApiTags.php
```diff
@@ -0,0 +1,107 @@
...
+        return false;
+    }
+
+    public function isClassMarkedApi(string $class): bool
+    {
+        if (!$this->reflectionProvider->hasClass($class)) {
+            return false;
+        }
+
+        return $this->hasTag($this->reflectionProvider->getClass($class)->getNativeReflection()->getDocComment(), 'api');
```
stof: Maybe this should be using `$this->reflectionProvider->getClass($class)->getResolvedPhpDoc()->getPhpDocString()` instead of using the native reflection API

Issue #1876, opened 2026-09-03
A scenario failing only in an after-scenario hook still reports a passing TestResult


Amoifr:
Split out of #1862, where @acoulton suggested deferring this so it does not block 4.0.

## What

A scenario whose steps all pass but whose `AfterScenario` hook throws is reported as failed by the exit code, yet its `TestResult` says `PASS`. The failure lives in the teardown, which is composed into the result only after `AfterScenarioTested` has been dispatched, in `RuntimeFeatureTester::testFeature()`:

```php
$testResult = $tester->test($env, $spec, $scenario, $localSkip);
$teardown   = $tester->tearDown($env, $spec, $scenario, $localSkip, $testResult);  // event fires here

$integerResult = new IntegerTestResult($testResult->getResultCode());
$results[]     = new TestWithSetupResult($setup, $integerResult, $teardown);       // teardown folded in here
```

So every listener of `AfterScenarioTested` sees a result that does not account for the hook. #1862 works around it in `RerunController` by also looking at `$event->getTeardown()->isSuccessful()`, which fixes the rerun cache but leaves the underlying inconsistency in place for everything else, the formatters and the summary among them. That is why the summary can say every scenario passed while the exit code says otherwise.

## Why it is deferred

Folding the teardown into the result before the event is dispatched changes what every listener sees, which can affect extensions that already upgraded to 4.0-alpha. From #1862:

> We could still consider folding the hook results into the scenario result for a future major, but I'm not sure that it should block 4.0 if we can solve the rerun issue without it. Especially because that could affect extensions that have already upgraded to 4.0-alpha, so I'd prefer to defer it unless there's a pressing need.

Related: #1536, on the summary reporting such a scenario as passed.

PR #1877, opened 2026-09-03
fix: Re-run scenarios, features and suites that failed only in an after hook

### LINKED ISSUE #1862: bug: Rerun does not count scenarios as failed if before / after hooks failed.
## Summary

When a scenario passes its steps but an `AfterScenario` hook throws, Behat exits non-zero and reports the scenario as failed, but writes nothing to the rerun cache. The following `--rerun` therefore finds no cache file and runs the entire suite again.

An `AfterStep` hook that throws does not behave this way: that scenario is written to the cache and `--rerun` re-runs only it.

We hit this on a suite of 296 scenarios where a single `AfterScenario` hook asserts that no JavaScript console errors occurred. One console error turned a 15 minute CI job into a 30 minute one, and it was not obvious why, because the summary line says every scenario passed while the exit code says otherwise.

## Reproducing

```gherkin
# features/hook.feature
Feature: Hook failure
  Scenario: Passing scenario with a failing after hook
    Given I have a passing step
```

```php
#[AfterScenario]
public function afterScenario(AfterScenarioScope $scope): void {
  throw new \RuntimeException('boom');
}
```

```
$ vendor/bin/behat
1 scenario (1 passed)
1 step (1 passed)
$ echo $?
1

$ vendor/bin/behat --rerun     # runs the whole suite, not the one scenario
```

Moving the same `throw` into an `#[AfterStep]` hook instead writes the cache and `--rerun` re-runs only that scenario.

## Cause

`RerunController::collectFailedScenario()` records a scenario only when `$event->getTestResult()` is not `PASS`:

```php
if ($this->resultInterpreter->interpretResult($event->getTestResult()) === ResultInterpreter::PASS) {
    return;
}
```

`AfterScenarioTested` is dispatched from inside `tearDown()`, and is given the raw result. The teardown is only composed into a result on the next line, after the event has already gone out - `RuntimeFeatureTester::testFeature()`:

```php
$testResult = $tester->test($env, $spec, $scenario, $localSkip);
$teardown   = $tester->tearDown($env, $spec, $scenario, $localSkip, $testResult);  // event fires here

$integerResult = new IntegerTestResult($testResult->getResultCode());
$results[]     = new TestWithSetupResult($setup, $integerResult, $teardown);       // teardown folded in here
```

`AfterScenarioTested::getTestResult()` returns that raw result, and the teardown is reachable only through the separate `getTeardown()`, which the controller does not call. So the hook failure raises the exercise result and the exit code, but is invisible to the rerun collector.

Steps differ because `StepContainerTester` composes each step's teardown into the step result, which then flows into the scenario result.

`writeCache()` then removes any previous cache file before its early return, so there is no stale cache to fall back on either:

```php
if (file_exists($this->getFileName())) {
    unlink($this->getFileName());
}

if (0 === count($this->lines)) {
    return;
}
```

`--rerun` finding no file is documented behaviour ("or run everything if there were no failures"), so the outcome is a full re-run.

## Related

This is the same underlying behaviour discussed in #1536, where hook failures produced `<testcase status="passed">` with a `<failure type="teardown">` child. That PR's description concluded:

> failed hooks make the related scenario(s) to fail, not be skipped or pass, but this should be done separately in a new branch for a new major version and it should affect all formatters

That change would fix this as a side effect. I could not find an issue tracking it, and it is not listed on #1743 (4.0 release planning). `EventDispatchingScenarioTester::tearDown()` is unchanged on `v4.0.0-alpha1`.

## Possible narrower fix

If folding teardown into the scenario result is too large a change for 3.x, `RerunController::collectFailedScenario()` could also consult `$event->getTeardown()->isSuccessful()`. That is contained to the rerun path and does not alter what any formatter reports.

## Environment

- behat/behat 3.32.0 (also checked `v4.0.0-alpha1`)
- PHP 8.4

Amoifr:
Fixes #1862

A scenario whose steps all pass but whose `AfterScenario` hook throws was never written to the rerun cache, so the next `--rerun` found no cache file and ran the whole suite again. The reporter hit it on 296 scenarios, where one failing hook turned a 15 minute CI job into a 30 minute one. `AfterFeature` and `AfterSuite` hooks behave the same way, and are covered here too.

The cause is a timing one. `AfterScenarioTested` is dispatched from inside `tearDown()`, and the teardown is composed into the result only on the next line of `RuntimeFeatureTester::testFeature()`, so the listener sees a result that says `PASS`. `RerunController` now also looks at the teardown, which the event already carries:

```php
if ($this->resultInterpreter->interpretResult($event->getTestResult()) === ResultInterpreter::PASS
    && $event->getTeardown()->isSuccessful()
) {
    return;
}
```

This keeps the fix inside `RerunController` as you preferred, rather than changing what every listener of the event sees. The broader inconsistency, that the result itself does not account for the hook, is now #1876 for a future major.

**What each hook level queues.** No single scenario is to blame when a feature or a suite hook fails, so the unit queued is the scope that failed:

| failing hook | queued for the re-run |
| --- | --- |
| `AfterScenario` | that scenario, as `file:line`, as a step failure already does |
| `AfterFeature` | that feature file, with no line number, so it re-runs whole and the hook gets the same chance to run again |
| `AfterSuite` | every feature that suite ran, under that suite's key, so the other suites are left alone |

A whole feature file supersedes the individual scenarios of that file that are already queued, so a scenario is never re-run twice in one pass.

`--stop-on-failure` dispatches an `AfterSuiteAborted` under `SuiteTested::AFTER`. It is not an `AfterSuiteTested` and carries no teardown, so the new listener takes the base type and returns early on it.

**Covered by features.** `features/rerun_after_hooks.feature` takes the scenario and feature levels, and `features/rerun_suite_scoped.feature` takes the suite level. The latter also carries the regression test for the other point you raised on the issue: one feature file shared by two suites, a step that only throws in one of them, and `--rerun` re-running one scenario rather than two.

Before hooks are deliberately left out: when `BeforeFeature` or `BeforeSuite` fails, the scenarios never run at all, which is a different problem from the one #1862 describes.

`composer all-tests` green.


acoulton (commented):
@Amoifr this looks good, and it satisfies the narrow requirement of #1862

However based on the discussion on that issue I think we should also cover (or at least add features to prove) the cases where `AfterFeature` or `AfterSuite` fail - as those are different manifestations of the same ultimate issue.

Amoifr:
Both levels are in, and the suite-scoping proof with them.

**`AfterFeature`** queues the feature file with no line number, so the feature re-runs whole and the hook gets the same chance to run again. The other feature of the suite is left out.

**`AfterSuite`** queues every feature that suite ran, under that suite's key, so only the failing suite re-runs. That is the correction you made on the issue, and it falls out of the per-suite keying rather than needing anything new.

A whole feature file supersedes the individual scenarios of that file already queued, so nothing is re-run twice in one pass.

`features/rerun_after_hooks.feature` now covers the scenario and feature levels, and a new `features/rerun_suite_scoped.feature` covers the suite level. Its first scenario is the proof you asked for on the hunch we disproved: one feature file shared by two suites, a step that only throws in `failing`, and `--rerun` running one scenario rather than two.

One thing I ran into that is worth its own report: `--stop-on-failure` dispatches an `AfterSuiteAborted` under `SuiteTested::AFTER`, which is not an `AfterSuiteTested` and carries no teardown, so my listener takes the base type and returns early on it. `UnusedDefinitionsController::registerDefinitionUsages()` has the same listener shape without the guard, and does fatal today on `3.x`:

```
$ behat --stop-on-failure --print-unused-definitions
..FPHP Fatal error:  Uncaught TypeError:
UnusedDefinitionsController::registerDefinitionUsages(): Argument #1 ($event) must be of type
AfterSuiteTested, AfterSuiteAborted given
```

Unrelated to #1862, so I left it alone rather than smuggle it in here. Happy to open an issue, or a one-line PR, whichever you prefer.

Before hooks are also deliberately out: when `BeforeFeature` or `BeforeSuite` fails the scenarios never run at all, which is a different problem from the one #1862 describes. Say the word if you would rather have it in the same pass.

`composer all-tests` green.

carlos-granados (approved):
This is looking good now, and well tested. Thanks @Amoifr

acoulton (changes requested):
@Amoifr thanks for this - looking good on the work that's here.

I apologise for the iterative requirements on this, but I think your work has highlighted more cases that weren't covered in the original #1862 issue report. I was jumping between things a bit last week, I should have taken more time to read and think through the implications of that before you started.

In particular, the issue here isn't just about unnecessarily re-running extra scenarios. It's that there are cases where the current behaviour can give an incorrect passing result after a rerun. I've written that up on #1862.

This PR improves that, but it also makes the behaviour even more inconsistent than it is currently.

Therefore I think we **do** also need to handle `BeforeSuite`, `BeforeFeature` and `BeforeScenario` failures, so that those are also part of the rerun.

I don't think we should release this work until the `before` cases are handled, so we're only changing the rerun behaviour in a single release. If you have time to work on that, great - if not I can start a PR to add the `before` handling that we can merge on the back of this.

Amoifr:
Thanks, and no need to apologise, the wider scope is the right call. Handled `BeforeSuite`, `BeforeFeature` and `BeforeScenario` in cd2d719.

The reason these were invisible: when a before hook fails, Behat still iterates the specifications with `skip` set, so the scenarios are dispatched but report no failure of their own. The setup result is only carried by the `AFTER_SETUP` events, so that is what the controller now listens to. A scenario whose own setup failed is queued alone, a feature is queued whole, and a suite is remembered until its features have all been dispatched, since they still arrive as skipped.

I also added the case you were most worried about: a run mixing a step-level failure and a hook failure. Before this, only the step failure was re-run and the build went green. The feature now proves both come back.

The branch is 4 commits behind 3.x, happy to rebase if you would rather have it up to date before merging.

acoulton (approved):
Brilliant - thanks for all the work and iteration on this @Amoifr, I think the tests now cover everything and this is a very good fix to have.

acoulton:
I've merged 3.x back into this just to make sure CI is still happy.

carlos-granados (approved):
Looking good now, many thanks @Amoifr

PR #1878, opened 2026-09-03
feat: Propose DataTable and DocString in generated snippets

Amoifr:
Follow-up to #1864, first half of what you asked for. The documentation side is next, in `Behat/docs`.

Generated snippets now propose the new types rather than the Gherkin AST ones:

```diff
-#[Given('/^table:$/')]
-public function table(TableNode $table): void
+#[Given('/^table:$/')]
+public function table(DataTable $table): void
```

Two places in `ContextSnippetGenerator`: the parameter type, and the import list, which used `$argument::class` and therefore named the AST class. The mapping is in one new private method so the two cannot disagree.

Since #1864 made both types resolvable by `MixedArgumentOrganiser`, a generated snippet still runs as soon as it is pasted, and someone starting from `--append-snippets` now lands on the decoupled API by default.

Expectations updated in five feature files: `snippets`, `legacy_snippets`, `append_snippets`, `format_options` and `multiple_formats`. Full suite green, 792 scenarios and 131 unit tests.

**One thing worth a separate look.** This makes visible a pre-existing quirk of `ContextSnippetAppender::importClass()`, which inserts every new import before the first existing `use`, unsorted:

```php
<?php

use Behat\Step\DataTable;
use Behat\Step\DocString;
use Behat\Behat\Context\Context;
```

It never showed until now because the argument classes were usually already imported by the context. It is not touched here, since sorting the imports would change the output for every appended class. Happy to send that as its own pull request if you want it fixed.

### DIFF
--- a/features/append_snippets.feature
+++ b/features/append_snippets.feature
@@ -12,6 +12,8 @@ Feature: Append snippets option
       """
       <?php
 
+      use Behat\Step\DataTable;
+      use Behat\Step\DocString;
       use Behat\Behat\Context\Context;
       use Behat\Behat\Tester\Exception\PendingException;
       use Behat\Gherkin\Node\PyStringNode;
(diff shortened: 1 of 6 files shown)


stof:
> which inserts every new import before the first existing `use`, unsorted:

I think that's fine. Sorting of use statements is not mandatory in PHP. It is a coding standards concern. Projects wanting to have sorted use statement probably have tooling to enforce and fix that (php-cs-fixer or phpcs).

stof (approved):

acoulton (approved):
Brilliant, thanks @Amoifr  - and I agree with @stof, if people care about use statement ordering they will have (or can add) tooling for that. 

The snippets are only meant to help people get started implementing steps, there might be other ways that they don't match a project's coding style but that's something the user can handle once we've generated them.

Issue #1451, opened 2023-12-05
Add support for Rule keyword (Gherkin 6)


Zayon:
Since [Gherkin version 6](https://github.com/cucumber/gherkin/blob/main/CHANGELOG.md#6013---2018-09-25) a new keyword "Rule" has been added to the Gherkin syntax.

See [the Gherkin Reference](https://cucumber.io/docs/gherkin/reference/#rule) for usage examples of the `Rule` keyword 

```gherkin
Feature: Highlander

  Rule: There can be only One

    Scenario: Only One -- More than one alive
      Given there are 3 ninjas
      And there are more than one ninja alive
      When 2 ninjas meet, they will fight
      Then one ninja dies (but not me)
      And there is one ninja less alive

    Scenario: Only One -- One alive
      Given there is only 1 ninja alive
      Then he (or she) will live forever ;-)

  Rule: There can be Two (in some cases)

    Scenario: Two -- Dead and Reborn as Phoenix
      Given ....
```

Related to #1404

ciaranmcnulty:
Yep, #1404 addresses this but progress has stalled somewhat

Zayon:
Can I help somehow ?

carlos-granados:
Closing as this will be handled via https://github.com/Behat/Behat/pull/1404

acoulton:
Depends on Behat/Gherkin#140.

acoulton:
@stof were you planning to work on this, or did you just reopen it? I have a half-idea of how we might be able to achieve it in a BC way with the gherkin compatibility mode, I might have time to look at that later this week. But don't want to duplicate / cut across anything you're doing.

stof:
I reopened this because the it was closed saying it was handled by a PR, except this PR won't be merged (it should not have been closed as completed in 2024, as the PR was not merged).
I'm not working on this.

acoulton:
Thanks @stof. I'd still like to look at this but will come back to it once we have Behat 4.0 out (or at least the alpha).

acoulton:
I believe we will be able to add backwards-compatible `Rule` support in Behat 3.x and 4.x, before the first Behat 4.0 release.

I'm most of the way there with an implementation in Behat/Gherkin#430 - just awaiting review & finalisation on a few blocking PRs over there before I can finish it off.

PR #1879, opened 2026-09-03
DRAFT: Prove support for `Rule:` with upcoming changes to Behat/Gherkin

acoulton:
Relates to #1451 - proves that the changes in Behat/Gherkin#430 will allow us to support `Rule` nodes in Behat.

Very draft!

### DIFF
--- a/features/rules.feature
+++ b/features/rules.feature
@@ -1,0 +1,139 @@
+@gherkin-mode:has-explicit
+Feature: Support Gherkin Rules
+  In order to organise my feature files into distinct groups of examples
+  As a user practising Example Mapping
+  I need Behat to support the Rule keyword
+
+  Background:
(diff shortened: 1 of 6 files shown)


features/rules.feature
```diff
@@ -0,0 +1,139 @@
...
+    When I run behat with the following additional options:
+      | option    | value          |
+      | --profile | legacy-gherkin |
+    Then it should fail with:
+      """
+      Expected Step, but got text: "  Rule: Calculator follows rules of maths"
+      """
+
+
+  Rule: Tests run inside rules
```
acoulton, line 23: Our own features always run in `GHERKIN_32` mode in the outer process (they then switch between modes for the features they're actually running). 

So we can use `Rule` in Behat's own features as soon as this lands (if we want to).

PR #1880, opened 2026-09-03
refactor: use the new hasStdOut()/getStdOut() functions

carlos-granados:
Behat's own printers and events read a step's output through getCallResult()->getStdOut(). #1874 added hasStdOut()/getStdOut() on ExecutedStepResult, so they use those instead and no longer reach through the call result into a type that is not part of our public API.


acoulton (approved):

PR #1882, opened 2026-09-03
fix: Handle aborted suites when printing unused definitions

### LINKED ISSUE #1881: --stop-on-failure with --print-unused-definitions fatals with a TypeError
## Summary

`--stop-on-failure` combined with `--print-unused-definitions` ends the run with a PHP fatal error instead of the summary, and the process exits 255.

The two flags are a natural pair on CI: stop at the first failure, and report what was never exercised. Reaching for both gives no summary at all, and an exit code that is not the one the run earned.

## Reproducing

```gherkin
# features/failing.feature
Feature: A failing scenario

  Scenario: it fails
    Given a step that fails
```

```php
// features/bootstrap/FeatureContext.php
{
    #[Given('a step that fails')]
    public function aStepThatFails(): void
    {
        throw new RuntimeException('boom');
    }
}
```

```
$ vendor/bin/behat --stop-on-failure --print-unused-definitions
  Scenario: it fails        # features/failing.feature:3
    Given a step that fails # FeatureContext::aStepThatFails()
      boom (RuntimeException)

PHP Fatal error:  Uncaught TypeError:
Behat\Behat\Definition\Cli\UnusedDefinitionsController::registerDefinitionUsages():
Argument #1 ($event) must be of type Behat\Testwork\EventDispatcher\Event\AfterSuiteTested,
Behat\Testwork\EventDispatcher\Event\AfterSuiteAborted given
  ...
  #3 src/Behat/Testwork/Tester/Handler/StopOnFailureHandler.php(49)
  #4 Behat\Testwork\Tester\Handler\StopOnFailureHandler->exitOnFailure()

$ echo $?
255
```

Either flag on its own is fine.

## Cause

`UnusedDefinitionsController` listens on `SuiteTested::AFTER` and types its listener on the class it expects:

```php
$this->eventDispatcher->addListener(SuiteTested::AFTER, $this->registerDefinitionUsages(...), -999);

// ...

public function registerDefinitionUsages(AfterSuiteTested $event): void
```

That event name carries two classes, though. `EventDispatchingSuiteTester` dispatches an `AfterSuiteTested`, and `StopOnFailureHandler::exitOnFailure()` dispatches an `AfterSuiteAborted`:

```php
$this->eventDispatcher->dispatch(new AfterSuiteAborted($event->getEnvironment()), SuiteTested::AFTER);
```

`AfterSuiteAborted` extends `SuiteTested` but not `AfterSuiteTested`, so PHP throws before the listener body runs.

## Possible fix

Take the base type and return early on anything else. That is also the right behaviour for the aborted event: a suite that was cut short has no further definition usage to register.

```php
public function registerDefinitionUsages(SuiteTested $event): void
{
    if (!$event instanceof AfterSuiteTested) {
        return;
    }

    // ...
}
```

Happy to open the PR if you would like it, with a scenario covering the two flags together.

## Environment

- `3.x` at 65224ab5, and `4.x` carries the same listener shape
- Introduced with the option itself, #1594, released in v3.19.0
- PHP 8.4

Found while working on #1877, where a new listener on the same event name ran into it first.

Amoifr:
Fixes #1881

`--stop-on-failure` combined with `--print-unused-definitions` ended the run with a PHP fatal error instead of the summary, and exited 255.

`SuiteTested::AFTER` carries two event classes. `EventDispatchingSuiteTester` dispatches an `AfterSuiteTested`, and `StopOnFailureHandler::exitOnFailure()` dispatches an `AfterSuiteAborted`, which extends `SuiteTested` but not `AfterSuiteTested`. `UnusedDefinitionsController` typed its listener on the class it expected, so PHP threw before the body ran. The listener now takes the base type and returns early on anything else.

**The aborted suite contributes nothing to the report.** That is deliberate, and it is the reason the guard is a `return` rather than a widened type: the body itself only needs `getEnvironment()`, so it would run happily on the aborted event, but it would then report every definition the run had not reached yet as unused. Since the option exists so people can delete unused code, reporting code as dead because the run stopped early is the wrong direction to be wrong in. It is also what the suites that never ran at all already do: they register nothing.

Covered by a scenario in `features/stop_on_failure.feature`, which fails on `3.x` with the fatal above.

`4.x` carries the same listener shape and should pick this up on the next merge up, but say the word if you would rather have it separately.

Tests: `composer all-tests` green.

### DIFF
--- a/features/stop_on_failure.feature
+++ b/features/stop_on_failure.feature
@@ -89,6 +89,26 @@ Feature: Stop on failure via config
       7 steps (5 passed, 1 failed, 1 skipped)
       """
 
+  Scenario: Print unused definitions while stopping on the first failure
+    When  I run behat with the following additional options:
+      | option                     | value |
+      | --stop-on-failure          |       |
+      | --print-unused-definitions |       |
+      | features/failing.feature   |       |
+    Then it should fail with:
+      """
+      --- Failed scenarios:
+
+          features/failing.feature:13 (on line 15)
+
+      2 scenarios (1 passed, 1 failed)
+      7 steps (5 passed, 1 failed, 1 skipped)
+      """
+    And the output should contain:
+      """
+      --- No unused definitions
+      """
+
   Scenario: Do not stop on undefined steps by default
     When  I run behat with the following additional options:
       | option                        | value           |
--- a/src/Behat/Behat/Definition/Cli/UnusedDefinitionsController.php
+++ b/src/Behat/Behat/Definition/Cli/UnusedDefinitionsController.php
@@ -68,8 +68,14 @@ final class UnusedDefinitionsController implements Controller
         return null;
     }
 
-    public function registerDefinitionUsages(AfterSuiteTested $event): void
+    public function registerDefinitionUsages(SuiteTested $event): void
     {
+        // A suite cut short by --stop-on-failure is dispatched under the same name as an
+        // AfterSuiteAborted, and has no further definition usage to register.
+        if (!$event instanceof AfterSuiteTested) {
+            return;
+        }
+
         $environmentDefinitions = $this->definitionRepository->getEnvironmentDefinitions($event->getEnvironment());
         foreach ($environmentDefinitions as $definition) {
             if ($definition instanceof RuntimeDefinition) {


features/stop_on_failure.feature
```diff
@@ -89,6 +89,29 @@ Feature: Stop on failure via config
       7 steps (5 passed, 1 failed, 1 skipped)
       """
 
+  Scenario: Print unused definitions while stopping on the first failure
+    # The suite is cut short, so it contributes nothing to the report, exactly as the suites that
```
carlos-granados: We don't usually add comments to our test cases, we prefer to keep them to the actual code. I believe that the comment in the unused definitions controller is enough to understand this fix, so I think this other comment is not really needed

carlos-granados (changes requested):

features/stop_on_failure.feature
Amoifr: Removed, thanks. You are right that the controller comment carries the reasoning on its own.

carlos-granados (approved):
Looking good now, many thanks @Amoifr

acoulton (approved):
Thanks @Amoifr

PR #1887, opened 2026-09-07
feat: throw an exception if too many arguments are used

carlos-granados:
This is the second part of issue  #1691 where in 4.0 we throw an exception instead of showing a deprecation warning


acoulton (approved):
Thanks @carlos-granados

PR #1888, opened 2026-09-07
feat: remove the PHPUnit exception stringer

carlos-granados:
Now that https://github.com/Behat/Behat/pull/1872 has been merged into 4.x, we can do the second step to fix https://github.com/Behat/Behat/issues/1746 which is to remove the PHPUnit exception stringer


acoulton (approved):
Thanks @carlos-granados - should we try to create the extension before we merge this, to provide an upgrade path for anyone already using 4.x-dev? I might have time to look at that at the start of next week, unless you want to do it.

carlos-granados:
> Thanks @carlos-granados - should we try to create the extension before we merge this, to provide an upgrade path for anyone already using 4.x-dev? I might have time to look at that at the start of next week, unless you want to do it.

@acoulton it would be great if you could work on this at some point as I expect to be quite busy in the next couple of weeks

acoulton:
I've made a good start on the extension - not quite ready for review but I should be able to tidy it up tomorrow.  https://github.com/Behat/PHPUnitAssertionsExtension/issues/5 tracks progress.

stof:
should that extension live in the behat organization or in friends-of-behat ? We moved most extensions away

acoulton:
@stof fair question. IMO it should be in Behat because:

* it's a fairly light extension
* it's been split directly out of Behat core
* we, as Behat core, are going to be officially recommending people use it as part of the 4.0 upgrade (if they use phpunit assertions as previously suggested by our docs).
* it's been created by / will be maintained by people who are core maintainers.

If you feel strongly it should be in friendsofbehat then we could put it there.

I don't have any maintainer access in that org, or authority to submit from packagist from there - so I'd need to either be made a friendsofbehat maintainer or to work with someone else to get the repository properly configured and the composer package published. 

That might include needing access to set up a GitHub app for automated updates similar to what we have in behat/gherkin (I'm considering pinning the maximum phpunit versions so that we can make sure the extension continues to work with future phpunit changes). That's not for discussion here, but just flagging that there might be a need for org-level settings access (or to work with someone that has it).

I'd also be nervous of putting it in friendsofbehat without a clear plan for future maintainership. I don't want to merge my own PRs without another reviewer, but it seems like friendsofbehat often has a bit of a challenge for maintainer capacity and I don't want to worsen that. IMO the easiest way to avoid that is to keep this within core so that @carlos-granados, you and I can collaborate as we do on other official packages.

Bear in mind as well that this is almost the last blocker for 4.0 on our side, so I am keen to keep it moving - and keeping it in Behat feels like the quickest route to me.

carlos-granados:
> I'd also be nervous of putting it in friendsofbehat without a clear plan for future maintainership. I don't want to merge my own PRs without another reviewer, but it seems like friendsofbehat often has a bit of a challenge for maintainer capacity and I don't want to worsen that. IMO the easiest way to avoid that is to keep this within core so that @carlos-granados, you and I can collaborate as we do on other official packages.
> 
Agree with @acoulton that it would be best to keep it in the behat namespace

acoulton:
v1.0 of the extension is released - https://github.com/Behat/PHPUnitAssertionsExtension/releases/tag/v1.0.0 - we just need to get it published on packagist and then I'll merge this.

PR #1889, opened 2026-09-10
Explain the relationship between --format and --out

### LINKED ISSUE #1409: Option help text and error messages do not clearly explain the (indexed) relationship between --format and --out
```
behat --format=pretty --format=junit --out=some/dir
```

This yields an exception because the `--out` will actually apply to the first formatter `pretty`, not the `junit` one, creating UX issues and confusion.

An alternative is to merge the params like Docker does:

```
behat --format=pretty --format=junit,target=some/dir
```

Amoifr:
Fixes #1409, scoped to the two points @acoulton reopened it for: the option help and the error message.

Reproducing it first turned up something the issue does not say. A single `--out` is not applied to the first formatter, it is applied to **every** one, through `setFormattersParameter()`. The index pairing only starts once there are two or more `--out`. So in the reported command the console formatter receives the directory meant for junit, and complains about a path the user never intended for it.

```
$ behat --format=pretty --format=junit --out=some/dir

  Filename expected as `output_path` parameter, but got `/app/some/dir`.
```

The message now says where that directory came from, and adopts the wording the two sibling factories already use:

```
  A file name expected for the `output_path` option, but a directory was given.
  Note that a single `--out` is applied to every `--format`.
  Pass one `--out` per `--format`, in the same order, to send them to different places.
```

It is split over three lines because the Symfony error box does not wrap: as a single sentence it rendered as a box 230 characters wide. The path is no longer interpolated, matching `FileOutputFactory` and `FilesystemOutputFactory`, and it is still on the exception through `getPath()`.

The help of `--out` now states the rule, and `--format` points at it, which is what @stof asked for:

```
  -f, --format=FORMAT   ...
                        You can use multiple formats at the same time. See --out
                        to send each of them to a different place.
  -o, --out=OUT         Write format output to a file/directory instead of
                        STDOUT (output_path). A single --out is applied to every
                        --format. To send formats to different places, pass one
                        --out per --format, in the same order, using std
                        for those that stay on STDOUT.
                        This option is mandatory for the junit formatter.
```

A scenario in `features/multiple_formats.feature` pins the message, using the reported command. Worth knowing for whoever reads it: `-f pretty -f progress --out=file` does *not* fail, because two console formatters both write to the file and the second simply overwrites the first. Only mixing a console formatter with junit trips it, since junit creates the directory that the console formatter then chokes on.

#### One thing I left alone

The symmetric case is worse and I did not touch it, since it is behaviour rather than wording:

```
$ behat --format=junit --format=pretty --out=report.xml
```

junit creates a **directory** named `report.xml` containing `default.xml`, and then pretty fails saying a file name was expected. The user gave a file name, so the message contradicts what they typed. Happy to open a separate issue if you think that deserves one.

### DIFF
--- a/features/multiple_formats.feature
+++ b/features/multiple_formats.feature
@@ -278,6 +278,17 @@ Feature: Multiple formats
       25 steps (20 passed, 2 failed, 3 undefined)
       """
 
+  Scenario: a single output is applied to every format
+    When I run "behat -f pretty -f junit --out=apples-out"
+    Then it should fail with:
+      """
+      In ConsoleOutputFactory.php line XX:
+
+        A file name expected for the `output_path` option, but a directory was given.
+        Note that a single `--out` is applied to every `--format`.
+        Pass one `--out` per `--format`, in the same order, to send them to different places.
+      """
+
   Scenario: 2 formats, write second to file
     When I run "behat -f pretty -o std --format=progress --out=apples.progress --format-settings='{\"multiline\": false, \"paths\": false}'"
     Then it should fail with:
--- a/src/Behat/Testwork/Output/Cli/OutputController.php
+++ b/src/Behat/Testwork/Output/Cli/OutputController.php
@@ -42,15 +42,19 @@ class OutputController implements Controller
                 InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY,
                 'How to format tests output. <comment>pretty</comment> is default.' . PHP_EOL .
                 'Available formats are:' . PHP_EOL . $this->getFormatterDescriptions() .
-                'You can use multiple formats at the same time.'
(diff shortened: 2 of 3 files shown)


acoulton (approved):
Thanks @Amoifr 

On this point:

> junit creates a directory named report.xml containing default.xml, and then pretty fails saying a file name was expected. The user gave a file name, so the message contradicts what they typed. Happy to open a separate issue if you think that deserves one.

I think we leave that. Although it could be confusing, the user didn't necessarily give a file name (`report.xml` is a valid directory name, even if it looks like a file). The underlying issue is the design of the `--out` API and we could get into quite a lot of complexity trying to work out the most meaningful exception message for every possible usage.

The main thing is that the documentation of the `--out` parameter is now much clearer, so if users do have errors that should allow them to understand what's gone wrong.

carlos-granados (approved):
Thanks @Amoifr

PR #1890, opened 2026-09-13
Add support for Symfony 8.0

thetrueares:


acoulton:
@thetrueares thanks for the contribution but unfortunately it isn't possible to support symfony 8 with Behat 3.x due to BC breaks in the symfony API. If it was just a composer.json update we would have done that when symfony 8 was released.

Symfony8 is supported with Behat 4.x - check the pinned issue #1743 and the release notes for the full details - but the short version is 4.0.0-alpha1 is considered stable for end users, and is increasingly supported by Behat extensions.

It is only an alpha in case we have to make any small internal changes as part of the process of extension authors updating their code, but these should not affect projects that are only using Behat as a test/feature runner.

thetrueares:
Umm, it's possible. If it's a bigger issue, just say and I can create the pull request. It's irresponsible to advise someone to use alpha stability in their testing infrastructure.

thetrueares:
This is an actual issue, support 8.0 in 3! Otherwise, Behat has to be ditched. You don't have the resources to maintain this 4.x branch to say that we should just upgrade to use it. 8.0 and 8.1 are out, 8.2 will be out and probably 8.3 by the time you release 4.x.

thetrueares:
Literally commit https://github.com/Behat/Behat/commit/96ae2e07bd65681b5c458d0249535b974857897b that to main and we're good to go?

acoulton:
Wow, that is possibly the most unfriendly and entitled comment I've had on an issue in 20 years of maintaining open source packages.

First, literally nobody owes you anything. 

You don't pay  me, or anyone else on the Behat team as far as I know. And yet I answered you straight away on a Sunday to point you in the direction of a solution. Despite the issue you are encountering being already fixed, and [already documented](https://docs.behat.org/en/latest/releases.html).
 
Second, if you took a moment to look at the issues or PRs in the history of the repo, or indeed to think through the impact of your own PR, or to run the CI tasks including the BC checker, you would understand the problem. 

Merging your PR would add return type hints to the interface that is used by every Behat extension. That would in turn break every single Behat extension, and a lot of end-user projects - regardless of the symfony version they have installed - until they themselves add the return type. That is therefore a BC break in our own API and cannot go into a 3.x release. 

> It's irresponsible to advise someone to use alpha stability in their testing infrastructure.

Then wait for 4.0 or use something else to run your tests. As above, nobody owes you anything.

4.0 stable is not far away, and we've done a lot of work to make sure that future symfony releases are decoupled from the Behat API the current maintainers inherited to avoid this happening again. 8.1 and further minor 8.x releases will be supported from day 1.

It would be far more irresponsible to break everyone's projects with an API change in a 3.x release, even if we did it to stop someone calling us names on the internet.

Personally I completely disagree with your assessment:
* We're not suggesting using the alpha in prod.
* We have very clearly documented that from an end user perspective the 4.0-alpha is essentially finished and stable, and explained why we've given it an alpha label.
* We have guaranteed that there will be no breaking changes to 4.0 stable that affect end users, except for the planned changes that are listed in the release notes (and which only affect already-deprecated features and behaviours).

And if we are wrong (or "lying" 🤦‍♂️) about any of that then literally the only thing that will happen is your CI will fail until you pin back to the working alpha, or yell at us some more to fix it...


But seriously, if you're not happy with my reply do feel free not to use Behat.

thetrueares:
> Wow, that is possibly the most unfriendly and entitled comment I've had on an issue in 20 years of maintaining open source packages.

Please learn was entitled means.

> First, literally nobody owes you anything.

And the same right back to you. But if you've been in open source for 20 years then you should have learned how to communicate with the truth and not hit people with complete nonsense. Maybe you've not really been doing code review, because that was nowhere near that unfriendly. But I do think people owe people basic respect. 

> You don't pay me, or anyone else on the Behat team as far as I know. And yet I answered you straight away on a Sunday to point you in the direction of a solution. Despite the issue you are encountering being already fixed, and [already documented](https://docs.behat.org/en/latest/releases.html).

The issue is not fixed. No one can use a stable version of behat with Symfony 8.

No one was demanding you do anything other be honest. I wasn't expecting you to add it, I was doing the adding. Literally, I'm doing the work, so fuck right off with your entitled bullshit you're trying to spread. "I'll just say they're entitled and people will back me up."

> Merging your PR would add return type hints to the interface that is used by every Behat extension. That would in turn break every single Behat extension, and a lot of end-user projects - regardless of the symfony version they have installed - until they themselves add the return type. That is therefore a BC break in our own API and cannot go into a 3.x release.

Yea, I was literally fixing that. It's not a big deal. It's basically https://github.com/thetrueares/Behat/commit/6a748c05f800f56792f8f2383d353852b741360a and that is how you deal with these sort of issues in a library/project like this. This is why you should not have your head up your ass and talk to others and you'll learn that by collaboration you'll find out how to solve problems. 

> Then wait for 4.0 or use something else to run your tests. As above, nobody owes you anything. 

Yea, this is the big problem. Actually, it's just your attitude. The tech problem is minor. You've not realised that some people need to work in professional environments and not just hack and slash agencies. So you have projects with hundreds of tests and using Symfony 7.4, they want to upgrade to 8 and the only issue could be as is in mines Behat. So in your hack and slash agencies everything is new and it doesn't matter so much if you use something else quickly. But in actual production setups that isn't possible. So in your worldm people have to wait multiple versions of Symfony to be released before they can upgrade behat or throwaway all their tests.

> We're not suggesting using the alpha in prod.

Yes you are. Into a production testing env. Please learn what prod is for each kind of tool.

> We have very clearly documented that from an end user perspective the 4.0-alpha is essentially finished and stable, and explained why we've given it an alpha label.

Please learn what alpha means. You're mistaking alpha with release candidate.

> We have guaranteed that there will be no breaking changes to 4.0 stable that affect end users, except for the planned changes that are listed in the release notes (and which only affect already-deprecated features and behaviours).

Haha, you're building a new version that has bc breaks in it and you're guaranteeing that it will has no BCs while literally saying it has BCs. You're locking down the contracts at Alpha stage? You do know alpha is a private internal testing?

> And if we are wrong (or "lying" 🤦‍♂️) about any of that then literally the only thing that will happen is your CI will fail until you pin back to the working alpha, or yell at us some more to fix it...

WTF is this comment? Yell at you to fix it some more? I don't want to yell at anyone, I want to send pull requests and contribute but here we are, with a fixed issue not being fixed again so production users can't get that fix... In my 15 years of maintaining production systems which is more than yours made up 20 years of maintaining. Dude, we can literally look people up so there is no need to lie. And your inability to use terminology correct is a clear flag you're not as senior was you want to portray. I've never actually had such problems getting an already fixed fix in. And you want people to yell to fix it but you're not fixing this one because, it's hard? Or you don't care? 

> But seriously, if you're not happy with my reply do feel free not to use Behat.

Maybe don't be an arsehole and merge a fix that allows people to use behat with Symfony 8.x? It's not that hard to do it. 

You don't owe me anything, and I owe you nothing so when it came to holding back I didn't owe you it. So what you need to do is start thinking about how highend comptent software engineers working in a professional environment will need from their tooling and stop focusing on hack and slash developers. It'll take you quite sometime to get to that level of understanding. But I don't care if you do it or not, it's you that looks stupid when you come out with nonsense like this. Straight up lies.

Anyways, I get it, it's too hard for you. I'll get my fork working and then see about hiring some developers to take over Behat and have the project act like a professional tool for professionals.

But seriously, codeception is probably going to need to be used. They actually maintain their shit and fix stuff.

acoulton:
Ok. I nearly locked this originally but left it open in the hope you might review the way you approached this and come back with a more pleasant response.

I'm disheartened you've instead doubled down. I will lock this now. Life is too short for arguing with strangers on the internet.

Interactions like this are a big part of why it's hard to get people to volunteer to maintain open source.

If you would like to contribute in a constructive way in future you will be welcome.

If not you will be blocked.

I wish you success with your fork.

acoulton:
For what it's worth (mostly for anyone that stumbles across this and finds the genuine question lurking in the abuse), there are indeed reasons why we've gone with a new major rather than ship a sniff/hack to switch between two different Extension interfaces at runtime.

* Historically, Behat did not enforce (or even advise) that extension authors specified the symfony version(s) they support. A lot of code assumes that the symfony interfaces and classes provided by Behat are stable.
* By default then if we'd added a workaround and a project supports symfony 8 and Behat 3, `composer update` would give it symfony 8. Our sniffed Extension class will then be fine, but any extensions the project uses will likely themselves fatal on the missing return types.
* There are also other BC breaks in symfony 8 that could well affect extensions and Behat users even though those are not visible in our own interface signatures - e.g. in the symfony classes that we pass to an Extension at runtime. Again, if we claim Behat 3 supports symfony 8, all of that code will break on a composer update.

Bear in mind in some cases the project itself will have no direct symfony dependency, and the symfony 8 update will happen whenever composer is able to resolve a set of third-party dependencies that all support symfony 8 (or do not have a symfony constraint).

IMO a BC break in a minor release is bad. A BC break that may trigger at some arbitrary later date depending on the dependencies of dependencies would be extremely bad practice.

Therefore, it felt like the responsible solution was not to make promises in composer.json that Behat 3.x can't keep, and instead to ship a new major as an opportunity for the ecosystem to check/add symfony8 support in a managed and predictable way.

The unconventional use of an alpha tag was also intentional, to encourage people to really read the release notes and decide if the stability/promises about further changes were a reasonable tradeoff for their usecase - rather than seeing an RC and assuming it would be fine.

People may disagree with any of those decisions, but they came from a place of care for Behat users and in particular the impact on people who depend on the stability of their long-standing Behat suites.

And the good news is that as articulated previously, 4.0.0-alpha is 100% ready to use in CI if you read the release notes. And 4.0.0-stable is now very close.

PR #1894, opened 2026-09-18
test: Upgrade to PHPUnit 11

### LINKED ISSUE #1828: Upgrade to a newer/supported PHPUnit
We are currently on phpunit 9, which is on life support.

CI has just been blocked by a security vulnerability that affects the 9.x series (see #1827). While this one is relatively safe to ignore, it's likely we'll have similar issues to review / ignore in future given vulnerability reporting is now more prevalent and composer is enforcing those at install time by default.

Therefore we should update to a supported version of phpunit sooner rather than later.

This will involve updating our FeatureContext not to use PHPUnit assertions, due to #1618 - this was probably already in scope for #1746.

acoulton:
Now that our own feature coverage don't use PHPUnit assertions, we can upgrade to a newer version. We can't yet go beyond version 11, as we need to be able to run tests on PHP 8.2 and version 12 is >= 8.3.

As part of this branch, I have enabled Rector on the unit tests directory (it wasn't included before) which has included a couple of CS fixes.

I've also resolved use of PHPUnit deprecated features (use attributes instead of annotations for data providers etc, and use `$this->willReturnValueMap()` instead of `$this->returnValueMap()`).

I also had to update the feature & test fixtures for Behat stringifying PHPUnit assertion failures to work with the now-installed PHPUnit version. Note that nothing in the actual stringification implementation has changed, only our tests for it - so there is no change to how anything works for end-user projects using older PHPUnit versions. These fixes will obviously only be relevant for 3.x, as 4.0 removes built-in support for stringifying PHPUnit exceptions.

Fixes #1828.


features/phpunit_exceptions.feature
```diff
@@ -10,14 +10,14 @@ Feature: Stringifying PHPUnit exceptions
       | --no-colors |          |
       | --format    | progress |
 
-  Scenario: With PHPUnit 9 working correctly
-    When I run "behat features/with_phpunit_9.feature"
+  Scenario: With PHPUnit 11 working correctly
```
carlos-granados: Should we continue naming a specific version? I think it's best to just say "with PHP unit working correctly" so that we don't need to continue updating this feature as we increase the version used

carlos-granados (commented):
Great job @acoulton, just a question

features/phpunit_exceptions.feature
acoulton: Good shout - I've updated the fixture feature file to be `with_supported_phpunit` and reworded. On reflection it's not really about whether PHPUnit is "working correctly" - we're outside the supported use of PHPUnit, so PHPUnit could be working "correctly" but just not how we expected it to... 

So I've avoided that wording and gone for "With a version of PHPUnit that we support, working as expected".

carlos-granados (approved):
Perfect now @acoulton

